PR #8, opened 2021-12-15
Change val keyword to let.

dabrahams:
Not completely tested because of https://github.com/val-lang/val/issues/9


dabrahams:
This passes, though:

```
VAL_HOME=$(pwd) swift test --skip testEval
```

kyouko-taiga:
The interpreter is broken at the moment, because of some changed in VIL representation. Your branch passes the type checker tests, so all is good.

PR #24, opened 2022-04-12
Omit needless and misleading words, etc.

dabrahams:
1. It isn't a Collection
2. The thing you get by subscripting it is not the module declaration
3. That it “represents” an AST isn't any different from saying it *is* an AST.
4. Notes to the implementor don't go in doc comments
5. Saying that an array is "a collection with" something is needless.  Saying the modules are “of the AST” is needless; it's a property!
6. Calling the source range of each node a “source range annotation” is needless.
7. Referring to “the AST” is a convention that doesn't scale.  What happens when there's a parameter of `Self` type? 
8. Referring to “the node,” ditto.

If tooling isn't telling us whether a subscript is just for reading or for read/write, without reading a comment, that's pretty sad.

### DIFF
--- a/Sources/Compiler/AST/AST.swift
+++ b/Sources/Compiler/AST/AST.swift
@@ -1,33 +1,32 @@
-/// A collection of module declarations representing an abstract syntax tree.
+/// An abstract syntax tree.
 public struct AST {
 
-  /// The nodes in the AST.
-  ///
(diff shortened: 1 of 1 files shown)


kyouko-taiga (approved):
Great changes!

dabrahams:
instructive I hope.  The doc comments are definitely getting better; these are things to consider when writing them.

PR #26, opened 2022-04-16
Fix typos.

dan-zheng:


Sources/Compiler/TypeChecking/TypeChecker.swift
```diff
@@ -1082,6 +1082,7 @@ public struct TypeChecker {
   /// Realizes and returns the traits of the specified conformance list, or `nil` if at least one
   /// of them is ill-formed.
   ///
+  /// FIXME(kyouko-taiga): Finish `Note:` comment below.
   /// - Note: All return
```
dan-zheng, line 1086: Note: this doc comment is unfinished.

kyouko-taiga: Yep, I saw that. Haven't pushed yet. Thanks.

PR #29, opened 2022-04-19
Parser (WIP)

dabrahams:


kyouko-taiga:
I think this PR is obsolete.

The branch [is tagged](https://github.com/val-lang/val/tree/archive/feature/parser-generator) so that we may restore it in the future if we want to re-open investigation on the use of a different parser technology.

PR #51, opened 2022-10-20
Translate to C++ module and generate .hpp and .cpp file for the C++ module

lucteo:
**Why?**
To properly handle the C++ transpile cases, we should transform Val AST to CXX AST before emitting C++ code.
We shall start by translating Val's `ModuleDecl` into a corresponding `CXXDecl`.

**How?**
- described the approach of building CXX AST nodes in the internal Readme.md
- created `CXXModuleDecl` (along with a simple `CXXDecl` and `CXXNode`)
- defined methods in `CXXModuleDecl to emit the header and the implementation C++ code
- the generated code for the C++ header has: header guard, common includes and module namespace
- the generated code for the C++ header has: include to the header, module namespace
- changed the main code to transpile through `CXXModuleDecl`

**Testing**
- transpiled `factorial.val` and inspected the generated `.hpp` and `.cpp` files to match the above content
- note: top-level decls are not emitted yet


Sources/Compiler/CXX/CXXModuleDecl.swift
```diff
@@ -0,0 +1,51 @@
+/// A C++ module.
+public struct CXXModuleDecl : CXXDecl {
+
+  /// The corresponding Val AST node
+  let original: ModuleDecl
+
+  /// The top-level declarations in the module
+  var topLevelDecsls: [CXXDecl] = []
```
dabrahams, line 8:
```suggestion
  var topLevelDecls: [CXXDecl] = []
```

Sources/Compiler/CXX/CXXNode.swift
```diff
@@ -0,0 +1,7 @@
+/// A protocol describing the API of a CXX AST node.
+public protocol CXXNode {
```
dabrahams, line 2: This protocol seems premature, especially since it has no requirements.  How do we know that CXX AST nodes have any common API?

lucteo: Agree. Will remove

PR #52, opened 2022-10-22
Progress on the type checker

kyouko-taiga:


Sources/Compiler/TypeChecking/ConstraintSolver.swift
```diff
@@ -555,7 +555,7 @@ struct ConstraintSolver {
       diagnostics: diagnostics)
 
     for c in stale {
-      s.diagnostics.append(.staleConstraint(constraint: c.constraint, range: range(of: c.location)))
+      s.diagnostics.append(.staleConstraint(constraint: c.constraint, at: range(of: c.location)))
```
dabrahams, line 558: FWIW, `range(of: c.location)` looks like boilerplate we could do without.

PR #55, opened 2022-10-28
Remove 'converting:' label on initializers.

dabrahams:
These are standard value-preserving conversions just like `Int.init?(_: String)`, for which the explicitly established convention is "no parameter label."

### DIFF
--- a/Sources/Compiler/AST/Decl/FunDecl.swift
+++ b/Sources/Compiler/AST/Decl/FunDecl.swift
@@ -85,7 +85,7 @@ public struct FunDecl: GenericDecl, GenericScope {
   /// The declaration of the implicit receiver parameter, if any.
   public var implicitReceiverDecl: NodeID<ParameterDecl>? {
     if let parameter = implicitParameterDecls.first, parameter.name == "self" {
-      return NodeID(converting: parameter.decl)
+      return NodeID(parameter.decl)
     } else {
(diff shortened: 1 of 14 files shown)


kyouko-taiga (commented):
The change makes sense in Swift. One thing to consider, however, is that I have generally tried to write code as it would be written in Val, as an exercise to get a first sense of what Val's constraint entail in a non-trivial code base.

There is no way to overload an initializer like `AnyScopeID.init` so that it would behave as the change proposes in Val.

dabrahams:
Uh-huh.  I don't think we should constrain ourselves that much when writing Swift code, and I think we will need an exception for initializers or *some* other standard for value-preserving conversions, because types will often have many of these.  Both of the inits of `AnyScopeID` are "converting".

For all of these I would have used `downcasting` because it's more specific.

dabrahams:
In Val these things will end up being `.asNodeID` accessors I think.  I'll try that change instead.

dabrahams:
Nope, works badly for the generic conversions.  The problem with `converting:` as a convention, aside from being ugly, is that it doesn't scale.  If you need to convert from two types, now you need another label.

dabrahams:
I think we can agree that the ban is going away, yes?

kyouko-taiga:
Yep

PR #57, opened 2022-10-30
NodeMap cleanup and rename

dabrahams:


kyouko-taiga (approved):

kyouko-taiga:
Shouldn't we also change `DeclMap` and `ExprMap` then?

dabrahams:
Probably, if they are similar.  I don't yet have a bird's eye view of the codebase.

PR #58, opened 2022-10-30
Encapsulate stored properties

dabrahams:


Sources/Compiler/AST/Attribute.swift
```diff
@@ -11,10 +11,10 @@ public struct Attribute: Codable {
   }
 
   /// The name of the attribute.
-  var name: SourceRepresentable<String>
+  let name: SourceRepresentable<String>
```
kyouko-taiga, line 14: Why make that property a constant? I generally think that simple data aggregates that only have non-private stored members should be completely mutable.

kyouko-taiga (commented):
I don't understand the benefit of making most of these stored properties immutable. I generally think that simple data aggregates (e.g., a pair) should have mutable stored properties to avoid more verbose and potentially less efficient functional updates.

Sources/Compiler/AST/Attribute.swift
dabrahams: Isn't there a relationship between the name and the arguments?  If I change the name without changing the arguments, is it always valid?  Is any string at all acceptable as a name?

kyouko-taiga: As of now, there's isn't any relationship encoded anywhere, and no restriction on the name. Otherwise I would probably turn the whole type into an enum, as we have for introducers.

dabrahams:
The problem, when there hasn't been enough discipline about access control, is I can't tell which properties are intended to be part of the public API, nor which are intended to publicly include mutation, so I just made everything as restricted as the current code would allow.  If you have specific objections, tell me which and we can look at rolling them back.

kyouko-taiga:
Okay, sounds like a plan. I have objections for:
- Attribute.swift
- DelcID.swift (do not make `base` private, so as to be consistent with all other node IDs)
- PatternID.swift (same as above)
- StmtID.swift (same as above)

Sources/Compiler/AST/Attribute.swift
dabrahams: My point is, if I just change the name without changing the arguments, it potentially becomes an invalid/unrecognized attribute, right?

kyouko-taiga: I think it is premature to enforce any restriction here, because there is no restriction documented anywhere. The parser is happy to parse `@foo(bonjour)` and let such an unrecognized attribute get its way into the AST.

PR #60, opened 2022-10-30
nix implicitly-unwrapped-optionals

dabrahams:


Sources/Compiler/TypeChecking/TypeChecker.swift
```diff
@@ -2868,4 +2868,10 @@ public struct TypeChecker {
     return (transformed, [])
   }
 
+  /// Resets `self` to an empty state, returning `self`'s old value.
+  mutating func release() -> Self {
```
kyouko-taiga, line 2872: Do we need this method already? It's not called anywhere if I'm not mistaken.

I think we could use a different finalization method for the type checker. I defined a `run` method (which is no longer used in the current code) that would produce a typed program by consuming the checker's property maps.

dabrahams: It is used on line 1408 of this file.

kyouko-taiga: Okay then. Let's go with this version for the time being. We'll change if necessary when we have a better idea of what the final version of the driver should look like.

PR #62, opened 2022-10-31
Transpiler infrastructure

kyouko-taiga:


dabrahams:
I guess you decided I wasn't going to make progress on that soon enough… 😰

kyouko-taiga:
I'm just laying out the foundations.

dabrahams:
That's great; I'm sure it will be a big help.

Sources/Compiler/CXX/CXXIdentifier.swift
```diff
@@ -1,23 +1,23 @@
...
+  public let description: String
 
-  init(_ identifier: String) {
+  public init(_ identifier: String) {
     description = CXXIdentifier.sanitize(identifier)
   }
 
   /// Sanitizes `identifier` and returns a valid C++ identifier.
-  static func sanitize(_ identifier: String) -> String {
+  public static func sanitize(_ identifier: String) -> String {
```
dabrahams, line 12: I assume at some point we'll need to add some smarts to ensure the result is unique

Sources/Compiler/CXX/CXXModule.swift
```diff
@@ -0,0 +1,116 @@
...
+
+  // MARK: Serialization
+
+  /// Emits the C++ header of the module.
+  public func emitHeader() -> String {
+    var output: String = ""
+
+    // Emit the header guard.
+    output.write("#ifndef __VAL_\(name.uppercased())__\n")
+    output.write("#define __VAL_\(name.uppercased())__\n")
```
dabrahams, line 72: Everybody uses underscores this way but they  [are all wrong ](https://stackoverflow.com/a/228797).
 
```suggestion
    output.write("#ifndef VAL_\(name.uppercased())_\n")
    output.write("#define VAL_\(name.uppercased())_\n")
```

Sources/Compiler/CXX/CXXIdentifier.swift
kyouko-taiga: I'm hoping we can come up with a generator that guarantees uniqueness. Val names are guaranteed unique by the compiler, so that should be possible.

PR #63, opened 2022-10-31
Nontrivial immutable AST

dabrahams:


Sources/Compiler/Parse/Parser.swift
```diff
@@ -2465,14 +2465,15 @@ public enum Parser {
...
-          range: head.range)
+        var labels: [String?] = []
 
         if context.currentCharacter == "(" {
-          var labels: [String?] = []
           let backup = context.backup()
           _ = context.take()
+          var closeParenFound = false
+          defer {
+            if !closeParenFound || labels.isEmpty { context.restore(from: backup) }
```
dabrahams, line 2475: @kyouko-taiga I don't really know what's up with this `context` thing, and the logic for restoration is curious.  I can kinda intuit why you'd want to not restore if the close paren is found but the fact that context is always restored when there are no labels is…somewhat non-obvious.  If you could add a suggested change with a comment that explains, I'd be grateful.

Sources/Compiler/Parse/Parser.swift
```diff
@@ -2465,14 +2462,15 @@ public enum Parser {
...
```
kyouko-taiga, line 2472:
```suggestion
            // Backtrack if we didn't find a closing parenthesis. That will let the argument-list
            // parser process the stream after the identifier to either catch a parse error if the
            // former case or parse an empty argument list in the latter.
            if !closeParenFound || labels.isEmpty { context.restore(from: backup) }
```

Note: I think errors after the opening parenthesis should be handled by the argument list parser. I expect ill-formed argument lists to be more likely than ill-formed labeled named. For example, if we see `foo(bar:???`, I think `???` the error should be diagnosed expecting `???` to be an expression, not a label.

dabrahams: Don't understand your note, but thanks for the suggestion!

dabrahams: @kyouko-taiga unfortunately, the suggestion doesn't address the problem, because it doesn't explain why `labels.isEmpty` should be causing backtracking.  Could you try another?

kyouko-taiga:
```suggestion
            // Backtrack if we didn't find a closing parenthesis or if there are no labels. That
            // will let the argument-list parser pickup after the identifier to either catch a
            // parse error in the former case (no closing parenthesis) or parse an empty argument
            // list in the latter (no labels).
            // Note: `foo()` is *not* a valid name, it's a function call.
            if !closeParenFound || labels.isEmpty { context.restore(from: backup) }
```

kyouko-taiga: To give more context in case you want to refine the comment yourself, the rule in the specification is defined as follow

```ebnf
function-entity-identifier ::= (token)
  identifier '(' argument-label+ ')'
```

The subtlety here is that `argument-label` should appear one or more times, not zero or more. So we backtrack when we found none. That's necessary to disambiguate a function call with no argument, as suggested in the comment.

Sources/Compiler/Parse/Parser.swift
```diff
@@ -2465,14 +2464,23 @@ public enum Parser {
...
-          range: head.range)
+        var labels: [String?] = []
 
         if context.currentCharacter == "(" {
-          var labels: [String?] = []
           let backup = context.backup()
           _ = context.take()
+          var closeParenFound = false
+          defer {
+            // Backtrack if we didn't find a closing parenthesis. That will let the argument-list
```
kyouko-taiga, line 2474: Oh, seems like my first suggestion hasn't been replaced.

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
```diff
@@ -565,17 +565,17 @@ struct ConstraintGenerator: ExprVisitor {
   mutating func visit(tuple id: NodeID<TupleExpr>) {
     defer { assert(inferredTypes[id] != nil) }
 
-    var expr = checker.ast[id]
+    var xxxx = checker.ast[id].elements
```
kyouko-taiga, line 568: Is this supposed to be temporary? `xxxx` is a strange identifier.

PR #68, opened 2022-11-04
Program abstraction

kyouko-taiga:


Sources/Compiler/TypeChecking/ConstraintGenerator.swift
```diff
@@ -3,7 +3,7 @@ import Utils
 /// A visitor that generates constraints based on the structure of the AST.
 struct ConstraintGenerator {
 
-  /// The result of constraint generation.
+  /// A data structure aggregating the results of constraint generation.
```
dabrahams, line 6: I don't think this is an improved comment.  Every type is "a data structure."  Instead, characterize the results.  Are they constraints?

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
```diff
@@ -241,16 +248,16 @@ struct ConstraintGenerator {
...
     switch checker.program.ast[id].failure {
     case .expr(let elseExpr):
-      assume(typeOf: id, equals: inferredType!)
+      assume(typeOf: id, equals: inferredType!, at: checker.program.ast.ranges[id])
       expectedTypes[elseExpr] = inferredType
       visit(expr: elseExpr, using: &checker)
 
     case .block(let thenBlock):
-      assume(typeOf: id, equals: .void)
+      assume(typeOf: id, equals: .void, at: checker.program.ast.ranges[id])
```
dabrahams, line 256: We should talk about a better name for `ranges` :-)

Sources/Compiler/TypeChecking/ConstraintSolver.swift
```diff
@@ -42,47 +27,63 @@ struct ConstraintSolver {
   /// The score of the best solution computed so far.
   private var best = Solution.Score.worst
 
+  init(
```
dabrahams, line 30: Comment please!  Consider the client of this type who isn't looking at its private members and their documentation.  How is the client to know the meaning of the initializer arguments?

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: I could write:

> The types inferred by constraint generation for the visited expressions, along with the constraints between these types and the diagnostics of the error encountered.

But I feel I'm only repeating what you get from reading the 10 lines that follow. What would you do?

Sources/Compiler/TypeChecking/ConstraintSolver.swift
```diff
@@ -218,14 +220,15 @@ struct ConstraintSolver {
       }
 
     default:
-      diagnostics.append(.incompatibleTypes(l, r, at: range(of: location)))
+      diagnostics.append(.incompatibleTypes(l, r, at: location.range))
```
dabrahams, line 223: I don't know what a `location` is and why it has-a `range`, but I think it would be better to create a static initialization function in whatever type is being appended here to eliminate the boilerplate of reaching in to get the range each time.

Sources/Compiler/TypeChecking/ConstraintSolver.swift
```diff
@@ -413,27 +418,35 @@ struct ConstraintSolver {
     fatalError("not implemented")
   }
 
-  private mutating func solve(disjunction: LocatableConstraint) -> Solution? {
+  private mutating func solve(
```
dabrahams, line 421: doc comment please.

Sources/Compiler/TypeChecking/ConstraintSolver.swift
```diff
@@ -413,27 +418,35 @@ struct ConstraintSolver {
...
+      using: &checker,
       insertingConstraintsWith: { (subsolver, branch) -> Void in
         for constraint in branch.constraints {
           subsolver.fresh.append(LocatableConstraint(constraint, location: disjunction.location))
         }
       })?.solution
   }
 
-  private mutating func solve(overload: LocatableConstraint) -> Solution? {
+  private mutating func solve(
```
dabrahams, line 438: doc comment please.  What does it mean to solve a constraint?

Sources/Compiler/TypeChecking/LocatableConstraint.swift
```diff
@@ -7,6 +7,9 @@ struct LocatableConstraint {
     /// The AST node to which the constraint is attached, if any.
     var node: AnyNodeID?
 
+    /// The source range from which the constraint originates, if any.
+    var range: SourceRange?
```
dabrahams, line 11: How about "`origin`?"

Sources/Compiler/TypeChecking/TypeChecker.swift
```diff
@@ -1358,37 +1362,29 @@ public struct TypeChecker {
...
 
     // Solve the constraints.
     var solver = ConstraintSolver(
-      checker: self,
       scope: scope,
-      fresh: constraints,
+      fresh: constraints + constraintGeneration.constraints,
       initialDiagnostics: constraintGeneration.diagnostics)
-    let solution = solver.solve()!
+    let solution = solver.apply(using: &self)
```
dabrahams, line 1378: Hmm, "using" seems a little weak semantically.  What mutation occurs in `self`?

dabrahams (commented):
Some notes

Sources/Compiler/TypeChecking/ConstraintSolver.swift
kyouko-taiga: Sorry I don't understand.

`incompatibleTypes` is a static initialization function of `Diagnostic`, which is what's being appended here.

dabrahams: In that case, I suggest you have it take a location parameter instead of a range (or overload as necessary) so you're not repeating `location.range` all over the place.  All code repetition is an opportunity for simplification.

kyouko-taiga: Hmm, I don't know. That's a lot of boilerplate someplace else. There are many static initialization function for the diagnostics. I'm not convinced this change will make the code simpler, and not all take a constraint location.

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
dabrahams: Documentation is _supposed_ to be somewhat redundant with the code, so that's not an issue.  What you want to do is express things at a higher level of abstraction if possible.  From the doc comment, I want to be able to answer the question, “what is this thing?” at some meaningful level without looking at the details.  If I don't know what "frobnicating" actually means, saying “it's the result of frobnicating” isn't useful.  IMO, your “redundant” comment is much better than not having it.

Note that, if the description makes you uncomfortable, it may be because it sounds a bit like a bag of properties.  When the human-level description is awkward it can be a clue that your API isn't doing a good job of representing the abstraction inherent in the problem.  However, while you're making code better it's often not obvious how to improve the abstraction and you may want to live with the discomfort for a while before you discover how to restructure things.


you have trouble describing it at a higher level of abstraction, that can be a clue that you've designed the wrong API.

Sources/Compiler/TypeChecking/TypeChecker.swift
dabrahams: Well, at least `updating` would be a little more descriptive.

Since the type checker itself stores deduced type information and you're modifying it here, though, why are you stashing that information in `solution`?  This gets to how `Solution` may not have the right API.

Sources/Compiler/TypeChecking/ConstraintSolver.swift
dabrahams: You and I must have very different definitions of “boilerplate” then.  For me it's always something that's repeated.  I'm suggesting you not repeat `location.range` at all in the context of creating an `.incompatibleTypes` error.

Sources/Compiler/TypeChecking/TypeChecker.swift
kyouko-taiga: It's a little more complex than that.

There is information that can be deduced directly from the source code, which is what type realization and name binding do (in most cases). Other information can only be inferred by solving constraints, which is what constraint solving does.

When you start solving a constraint set, you don't know whether the substitution you are making are actually correct. You cannot store your assumptions in the property maps yet because solving the constraint set might make you explore different paths (because of all forms of polymorphism). So you stash them in a `solution` until you're confident you found the best one given the constraint you had.

I can comment on that in the API of solution, but really that is more a high-level explanation of how type inference work in general.

dabrahams: I might choose a different name for `Solution` in that case.  IIUC, it's a solution to the constraints, but then there's an ordering applied to solutions to choose the best one, and _that's_ the solution of the larger problem.  And if you're carrying diagnostics around in this thing, I'm guessing it's not even necessarily a solution to the constraints, but a _potential_ solution.  Is it a `substitution`?  is it a `TypeVariableAssignment`?

kyouko-taiga: It's more than that. It's a substitution map (from type variables to types) as well as a table mapping identifiers to their resolved declaration in case they were overloaded.

kyouko-taiga: I also store the relative cost of that solution, which is computed by looking at the number of non-trivial type substitutions it involves, and used for ordering.

All solutions "satisfy" a constraint set in the sense that they provide a way to understand how it may apply to the AST. If there are errors in the substitutions, that's because there's no way to apply the constraints without surfacing those errors.

dabrahams: I'd be inclined to call it something along the lines of `TypeVariableAssignment` then.  It carries auxiliary information about resolved declarations.  That might even be non-salient.  It's at least redundant for a given context.

kyouko-taiga: I disagree. There's no redundant information in a solution. Nothing that's placed here has yet been written in the property maps of the program. That's all "tentative" information until we decide to apply the solution. The value of a solution is defined by its two tables, its score, and the diagnostics it carries.

IMO, `Solution` makes sense as an abstraction. You run a type solver and you get several solutions, each of them potentially substituting type variables by different types, mapping identifiers to different declarations, and carrying different diagnostics related to these choices.

I don't think `TypeVariableAssignment` is an improvement over `Solution`. I may be inclined to place `Solution` in `ConstraintSolver` so that its qualified name carries more information.

Sources/Compiler/TypeChecking/ConstraintSolver.swift
kyouko-taiga: Let's opt for the status quo for the time being. We can discuss how to refine the diagnostics API later. That's not directly related to this PR.

Sources/Compiler/TypeChecking/TypeChecker.swift
kyouko-taiga: Circling back to the label `using` or `updating`, I don't think the latter is more descriptive. We are *using* the type checker to perform the action, not simply updating it. The fact that we must pass it inout suggests that it will be updated.

I suggest we preserve the status quo here.

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: Maybe it'd be best to discuss suggestions in an issue. I use this term consistently across all source files.

Sources/CLI/main.swift
```diff
@@ -92,7 +92,7 @@ struct CLI: ParsableCommand {
     /// The name of the product being built.
     let productName = "main"
     /// The AST of the program being compiled.
-    var rawProgram = AST()
+    var ast = AST()
```
dabrahams, line 95:
```suggestion
    var syntax = AST()
```

Just a thought.

dabrahams:
As I'm poking into the transpilation code, I'd like if each `Program` contained an (`Optional`) reference to its entry point, the `main` function.

Sources/CLI/main.swift
kyouko-taiga: Hmm, I don't think it's a big improvement. This value is indeed an abstract syntax tree and "AST" is a pervasive-enough acronym that I confident using it liberally.

kyouko-taiga:
> As I'm poking into the transpilation code, I'd like if each `Program` contained an (`Optional`) reference to its entry point, the `main` function.

Sounds like a reasonable request. How would you deal with ill-formed programs that declare multiple entry points? Just return the first now we found?

dabrahams:
> Sounds like a reasonable request. How would you deal with ill-formed programs that declare multiple entry points? Just return the first we found?

That, or `nil`.  It doesn't really matter I think.

PR #69, opened 2022-11-06
Fact return type int

lucteo:
**Why**
Make progress on transpiler the factorial example. In this step we translate the factorial function (without parameters).

**How**
* treated the case of product types in CXXTypeExpr
* if the Val type is `Int`, transform it to C++ `int`
* add test for functions returning `Int`

**Testing**
* transpiling factorial.val should not crash -> manual inspection
* add test case for function returning `Int`

### DIFF
--- a/Sources/Compiler/CXX/CXXTypeExpr.swift
+++ b/Sources/Compiler/CXX/CXXTypeExpr.swift
@@ -16,6 +16,14 @@ public struct CXXTypeExpr: CustomStringConvertible {
     case .void:
       description = isReturnType ? "void" : "std::monostate"
 
+    case .product(let productType):
+      switch productType.name.value {
+      case "Int":
+        description = "int"
+      default:
+        description = productType.name.value
+      }
+
     default:
       return nil
     }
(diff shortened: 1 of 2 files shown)


dabrahams (approved):
LGTM

Sources/Compiler/CXX/CXXTypeExpr.swift
```diff
@@ -16,6 +16,14 @@ public struct CXXTypeExpr: CustomStringConvertible {
     case .void:
       description = isReturnType ? "void" : "std::monostate"
 
+    case .product(let productType):
+      switch productType.name.value {
```
kyouko-taiga, line 20: That won't work because one could write a product type named `Int` in any scope, thereby shadowing the standard library. `AST.coreType(named: "Int")` should be used instead.

kyouko-taiga (commented):
Name shadowing will break tests based on unqualified names.

Sources/Compiler/CXX/CXXTypeExpr.swift
```diff
@@ -17,10 +17,11 @@ public struct CXXTypeExpr: CustomStringConvertible {
       description = isReturnType ? "void" : "std::monostate"
 
     case .product(let productType):
-      switch productType.name.value {
-      case "Int":
+      // TODO: we should translate this to an "int" struct
+      let intType = ast.coreType(named: "Int")
+      if productType == intType {
```
kyouko-taiga, line 22:
```suggestion
      if productType == ast.coreType(named: "Int") {
```

kyouko-taiga (approved):
LGTM

PR #77, opened 2022-11-09
Transpile to C++ function parameters

lucteo:
**Why**
Make progress on transpiler the factorial example. In this step we translate the factorial function parameters.

**How**
- translate Val function parameter types to CXX types
- take the bare type from the parameter type when translating a parameter type to C++
- currently ignore the passing convention

**Testing**
- transpiling factorial.val should produce the right declaration for the factorial function
- add test case for function taking an `Int`

### DIFF
--- a/Sources/Compiler/CXX/CXXModule.swift
+++ b/Sources/Compiler/CXX/CXXModule.swift
@@ -31,11 +31,8 @@ public struct CXXModule {
 
     assert(program.isGlobal(valFunDecl))
 
-    // The actual fun decl
-    let funDeclAst = program.ast[valFunDecl]
-
     /// The identifier of the function.
-    let identifier = CXXIdentifier(funDeclAst.identifier?.value ?? "")
+    let identifier = CXXIdentifier(program.ast[valFunDecl].identifier?.value ?? "")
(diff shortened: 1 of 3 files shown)


Sources/Compiler/CXX/CXXModule.swift
```diff
@@ -31,10 +31,13 @@ public struct CXXModule {
 
     assert(program.isGlobal(valFunDecl))
 
+    // The actual fun decl
+    let funDeclAst = program.ast[valFunDecl]
```
kyouko-taiga, line 35: I generally find it harder to trace back the value of this kind of local variables. I think it's better to keep the full expression (here `program.ast[valDunDecl]`) when it is not prohibitively too long.

Also, AFAIK, the name convention in Swift is to write all letters of an abbreviation in upper case. So the variable would be `funDeclAST`.

The `AST` suffix doesn't contribute very valuable information. If you want to insist on creating a local variable, I would call it `funDeclNode`.

kyouko-taiga (commented):
See my comment.

kyouko-taiga (approved):

Issue #70, opened 2022-11-07
Human-readable representation of serialized ASTs


kyouko-taiga:
The AST implements `Codable` to get serialized as JSON:

```bash
valc --emit raw-ast -o main.json main.val
```

The resulting json file is barely usable by a human, though. It's just an array of quite shallow data structures. It would be interesting to transform that representation to something human-readable (in JSON or another format).

kyouko-taiga:
A 2 digit issue! Ah, simpler times.
I hadn't thought about using JSON navigation tools, that is a great idea. I think we can close the issue then.

PR #78, opened 2022-11-09
Also write the description of the NodeKind when serializing to json

lucteo:
**Why**
Better readability (by humans) of the generated AST of Val nodes

**How**
- Also write the `description` field of NodeKind
- Also write the source text corresponding to a source range

**Testing**
Manually inspected the generated ast.json

### DIFF
--- a/Sources/Compiler/AST/NodeIDs/NodeKind.swift
+++ b/Sources/Compiler/AST/NodeIDs/NodeKind.swift
@@ -15,6 +15,17 @@ public struct NodeKind: Hashable, Codable {
     self.rawValue = rawValue
   }
 
+  enum MyCodingKeys: String, CodingKey {
+      case rawValue
+      case description
+  }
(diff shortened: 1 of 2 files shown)


kyouko-taiga:
Does this change have an impact on deserialization?

lucteo:
> Does this change have an impact on deserialization?

As far as I can't tell (not very sure though, as this is my first exposure to Swift serialization), the answer is NO. We only add a new field when we encode things, but we don't read anything new.

All the tests seem to pass ok

kyouko-taiga:
There are no tests for the deserialization. It would be great if you could conform.

To be honest I am not sure about this change. I don't think the current serialization is useful for humans. I opened an issue describing what I think should be done for that purpose: (see #70)

kyouko-taiga:
FYI. The plan was to use the serialization to store compiled modules.

PR #80, opened 2022-11-10
Node bundle proof of concept

dabrahams:
To get a sense of what's possible…

### DIFF
--- a/Sources/Compiler/TypedProgram.swift
+++ b/Sources/Compiler/TypedProgram.swift
@@ -36,5 +36,87 @@ public struct TypedProgram: Program {
     self.exprTypes = exprTypes
     self.referredDecls = referredDecls
   }
+}
 
+public typealias ValNode = Node
+extension AST {
+  public typealias Node = ValNode
+}
+
+extension TypedProgram {
+  @dynamicMemberLookup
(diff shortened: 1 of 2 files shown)


Sources/Compiler/TypedProgram.swift
```diff
@@ -36,5 +36,148 @@ public struct TypedProgram: Program {
...
+  typealias AnyScope = SomeNode<AnyScopeID>
+  typealias AnyDecl = SomeNode<AnyDeclID>
+
+  typealias Node<T: AST.Node> = SomeNode<NodeID<T>>
+}
+
+extension TypedProgram.SomeNode where ID: ConcreteNodeID {
+
+  /// The corresponding AST node.
+  private var astPart: ID.Subject {
```
kyouko-taiga, line 68: I first read this as "asPart". Maybe that projection could be named just `ast` or `syntax`?

kyouko-taiga (commented):
Looks very promising. I think the same approach could be done for all "phases" of compilation represented by a `***Program`.

Sources/Compiler/TypedProgram.swift
dabrahams: +1

PR #81, opened 2022-11-10
Refactor fun decl

kyouko-taiga:
This PR splits the `FunDecl` into three types:
- `FunctionDecl`, which represents global, local, and member function declarations
- `MethodDecl`, which represents method bundle declarations
- `InitializerDecl`, which represents initializer declarations

Along with this change, the PR refactors a significant part of the parser so that AST invariants can be better enforced. Specifically, it uses continuation passing style to gather all information necessary to build a node before inserting it in the AST. As a result, declaration nodes no longer need to be modified after having being constructed, eliminating the need for most of the `incorporate` methods.

Finally, the PR refactors makes the API for generating diagnostics more uniform.


dabrahams:
If you want effective review, it helps to keep each PR to one primary code transformation.  In the narrative I see three different major transformations.  Please consider doing each one as a separate step.

Sources/CLI/main.swift
```diff
@@ -79,9 +79,11 @@ struct CLI: ParsableCommand {
     transform: URL.init(fileURLWithPath:))
   var inputs: [URL]
 
+  private var noteLabel   : String { "note: ".styled([.bold, .cyan]) }
+
   private var warningLabel: String { "warning: ".styled([.bold, .yellow]) }
 
-  private var errorLabel: String { "error: ".styled([.bold, .red]) }
+  private var errorLabel  : String { "error: ".styled([.bold, .red]) }
```
dabrahams, line 86:
```suggestion
  private var errorLabel : String { "error: ".styled([.bold, .red]) }
```

kyouko-taiga:
> please consider doing each one as a separate step

Point taken. Normally I would try to keep PR focused. The problem here was that refactoring `FunDecl` implied making heavy changes to the compiler.

Sources/Compiler/AST/AST.swift
```diff
@@ -67,32 +73,45 @@ public struct AST: Codable {
...
+    withFiles(in: ValModule.core!, { (sourceURL) in
       if sourceURL.pathExtension != "val" { return true }
 
-      let sourceFile = try SourceFile(contentsOf: sourceURL)
-      let (decls, diagnostics) = Parser.parse(sourceFile, into: corelib!, in: &self)
-      if (decls == nil) || !diagnostics.isEmpty {
-        throw CompilerError(description: "parser failed", diagnostics: diagnostics)
+      // Parse the file.
+      do {
+        let sourceFile = try SourceFile(contentsOf: sourceURL)
```
dabrahams, line 91: There's a suspicious amount of error handling mechanics here. Maybe we should talk this over

Sources/Compiler/AST/AnyNode.swift
```diff
@@ -28,10 +28,12 @@ struct AnyNode: Codable {
     case .builtinDecl               : node = try container.decode(BuiltinDecl.self)
     case .conformanceDecl           : node = try container.decode(ConformanceDecl.self)
     case .extensionDecl             : node = try container.decode(ExtensionDecl.self)
-    case .funDecl                   : node = try container.decode(FunDecl.self)
```
dabrahams, line 31: Just a general concern that there's an awful lot of switches mapping between enum values and types.  Can we just use type values in place of the enum values?  Failing that, can you do all of this with a couple of dictionaries?

Sources/Compiler/AST/Decl/AssociatedValueDecl.swift
```diff
@@ -3,6 +3,9 @@ public struct AssociatedValueDecl: SingleEntityDecl {
 
   public static let kind = NodeKind.associatedValueDecl
 
+  /// The source range of the declaration's introducer, if any.
+  public let introducerRange: SourceRange?
```
dabrahams, line 7:
```suggestion
  public let introducer: SourceRange?
```

Sources/Compiler/AST/Decl/BindingDecl.swift
```diff
@@ -4,53 +4,39 @@ public struct BindingDecl: Decl {
...
   public let pattern: NodeID<BindingPattern>
 
   /// The initializer of the declaration, if any.
   public let initializer: AnyExprID?
 
-  /// Creates an undecorated instance with the given `attributes`, `pattern`, and `initializer`.
-  ///
-  /// Decorations may be added later via a call to `decorate`.
+  /// Creates an instance with the given properties.
   public init(
```
dabrahams, line 22: An aside: wondering if too much stuff is public.

Sources/Compiler/AST/Decl/FunctionDecl.swift
```diff
@@ -91,31 +71,32 @@ public struct FunDecl: GenericDecl, GenericScope {
     }
   }
 
+  /// Creates an instance with the given properties.
   public init(
-    introducer: SourceRepresentable<Introducer>,
+    introducerRange: SourceRange?,
```
dabrahams, line 76:
```suggestion
    introducer: SourceRange?,
```

Sources/Compiler/AST/Decl/FunctionDecl.swift
```diff
@@ -176,4 +134,19 @@ public struct FunDecl: GenericDecl, GenericScope {
     self.implicitParameterDecls = implicitParameterDecls
   }
 
+  public func isWellFormed(in ast: AST) -> SuccessOrDiagnostics {
```
dabrahams, line 137:
```suggestion
  public func validateForm(in ast: AST) -> SuccessOrDiagnostics {
```

Sources/Compiler/AST/Decl/InitializerDecl.swift
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /// The introducer of the declaration.
+  public let introducer: SourceRepresentable<Introducer>
+
+  /// The attributes of the declaration, if any.
+  public let attributes: [SourceRepresentable<Attribute>]
+
+  /// The access modifier of the declaration, if any.
+  public let accessModifier: SourceRepresentable<AccessModifier>?
```
dabrahams, line 24: Isn't there a default access modifier?  Can't we synthesize it with an empty source range and avoid the optional?

Sources/Compiler/AST/Decl/MethodDecl.swift
```diff
@@ -0,0 +1,74 @@
+/// A method declaration.
+public struct MethodDecl: GenericDecl, GenericScope {
+
+  public static let kind = NodeKind.methodDecl
+
+  /// The source range of the `fun` introducer, if any.
+  public let introducerRange: SourceRange?
+
+  /// The attributes of the declaration, if any.
```
dabrahams, line 9:
```suggestion
  /// The attributes of the declaration.
```

Sources/Compiler/AST/Decl/MethodDecl.swift
```diff
@@ -0,0 +1,74 @@
+/// A method declaration.
+public struct MethodDecl: GenericDecl, GenericScope {
+
+  public static let kind = NodeKind.methodDecl
+
+  /// The source range of the `fun` introducer, if any.
+  public let introducerRange: SourceRange?
```
dabrahams, line 7:
```suggestion
  /// The `fun` introducer, if any.
  public let introducer: SourceRange?
```

Sources/Compiler/AST/Decl/NamespaceDecl.swift
```diff
@@ -3,31 +3,38 @@ public struct NamespaceDecl: SingleEntityDecl, LexicalScope {
...
 
   public var name: String { identifier.value }
 
-  /// Incorporates `accessModifier` into `self`.
-  ///
-  /// - Precondition: `self.accessModifier == nil`
-  internal mutating func incorporate(_ accessModifier: SourceRepresentable<AccessModifier>) {
-    precondition(self.accessModifier == nil)
-    self.accessModifier = accessModifier
+  public func isWellFormed(in ast: AST) -> SuccessOrDiagnostics {
```
dabrahams, line 33:
```suggestion
  public func validateForm(in ast: AST) -> SuccessOrDiagnostics {
```

or even

```suggestion
  public func checkForm(in ast: AST) -> SuccessOrDiagnostics {
```

Sources/Compiler/AST/Decl/SubscriptImplDecl.swift
```diff
@@ -13,18 +14,18 @@ public struct SubscriptImplDecl: Decl, LexicalScope {
 
   }
 
-  /// The introducer of the method.
+  /// The introducer of the subscript.
   public let introducer: SourceRepresentable<ImplIntroducer>
 
   /// The declaration of the implicit receiver parameter, if any.
```
dabrahams, line 20: If it's implicit, how does it have a declaration?

Sources/Compiler/AST/NodeIDs/NodeKind.swift
```diff
@@ -44,36 +44,40 @@ public struct NodeKind: Hashable, Codable {
      5 << 16 | decl.rawValue | genericScope.rawValue)
   public static let extensionDecl = NodeKind(
      6 << 16 | decl.rawValue | genericScope.rawValue)
```
dabrahams, line 46: This kind of bit manipulation seems like it should be unnecessary and is probably at the very least premature optimization.

dabrahams (changes requested):
This PR is too big to review in a reasonable amount of time.  Please break it up into changes that make a single kind of code transformation and can be evaluated on their own.

Sources/CLI/main.swift
```diff
@@ -294,25 +302,28 @@ struct CLI: ParsableCommand {
...
-    }
-    write(": ")
-
+  func log(diagnostic: Diagnostic, asChild isChild: Bool = false) {
     // Log the location, if available.
     if let location = diagnostic.location {
       let path = location.source.url.relativePath
       let (line, column) = location.source.lineAndColumnIndices(at: location)
-      write("\(path):\(line):\(column): ")
+      write("\(path):\(line):\(column): ".styled([.bold]))
```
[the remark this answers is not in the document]
kyouko-taiga: I'm just replicating the aesthetics decisions of clang and swiftc.

Sources/Compiler/AST/AST+Invariants.swift
```diff
@@ -0,0 +1,209 @@
...
+  /// Otherwise, returns `.failure` with the diagnostics of the broken invariants.
+  func isValidGlobalScopeMember(
+    _ decl: AnyDeclID,
+    atTopLevel: Bool
+  ) -> SuccessOrDiagnostics {
+    var ds: [Diagnostic] = []
+
+    switch decl.kind {
+    case .associatedTypeDecl:
+      ds.append(.diagnose(unexpectedAssociatedTypeDecl: self[NodeID(rawValue: decl.rawValue)]))
```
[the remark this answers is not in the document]
kyouko-taiga: Maybe, but I couldn't find anything better.
- I tried using protocols but that is a poor solution because it forces you to have only one notional visit function per type.
- More promisingly, I tried using dispatch tables (i.e., dictionaries `[NodeKind: (AnyNodeID) -> Result]`) but it increased verbosity.

Sources/Compiler/AST/AST.swift
```diff
@@ -25,7 +25,11 @@ public struct AST: Codable {
   public var builtinDecl: NodeID<BuiltinDecl> { NodeID(rawValue: 0) }
 
   /// Inserts `n` into `self`.
-  public mutating func insert<T: Node>(_ n: T) -> NodeID<T> {
+  public mutating func insert<T: Node>(_ n: T) throws -> NodeID<T> {
```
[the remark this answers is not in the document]
kyouko-taiga: It is the single point where (syntactic) well-formedness has to be upheld, as the subtree rooted by `n` is immutable once it's been inserted. The convenience of automating the check in `insert(_:)` is extremely compelling.

Sources/Compiler/AST/AnyNode.swift
kyouko-taiga: I don't understand the remark nor the question.

Sources/Compiler/AST/Decl/AssociatedValueDecl.swift
kyouko-taiga: This change would break consistency with other nodes, where `introducer` is a `SourceRepresentable<SomeIntroducerEnum>`

Sources/Compiler/AST/Decl/BindingDecl.swift
kyouko-taiga: Might be too early to decide. I'm fairly certain AST nodes must be public,

Sources/Compiler/AST/Decl/FunctionDecl.swift
kyouko-taiga: Same answer as before.

Sources/Compiler/AST/Decl/InitializerDecl.swift
kyouko-taiga: Yes we could. That's a good idea.

Sources/Compiler/AST/Decl/AssociatedValueDecl.swift
dabrahams: 1. You could make it consistent by giving it an empty source range when it's implicit.
2. IMO it doesn't matter if it's consistent.  It's more important that the thing's name reflect its logical role.

Sources/Compiler/AST/Decl/SubscriptImplDecl.swift
kyouko-taiga: It gets synthesized.

Sources/Compiler/AST/NodeIDs/NodeKind.swift
kyouko-taiga: The code already exists, is not so hard to maintain, and I am not sure another form would make it obviously easier to read and/or maintain.

Sources/Compiler/AST/AST+Invariants.swift
```diff
@@ -0,0 +1,209 @@
+import Utils
```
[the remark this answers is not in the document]
kyouko-taiga: OK, will change after merge.

Sources/Compiler/AST/AnyNode.swift
dabrahams: Suppose `extension NodeID { typealias Kind = Node.Type }`?  Then couldn't you get rid of all these switch statements?

kyouko-taiga:
Sorry, I won't split the PR. I understand it is too big to get a proper review, but we need these changes to progress the AST refactoring process.

Many of your remarks relate to choices that have only been carried out. We can gradually improve on those once we have an AST that reflects the changes we made recently about the grammar (several changes on the syntax, plus the merge of TypeExpr and Expr).

I suggest you open issues so that we can keep track of the design choices you'd like to revisit.

Sources/Compiler/AST/AST.swift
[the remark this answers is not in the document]
dabrahams: May I suggest this, then?

```
try program.insert(wellFormed: n)
```

Sources/Compiler/AST/AST+Invariants.swift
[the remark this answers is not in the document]
dabrahams: I'm talking about lines almost exactly like the one I'm commenting on:

**ds.append(.diagnose(**unexpectedAssociatedTypeDecl: **self[NodeID(rawValue: decl.rawValue)]))**

The only really salient information in that line is the name of the error being diagnosed and it's buried in the middle.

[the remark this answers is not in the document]
kyouko-taiga: I tried to improve diagnostic reporting by transferring as much as the logic I could into the `diagnose` family of functions. I'm open to suggestions to make it even better.

`NodeID(rawValue otherID.rawValue)` is indeed boilerplate, but that's also the most efficient way to convert a node ID at call site. We could decide to use `NodeID(otherID)!` to make it shorter, at the cost of the forced unwrap.

dabrahams:
> Sorry, I won't split the PR. I understand it is too big to get a proper review, but we need these changes to progress on the AST refactoring process.

Sure, we need them (or some large subset of them).  But you can interactively rebase and organize your changes so even a large change makes a single kind of code transformation that can be quickly evaluated.  If you need some help doing that with Git, I'll be happy to show you how it's done.

> Many of your remarks relate to choices that have only been carried out. In a sense, you are reviewing the code I already wrote before that PR. 

Yes, those are just thoughts and meant to spark some discussion, not to meant to block acceptance of your change.

> We can gradually improve on those design choices once we have an AST that reflects the changes we made recently about the grammar (several changes on the syntax, plus the merge of TypeExpr and Expr).
> 
> I suggest you open issues so that we can keep track of the choices you'd like to revisit.

I'll try.

Sources/Compiler/AST/AST.swift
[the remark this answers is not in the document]
kyouko-taiga: Looks good. I'll make the change.

dabrahams:
I'll have to wait to see which remarks you decide to address before I can tell which things should become issues.

Sources/Compiler/AST/Decl/BindingDecl.swift
dabrahams: But do they need to be publicly initializable?

Sources/Compiler/AST/NodeIDs/NodeKind.swift
dabrahams: IMO `typealias NodeKind = Node.Type` would probably make all of this unnecessary.

Sources/Compiler/AST/Decl/BindingDecl.swift
kyouko-taiga: Yes, because I think we'll eventually expose Val compiler as a library that lets you programmatically create ASTs.

Sources/Compiler/AST/NodeIDs/NodeKind.swift
kyouko-taiga: Unfortunately no.

I originally designed it that way. Then I ended up writing a lot of code that was testing for nodes that belong to the same syntactic category (e.g., all lexical scopes, all type declarations, etc.). So I defined an additional bit to implement the `<=` operator.

Sources/Compiler/AST/Decl/AssociatedValueDecl.swift
kyouko-taiga: That's a good topic for an issue.

Sources/Compiler/AST/Decl/FunctionDecl.swift
kyouko-taiga: Will change.

Sources/Compiler/AST/AnyNode.swift
kyouko-taiga: Hmm, you're probably right. Good topic for an issue.

Sources/Compiler/AST/Decl/SubscriptImplDecl.swift
dabrahams:
```suggestion
  /// The synthesized declaration of the implicit receiver parameter, if any.
```

kyouko-taiga:
I made some of the suggested changes. I think other suggestions could be discussed as issues (I already opened a couple).

Sources/Compiler/AST/NodeIDs/NodeKind.swift
dabrahams: What `<=` should mean on a node kind is non-obvious to me.  What's wrong with `kind is Scope.Type` or `kind as? Scope.Type` for testing categories?

Sources/Compiler/AST/AST+Invariants.swift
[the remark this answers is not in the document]
dabrahams: I think you're missing my point.  There's gotta be a way to refactor so you're just writing something like

```
unexpectedAssociatedTypeDecl(decl)
```

instead of all that repetitive stuff I've bolded.

Sources/Compiler/AST/NodeIDs/NodeKind.swift
kyouko-taiga: The kind of `ProductTypeDecl` is not `LexicalScope`, it is `ProductTypeDecl`. So you cannot test whether `someTypeErasedID` (which might be a product type) is a lexical scope using that approach.

If you use the meta type of the referred node as its kind, then AFAICT there's no obvious way to type `AnyNodeID.kind`.

kyouko-taiga: Operator `<=` is implementing a sort of subtyping check.

```swift
if decl.kind <= .lexicalScope {
  // decl is an ID of any type that is a lexical scope.
}
```

Sources/Compiler/AST/AST+Invariants.swift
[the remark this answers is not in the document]
kyouko-taiga: I think you already suggested that approach, and I'm not sure I like it.

`diagnose(unexpectedAssociatedTypeDecl:)` might be called in different types, not only `AST`. So to make your call work we'll have to define `unexpectedAssociatedTypeDecl(_:)` in `AST` and any other type that may produce this diagnostic.

That, I believe, is also boilerplate that will have to be maintained every time we want to change the API of the diagnostic generator.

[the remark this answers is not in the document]
dabrahams: Still missing my point.  I'm not attached to that call form.  I mean you should increase the salient information density.

Sources/Compiler/AST/NodeIDs/NodeKind.swift
dabrahams: I don't follow.  `ProductTypeDecl` does conform to `LexicalScope`.  If you have a `someTypeErasedID` storing `ProductTypeDecl.self` as its `.kind` of static type `Node.Type`, you can write `someTypeErasedID.kind is LexicalScope.Type` and get your answer.  What am I missing?

kyouko-taiga: OK, didn't know you could use the `is` operator that way. That's a great idea then. Can you make the change?

Sources/Compiler/AST/AST+Invariants.swift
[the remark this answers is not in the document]
kyouko-taiga: If you show me how, I am interested. Otherwise, I think the current design is the best I could do so far.

PR #82, opened 2022-11-10
Start emitting function bodies

lucteo:
- decouple function body from CXXFunDecl
- CXXFunDecl now can only write its signature (it can't write its definition)
- CXXModule keeps track or the bodies of each associated function
- add new entity CXXRepresentable protocol to represent any type of CXX code
- add CXXComment for introducing comments into the generated C++ code


lucteo:
For factorial example, generates:
```
int factorial(int n) {
// expr
}
int main() {
// block
}
```

kyouko-taiga (approved):
LGTM

PR #83, opened 2022-11-11
Immutable AST continued

kyouko-taiga:


dabrahams (commented):
This PR has a lot of changes that seem to be unrelated to making the AST immutable, which makes review difficult.

dabrahams:
I think part of the problem is that the base branch of the PR ought to have been refactor-FunDecl rather than main.

Sources/Compiler/AST/AST+Invariants.swift
```diff
@@ -0,0 +1,209 @@
...
+
+extension AST {
+
+  /// Returns `.success` if `decl` is a well-formed top-level or namespace member declaration.
+  /// Otherwise, returns `.failure` with the diagnostics of the broken invariants.
+  func isValidGlobalScopeMember(
+    _ decl: AnyDeclID,
+    atTopLevel: Bool
+  ) -> SuccessOrDiagnostics {
+    var ds: [Diagnostic] = []
```
dabrahams, line 11: The name `ds` is a weird grammatical mix. I read it as “Ds,” i.e. a collection of D objects, whatever D is supposed to mean.  What is it about being a collection that warrants giving it special status and injecting type information into its name?

```suggestion
    var report: [Diagnostic] = []
```

or even 

```suggestion
    var log: [Diagnostic] = []
```

kyouko-taiga: OK, reasonable request. I'll create an issue.

kyouko-taiga:
Ah, my mistake. It was supposed to be.

Sources/Compiler/AST/Expr/SequenceExpr.swift
```diff
@@ -1,79 +1,50 @@
...
-      switch self {
-      case .node(let lhsOperator, let lhsLeft, var lhsRight):
-        if let l = lhsOperator.precedence {
-          if let r = operator_.precedence {
-            // Both operators are in groups.
-            if (l < r) || (l == r && l.associativity == .left) {
-              self = .node(operator: operator_, left: self, right: .leaf(rhs))
-              return
-            }
+  /// Creates an instance with the given properties.
```
dabrahams, line 28: See, you haven't told me the significance of `tail`, so I can't know what it means to create an instance with the given `tail` property.

Sources/Compiler/AST/Expr/SequenceExpr.swift
```diff
@@ -1,79 +1,50 @@
...
-          }
-        } else if operator_.precedence != nil {
-          // Only right operator is in a group. Assume higher precedence.
-          lhsRight.append(operator: operator_, rhs: rhs)
-          self = .node(operator: lhsOperator, left: lhsLeft, right: lhsRight)
-        } else {
-          // Neither operator is in a group. Assume left associativity.
-          self = .node(operator: operator_, left: self, right: .leaf(rhs))
-        }
+  public func isWellFormed(in ast: AST) -> SuccessOrDiagnostics {
```
dabrahams, line 34:
```suggestion
  public func wellFormedness(in ast: AST) -> SuccessOrDiagnostics {
```
Anything but an `is` prefix for a non-`Bool` returning function.

Sources/Compiler/AST/Expr/SequenceExpr.swift
```diff
@@ -1,79 +1,50 @@
...
+    var ds: [Diagnostic] = []
```
dabrahams, line 35: We have a better name for `ds` now, right?

Sources/Compiler/AST/Name.swift
```diff
@@ -59,6 +59,11 @@ public struct Name: Hashable, Codable {
     self.notation = notation
     self.introducer = introducer
   }
+
+  /// Returns a textual descriptions of `labels`.
```
dabrahams, line 63:
```suggestion
  /// Returns a textual description of `labels`.
```

Sources/Compiler/AST/Name.swift
```diff
@@ -59,6 +59,11 @@ public struct Name: Hashable, Codable {
...
+  public static func describe(labels: [String?]) -> String {
+    labels.reduce(into: "", { (s, l) in s += (l ?? "_") + ":" })
```
dabrahams, line 65: When building strings, you will get better performance and your code will come out a lot clearer if you lazily map things into strings and then `.join` them, than if you try to do reduction with append.

Even when the strings to be joined are dynamic, it will be much much clearer to map them and then `.join`, and we'll probably never notice a difference in performance.

You can also reduce noise in the code that does `message.write("...\n")` over and over by `message.append("...")` and then using `.join(separator: "\n")`

Sources/Compiler/AST/PassingConvention.swift
```diff
@@ -25,4 +26,12 @@ public enum PassingConvention: Codable {
     }
   }
 
+  /// Creates a convnetion matching the given receiver effect.
```
dabrahams, line 29:
```suggestion
  /// Creates a convention matching the given receiver effect.
```

Sources/Compiler/AST/PassingConvention.swift
```diff
@@ -16,6 +16,7 @@ public enum PassingConvention: Codable {
   /// Yielded.
   case yielded
 
+  /// Creates a convention matching the given remote type capability.
```
dabrahams, line 19: What does it mean to create a convention matching a remote type capability?   A doc comment should give me enough information to write a test.

Also, please don't use "the given _\<type rewritten in english\>_."  It's a silly convention used by DevPubs that takes up valuable words that could be used to give real salient information.

```suggestion
  /// Creates an instance matching `capability`.
```

But then use the space you've won to write something meaningful :-)

Sources/Compiler/AST/Name.swift
```diff
@@ -59,6 +59,11 @@ public struct Name: Hashable, Codable {
     self.notation = notation
     self.introducer = introducer
   }
+
+  /// Returns a textual descriptions of `labels`.
+  public static func describe(labels: [String?]) -> String {
```
dabrahams, line 64: You sure this should be public?

Sources/Compiler/IR/Emitter.swift
```diff
@@ -68,18 +68,29 @@ public struct Emitter {
     // Configure the locals.
     var locals = DeclProperty<Operand>()
 
-    for (i, capture) in program.ast[declID].implicitParameterDecls.enumerated() {
-      locals[capture.decl] = .parameter(block: entryID, index: i)
+    let explicitCaptures = program.ast[declID].explicitCaptures
```
dabrahams, line 71: It's a bit hard to tell what's going on here; I get the feeling shorter local variable names might make it easier, and I also get the sense the function is getting a bit long…which might be helped by shorter names.  Not sure, so do with this remark what you will.

Sources/Compiler/IR/Emitter.swift
```diff
@@ -666,8 +665,192 @@ public struct Emitter {
...
-    return emitR(expr: root, into: &module)
+    emit(.sink, foldedSequence: program.foldedSequenceExprs[expr]!, into: &module)
+  }
+
+  private mutating func emit(
+    _ convention: PassingConvention,
+    foldedSequence expr: FoldedSequenceExpr,
+    into module: inout Module
+  ) -> Operand {
+    switch expr {
```
dabrahams, line 676: I'm not really able to evaluate most of this code.  It looks plausible, but I don't know the codebase well enough to really tell for sure.  If it's well-covered by tests in this PR, then I think we can be pretty confident.  If not, please write some more tests :-)

dabrahams (changes requested):
Most of the comments are in the parts before IR.  Reviewing got tiring at some point and I was unable to usefully continue.  I hope what I've written is of some use.

Sources/Compiler/AST/Expr/FoldedSequenceExpr.swift
```diff
@@ -0,0 +1,47 @@
...
+
+  /// An operator with its precedence.
+  public typealias Callee = (expr: NodeID<NameExpr>, precedence: PrecedenceGroup?)
+
+  case node(callee: Callee, left: FoldedSequenceExpr, right: FoldedSequenceExpr)
+
+  case leaf(AnyExprID)
+
+  /// Appends the RHS of a binary operation to `self`.
+  mutating func append(callee: Callee, right: AnyExprID) {
```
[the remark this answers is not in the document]
kyouko-taiga: I think we need to discuss this. I didn't understand anything.

Sources/Compiler/IR/Emitter.swift
kyouko-taiga: This part needs work. But I think we should rethink how captures work in general. That would have the potential to simplify IR lowering.

Sources/Compiler/AST/Name.swift
kyouko-taiga: I really don't see the harm in exposing a helper function that implements boilerplate. But we can make it internal if you think it's necessary.

IMO it doesn't hurt local reasoning. Nothing is mutating and from the docs it is clear that the function only serves getting human-readably descriptions.

Sources/Compiler/AST/Expr/FoldedSequenceExpr.swift
```diff
@@ -0,0 +1,47 @@
+/// A folded sequence expression.
+public indirect enum FoldedSequenceExpr {
+
+  /// An operator with its precedence.
+  public typealias Callee = (expr: NodeID<NameExpr>, precedence: PrecedenceGroup?)
```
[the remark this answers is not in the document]
kyouko-taiga: The operator is the callee of a binary operation. But your suggestion makes sense.

Sources/Compiler/AST/Expr/SequenceExpr.swift
```diff
@@ -1,79 +1,50 @@
...
-  public typealias UnfoldedTail = [TailElement]
+  public static let kind = NodeKind.sequenceExpr
 
   /// The operator and right operand in an unfolded sequence of infix expressions.
   public struct TailElement: Codable {
 
-    /// The name of the operator.
-    public var operatorName: SourceRepresentable<Identifier>
+    /// The operator.
+    public var `operator`: NodeID<NameExpr>
```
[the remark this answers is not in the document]
kyouko-taiga: `operatorName` repeats the information that's already in the type `NameExpr`. I'm trying to apply your advice here.

Sources/Compiler/AST/Expr/SequenceExpr.swift
```diff
@@ -1,79 +1,50 @@
...
       self.operand = operand
     }
 
   }
 
-  /// A tree encoding sub-expression ordering of an unfolded `SequenceExpr`.
-  internal indirect enum Tree {
-
-    typealias Operator = (name: SourceRepresentable<Identifier>, precedence: PrecedenceGroup?)
+  /// The first operand of the expression.
```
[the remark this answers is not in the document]
kyouko-taiga: Yep, will add to the docs.

Sources/Compiler/AST/Name.swift
kyouko-taiga: Not entirely convinced `labels.map({ "\($0 ?? "_"):" }).joined()` is easier to read, but OK.

kyouko-taiga: Because I was intrigued I ran a micro benchmark. FWIW, `reduce` is 25% faster than `map`.

Sources/Compiler/AST/PassingConvention.swift
kyouko-taiga: I think I am getting unconvinced that all functions must be 100% understandable to the point of someone being able to write unit tests by their single-line documentation. That is a fine principle, really, but here I don't think any comment can be clearer than the 6-lines body.

I'll be honest. Have to scratch my head trying to come up with a good comment for this kind of function feels like a disincentive to refactor the boilerplate that it removed elsewhere.

Sources/Compiler/AST/Name.swift
dabrahams: The less you expose publicly, the clearer the abstraction you intend clients to see becomes for the reader.

Sources/Compiler/AST/Name.swift
kyouko-taiga: lazy map is much, much worse, about 10x.

Sources/Compiler/AST/Name.swift
kyouko-taiga: I don't disagree with this principle. I'm saying that I genuinely do not believe making this function internal makes the abstraction any clearer.

Sources/Compiler/AST/Expr/SequenceExpr.swift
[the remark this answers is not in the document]
dabrahams: I appreciate that, thank you.  It doesn't work 100% of the time, so one needs to be judicious.  In this case, I'm just asking you to think it through one more time.   Also, if you take `Name` out of the name, maybe ought to leave it in the comment, _if_ there's any useful distinction at all between an operator and and its name at this level.

Sources/Compiler/AST/Name.swift
dabrahams: :shrug: It's—at _least_—the difference between a smaller API and a larger one.  The smaller one is always easier to understand.

Sources/Compiler/AST/Expr/FoldedSequenceExpr.swift
```diff
@@ -0,0 +1,66 @@
...
+/// represented by `self`.
+public indirect enum FoldedSequenceExpr {
+
+  /// The expression of an operator in the AST together with its precedence.
+  public typealias Operator = (expr: NodeID<NameExpr>, precedence: PrecedenceGroup?)
+
+  /// A parent node representing the application of `operator` on `left` and `right`.
+  case parent(operator: Operator, left: FoldedSequenceExpr, right: FoldedSequenceExpr)
+
+  /// A leaf node representing some expression.
```
dabrahams, line 15:
```suggestion
  /// An expression that is not, at the top level, an operator application.
```

Or something.  What you wrote is obvious from the declaration.  Use the opportunity provided by the comment to say something more.

Sources/Compiler/AST/Expr/SequenceExpr.swift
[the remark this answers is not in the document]
kyouko-taiga: Can you have a look at the updated file and tell me if you still feel uncomfortable with the names of those properties? IMO `operator` is the name that best reflects the role of this property in the sub-tree represented by `TailElement`.

Sources/Compiler/AST/Expr/FoldedSequenceExpr.swift
kyouko-taiga: Does the precision adds anything useful to the reader?

What you wrote is wrong because `AnyExprID` can represent, well any expression. That includes another `SequenceExpr`, which doesn't violate well-formedness if it was obtained from`a * (b + c)`. Further `!a` is also an operator application that certainly can be a leaf node.

I feel like adding qualification to exclude this cases will just make the comment heavier.

Sources/Compiler/AST/Expr/FoldedSequenceExpr.swift
```diff
@@ -0,0 +1,66 @@
...
+/// once the precedence groups of its operators have been determined. A tree is created by calling
+/// `append(operator:right:)` to append an operator and its right operand to the sub-sequence
+/// represented by `self`.
+public indirect enum FoldedSequenceExpr {
+
+  /// The expression of an operator in the AST together with its precedence.
+  public typealias Operator = (expr: NodeID<NameExpr>, precedence: PrecedenceGroup?)
+
+  /// A parent node representing the application of `operator` on `left` and `right`.
+  case parent(operator: Operator, left: FoldedSequenceExpr, right: FoldedSequenceExpr)
```
dabrahams, line 13:
```suggestion
  /// The application of an operator to its `left` and `right` operands.
  case operator(Operator, left: FoldedSequenceExpr, right: FoldedSequenceExpr)
```

perhaps?

Sources/Compiler/AST/Name.swift
kyouko-taiga: OK. Can you suggest the change?

Sources/Compiler/AST/PassingConvention.swift
dabrahams: Don't view scratching your head to come up with a good comment as a burden; it's an opportunity. It is part of the process of understanding your own code and the problem you're solving, and eventually arriving at the right API.  The comment is for the consumer of the function who doesn't see the implementation, not the person who wrote it.  I don't know what a “remote type capability” is.  Should there even be a type for it distinct from `PassingConvention`?

Sources/Compiler/AST/Name.swift
dabrahams, line 64:
```suggestion
  private static func describe(labels: [String?]) -> String {
```

Sources/Compiler/AST/Name.swift
dabrahams: Huh, we should do better in Val! Can you post a gist link to your microbenchmark?
My argument is about clarity, but if the performance difference is significant I withdraw the suggestion.

Sources/Compiler/AST/Name.swift
kyouko-taiga: Can't be `private`, must be internal. I'll make the change.

Sources/Compiler/AST/Expr/FoldedSequenceExpr.swift
dabrahams: > Does the precision adds anything useful to the reader?

Absolutely, yes.

> What you wrote is wrong

Which goes to show that I don't understand the abstraction being documented here.

Sources/Compiler/AST/Name.swift
kyouko-taiga: Here's the microbenchmark: https://gist.github.com/kyouko-taiga/91223b1c76ff4af8e2f19f4dce72e65f

Sources/Compiler/AST/PassingConvention.swift
kyouko-taiga: Perhaps, although I am not sure what should be done for `yielded`. The current code predates the latest iteration of remote types.

I'll keep the status quo until remote types are better specified.

kyouko-taiga:
I opened an issue for the suggestions that I thought were out of scope for the PR.

PR #91, opened 2022-11-16
Enable warnings-as-errors; fix two warnings

nickpdemarco:
You may find [this discussion](https://github.com/apple/swift/issues/48323) interesting - this warning is a bit controversial, but I think the case I fixed here is generally-agreed to be warning-worthy.


kyouko-taiga (approved):
LGTM

PR #92, opened 2022-11-16
Add a simple test case for the AST's conformance to 'Codable'

kyouko-taiga:
There is absolutely no test for the AST's conformance to `Codable`. This PR creates a simple one that just serializes and deserializes an AST.


dabrahams (approved):
Nice.  IMO it will be important to test `TypedProgram` et. al. in a similar way, and those may be interesting because the property maps contain non-whole-part node IDs

PR #94, opened 2022-11-16
NodeKind cleanup

dabrahams:


Package.resolved
```diff
@@ -23,8 +23,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-collections.git",
       "state" : {
-        "revision" : "f504716c27d2e5d4144fa4794b12129301d17729",
-        "version" : "1.0.3"
+        "revision" : "48254824bb4248676bf7ce56014ff57b142b77eb",
+        "version" : "1.0.2"
```
kyouko-taiga, line 27: It seems odd that we're going back here. Any particular reason?

Sources/Compiler/AST/DeclLocator.swift
```diff
@@ -51,7 +51,7 @@ public struct DeclLocator: Hashable {
           self = .lambda(decl)
         }
 
-      case .initializerDecl:
+        case InitializerDecl.self:
```
kyouko-taiga, line 54:
```suggestion
      case InitializerDecl.self:
```

Sources/Compiler/AST/DeclLocator.swift
```diff
@@ -64,7 +64,7 @@ public struct DeclLocator: Hashable {
 
         self = .function(name: "init", labels: labels, notation: nil)
 
-      case .methodDecl:
+        case MethodDecl.self:
```
kyouko-taiga, line 67:
```suggestion
      case MethodDecl.self:
```

Sources/Compiler/AST/DeclLocator.swift
```diff
@@ -78,11 +78,11 @@ public struct DeclLocator: Hashable {
         let name = program.ast[decl].identifier.value
         self = .function(name: name, labels: labels, notation: program.ast[decl].notation?.value)
 
-      case .methodImplDecl:
+        case MethodImplDecl.self:
```
kyouko-taiga, line 81:
```suggestion
      case MethodImplDecl.self:
```

Sources/Compiler/AST/DeclLocator.swift
```diff
@@ -78,11 +78,11 @@ public struct DeclLocator: Hashable {
...
         let decl = NodeID<MethodImplDecl>(rawValue: decl.rawValue)
         self = .methodImpl(program.ast[decl].introducer.value)
 
-      case .productTypeDecl:
+        case ProductTypeDecl.self:
```
kyouko-taiga, line 85:
```suggestion
      case ProductTypeDecl.self:
```

Sources/Compiler/IR/Emitter.swift
```diff
@@ -538,7 +538,7 @@ public struct Emitter {
             at: insertionPoint!)[0]
         }
 
-      case .member(let calleeDecl) where calleeDecl.kind == .functionDecl:
+        case .member(let calleeDecl) where calleeDecl.kind == FunctionDecl.self:
```
kyouko-taiga, line 541:
```suggestion
      case .member(let calleeDecl) where calleeDecl.kind == FunctionDecl.self:
```

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
```diff
@@ -359,12 +359,12 @@ struct ConstraintGenerator {
           fatalError("not implemented")
         }
 
-      case .traitDecl:
+        case TraitDecl.self:
```
kyouko-taiga, line 362:
```suggestion
      case TraitDecl.self:
```

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
```diff
@@ -359,12 +359,12 @@ struct ConstraintGenerator {
...
         let trait = TraitType(decl: NodeID(d)!, ast: checker.program.ast)
         diagnostics.append(.diagnose(cannotConstructTrait: trait, at: checker.program.ast.ranges[callee]))
         assignToError(id)
 
-      case .typeAliasDecl:
+        case TypeAliasDecl.self:
```
kyouko-taiga, line 367:
```suggestion
      case TypeAliasDecl.self:
```

Sources/Compiler/TypeChecking/TypeChecker.swift
```diff
@@ -1580,7 +1582,7 @@ public struct TypeChecker {
         return .tuple(TupleType(elements))
       }
 
-    case .wildcardPattern:
+      case WildcardPattern.self:
```
kyouko-taiga, line 1585:
```suggestion
    case WildcardPattern.self:
```

Tests/ValTests/ASTTests.swift
```diff
@@ -52,7 +52,7 @@ final class ASTTests: XCTestCase {
     let deserialized = try decoder.decode(AST.self, from: serialized)
 
     // Deserialized AST should containt a function `foo`.
-    XCTAssertEqual(deserialized[source].decls.first?.kind, .functionDecl)
+    XCTAssertEqual(deserialized[source].decls.first?.kind, .init(FunctionDecl.self))
```
kyouko-taiga, line 55: I find this use of `.init` to be not as clear as Using `NodeKind`. The reader has to determine the type of `XCTAssertEqual`'s first argument to understand what type is being built.

I don't think it's the same as omitting the name of an enum when referring to its case, as this practice is more idiomatic in Swift AFAIK.

Tests/ValTests/ParserTests.swift
```diff
@@ -492,15 +492,15 @@ final class ParserTests: XCTestCase {
     let signature = try XCTUnwrap(try apply(Parser.functionDeclSignature, on: input).element)
     XCTAssertEqual(signature.parameters.count, 1)
     XCTAssertNil(signature.receiverEffect)
-    XCTAssertEqual(signature.output?.kind, .nameTypeExpr)
+    XCTAssertEqual(signature.output?.kind, .init(NameTypeExpr.self))
```
kyouko-taiga, line 495: Same remark as https://github.com/val-lang/val/pull/94/files#r1025247457

Sources/Compiler/AST/NodeIDs/NodeKind.swift
```diff
@@ -1,315 +1,196 @@
...
-
-    default                         : return("Unknown")
-    }
-  }
+    MatchCase.self,
+    TopLevelDeclSet.self
+  ]
 
+  static let indices = Dictionary(
+    uniqueKeysWithValues: allValues.enumerated().lazy.map { (n, t) in (NodeKind(t), n) })
```
kyouko-taiga, line 195:
```suggestion
    uniqueKeysWithValues: allValues.enumerated().map { (n, t) in (NodeKind(t), n) })
```

`lazy` is not as efficient as you seem to think. In fact, I believe it makes things slower unless the collection is really large.

I profiled this line; it's faster without the laziness.

kyouko-taiga (commented):
Nice work.

It seems you broke indentation in a few places. I provided suggestions where I could spot them but the PR might deserve a second check.

I try to consistently add one empty line at the begin and another at end of a type declaration. e.g. That might be a silly styling rule, but I like it. If you don't mind, I'd appreciate if you could make sure it's upheld in your changes.

I profiled your use of `lazy` and I think it makes things worse in general. I would recommend we always default to eager evaluation it until we decide to dig into performance.

Sources/Compiler/AST/Node.swift
```diff
@@ -1,9 +1,8 @@
+import Utils
```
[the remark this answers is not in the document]
dabrahams: It's not anymore.  Good catch.

```suggestion
```

Tests/ValTests/ParserTests.swift
dabrahams:
```suggestion
    XCTAssertEqual(signature.output?.kind, NodeKind(NameTypeExpr.self))
```

Sources/Compiler/AST/NodeIDs/NodeKind.swift
dabrahams:
```suggestion
    uniqueKeysWithValues: allValues.enumerated().map { (n, t) in (NodeKind(t), n) })
```

dabrahams:
> It seems you broke indentation in a few places. I provided suggestions where I could spot them but the PR might deserve a second check.

I'll look again, thanks.  I was using an experimental emacs configuration that I've since dumped.
`DeclLocator.swift` has pre-existing indentation problems (starting at `case InitializerDecl.self:`), and it may not be the only one.  I'll submit a separate PR to fix that one.

Constructs like this clash with my auto-indentation and also with my (and Swift-Format's) central formatting rule: that a line containing an open delimiter not matched on the same line must end with that open delimiter (with an exception for closure signatures, which idiomatically extend beyond the closure's opening `{`).

```swift
        // here -----V
        result.append(character.utf16.reduce(into: "u", { (u, point) in
          u += String(point, radix: 16)
        }))
```

Any of these would work better:

```swift
        result.append(
          character.utf16.reduce(
              into: "u", { (u, point) in u += String(point, radix: 16) }
          ))
```

```swift
        result.append(
          character.utf16.reduce(
              into: "u", { (u, point) in u += String(point, radix: 16) }))
```

```swift
        result.append(
          character.utf16.reduce(into: "u", { (u, point) in u += String(point, radix: 16) }))
```

> I try to consistently add one empty line at the begin and another at end of a type declaration. e.g. That might be a silly styling rule, but I like it. If you don't mind, I'd appreciate if you could make sure it's upheld in your changes.

I did try to uphold it everywhere, and I just rechecked the PR; can't find a single instance where I failed except for 

```swift
public protocol TypeDecl: SingleEntityDecl {}
```

How many blank lines would you like in a case like that?  BTW, it doesn't seem silly, but without any rationale it does seem arbitrary and inconsistent to do it for types and not all other blocks.

kyouko-taiga:
> `DeclLocator.swift` has pre-existing indentation problems (starting at case `InitializerDecl.self:`), and it may not be the only one. I'll submit a separate PR to fix that one.

Great, thanks a lot!

> Constructs like this clash with my auto-indentation 

Okay, I'll use a different style then. This one looks perfectly fine:
```swift
result.append(
  character.utf16.reduce(
    into: "u", { (u, point) in u += String(point, radix: 16) }))
```

What would happen if the closure is not a one-liner? Would this style work with your auto-indent setup?

```swift
result.append(
  character.utf16.reduce(
    into: "u", { (u, point) in
      print(point)
      u += String(point, radix: 16)
    }))
```

> I did try to uphold it everywhere, and I just rechecked the PR; can't find a single instance where I failed except for

I spotted the extension of `NodeKind` in `NodeKind.swift` at L94.

> How many blank lines would you like in a case like that? BTW, it doesn't seem silly, but without any rationale it does seem arbitrary and inconsistent to do it for types and not all other blocks.

I would keep this type just like it is.

I add one blank line before the first declaration and one blank line before the last. The blank line after the last statement is mostly for symmetry. The reason is that I think the separation helps distinguishing a member from the head of the declaration in general. It also helps in my IDE when I hide the bodies. For example, taken from Durian:

```swift
public struct Choose<First: Combinator, Second: Combinator>: Combinator
where First.Context == Second.Context,
      First.Element == Second.Element
{

  public typealias Context = First.Context
  ...
```

But I'll gladly admit that my main reason is that I've grown used to this style. I would not strongly defend it if it had to be questioned.

Sources/Compiler/AST/Node.swift
```diff
@@ -14,4 +12,6 @@ extension Node {
 
   public func validateForm(in ast: AST) -> SuccessOrDiagnostics { .success }
 
+  /// A unique identifier denoting the type of this node.
+  static var kind: NodeKind { NodeKind(Self.self) }
```
kyouko-taiga, line 16:
```suggestion
  static var kind: NodeKind { NodeKind(Self.self) }
  
```

dabrahams:
> What would happen if the closure is not a one-liner? Would this style work with your auto-indent setup?

Yes, that would work.

kyouko-taiga (approved):
Magnifique!

PR #98, opened 2022-11-21
Tests improvements

lucteo:
**Why?**
Improve CXX transpiler tests, to ease the development of CXX transpiler

**What?**
- display the name of the test file when a test fails
- separate the `cpp` annotations into two annotations: one for source files, and one for headers
- add test for factorial transpilation -- this may be brittle at the beginning, but it will help in the development

**Testing**
- all unit tests must pass
- manual testing to ensure testing framework work as expected

### DIFF
--- a/Tests/ValTests/CXXTests.swift
+++ b/Tests/ValTests/CXXTests.swift
@@ -1,10 +1,14 @@
 import XCTest
 import Compiler
 
-func check(_ haystack: String, contains needle: String.SubSequence) {
+func check(_ haystack: String, contains needle: String.SubSequence, for testFile: String) {
   XCTAssert(
     haystack.contains(needle),
     """
+
+
+    Test file: \(testFile)
+    ===========\(String(repeating: "=", count: testFile.count))
 
     \(String(reflecting: needle))
     not found in source
@@ -71,9 +75,9 @@ final class CXXTests: XCTestCase {
           let specifications = annotation.argument?.split(separator: "\n", maxSplits: 1) ?? []
           assert((1...2).contains(specifications.count))
           if specifications.count > 1 {
-            check(cxxHeader, contains: specifications.first!)
+            check(cxxHeader, contains: specifications.first!, for: tc.name)
           }
-          check(cxxSource, contains: specifications.last!)
+          check(cxxSource, contains: specifications.last!, for: tc.name)
 
         default:
           XCTFail("\(tc.name): unexpected test command: '\(annotation.command)'")
--- a/Tests/ValTests/CXXTests.swift
+++ b/Tests/ValTests/CXXTests.swift
@@ -88,7 +88,7 @@ final class CXXTests: XCTestCase {
         case "cpp":
           check(cxxSource, contains: code, for: tc.name)
 
-        case "hpp":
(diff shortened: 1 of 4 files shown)


Tests/ValTests/CXXTests.swift
```diff
@@ -14,6 +18,15 @@ func check(_ haystack: String, contains needle: String.SubSequence) {
     """)
 }
 
+/// Remove trailing newlines from the given string subsequence
```
kyouko-taiga, line 21:
```suggestion
/// Removes trailing newlines from the given string subsequence.
func removingTrailingNewlines(_ s: Substring) -> Substring {
```

- I'm suggesting `removingTrailingNewlines` because We've been trying to avoid abbreviated names for functions.
- I'm suggesting `Substring` rather than `String.SubSequence` because I think it's more straightforward.

Note that I would typically implement this function as a method of `StringProtocol` so that it can apply to both strings and substrings without polluting the global namespace:
```
extension StringProtocol {

  func removingTrailingNewlines() -> Self.SubSequence {
    if let i = self.lastIndex(where: { !$0.isNewline }) {
      return self.prefix(through: i)
    } else {
      return self.prefix(upTo: startIndex)
    }
  }

}
```

Tests/ValTests/TestCases/CXX/EmptyMain.val
```diff
@@ -1,5 +1,7 @@
-/*! cpp
+/*! hpp
```
kyouko-taiga, line 1: Do we name our headers with a `.hpp` extension? IIRC I chose used `.h` because it was what's suggested in the core guidelines. In that case I would consider mirroring this choice here.

kyouko-taiga (changes requested):
Please chose a non-abbreviated name for `rtrimm` and consider declaring it in an extension of `Substring` or `StringProtocol`.

Tests/ValTests/TestCases/CXX/EmptyMain.val
lucteo: This is just a convention; it seems that people would have different expectations. I've seen .h, .hpp, .hxx, and probably others. Myself, I've used .h for a long time; recently I've switched to .hpp. The reason is to distinguish between headers that are C compatible, and headers that need a C++ compiler.

If I were to choose, I would use .hpp, but I won't object if you would want to use simple .h

kyouko-taiga: I will defer that choice to C++ experts, yourself included. Feel free to ignore my comment.

lucteo: @dabrahams : which one do you prefer?

kyouko-taiga: I prefer `.h`

lucteo: @kyouko-taiga : if core guideline says to use `.h`, I'll also change my preference to use `.h`. Will change this in the code to match the header extension.

kyouko-taiga (approved):
LGTM

PR #101, opened 2022-11-22
Accept expr wherever typeExpr appears in the grammar.

dabrahams:


Tests/ValTests/ParserTests.swift
```diff
@@ -1648,6 +1648,12 @@ final class ParserTests: XCTestCase {
     XCTAssertEqual(list.count, 2)
   }
 
+  func testTypeExprUnificationWithExpr() throws {
+    let input = SourceFile(contents: "<T, @value 40 + two()>")
```
kyouko-taiga, line 1652: Do tests succeed without the `@value`?

kyouko-taiga (changes requested):
Could you create a test that doesn't have `@value` in front of something that cannot parse as a type expression.

dabrahams:
Yes!

dabrahams:
I can, but it's not a very specific request. Is this test supposed to fail or succeed? What grammar symbol am I trying to parse.

kyouko-taiga:
The test is supposed to succeed. I believe I suggested:

```
func testStaticArgumentList() throws {
  let input = SourceFile(contents: "<T, 40 + two()>")
  let list = try XCTUnwrap(try apply(Parser.staticArgumentList, on: input).element)
  XCTAssertEqual(list.count, 2)
}
```

The goal is to determine whether the parser can accept an expression where it used to only parse type expressions without any "hint".

kyouko-taiga:
OK, good enough for now I guess.

kyouko-taiga (approved):

PR #106, opened 2022-11-24
Simplify types

kyouko-taiga:
This PR replaces the big enum that was used to represent types by an existential container.

**Alternative considered:**

I considered using `any TypeProtocol` instead of defining a custom `AnyType` container. The problem with that approach is that it prevents synthesizing conformance to `Equatable` and `Hashable`, thus introducing a lot of tedious and error-prone handwritten equality and hash methods.

I considered not making `TypeProtocol` refine `Hashable` and use `AnyHashable` or `Utils.HashableBox` to let Swift synthesize conformance to `Equatable` and `Hashable`. This approach has the advantage to leverage existing facilities, but having a dedicated container (i.e., `AnyType` rather than `AnyHashable`) allows us to define more convenient APIs.


Sources/Compiler/Types/TypeProtocol.swift
```diff
@@ -1,16 +1,69 @@
 import Utils
 
-/// A protocol describing the API of a type's associated value.
-public protocol TypeProtocol {
+/// A protocol describing the API of a Val type.
+public protocol TypeProtocol: Hashable {
```
dabrahams, line 4:
```suggestion
public protocol Type: Hashable {
```

Sources/Compiler/Parse/SourceLocation.swift
```diff
@@ -55,3 +55,18 @@ extension SourceLocation: Codable {
   }
 
 }
+
+extension SourceLocation: CustomReflectable {
```
dabrahams, line 59: Better saved for a separate PR.

Sources/Compiler/Parse/SourceLocation.swift
```diff
@@ -55,3 +55,18 @@ extension SourceLocation: Codable {
...
+
+  public var customMirror: Mirror {
```
dabrahams, line 61: I've [experimented](https://github.com/val-lang/Lotsawa/blob/8f4ab5b006e1d65cee5e0fe44511eee1ba19eb4f/Sources/Lotsawa/Chart.swift#L291-L320) with mirrors recently, and the debugger integration is so weak.  A Custom[Debug]StringRepresentable conformance is probably a better use of everyone's time.

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
```diff
@@ -341,7 +341,7 @@ struct ConstraintGenerator {
           // Propagate the type of the constructor down.
           let outputType = propagateDown(
             callee: callee,
-            calleeType: LambdaType(candidates[0].type)!,
```
dabrahams, line 344: FWIW, I think it would be easy to extend `TypeProtocol` so this terser syntax still works.

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
```diff
@@ -892,26 +892,26 @@ struct ConstraintGenerator {
...
       }
     } else {
-      inferredTypes[id] = inferredType
+      inferredTypes[id] = AnyType(inferredType)
     }
   }
 
-  private mutating func assignToError<T: ExprID>(_ id: T) {
-    inferredTypes[id] = .error(ErrorType())
+  private mutating func assignToError<ID: ExprID>(_ id: ID) {
```
dabrahams, line 913: You can come up with a better name than `id`/`ID` someday.

Sources/Compiler/TypeChecking/ConstraintSolver.swift
```diff
@@ -201,21 +201,21 @@ struct ConstraintSolver {
       if let lambda = LambdaType(letImplOf: l) {
         minterms.append(
           .init(
-            constraints: [EqualityConstraint(.lambda(lambda), r, because: constraint.cause)],
+            constraints: [EqualityConstraint(AnyType(lambda), r, because: constraint.cause)],
```
dabrahams, line 204:
```suggestion
            constraints: [EqualityConstraint(^lambda, r, cause: constraint.cause)],
```

Just a thought.

Sources/Compiler/TypeChecking/Constraints/ConstraintSet.swift
```diff
@@ -9,7 +9,7 @@ public struct ConstraintHashableWitness: HashableWitness {
     constraint.hash(into: &hasher)
   }
 
-  public static func isEqual(_ left: Element, to right: Element) -> Bool {
+  public static func isEqual(_ left: Constraint, to right: Constraint) -> Bool {
```
dabrahams, line 12: FWIW, `x` and `y` would be better than `left`/`right`, because it's a symmetric relation and handedness is irrelevant.

Sources/Compiler/TypeChecking/Solution.swift
```diff
@@ -59,16 +59,16 @@ struct Solution {
   var score: Score { Score(errorCount: diagnostics.count, penalties: penalties) }
 
   /// Reifies the given type, substituting each free variable by its corresponding binding.
```
dabrahams, line 61:
```suggestion
  /// Substitutes each free variable for its corresponding binding.
```

Telling me that `reify(x, ...)` “reifies x” is not valuable.
"By" is the wrong word, and "substitutes" is a bit murky in terms of which thing is new/old.  Can we phrase this as "Replaces _oldthing_ with _newthing_?"

Sources/Compiler/Types/AnyType.swift
```diff
@@ -0,0 +1,207 @@
+import Utils
+
+/// A box wrapping a type.
+fileprivate protocol TypeBox {
```
dabrahams, line 4: Intuitively, this amount of manual type erasure machinery seems needless. Maybe we should discuss?

Sources/Compiler/Types/AssociatedValueType.swift
```diff
@@ -1,33 +1,36 @@
...
 
-  public let flags: TypeFlags = .isCanonical
-
-  public init(decl: NodeID<AssociatedValueDecl>, domain: Type, ast: AST) {
-    switch domain {
-    case .associatedType, .conformanceLens, .genericTypeParam:
+  /// Creates an instance denoting the associated value declared by `decl` as a member of `domain`.
+  ///
+  /// - Requires: `domain` is an associated type, conformance lens, or generic type parameter.
+  public init(decl: NodeID<AssociatedValueDecl>, domain: AnyType, ast: AST) {
```
dabrahams, line 20: Decided to keep the type erasure after all?

Sources/Compiler/Types/BoundGenericType.swift
```diff
@@ -38,20 +42,39 @@ public struct BoundGenericType: TypeProtocol, Hashable {
     self.flags = flags
   }
 
+  public func transformParts(_ transformer: (AnyType) -> TypeTransformAction) -> Self {
```
dabrahams, line 45: Doc please.

Sources/Compiler/Types/BuiltinSymbols.swift
```diff
@@ -83,11 +83,11 @@ public enum BuiltinSymbols {
 
 extension LambdaType {
 
-  fileprivate init(from inputs: (PassingConvention, BuiltinType)..., to output: Type) {
+  fileprivate init(from inputs: (PassingConvention, BuiltinType)..., to output: AnyType) {
     self.init(
       inputs: inputs.map({ (convention, type) in
         CallableTypeParameter(
-          type: .parameter(ParameterType(convention: convention, bareType: .builtin(type))))
+          type: AnyType(ParameterType(convention: convention, bareType: .builtin(type))))
```
dabrahams, line 90:
```suggestion
          type: ^ParameterType(convention: convention, bareType: .builtin(type))
```

Sources/Compiler/Types/GenericTypeParamType.swift
```diff
@@ -27,6 +26,8 @@ public struct GenericTypeParamType: TypeProtocol, Hashable {
     }
   }
 
+  public var flags: TypeFlags { [.isCanonical, .hasGenericTypeParam] }
```
dabrahams, line 29: Doc please

Sources/Compiler/Types/GenericValueParamType.swift
```diff
@@ -26,6 +25,8 @@ public struct GenericValueParamType: TypeProtocol, Hashable {
     }
   }
 
+  public var flags: TypeFlags { [.isCanonical, .hasGenericValueParam] }
```
dabrahams, line 28: Doc please

Sources/Compiler/Types/SubscriptType.swift
```diff
@@ -38,26 +39,34 @@ public struct SubscriptType: TypeProtocol, Hashable {
...
   public var captures: [TupleType.Element] {
-    if case .tuple(let type) = environment {
-      return type.elements
-    } else {
-      return []
-    }
+    (environment.base as? TupleType)?.elements ?? []
+  }
+
+  public func transformParts(_ transformer: (AnyType) -> TypeTransformAction) -> Self {
```
dabrahams, line 45: Doc please.

Sources/Compiler/Types/TraitType.swift
```diff
@@ -1,21 +1,22 @@
...
 
-  public let flags: TypeFlags = .isCanonical
-
+  /// Creates an instance denoting the product type declared by `decl`.
   public init(decl: NodeID<TraitDecl>, ast: AST) {
     self.decl = decl
     self.name = Incidental(ast[decl].name)
   }
 
+  public var flags: TypeFlags { .isCanonical }
```
dabrahams, line 18: Doc please.

Sources/Compiler/Types/TypeProtocol.swift
```diff
@@ -1,16 +1,69 @@
...
 
 extension TypeProtocol {
 
   /// Returns whether the specified flags are raised on this type.
   public subscript(fs: TypeFlags) -> Bool { flags.contains(fs) }
 
+  /// Returns this type transformed with `transformer`.
+  ///
+  /// This method visits the structure of the type and calls `transformer` on each type composing
+  /// that structure. The result of the call substitutes the visited type. If `transformer` returns
```
dabrahams, line 22:
```suggestion
  /// that structure. The payload of the result is substituted for the visited type. If `transformer` returns
```

Except even that isn't quite accurate, because in the case of `stepInto` it's the recursively-transformed payload.  Happy to work with you on this.

Sources/Compiler/Types/TypeProtocol.swift
```diff
@@ -1,16 +1,69 @@
...
+
+  /// Applies `TypeProtocol.transform(_:)` on the types that are part of `self`.
+  public func transformParts(_ transformer: (AnyType) -> TypeTransformAction) -> Self {
+    // Default implementation is the identity.
+    self
+  }
+
+  /// The skolemized form of this type.
+  ///
+  /// A type is skolemized when all its generic parameters have been substituted for skolems.
```
dabrahams, line 42: Telling me that `x.skolemized`, where I know `x` is a type, is “the skolemized form of this type,” is a waste of a good summary.

```suggestion
  /// `self` with all generic parameters replaced by skolems.
```

Sources/Compiler/Types/TypeTransformAction.swift
```diff
@@ -0,0 +1,8 @@
+/// The result of a call to a closure passed to `TypeProtocol.transform(_:)`.
+public enum TypeTransformAction {
+
+  case stepInto(AnyType)
```
dabrahams, line 4: Docs for the cases please.

dabrahams (changes requested):
Mostly awesome.  A few docs are missing or need adjustment.  I have questions about the amount of code required for `AnyType`

Sources/Compiler/AST/DeclLocator.swift
```diff
@@ -38,8 +38,8 @@ public struct DeclLocator: Hashable {
         let decl = NodeID<FunctionDecl>(rawValue: decl.rawValue)
 
         let labels: [String]
-        switch program.declTypes[decl]! {
-        case .lambda(let type):
+        switch program.declTypes[decl]!.base {
```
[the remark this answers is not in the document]
kyouko-taiga: That makes a lot of sense

Sources/Compiler/Parse/SourceLocation.swift
kyouko-taiga: It gave me minor improvements in Xcode debuggers, but I can definitely implement `CustomDebugStringRepresentable` too. I'll make a separate PR

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: Hmm, I see what you mean. OK I'll experiment.

Sources/Compiler/TypeChecking/ConstraintSolver.swift
kyouko-taiga: The operator is implemented in #107, I'll open an issue to improve the labels of the constraints.

Sources/Compiler/TypeChecking/Solution.swift
kyouko-taiga: Yes, we can.

Sources/Compiler/Types/BoundGenericType.swift
kyouko-taiga: The docs comes from the protocol conformance.

Sources/Compiler/Types/AssociatedValueType.swift
kyouko-taiga: Thought it'd be best in a separate PR.

Sources/Compiler/Types/GenericTypeParamType.swift
kyouko-taiga: Docs come with protocol conformance.

Sources/Compiler/Types/AnyType.swift
kyouko-taiga: I took inspiration from `AnyHashable`. The `TypeBox` protocol is a trick that I use to avoid requiring extra methods on `TypeProtocol` forward the methods that use `Self` requirements, such as `==`

PR #107, opened 2022-11-24
Add a custom operator to wrap type instances into 'AnyType'

kyouko-taiga:


Sources/Compiler/IR/LoweredType.swift
```diff
@@ -12,7 +12,7 @@ public struct LoweredType: Hashable {
   /// - Requires: `astType` must be canonical.
   private init<T: TypeProtocol>(astType: T, isAddress: Bool) {
     precondition(astType[.isCanonical], "source type is not canonical")
-    self.astType = AnyType(astType)
+    self.astType = ^astType
```
dabrahams, line 15: Can we find a better name for this?  [Your comment](https://github.com/val-lang/val/pull/80#discussion_r1018645957) applies.

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
```diff
@@ -317,7 +317,7 @@ struct ConstraintGenerator {
           let ctor = (initializer.type.base as! LambdaType).ctor()!
 
           if labels.elementsEqual(ctor.labels) {
-            let (ty, cs) = checker.open(type: AnyType(ctor))
+            let (ty, cs) = checker.open(type: ^ctor)
```
dabrahams, line 320: I'm not sure how something named `ctor` is a type (assuming I'm reading the abbrev correctly).  FWIW

Sources/Compiler/Types/AnyType.swift
```diff
@@ -205,3 +203,8 @@ extension AnyType: CustomReflectable {
   public var customMirror: Mirror { Mirror(reflecting: base) }
 
 }
+
+/// Creates a type-erased container wrapping the given instance.
+public prefix func ^ <T: TypeProtocol>(_ base: T) -> AnyType {
```
dabrahams, line 208: I think this should be a static member in an extension of `TypeProtocol`, neh?

dabrahams (approved):
Fantastic improvement.  Please see the one question in AnyType.swift

Sources/Compiler/IR/LoweredType.swift
kyouko-taiga: Good observation.
What do you think about `highLevelType`?

I'm struggling a bit to describe the difference between lowered types and "high-level" types. The former are carry additional information, because convey the "l-valueness" of a an expression.

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: Would it help if it was named `ctorType`?

Sources/Compiler/Types/AnyType.swift
kyouko-taiga: For some reason that I don't fully understand, that causes calls to `^` to be ill-typed when the operand is `any TypeProtocol`:
> Type 'any TypeProtocol' cannot conform to 'TypeProtocol'

I guess the error makes sense. A protocol cannot conform to itself, but I don't get why it works if the operator is declared as a free function.

I can go ahead and make the change, though. There's only one case where it would break the code, and I can use the long form to fix it.

kyouko-taiga: To be precise, making the operator a static member of `TypeProtocol` breaks this line: https://github.com/val-lang/val/blob/d8b842bbac87efd41e69616ab25a82fa4a819a6a/Sources/Compiler/TypeChecking/ConstraintGenerator.swift#L461

Sources/Compiler/IR/LoweredType.swift
dabrahams: Let's start here: which one is formally going to be called "type" in the spec?

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
dabrahams: I don't know.  What is a "ctorType?"

Sources/Compiler/Types/AnyType.swift
dabrahams: > I don't get why it works if the operator is declared as a free function.

Oh… g*d.  It's technically a bug, but I'm sure nobody will fix it because lots of people want Swift to automatically open existentials.  We should think carefully about whether we want to exploit it.

dabrahams: Err, nevermind, it's not a bug: https://github.com/apple/swift-evolution/blob/main/proposals/0352-implicit-open-existentials.md

dabrahams: Well, at least it appears they implemented:
```
func h<T>(a: T) -> some Collection<T> {
  [a]
}
```
🍾 !!

I think they still have no way to express `some Collection where .Element: Equatable`.

PR #108, opened 2022-11-24
Step 2 of @value elimination

dabrahams:
@kyouko-taiga Please take a look at the failure; it's not something I can just fix in the parser.

### DIFF
--- a/Sources/Compiler/Parse/Parser.swift
+++ b/Sources/Compiler/Parse/Parser.swift
@@ -2843,12 +2843,8 @@ public enum Parser {
       })
   )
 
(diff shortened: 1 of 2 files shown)


kyouko-taiga:
Problem is caused by the code trying to create a bound generic type with a value expression. I don't know how to deal with this case yet, so I guess we're blocked until I can figure out.

kyouko-taiga:
I think this set of changes has been made obsolete by #121. Is that correct?

kyouko-taiga (commented):
I think this set of changes has been made obsolete by #121. Is that correct?

PR #110, opened 2022-11-24
Step 4 of @value elimination

dabrahams:


dabrahams:
@kyouko-taiga PTAL at the failure; we now have no hard distinction between generic value parameters and generic type parameters: _id-1_`:` _id-2_ could be either one.  Can you suggest an approach?

kyouko-taiga (commented):
I think this set of changes has been made obsolete by #121. Is that correct?

dabrahams:
#121 has so much in it that it’s hard for me to tell for sure, but it seems likely.

PR #115, opened 2022-11-25
Refactor type expr

kyouko-taiga:
This PR associates meta types to type declarations rather than the type they are declaring. In other words, it makes it so that the expression `A` has type `MetaType<A>` in the following program:

```val
type A {}
public fun main() {
  let _ = A
}
```


dabrahams:
You might want to rebase this to get my commit out of the history.

kyouko-taiga:
Seems like all I managed to do with this rebase is to create a zillion conflicts :(

PR #117, opened 2022-11-26
Encapsulate node origin

kyouko-taiga:
Encapsulate the origin (formerly "range") of an AST node in its representation and rename properties denoting source origins (from `range` to `origin`).


dabrahams (approved):
It's better to separate simple renames from all other moves in a PR.  If the PR consists only of renames I will usually approve without looking at it and not feel like I missed anything.  In this case, I don't want to dig for the one or two places of interest, so I'm having that feeling of probably missing something.

I can show you how to do this kind of commit shuffling with Magit and interactive rebase.

PR #118, opened 2022-11-26
Start borrowed local binding

lucteo:
*Why*
Make progress in implementing the factorial example.

*What*
- start transpilng local var bindings
- add void-cast operation to represent operations for which return value is ignored

*Testing*
Unit tests pass


kyouko-taiga (approved):
Great work!

PR #122, opened 2022-11-27
Add an downcasting initializer to 'TypeProtocol'

kyouko-taiga:


dabrahams:
Looks like this was a big improvement!

kyouko-taiga:
It was, but it also introduced a very insidious bug by changing the meaning of `MetatypeType.init(_:)`. Took me a good hour to find it.

Should be fine for new code though.

dabrahams:
The way you avoid that problem is to first make a commit changing the signature of `MetatypeType.init()`, get that to compile, then add your new initializer.

kyouko-taiga:
That's what I did, and then I got bitten by a merge 😅

PR #130, opened 2022-12-01
Simplify usage of typed-checked nodes

lucteo:
**Why?**
It is not straightforward to build logic after type-checking the Val program. The user needs to keep around a `TypedProgram` object and use it in parallel with walking over the AST. This creates a lot of boliderplate code

We want to remove the use of boilerplate code, and make the code that reasons on typed-checked node as easy as possible to reason about.

**How?**
- add *type-checked* nodes that reflect all the properties of the original nodes, but keeps the type information also around.
- allow clients to use `TypedXXX` to represent type-checked nodes

**Testing**
- all existing unit tests pass


lucteo:
Dropping in the favour of #131

PR #131, opened 2022-12-01
Simplify usage of typed-checked nodes (CXX transipilation part)

lucteo:
It is not straightforward to build logic after type-checking the Val program. The user needs to keep around a `TypedProgram` object and use it in parallel with walking over the AST. This creates a lot of boliderplate code

We want to remove the use of boilerplate code, and make the code that reasons on typed-checked node as easy as possible to reason about.

In this PR:
- add *type-checked* nodes that reflect all the properties of the original nodes, but keeps the type information also around.
- allow clients to use `TypedXXX` to represent type-checked nodes
- change CXX transpiler to use typed nodes instead of node IDs

Testing:
- all existing unit tests pass
- no new unit tests created, as the functionality should be covered indirectly by using the typed nodes


lucteo:
attn: @dabrahams , @kyouko-taiga : Using inner enums in the AST nodes would require extra hoops if we translate them into typed nodes.

I'm not sure if there is a clean way to generate this automatically with key paths; @dabrahams do you know of a such a method?

Sources/Compiler/TypedChecked.swift
```diff
@@ -0,0 +1,101 @@
+/// Helper protocol that allows its implementers to have easy access to type aliases for typed nodes.
+public protocol TypedChecked {}
```
kyouko-taiga, line 2: I don't think his protocol is necessary. Having to maintain the list of type aliases is a heavy burden. I think it's better to use `TypedProgram.Node<T>` and rely on contextual type inference to reduce verbosity whenever possible.

Note that the aliases would have be duplicated if we want to have those augmented nodes for different phases of compilation (e.g., after scope construction, after borrow checking, ...)

kyouko-taiga (changes requested):
Fantastic work! Most things look very good, except the TypedProtocol and its type aliases (see comments).

I wouldn't mind about testing. It's a cross-cutting feature that will get thoroughly tested indirectly.

Sources/Compiler/TypedChecked.swift
lucteo: This comes from @dabrahams 's suggestion on https://github.com/lucteo/val/commit/0debaebf1ca3844b0ec1ff3f5f3d3d7902769639#r91233054

I agree that it's not nice to duplicate the lists, but I feel it pays off in terms of readability. Do you know of any way that I can improve on the readability (i.e., don't add a lot of boilerplate code when mentioning typed nodes), and get rid of this ugly list?

lucteo: Maybe, as a compromise between the two alternatives we can use something like: `typealias Typed<T: AST.Node> = SomeProgram.Node<T>`

Then, we we would use `Typed<FunctionDecl>` instead of `TypedFunctionDecl`. We would also do it from anywhere, not only from classes implemented `TypedChecked`.

How does that sound to you?

dabrahams: Works for me!

dabrahams: Aside from contextual type inference, for lines like:

```swift
return emit(returnStmt: TypedReturnStmt(stmt)!)
```

we have a prefix `^` operator convention that Dimi introduced.

```swift
return emit(returnStmt: (^stmt)!)
```

lucteo: I'm now starting to remove `TypedChecked`...

Looking at the definition of `^`, I don't see how I can do what you mention above.
```
/// Creates a type-erased container wrapping the given instance.
public prefix func ^ <T: TypeProtocol>(_ base: T) -> AnyType {
  AnyType(base)
}
```

Besides this being specialized from the wrong type of protocol, I need the reverse of this operation.
For clarity, I would just use the full variant, i.e., `Typed<ReturnStmt>(stmt)!`

lucteo: Patch is up.

Now I'm thinking of moving `TypedNode` (a typed node constructed from an ID) and `Typed` (constructed from an AST node) outside of `TypedProgram` (i.e., instead of `TypedProgram.SomeNode<>` and `TypedProgram.Node<>`. It will display better type names in the compiler, and will separate out the concerns a bit more (`TypedNode` would still depend on `TypedProgram`)

Sources/Compiler/AST/NodeIDs/NodeID.swift
```diff
@@ -20,6 +25,8 @@ public struct NodeID<Subject: Node>: NodeIDProtocol {
   /// The dynamic type of node being referred to.
   public var kind: NodeKind { NodeKind(Subject.self) }
 
+  public init(_ source: Self) { self = source }
```
kyouko-taiga, line 28: What is the purpose of this initializer? It seems its can only be used to create copies of `Self`.

Sources/Compiler/CXX/CXXModule.swift
```diff
@@ -17,34 +17,33 @@ public struct CXXModule {
...
   /// - Requires: `valFunctionDecl` must be declared in `self.decl`.
   public mutating func getOrCreateFunction(
-    correspondingTo valFunctionDecl: NodeID<FunctionDecl>,
-    program: TypedProgram
+    correspondingTo valFunctionDecl: Typed<FunctionDecl>
   ) -> CXXFunctionDecl.ID {
     if let cxxFunctionDecl = valToCXXFunction[valFunctionDecl] { return cxxFunctionDecl }
 
-    assert(program.isGlobal(valFunctionDecl))
+    assert(program.isGlobal(valFunctionDecl.id))
```
kyouko-taiga, line 35: _(non-blocking)_

I think it should be possible to implement `isGlobal` as a property of typed declarations.

Sources/Compiler/TypedNode.swift
```diff
@@ -0,0 +1,188 @@
+// Renaming dance
+
+public typealias ValNode = Node
+extension AST {
```
kyouko-taiga, line 4:
```suggestion
public typealias ValNode = Node

extension AST {
```

Sources/Compiler/TypedNode.swift
```diff
@@ -0,0 +1,188 @@
...
+  public typealias Node = ValNode
+}
```
kyouko-taiga, line 6:
```suggestion
public typealias ValNode = Node

extension AST {

  public typealias Node = ValNode
  
}
```

Sources/Compiler/TypedNode.swift
```diff
@@ -0,0 +1,188 @@
...
+
+/// A projection from a `TypedProgram` of an AST node along with all the non-syntax information
+/// related to that node.
+@dynamicMemberLookup
+public struct TypedNode<ID: NodeIDProtocol> : Hashable {
+  /// The whole program of which this node is a notional part.
```
kyouko-taiga, line 12:
```suggestion
public struct TypedNode<ID: NodeIDProtocol> : Hashable {

  /// The whole program of which this node is a notional part.
```

Sources/Compiler/TypedNode.swift
```diff
@@ -0,0 +1,188 @@
...
+
+  /// The node's identity in `whole.ast`.
+  let id: ID
+
+  /// Equality comparison; only check the node ID.
+  public static func == (lhs: TypedNode<ID>, rhs: TypedNode<ID>) -> Bool {
+      lhs.id == rhs.id
+  }
+
+  /// Hash function; only consider the node ID.
```
kyouko-taiga, line 23:
```suggestion
  /// Hashes `id` into `hasher`.
```

Sources/Compiler/TypedNode.swift
```diff
@@ -0,0 +1,188 @@
...
+  public func hash(into hasher: inout Hasher) {
+      hasher.combine(id)
+  }
+}
+
+/// An AST node with type information.
+public typealias Typed<T: AST.Node> = TypedNode<NodeID<T>>
+
+extension TypedProgram {
+  /// Get the typed node for the given node ID.
```
kyouko-taiga, line 33:
```suggestion
  /// Bundles `id` together with `self`.
```

Sources/Compiler/TypedNode.swift
```diff
@@ -0,0 +1,188 @@
...
+    whole = s.whole
+    id = .init(myID)
+  }
+}
+
+extension TypedNode {
+  
+  /// The corresponding node kind.
+  var kind: NodeKind { id.kind }
+}
```
kyouko-taiga, line 94:
```suggestion
  var kind: NodeKind { id.kind }

}
```

kyouko-taiga (changes requested):
I left a couple of comments.

Also, could you add a white space before and after the first and last declarations (respectively) in type declarations? For example:
```swift
// Good
struct Foo {

  /// Some comment.
  let x: Int

  /// Some comment.  
  let y: String

}

// Bad
struct Foo {
  /// Some comment.  
  let x: Int

  /// Some comment.  
  let y: String
}
```

Sources/Compiler/AST/NodeIDs/NodeID.swift
dabrahams: Probably a C++ instinct.  You don't need these in Swift; everything is implicitly copyable.

Sources/Compiler/CXX/CXXModule.swift
lucteo: I would prefer to leave this out of this PR.

lucteo:
> Also, could you add a white space before and after the first and last declarations (respectively) in type declarations? For example:

I got so used to clang-format, that I can't no longer format the code without a formatting tool. I tried to find a similar thing in XCode but I didn't see it. Do you have an automatic formatter that you use?

Sources/Compiler/AST/NodeIDs/NodeID.swift
lucteo: If I remove this, the optional initializer below fails:
> Non-failable initializer requirement 'init(_:)' cannot be satisfied by a failable initializer ('init?')

kyouko-taiga:
Sadly I don't have good formatter to recommend. I think @dabrahams mentioned something about swift-format?

kyouko-taiga (approved):
LGTM. Let's not worry too much about code format. I'm sure things will eventually reached a fixed point.

PR #132, opened 2022-12-05
Refactor the expression parser

kyouko-taiga:


Sources/Compiler/Parse/Parser.swift
```diff
@@ -1538,562 +1531,1066 @@ public enum Parser {
...
-        try state.ast.insert(wellFormed: IntegerLiteralExpr(
-          value: state.lexer.source[token.origin].filter({ $0 != "_" }),
-          origin: token.origin))
-      })
-  )
+    switch head.kind {
+    case .bool:
+      // Boolean literal.
+      _ = state.take()
+      let expr = try state.ast.insert(
```
dabrahams, line 1757: Lots of boilerplate repetition in this hunk.  Can we do something generic parameterized on an expr type?

Sources/Compiler/Parse/Parser.swift
```diff
@@ -1538,562 +1531,1066 @@ public enum Parser {
...
-    take(.string)
-      .map({ (state, token) -> NodeID<StringLiteralExpr> in
-        try state.ast.insert(wellFormed: StringLiteralExpr(
-          value: String(state.lexer.source[token.origin].dropFirst().dropLast()),
-          origin: token.origin))
-      })
-  )
+    case .any:
+      // Existential type expression.
+      return try parseExistentialTypeExpr(in: &state).map(AnyExprID.init)
```
dabrahams, line 1804: Nice; it's getting smaller, but still looks like this same pattern is repeated over and over so it'd be nice to factor it out.

Sources/Compiler/Parse/Parser.swift
```diff
@@ -1538,562 +1531,1066 @@ public enum Parser {
...
-          origin: state.ast[tree.0.0].origin!.extended(upTo: state.currentIndex)))
-      })
-  )
+  private static func parseNameExprComponent(
+    in state: inout ParserState
+  ) throws -> NameExprComponent? {
+    // Parse the name of the component.
+    guard let name = try parseEntityName(in: &state) else { return nil }
+
+    // If the next token is a left angle bracket, without any leading whitespace, parse a static
```
dabrahams, line 1904:
```suggestion
    // If the next token is a left angle bracket without any leading whitespace, parse a static
```

Sources/Compiler/Parse/Parser.swift
```diff
@@ -3033,6 +3319,22 @@ public enum Parser {
...
   }
 
+  private static func expected(
+    _ closerDescription: String,
+    matching opener: Token,
+    in state: ParserState
+  ) -> Diagnostic {
+    expected(
+      "'\(closerDescription)'",
+      at: state.currentLocation,
```
dabrahams, line 3329: This is a great example why the "if you split the list everything gets its own line" pattern doesn't work so well..  You end up with these tall columns and things that are closely associated, like `x` and `y` parameters or phrases

```swift
   expected(
      "'\(closerDescription)'", at: state.currentLocation,
```

lose their visible associations.

Sources/Compiler/Parse/Parser.swift
```diff
@@ -3033,6 +3319,22 @@ public enum Parser {
     .error("expected \(subject)", range: location ..< location, children: children)
   }
 
+  private static func expected(
+    _ closerDescription: String,
+    matching opener: Token,
+    in state: ParserState
+  ) -> Diagnostic {
```
dabrahams, line 3326:
```suggestion
  private static func expected(_ closerName: String,  matching opener: Token, in state: ParserState) -> Diagnostic {
```

or

```suggestion
  private static func expected(
    _ closerDescription: String,  matching opener: Token, in state: ParserState) -> Diagnostic
  {
```

Sources/Compiler/Parse/Parser.swift
```diff
@@ -3289,6 +3605,25 @@ where Base.Context == ParserState
...
+  init?(_ token: Token) {
+    switch token.kind {
+    case .infix:
+      self = .infix
+    case .prefix:
+      self = .prefix
+    case .postfix:
+      self = .postfix
+    default:
+      return nil
```
dabrahams, line 3621:
```suggestion
    case .infix: self = .infix
    case .prefix: self = .prefix
    case .postfix: self = .postfix
    default: return nil
```

Sources/Compiler/Parse/ParserState.swift
```diff
@@ -54,6 +54,9 @@ struct ParserState {
     self.currentIndex = lexer.source.contents.startIndex
   }
 
+  /// The character stream from the current index.
```
dabrahams, line 57: Can you use words other than “character stream” to describe this thing called `characterStream`?  Or better yet, can you find a name for it that doesn't simply describe its notional type?  Is this the `remainingText` or the `unlexedSource` or something?

Sources/Compiler/Parse/ParserState.swift
```diff
@@ -54,6 +54,9 @@ struct ParserState {
...
+  var characterStream: Substring { lexer.source.contents[currentIndex...] }
+
   /// Indicates whether the parser is at global scope.
   var atGlobalScope: Bool { atModuleScope || atNamespaceScope }
```
dabrahams, line 61:
```suggestion
  var isAtGlobalScope: Bool { isAtModuleScope || isAtNamespaceScope }
```

I realize you didn't do anything with this in this PR, but these names don't seem to follow the guidelines.  Also, I think this definition of “global” is a little surprising.  Are static members of types "at global scope?"  if not, why not.

Sources/Compiler/Parse/ParserState.swift
```diff
@@ -135,6 +138,20 @@ struct ParserState {
     return token
   }
 
+  /// Returns whether the next token has the specified kind, or `false` if the stream is empty.
```
dabrahams, line 141:
```suggestion
  /// Returns whether a token of the given `kind` is next in the input.
```

"specified" is just a 3-syllable way to say “given” and even that should usually be avoided.

[And it seems I'll never be comfortable reading “Returns whether” in documentation.  Grumpy old man!]

Sources/Compiler/Parse/ParserState.swift
```diff
@@ -135,6 +138,20 @@ struct ParserState {
...
+  mutating func isNext(_ kind: Token.Kind) -> Bool {
+    peek()?.kind == kind
+  }
+
+  /// Returns whether the next token satisfies `predicate`, or `false` if the stream is empty.
```
dabrahams, line 146:
```suggestion
  /// Returns whether a token satisfying `predicate` is next in the input.
```

Sources/Compiler/Parse/ParserState.swift
```diff
@@ -149,7 +166,7 @@ struct ParserState {
     return token
   }
 
-  /// Consumes and returns the next token, only if it has the specified kind.
+  /// Consumes and returns the next token it has the specified kind.
```
dabrahams, line 169:
```suggestion
  /// Consumes and returns the next token if it has the specified kind.
```

Sources/Compiler/Parse/ParserState.swift
```diff
@@ -167,7 +184,7 @@ struct ParserState {
     return take(kind)
   }
 
-  /// Consumes and returns the next token, if it has the specified predicate.
+  /// Consumes and returns the next token if it satisfies the specified predicate.
```
dabrahams, line 187:
```suggestion
  /// Consumes and returns the next token if it satisfies `predicate`.
```

Sources/Compiler/Parse/Token.swift
```diff
@@ -112,20 +112,24 @@ public struct Token {
 
   /// Indicates whether `self` may be in an operator
   ///
-  /// Use this property rather than testing `self.kind == .oper` as some operators (e.g., `==`) do
-  /// not have the kind `.oper`.
-  public var isOperatorToken: Bool {
+  /// Operators may be represented by one or more tokens of different kinds. For example, `<<` is
```
dabrahams, line 115:
```suggestion
  /// Operators may be represented by one or more tokens. For example, `<<` is
```

Sources/Compiler/TypeChecking/CaptureCollector.swift
```diff
@@ -559,12 +559,10 @@ struct CaptureCollector {
     into captures: inout FreeSet
   ) {
     switch ast[id].domain {
-    case .none:
+    case .none, .implicit:
```
dabrahams, line 562: Intended functionality change?

dabrahams (approved):
Looks good overall.  A couple of small things you should probably address.

Sources/Compiler/Parse/ParserState.swift
kyouko-taiga: That property indicates whether the parser is parsing at the global scope, which does not imply that things parsed in type scope can't end up being considered "global" (see `Program.isGlobal(_:)`).

Sources/Compiler/Parse/Parser.swift
kyouko-taiga: Fixed in another PR.

Sources/Compiler/TypeChecking/CaptureCollector.swift
kyouko-taiga: Yep. Previous version was a mistake.

PR #135, opened 2022-12-07
Remove probably-wrong conformances

dabrahams:
An empty cache should be equivalent to a full one, if anything, right?


kyouko-taiga (approved):

kyouko-taiga:
I don't understand the question.

dabrahams:
Before I realized I could delete the whole thing I was deleting the Hashable/Equatable conformances, because they made the contents of the cache significant.

PR #137, opened 2022-12-08
SubstitutionMap cleanup

dabrahams:
- Shouldn't be a collection; things whose order is not salient should _have_, rather than _be_ collections.  Let's not compound Swift's error.  Plus nobody's using the functionality.
- flattened() was badly named and documented; the flattening is below the level of the client's abstraction.


kyouko-taiga (approved):
LGTM

PR #140, opened 2022-12-09
Parser cleanup

kyouko-taiga:


Sources/Compiler/AST/Expr/LabeledArgument.swift
```diff
@@ -1,13 +1,13 @@
-/// An argument of a function or subscript call.
-public struct CallArgument: Codable {
+/// A static argument or an argument to a function or subscript call.
+public struct LabeledArgument: Codable {
```
dabrahams, line 2: Unless there are `UnlabeledArgument`s, I'm not sure this is an improved name.  Are we going to have `GenericArgument`s that are distinct?  If so, I like the old name.  Otherwise, I might opt for `Argument`.

Sources/Compiler/AST/Expr/LabeledArgument.swift
```diff
@@ -1,13 +1,13 @@
...
 
   /// The label of the argument, if any.
   public let label: SourceRepresentable<Identifier>?
 
   /// The value of the argument.
   public let value: AnyExprID
 
-  public init(label: SourceRepresentable<Identifier>? = nil, value: AnyExprID) {
+  public init(label: SourceRepresentable<Identifier>?, value: AnyExprID) {
```
dabrahams, line 10: Nice.

Sources/Compiler/Parse/Parser+Diagnostics.swift
```diff
@@ -0,0 +1,59 @@
+extension Diagnostic {
+
+  static func diagnose(assignOperatorRequiresWhitespaces token: Token) -> Diagnostic {
+    .error("infix operator requires whitespaces on both sides", range: token.origin)
```
dabrahams, line 4:
```suggestion
    .error("assignment operator requires whitespaces on both sides", range: token.origin)
```

Sources/Compiler/Parse/Parser+Diagnostics.swift
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  static func diagnose(expected kind: Token.Kind, at location: SourceLocation) -> Diagnostic {
+    .error("expected '\(kind)'", range: location ..< location)
```
dabrahams, line 8: Shouldn't this one just dispatch to the next one?

Sources/Compiler/Parse/Parser+Diagnostics.swift
```diff
@@ -0,0 +1,59 @@
...
+      at: state.currentLocation,
+      children: [
+        .error(
+          "to match this '\(state.lexer.source[opener.origin])'",
+          range: opener.origin)
+      ]
+    )
+  }
+
+  static func diagnose(infixOperatorRequiresWhitespacesAt range: SourceRange?) -> Diagnostic {
```
dabrahams, line 35: Can we stop calling these parameters `range` in new code at least?

Sources/Compiler/Parse/Parser+Diagnostics.swift
```diff
@@ -0,0 +1,59 @@
...
+
+  static func diagnose(assignOperatorRequiresWhitespaces token: Token) -> Diagnostic {
+    .error("infix operator requires whitespaces on both sides", range: token.origin)
+  }
+
+  static func diagnose(expected kind: Token.Kind, at location: SourceLocation) -> Diagnostic {
+    .error("expected '\(kind)'", range: location ..< location)
+  }
+
+  static func diagnose(
```
dabrahams, line 11: Not related to this PR, is the word `diagnose` really adding anything to any of these at the use site?  I kinda get the feeling that the first argument label should replace it.

Sources/Compiler/Parse/Parser+Diagnostics.swift
```diff
@@ -0,0 +1,59 @@
...
+
+  static func diagnose(separatedPrefixOperatorAt range: SourceRange?) -> Diagnostic {
+    .error("prefix operator cannot be separated from its operand", range: range)
+  }
+
+  static func diagnose(unexpectedToken token: Token) -> Diagnostic {
+    .error("unexpected token '\(token.kind)'", range: token.origin)
+  }
+
+  static func diagnose(unterminatedCommentEndingAt endLocation: SourceLocation) -> Diagnostic {
```
dabrahams, line 51: Seems like you probably want the start of these unterminated things as child diagnostics?

Sources/Compiler/Parse/Parser+Diagnostics.swift
dabrahams: Actually now that I look at the changes in `Parser.swift`, maybe it _is_ related to this PR.  Use sites are starting to look pretty complex and show signs of the Muffin Man.

Sources/Compiler/Parse/Parser.swift
```diff
@@ -259,7 +259,7 @@ public enum Parser {
       if prologue.isEmpty {
         return nil
       } else {
-        throw DiagnosedError(expected("declaration", at: state.currentLocation))
+        throw DiagnosedError(.diagnose(expected: "declaration", at: state.currentLocation))
```
dabrahams, line 262: I don't see how the addition of the word `diagnose` is improving this use site.  Is there any reason these shouldn't all be `init`s of `DiagnosedError`?

Sources/Compiler/Parse/Parser.swift
```diff
@@ -1329,11 +1291,6 @@ public enum Parser {
...
   )
 
-  static let parameterList = (
-    parameterDecl.and(zeroOrMany(take(.comma).and(parameterDecl).second))
-      .map({ (_, tree) -> [NodeID<ParameterDecl>] in [tree.0] + tree.1 })
-  )
-
   static let parameterDecl = (
     parameterInterface
       .and(maybe(take(.colon).and(parameterTypeExpr)))
```
dabrahams, line 1296: How did you decide which `.and`s to keep and which to drop?

Sources/Compiler/Parse/Parser.swift
```diff
@@ -2198,33 +2097,28 @@ public enum Parser {
       wellFormed: SpawnExpr(decl: decl, origin: state.ast[decl].origin))
   }
 
-  private static func parseLambdaTypeOrTupleExpr(in state: inout ParserState) throws -> AnyExprID? {
+  private static func parseLambdaTypeOrTupleExpr(
+    in state: inout ParserState
+  ) throws -> AnyExprID? {
+    // Expect a left parenthesis.
```
dabrahams, line 2103: Is this new functionality?  Perhaps it should (in principle) be in a separate PR?

dabrahams (approved):
Very nice reduction!  Please see inline comments.

Sources/Compiler/AST/Expr/LabeledArgument.swift
kyouko-taiga: I wanted `Argument`, but it clashes with `swift-argument-parser` causes annoying issues with the property wrapper that we use in `CLI`.

I believe `GenericArgument` is gone with that PR.

dabrahams:
> On Dec 9, 2022, at 6:01 PM, Dimi Racordon ***@***.***> wrote:
> 
> I wanted Argument, but it clashes with swift-argument-parser causes annoying issues with the property wrapper that we use in CLI.
> 
> 

Need `fileprivate import`!!

Sources/Compiler/Parse/Parser+Diagnostics.swift
kyouko-taiga: Makes sense

Sources/Compiler/Parse/Parser+Diagnostics.swift
kyouko-taiga: Sure, but most of these functions were just copy/pasted from a different file.

Sources/Compiler/Parse/Parser+Diagnostics.swift
kyouko-taiga: I used to have different function names, but that doesn't always work. Sometimes you really don't want a label for the first argument.

I'm open to suggestions but I prefer `diagnose(blah:blah:)` over `blah(_:blah:)`.

dabrahams: Look at the use-sites.  Are you really writing what you'd want to read there?

Sources/Compiler/Parse/Parser+Diagnostics.swift
kyouko-taiga: That would make sense. Probably a good suggestion for an issue about improving diagnostics of unterminated comments.

Sources/Compiler/Parse/Parser.swift
kyouko-taiga: That could be a solution yes. I don't mind `diagnose` but if you feel strongly against it, please open an issue so that we can do a PR that renames all uses.

Sources/Compiler/Parse/Parser.swift
kyouko-taiga: I want to remove all of them eventually.

Sources/Compiler/Parse/Parser+Diagnostics.swift
kyouko-taiga: I'm writing a lot of `diagnostics.append(.diagnose(blah: foo, blah: bar))` at call sites. Admittedly, that's little verbose, but I honestly think that's acceptable. As mentioned in another comment, I wouldn't mind using `diagnostics.append(.init(blah: foo, blah: bar))`.

Like I said, a previous version was designed do that you could write `diagnostics.append(.unexpected(token, at: currentLocation))`, which I guess you would prefer over the current approach. The problem was that it would not work well with things such as `diagnose(_:doesNotConformTo trait:at:because:)`(see `TypeChecker+Diagnostics.swift` for this example and others). I also felt it was starting to make `Diagnostic`'s namespace very messy. So ultimately I decided to keep the diagnostic constructors more uniform.

kyouko-taiga: Excellent suggestion!

PR #142, opened 2022-12-10
Improvements on the API and documentation of IR-related components

kyouko-taiga:


Sources/Compiler/IR/Module.swift
```diff
@@ -1,17 +1,17 @@
 import Utils
 
-/// A module lowered to Val IR.
+/// A collection of IR functions.
```
dabrahams, line 3: The old comment was better.  This is not just an arbitrary collection of IR functions.  It contains all the auxilliary data computed as part of lowering.  It has a name and an entry point.  No, this is a Val module in IR form.

Sources/Compiler/IR/Module.swift
```diff
@@ -1,17 +1,17 @@
...
 public struct Module {
 
   /// An array of functions.
   public typealias Functions = [Function]
 
+  /// The program defining the functions in `self`.
+  public let program: TypedProgram
```
dabrahams, line 10:
```suggestion
  /// The typechecked but un-lowered source from which `self` was derived.
  public let source: TypedProgram
  ```

Well now, `TypedProgram` isn't really a whole program, is it?

File: Sources/Compiler/IR/Module.swift [hunk 3]
```diff
@@ -1,17 +1,17 @@
 import Utils
 
-/// A module lowered to Val IR.
+/// A collection of IR functions.
 public struct Module {
 
   /// An array of functions.
```
dabrahams, line 6:
```suggestion
/// The form in which a `Module` exposes all its lowered functions.
```

Sources/Compiler/IR/Module.swift
```diff
@@ -21,14 +21,24 @@ public struct Module {
   /// The ID of the module's entry function, if any.
   public private(set) var entryFunctionID: Function.ID?
 
-  /// A map from function declaration its ID in the module.
+  /// A map from function declaration to its ID in `self.functions`.
```
dabrahams, line 24: “`ID` in the module” made sense.  If you're going to describe it in terms of `self.functions`, it should be `Index`.

Sources/Compiler/IR/Module.swift
```diff
@@ -21,14 +21,24 @@ public struct Module {
...
   private var loweredFunctions: [NodeID<FunctionDecl>: Function.ID] = [:]
 
-  public init(decl: NodeID<ModuleDecl>, name: String) {
+  /// Creates an IR module lowering `decl` using `emitter`.
```
dabrahams, line 27: It's gotta do something else; it modifies emitter somehow!  That should be described.

Sources/Compiler/IR/Function.swift
```diff
@@ -4,7 +4,7 @@ import Utils
 public struct Function {
 
   /// The ID of a Val IR function.
```
dabrahams, line 6:
```suggestion
/// The ID of a `Function` in its `Module`.
```

Sources/Compiler/IR/Module.swift
```diff
@@ -40,19 +40,19 @@ public struct Module {
     }
   }
 
-  /// Returns whether the module is well-formed.
-  public func check() -> Bool {
+  /// Returns whether the IR in `self` is well-formed.
```
dabrahams, line 56: If a `false` return indicates a bug in the compiler, you should say so in the doc.

dabrahams (approved):
Please see comments inline

Sources/Compiler/IR/Emitter.swift
```diff
@@ -25,15 +25,6 @@ public struct Emitter {
 
   // MARK: Declarations
 
-  /// Emits the Val IR of the module identified by `decl`.
-  public mutating func emit(module decl: NodeID<ModuleDecl>) -> Module {
```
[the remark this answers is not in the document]
dabrahams: I guess it doesn't really matter…

Sources/Compiler/IR/Block.swift
```diff
@@ -1,13 +1,13 @@
 import Utils
 
-/// A basic block in a Val IR function.
+/// A collection of instructions.
 public struct Block {
 
-  /// The ID of a basic block in a Val IR function.
+  /// The ID of a basic block.
```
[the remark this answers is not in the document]
kyouko-taiga: It was just shorter, but I think it bring confusion. I'll rename it in a different PR.

Sources/Compiler/IR/Block.swift
```diff
@@ -30,18 +30,18 @@ public struct Block {
 
   }
 
+  /// The address of an instruction in `self`.
```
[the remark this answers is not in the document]
kyouko-taiga: I'll skip this comment, as I think #143 addresses it (no pun intended).

Sources/Compiler/IR/Module.swift
kyouko-taiga: It isn't doing anything else, but your observation revealed that there's no need to reuse the same emitter to create different modules.

Sources/Compiler/IR/Function.swift
```diff
@@ -53,66 +64,73 @@ public struct Function {
 
 }
 
-extension Function: Sequence {
-
-  public typealias BlockAddress = DoublyLinkedList<Block>.Address
+extension Function {
 
-  public typealias Element = (block: BlockAddress, inst: Block.InstAddress)
+  /// A view of a function's instruction addresses.
```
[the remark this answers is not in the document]
kyouko-taiga: Not so easy to do it with `FlattenCollection`, because you need to synthesize values that aren't element of any of the sequences we have.

But I removed this property in #143 anyway, so we can forget about this piece of code.

Sources/Compiler/IR/Module.swift
kyouko-taiga: I wanted to mention the "collection of functions" because that seamed helpful to you when went over the code. I'll document that differently.

Note that I tried using `Incidental<T>` to highlight the notional value of the module, but it made using these incidental properties more difficult to use. 

Defining conformance to `Equatable` also involves implementing additional stuff that isn't actually necessary.

Sources/Compiler/IR/Module.swift
kyouko-taiga: Interesting. I would say that it is, and so was `ScopedProgram`. The additional information we gather is only inferred from what we already had.

dabrahams: Does every `TypedProgram` have a main?  If not, then it's just a module ersump'n.

Sources/Compiler/IR/Module.swift
dabrahams: It clearly was having an effect on the emitter, or it wouldn't be an `inout` parameter.

kyouko-taiga: Actually, no, but only because I designed it poorly.

The emitter isn't really an abstraction. It's a very big function (we have several of those actually, like the constraint generator). If we look closely at its "state", it's only temporary data that are cleaned up after the emitter "returns".

kyouko-taiga: Maybe it would help have a `callAsFunction` entry point for these "not really abstraction" things.

dabrahams: Prolly not because then you have to construct one.  IMO what you do is make them `fileprivate` and create a function at module scope that presents the actual abstraction… or, make it a method on one of the types normally used as input: `program.emit(into: &m)` or something.

PR #143, opened 2022-12-11
Use indices and addresses to insert IR instructions

kyouko-taiga:


Sources/Compiler/IR/InstructionIndex.swift
```diff
@@ -0,0 +1,31 @@
...
+///
+/// An index identifies a function, a basic block in this function, and a position in this block,
+/// including the "past-the-end" position that isn't valid for use as a subscript argument. Unlike
+/// an identity, an index is not *stable*: inserting or removing instructions from the containing
+/// block may invalidate existing indices.
+///
+/// - SeeAlso: `InstructionIdentity`
+public struct InstructionIndex: Hashable {
+
+  /// The index of the function in which the instruction resides.
```
dabrahams:
```suggestion
  /// The function containing this position.
```

Sources/Compiler/IR/InstructionIndex.swift
```diff
@@ -0,0 +1,31 @@
...
+  public let function: Module.Functions.Index
+
+  /// The address of the block in which the instruction resides.
```
dabrahams:
```suggestion
  /// The block containing this position.
```

Sources/Compiler/IR/InstructionIndex.swift
```diff
@@ -0,0 +1,31 @@
...
+  public let block: Function.Blocks.Address
+
+  /// The index of the instruction.
+  public let index: Block.Instructions.Index
```
dabrahams:
```suggestion
  /// This position relative to its block.
  public let indexInBlock: Block.Instructions.Index
```

Sources/Compiler/IR/InstructionIndex.swift
```diff
@@ -0,0 +1,31 @@
...
+
+  /// Creates a path with the given properties.
```
dabrahams: In this context we don't know what "a path" means.  Please try again.

Sources/Compiler/IR/InstructionID.swift
```diff
@@ -0,0 +1,26 @@
...
+  /// The index of the function in which the instruction resides.
+  public let function: Module.Functions.Index
+
+  /// The address of the block in which the instruction resides.
+  public let block: Function.Blocks.Address
+
+  /// The address of the instruction.
+  public let address: Block.Instructions.Address
+
+  /// Creates a path with the given properties.
```
dabrahams: In this context we don't know what "a path" means.  Please try again.

Sources/Compiler/IR/Module.swift
```diff
@@ -39,6 +39,24 @@ public struct Module {
   /// The module's name.
   public var name: String { program.ast[decl].name }
 
+  /// Accesses the function at given index.
+  public subscript(function index: Functions.Index) -> Function {
+    _read   { yield  functions[index] }
+    _modify { yield &functions[index] }
+  }
```
dabrahams:
```suggestion
  /// Accesses the function at `i`.
  public subscript(_ i: Functions.Index) -> Function {
    _read { yield functions[i] }
    _modify { yield &functions[i] }
  }
```

or even, since function indices _are_ stable,

```suggestion
  /// Accesses the given function.
  public subscript(_ f: Functions.Index) -> Function {
    _read { yield functions[f] }
    _modify { yield &functions[f] }
  }
```

Sources/Compiler/IR/Module.swift
```diff
@@ -39,6 +39,24 @@ public struct Module {
...
+
+  /// Accesses the basic block with the given identity.
+  public subscript(block id: Block.ID) -> Block {
+    _read   { yield  functions[id.function].blocks[id.address] }
+    _modify { yield &functions[id.function].blocks[id.address] }
+  }
```
dabrahams:
```suggestion
  /// Accesses the given block.
  public subscript(_ b: Block.ID) -> Block {
    _read   { yield  functions[b.function].blocks[b.address] }
    _modify { yield &functions[b.function].blocks[b.address] }
  }
```

Totally instinct question, and I'll accept a one-word answer: do we really want all of these subscripts to be public?

Sources/Compiler/IR/Module.swift
```diff
@@ -39,6 +39,24 @@ public struct Module {
...
+
+  /// Accesses the instruction with the given identity.
```
dabrahams: See comments on the two subscripts above.

dabrahams (changes requested):
Because you rebased while I was reviewing it, all my comments are marked outdated, and I have stopped reviewing so we have a chance of not losing my input.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
```diff
@@ -182,7 +181,7 @@ public struct DefiniteInitializationPass: TransformPass {
     // Interpret the function until we reach a fixed point.
     while let block = work.popFirst() {
       // Handle the entry as a special case.
-      if block == module[functionID].blocks.firstAddress {
+      if block == module[function: functionID].blocks.firstAddress {
```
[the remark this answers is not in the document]
kyouko-taiga: You can use subscript to access functions, blocks, and instructions. I want to make it explicit which kind of object you're accessing at the use site.

The rationale is the same as for overloaded visitors (e.g., `visit(functionCall)`).

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
```diff
@@ -704,7 +700,7 @@ public struct DefiniteInitializationPass: TransformPass {
   /// with a projection of `self` and the state summary of the object before returning `false`.
   private mutating func consume(
     localForKey key: FunctionLocal,
-    with consumer: InstID,
```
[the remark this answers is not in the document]
kyouko-taiga: In this particular case, I believe it makes sense to make the renaming part of the PR. There's a symmetry between `InstructionID` and `InstructionIndex` that is part of the suggested changes.

I think it's silly to create a new type and name it `InstIndex` only to push a PR right after to rename it.

Sources/Utils/DoublyLinkedList.swift
```diff
@@ -244,6 +244,19 @@ public struct DoublyLinkedList<Element> {
     return newAddress
   }
 
+  /// Inserts `newElement` at the given position and returns its address.
+  ///
+  /// The new element is inserted before the element currently at `position`. You can pass a
+  /// "past the end" index to append `newElement` at the end of `self`.
```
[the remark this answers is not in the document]
kyouko-taiga: Though it may be standard, it is not necessarily intuitive to everyone. It isn't for me and I always go back to `Array`'s documentation to remember how it should work.

Sources/Compiler/IR/InstructionID.swift
```diff
@@ -0,0 +1,26 @@
+/// A stable identity denoting an instruction in a module.
+///
+/// - SeeAlso: `InstructionIndex`
+public struct InstructionID: Hashable {
+
+  /// The index of the function in which the instruction resides.
```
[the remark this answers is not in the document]
kyouko-taiga: I prefer index here. `InstructionID` is not storing a function. Now that we have bundles, I think the distinction is relevant.

Sources/Compiler/IR/InstructionID.swift
```diff
@@ -0,0 +1,26 @@
...
+  public let function: Module.Functions.Index
+
+  /// The address of the block in which the instruction resides.
```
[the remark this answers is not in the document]
kyouko-taiga: The distinction is definitely relevant in for objects that could be either an address or an index.

Sources/Compiler/IR/InstructionID.swift
```diff
@@ -0,0 +1,26 @@
...
+  public let block: Function.Blocks.Address
+
+  /// The address of the instruction.
+  public let address: Block.Instructions.Address
```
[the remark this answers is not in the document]
kyouko-taiga: I know I am biased against multi-word identifiers because I really dislike came case, but somehow I feel this one is particularly unattractive.

Sources/Compiler/IR/InstructionID.swift
```diff
@@ -0,0 +1,26 @@
...
+
+  /// Creates a path with the given properties.
+  public init(
```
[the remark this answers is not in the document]
kyouko-taiga: Please don't assume I'm never considering use sites ;)
I did hear, and I thought the labels helped.

I'm not opposed to remove them but I'm not a fan of having an initial `path:`.

Sources/Compiler/IR/InstructionIndex.swift
```diff
@@ -0,0 +1,31 @@
+/// An index denoting an instruction in a module.
```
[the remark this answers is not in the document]
kyouko-taiga: I think we should own the fact that in Swift "index" and "position" are interchangeable terms. Given that observation, I'd be tempted to use "index" in the documentation because that's the name of `Collection`'s associated type.

Sources/Compiler/IR/Module.swift
kyouko-taiga: As mentioned earlier, I want to keep the label.

Sources/Compiler/IR/Module.swift
kyouko-taiga: yes

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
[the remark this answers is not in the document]
dabrahams: > I want to make it explicit which kind of object you're accessing at the use site.

Why?  Look at the use site; isn't it obvious that it's not helping?

> The rationale is the same as for overloaded visitors (e.g., visit(functionCall:)).

I disagree that it's the same at all.  The visitor functions typically have tens of overloads which makes finding the one you want difficult.  These subscripts don't have that problem.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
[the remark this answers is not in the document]
dabrahams: 🤷 Doing a good PR review is a lot of work.  I'd rather be writing code.  It isn't silly at all to group renames that don't otherwise change functionality or API separately if it helps reviewers.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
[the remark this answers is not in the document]
kyouko-taiga: It's obvious in this particular call site because the index is called `functionID`. This call site would not be that obvious:

```swift
module[continuation].inputs
```

Basic blocks and functions have inputs. `continuation` is a reasonable name for either of these entities..

Sources/Utils/DoublyLinkedList.swift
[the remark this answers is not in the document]
dabrahams: 🤷 Anything you can name is "not necessarily intuitive to everyone."  If we go out of our way to explain how every common idiom works every time we use it, we're making the code harder to read, not easier.

Sources/Compiler/IR/InstructionID.swift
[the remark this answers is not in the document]
dabrahams: The distinction is there in the type.  Documentation should explain semantic value at the level of the abstraction being defined.

Sources/Compiler/IR/InstructionID.swift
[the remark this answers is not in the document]
dabrahams: The distinction is there in the type. Documentation should explain semantic value at the level of the abstraction being defined.

Sources/Compiler/IR/InstructionID.swift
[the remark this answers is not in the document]
dabrahams: Yes, I don't like the aesthetics either.   By all means come up with something prettier.  But semantics outweigh aesthetics.

Sources/Compiler/IR/InstructionIndex.swift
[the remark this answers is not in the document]
dabrahams: > I think we should own the fact that in Swift "index" and "position" are interchangeable terms.

I do own it.  It's marginally helpful to use a different word even if the two words are normally interchangeable.  That's especially true because "index" has other valid meanings.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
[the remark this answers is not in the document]
kyouko-taiga: If I commit a new type named `InstIndex` *after* having merged your "de-abbrev `Inst`" PR, I can almost predict you'll comment:

> Didn't we agree we should not abbreviate "Inst"?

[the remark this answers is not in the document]
dabrahams: What's your point?

Sources/Utils/DoublyLinkedList.swift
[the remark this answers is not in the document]
kyouko-taiga: Let's agree that if it helps one of the main maintainer of the code be more efficient, then it's worth it ;)

Sources/Compiler/IR/InstructionIndex.swift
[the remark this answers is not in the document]
kyouko-taiga: My point is that these other meanings are mostly relevant outside of the Swift ecosystem.

Sources/Utils/DoublyLinkedList.swift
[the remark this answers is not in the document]
dabrahams: OK. Do you want me to explain this any time I write an insertion API using an index to denote position?  Just LMK.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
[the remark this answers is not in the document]
kyouko-taiga: It's a catch-22. Either I rename `InstID` and you need to stumble upon seemingly unrelated renames, or I define `InstIndex` and prompts you to request a name change.

Let's put it this way. The relationship between `InstructionID` and `InstructionIndex` is significant. It stems from the observations hat both concepts are necessary. The names of these types highlight the relationship. `InstID` would be semantically farther to `InstrunctionIndex` than `InstructionID` is. To me, it follows that I'm justified renaming `InstID` in this PR.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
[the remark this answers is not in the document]
dabrahams: In that code, you know you're getting the continuation out of the module and looking at its inputs. It absolutely doesn't matter whether it's a basic block or a function.  I object to putting type indications into non-type names because the idea that you need to do it basically implies reading generic programs is impossible.

Sources/Compiler/IR/InstructionIndex.swift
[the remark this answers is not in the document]
dabrahams: I don't understand why that matters.  I did say "marginally."

Sources/Utils/DoublyLinkedList.swift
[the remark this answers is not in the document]
kyouko-taiga: That will certainly be helpful to me.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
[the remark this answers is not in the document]
kyouko-taiga: I am surprised you're fighting against this so strongly. This is a form of type-based overloading, one of which we wanted to get rid of. The only way for the reader to infer what the type of `module[continuation].inputs` is is to know the different overloads of `module.[]`and the type of `continuation`.

It's fine though, I won't die on that hill. Let's remove the labels.

Sources/Compiler/IR/InstructionIndex.swift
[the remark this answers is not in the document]
kyouko-taiga: I think the use of the term "position" is confusing, because it suggests the existence of another kind of abstraction, different from an index. I guess the ship has sailed because we already use "position" and "index" interchangeably in many places of the code, but I believe it was a mistake.

Given the situation, however, I'm fine with your suggested edit. I'm sure the discussion about "position" and "index" will come back when we document Val's standard library.

Sources/Compiler/IR/InstructionID.swift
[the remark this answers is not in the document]
kyouko-taiga: I have another idea. I'll propose it is a different PR.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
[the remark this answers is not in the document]
dabrahams: I'm sorry, I don't understand. I got lost, which probably means I don't understand the situation as well as I thought. So I withdraw the comment.

dabrahams:
> On Dec 14, 2022, at 10:53 AM, Dimi Racordon ***@***.***> wrote:
> 
> I am surprised you're fighting against this so strongly. This is a form of type-based overloading, one of which we wanted to get rid of.

Until we finally relented.

I understand why it’s confusing to you, but there actually is a logic behind what I’m saying.  I’d rather not have type-based overloads because it hurts code writers.  However, readers take precedence over writers, and when the overloads form a logical unit (this particular unit could have been a single subscript that is generic over `T: ModuleComponentID` for example) it’s better to group the names rather than to inject type information at the use-site.

> The only way for the reader to infer what the type of module[continuation].inputs is is to know the different overloads of module.[]and the type of continuation.

Knowing the type is not that important.  You can understand the semantics of that statement without knowing its type: it gets you the inputs of the continuation.  I strongly believe that adding words to API just to describe the types of (sub) expressions is more harmful than helpful, as a rule.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
```diff
@@ -33,7 +33,7 @@ public struct DefiniteInitializationPass: TransformPass {
 
   public mutating func run(function functionID: Function.ID, module: inout Module) -> Bool {
```
dabrahams, line 34: Missing a doc

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
```diff
@@ -247,7 +247,7 @@ public struct DefiniteInitializationPass: TransformPass {
 
   /// Creates the before-context of the function's entry block.
```
dabrahams, line 248:
```suggestion
  /// Returns the before-context of the function's entry block.
```

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
```diff
@@ -279,9 +279,9 @@ public struct DefiniteInitializationPass: TransformPass {
   }
 
   private mutating func eval(block: Function.Blocks.Address, in module: inout Module) -> Bool {
```
dabrahams, line 281: Because of the missing doc here, it's hard to evaluate what is going on. The function signature _reads_ like this is part of an interpreter implementation.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
```diff
@@ -465,7 +465,7 @@ public struct DefiniteInitializationPass: TransformPass {
     deallocStack instruction: DeallocStackInstruction, id: InstructionID, module: inout Module
   ) -> Bool {
     // The location operand is the result an `alloc_stack` instruction.
-    let alloc = module[instruction: instruction.location.instruction!] as! AllocStackInstruction
+    let alloc = module[instruction.location.instruction!] as! AllocStackInstruction
```
dabrahams, line 468: it's hard to guess what `instruction.location.instruction!` might mean.  The whole line reads as a pile of types, suggesting there might be better names for these things.  But, hard to know with no doc.  Can you use role-based naming?

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
```diff
@@ -499,9 +499,9 @@ public struct DefiniteInitializationPass: TransformPass {
 
       // Apply the effect of the inserted instructions on the context directly.
       let consumer = InstructionID(
-        function: id.function,
-        block: id.block,
-        address: module[function: id.function][id.block].instructions.address(before: id.address)!)
+        id.function,
+        id.block,
+        module[id.function][id.block].instructions.address(before: id.address)!)
```
dabrahams, line 504: There ought to be a more expressive way to say this.  Something like

```swift
let consumer = id.prior(in: module)
```

or something?  It would be much clearer with a role-based name for `id`.

Sources/Compiler/IR/Analysis/DominatorTree.swift
```diff
@@ -36,7 +36,7 @@ struct DominatorTree {
 
   /// Creates the dominator tree of the specified function.
   init(function functionID: Function.ID, cfg: ControlFlowGraph? = nil, in module: Module) {
-    let function = module[function: functionID]
+    let function = module[functionID]
```
dabrahams, line 39:
```suggestion
    let subject = module[functionID]
```

Sources/Compiler/IR/Analysis/DominatorTree.swift
```diff
@@ -106,7 +106,7 @@ struct DominatorTree {
   func dominates(definition: InstructionID, use: Use, in module: Module) -> Bool {
     // If `definition` is in the same block as `use`, check which comes first.
     if definition.block == use.user.block {
-      for i in module[function: definition.function][definition.block].instructions.indices {
+      for i in module[definition.function][definition.block].instructions.indices {
```
dabrahams, line 109: If we're doing stuff like this a lot, maybe we should have `module[blockOf: definition]` or something.

Sources/Compiler/IR/InstructionID.swift
```diff
@@ -1,30 +1,30 @@
-/// A stable identity denoting an instruction in a module.
+/// The stable identity of an instruction in its module.
 ///
 /// - SeeAlso: `InstructionIndex`
 public struct InstructionID: Hashable {
 
-  /// The index of the function in which the instruction resides.
+  /// The function containing this identity.
```
dabrahams, line 6:
```suggestion
  /// The function containing the instruction.
```

or if you must,

```suggestion
  /// The function containing the identified instruction.
```

What you wrote here suggests that you can query the function and find this identity stored therein.

Sources/Compiler/IR/InstructionID.swift
```diff
@@ -1,30 +1,30 @@
...
   public let function: Module.Functions.Index
 
-  /// The address of the block in which the instruction resides.
+  /// The block containing this identity.
```
dabrahams, line 9:
```suggestion
  /// The block containing the instruction.
```

Sources/Compiler/IR/InstructionID.swift
```diff
@@ -1,30 +1,30 @@
...
   public let block: Function.Blocks.Address
 
-  /// The address of the instruction.
+  /// The identity of the instruction in its block.
```
dabrahams, line 12: "the instruction" works fine here; why not above?

Sources/Compiler/IR/InstructionID.swift
```diff
@@ -1,30 +1,30 @@
...
   public let address: Block.Instructions.Address
 
-  /// Creates a path with the given properties.
+  /// Creates an identity with the given properties.
```
dabrahams, line 15:
```suggestion
  /// Creates an instance with the given properties.
```
“creates an identity” makes me think it's doing more than just creating an instance of this type, and I wonder what else is going on.  When I see "Creates an instance” I am able to skip over it as the standard phrase for “makes one of these.”  Does that make sense?

Sources/Compiler/IR/InstructionID.swift
```diff
@@ -1,30 +1,30 @@
...
+    _ address: Block.Instructions.Address
   ) {
     self.function = function
     self.block = block
     self.address = address
   }
 
-  /// Creates an index with the given properties.
-  public init(block: Block.ID, address: Block.Instructions.Address) {
+  /// Creates an identity with the given properties.
```
dabrahams, line 26:
```suggestion
  /// Creates an instance with the given properties.
```

But I don't think this comment is accurate; it has no property of type `Block.ID`.

Sources/Compiler/IR/InstructionIndex.swift
```diff
@@ -8,28 +8,28 @@
 /// - SeeAlso: `InstructionIdentity`
 public struct InstructionIndex: Hashable {
 
-  /// The index of the function in which the instruction resides.
+  /// The function containing this index.
```
dabrahams, line 11: same problem.  A thing doesn't contain its own indices, in general.

```suggestion
  /// The function in the module where this position is located.
```
or

```suggestion
  /// The function part of this position.
```

Or even,

```suggestion
  /// The function containing this position.
```

It's subtle, but this is the difference between the abstraction “position” and “index,” which indicates a datum.  The former works in that context, but the latter does not.

Sources/Compiler/IR/InstructionIndex.swift
```diff
@@ -8,28 +8,28 @@
...
   public let function: Module.Functions.Index
 
-  /// The address of the block in which the instruction resides.
+  /// The block containing this index.
```
dabrahams, line 14:
```suggestion
  /// The block in the module where this position is located.
```
or

```suggestion
  /// The function part of this position.
```

Sources/Compiler/IR/InstructionIndex.swift
```diff
@@ -8,28 +8,28 @@
...
+    _ index: Block.Instructions.Index
   ) {
     self.function = function
     self.block = block
     self.index = index
   }
 
   /// Creates an index with the given properties.
-  public init(block: Block.ID, index: Block.Instructions.Index) {
+  public init(_ block: Block.ID, _ index: Block.Instructions.Index) {
```
dabrahams, line 32: comment is not quite accurate, as per similar remark above.

Sources/Compiler/IR/Module.swift
```diff
@@ -206,41 +208,48 @@ public struct Module {
   /// The instruction is inserted before the instruction currently at `position`. You can pass a
   /// "past the end" position to append at the end of a block.
   @discardableResult
-  mutating func insert<I: Instruction>(_ newInstruction: I, at position: InstructionIndex)
-    -> [Operand]
-  {
+  mutating func insert<I: Instruction>(
+    _ newInstruction: I,
+    at position: InstructionIndex
```
dabrahams, line 213:
```suggestion
    at insertionPoint: InstructionIndex
```

role-based naming

Sources/Compiler/IR/Module.swift
```diff
@@ -206,41 +208,48 @@ public struct Module {
...
       })
   }
 
   /// Inserts `newInstruction` before the instruction identified by `id` and returns the identities
   /// of its results.
   @discardableResult
-  mutating func insert<I: Instruction>(_ newInstruction: I, before id: InstructionID) -> [Operand] {
+  mutating func insert<I: Instruction>(
+    _ newInstruction: I,
+    before id: InstructionID
```
dabrahams, line 229:
```suggestion
    before successor: InstructionID
```

role-based naming

Sources/Compiler/IR/Module.swift
```diff
@@ -206,41 +208,48 @@ public struct Module {
...
       })
   }
 
   /// Inserts `newInstruction` after the instruction identified by `id` and returns the identities
   /// of its results.
   @discardableResult
-  mutating func insert<I: Instruction>(_ newInstruction: I, after id: InstructionID) -> [Operand] {
+  mutating func insert<I: Instruction>(
+    _ newInstruction: I,
+    after id: InstructionID
```
dabrahams, line 245:
```suggestion
    after predecessor: InstructionID
```

role-based naming

Sources/Compiler/IR/Operands/Instruction/CallInstruction.swift
```diff
@@ -61,7 +61,7 @@ public struct CallInstruction: Instruction {
         // Operand of a `let` parameter must be a borrow or a constant.
         switch operands[i] {
         case .result(let id, _):
-          if let instruction = module[instruction: id] as? BorrowInstruction {
+          if let instruction = module[id] as? BorrowInstruction {
```
dabrahams, line 64: Arguably `guard` can help readability in these situations by reducing nesting (also below).

dabrahams (approved):
Please see inline comments.  Some of these things may be English subtleties, but I think it's more about speaking in terms of the abstractions the reader is using.  Let me know if you have questions!

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
kyouko-taiga: The doc is inherited from the protocol to which the pass conforms.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
kyouko-taiga: That's because this pass is an interpreter ;)

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
kyouko-taiga: The first `instruction` is the name of function parameter. I guess we could rename it to `deallocStack`. `location` is the role of the property. The second `instruction` is an application of your pattern fro extracting payload out of enums.

Sources/Compiler/IR/Analysis/DefiniteInitializationPass.swift
kyouko-taiga: Sounds like a good idea. Can you open a issue?

Sources/Compiler/IR/InstructionID.swift
kyouko-taiga: `Block.ID` is just a bundle for the properties. I think that is fine. I really want a way to document member wise-like initializers without ceremony.

Sources/Compiler/IR/Operands/Instruction/CallInstruction.swift
kyouko-taiga: `guard` failed me more than I can count during debugging. We're not on good terms anymore.

PR #144, opened 2022-12-11
Use typed nodes in the emitter

lucteo:
We want to easily use the AST nodes after they are typed-checked. Going from the node IDs to their typed information is not as smooth as we would expect.

in this PR we use Typed nodes for the Emitter code.


Sources/Compiler/IR/Emitter.swift
```diff
@@ -67,31 +67,30 @@ public struct Emitter {
...
+    let explicitCaptures = decl.explicitCaptures
     for (i, capture) in explicitCaptures.enumerated() {
-      locals[capture] = .parameter(block: entryID, index: i)
+      locals[capture] = Operand.parameter(block: entryID, index: i)
     }
 
-    let implicitCaptures = program.implicitCaptures[declID]!
-    for (i, capture) in implicitCaptures.enumerated() {
-      locals[capture.decl] = .parameter(block: entryID, index: i + explicitCaptures.count)
+    for (i, capture) in decl.implicitCaptures!.enumerated() {
```
dabrahams, line 77: @kyouko-taiga I wonder if property maps exposing arrays should (in general) be rejiggered to yield an empty array when the property is not set.  That would doubtless cut down on some of the unwrapping.

Sources/Compiler/TypedNode.swift
```diff
@@ -138,9 +143,57 @@ extension TypedNode where ID: ExprID {
...
+
+    /// Domain is implicit; the expression denotes a type member.
+    case implicit
+
+    /// Domain is a value expression or a type identifier.
+    case expr(TypedNode<AnyExprID>)
+
+  }
+
+  /// The domain of the name, if it is qualified.
```
dabrahams, line 159: I realize this is a copy/paste, but is it accurate, @kyouko-taiga ? What does `none` mean if not unqualified?  This property isn't an optional!

Sources/Compiler/TypedNode.swift
```diff
@@ -138,9 +143,57 @@ extension TypedNode where ID: ExprID {
 }
 
 extension TypedNode where ID == NodeID<NameExpr> {
+  enum Domain: Equatable {
```
dabrahams, line 146: Needs a doc comment.
@kyouko-taiga should this be called `Qualification`?

Sources/Compiler/TypedNode.swift
```diff
@@ -138,9 +143,57 @@ extension TypedNode where ID: ExprID {
...
+
+    /// No domain.
```
dabrahams, line 148: This is a vacuous doc. It should tell us what it _means_ to have no domain. @kyouko-taiga can you help?

Sources/Compiler/TypedNode.swift
```diff
@@ -138,9 +143,57 @@ extension TypedNode where ID: ExprID {
...
+    case .none:
+      return .none
+    case .implicit:
+      return .implicit
+    case .expr(let expr):
+      return .expr(whole[expr])
+    }
+  }
+
+  /// A reference to a declaration.
```
dabrahams, line 171: This is a vacuous doc.  @kyouko-taiga can you help?
I realize this is a hard one, so I'm not sure what _I_ would do.

Sources/Compiler/TypedNode.swift
```diff
@@ -138,9 +143,57 @@ extension TypedNode where ID: ExprID {
...
+  enum DeclRef: Hashable {
```
dabrahams, line 172: @kyouko-taiga are we sure this wouldn't be better as a struct containing a `TypedNode<AnyDeclID>` along with an `isMember` property?

Sources/Compiler/TypedNode.swift
```diff
@@ -192,3 +245,12 @@ extension TypedNode where ID == NodeID<FunctionDecl> {
     }
   }
 }
+
+extension TypedNode where ID == NodeID<SequenceExpr> {
+
+  /// A map from (typed) sequence expressions to their evaluation order.
```
dabrahams, line 251: Is this an intentional change?  It doesn't seem like it's part of what this PR is doing.

Tests/ValTests/EmitterTests.swift
```diff
@@ -46,7 +46,7 @@ final class EmitterTests: XCTestCase {
 
       // Emit Val's IR.
       var emitter = Emitter(program: typedProgram)
-      var irModule = emitter.emit(module: module)
+      var irModule = emitter.emit(module: typedProgram[module])
```
dabrahams, line 49: Did we go backwards again?  This one is probably not that important, so I'm just asking.

dabrahams (approved):
This mostly looks like a great improvement!  Several notes inline.

I still regret seeing `Typed<FooDecl>` everywhere; in this domain, `FooDecl` is really the appropriate type name for that thing, and the `Typed<>` wrapper is just adding noise.  I have only one other suggestion, which is an extremely marginal improvement, but, we can add 

```swift
extension Node { typealias Typed = Typed<Node> }
```
and then can use `FooDecl.Typed`.  Something to consider, anyhow.

Sources/Compiler/IR/Emitter.swift
```diff
@@ -67,31 +67,30 @@ public struct Emitter {
...
+    var implicitParameterCount = explicitCaptures.count + decl.implicitCaptures!.count
+    if let receiver = decl.receiver {
+      locals[receiver] = Operand.parameter(block: entryID, index: implicitParameterCount)
       implicitParameterCount += 1
     }
 
-    for (i, parameter) in program.ast[declID].parameters.enumerated() {
-      locals[parameter] = .parameter(block: entryID, index: i + implicitParameterCount)
+    for (i, parameter) in decl.parameters.enumerated() {
+      locals[parameter] = Operand.parameter(block: entryID, index: i + implicitParameterCount)
```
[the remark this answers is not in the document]
lucteo: For some reason, while I was working on reducing the ton of errors after merge, this is how I got rid of the first error I've seen (something about ambiguous `.parameter`). Now it seems to pass fine; probably the error was coming from somewhere else.

I'll redo this, and all other similar places.

Sources/Compiler/IR/Emitter.swift
```diff
@@ -372,30 +360,30 @@ public struct Emitter {
...
+    for item in expr.condition {
       let success = module.createBasicBlock(atEndOf: functionID)
       let failure = module.createBasicBlock(atEndOf: functionID)
       alt = failure
 
       switch item {
       case .expr(let itemExpr):
         // Evaluate the condition in the current block.
-        var condition = emitL(expr: itemExpr, withCapability: .let, into: &module)
+        var condition = emitL(expr: program[itemExpr], withCapability: .let, into: &module)
```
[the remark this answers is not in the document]
lucteo: There are still a few cases (outside of AST nodes) that still refer to NodeIDs. This is one of them, as `itemExpr` is of type `AnyExprID` (as `ConditionItem` is not a node, and hasn't changed).

The syntax `typedProgram[nodeID]` goes from a `NodeID` to a `TypedNode<ID>`.

(I wanted to also create a new `ConditionItem` to make this typed too, but it feels like we are adding more code than we are actually simplifying)

Sources/Compiler/TypedNode.swift
```diff
@@ -121,6 +121,11 @@ extension TypedNode where ID: DeclID {
   var type: AnyType {
     whole.declTypes[id]!
   }
+
+  /// The implicit captures for the declared entity.
+  var implicitCaptures: [ImplicitCapture]? {
+    whole.implicitCaptures[id]
```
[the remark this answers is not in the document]
lucteo: A great person before me named in `whole`, and I didn't want to interfere 😛 

Indeed sometimes this feels the wrong name, but I can also see where it's coming from: it represent the whole typed program, from which we are only viewing a slice. Maybe a better name would be `wholeProgram` or `program`.

Sources/Compiler/TypedNode.swift
lucteo: the following code uses it:
```
  private mutating func emitR(
    sequence expr: Typed<SequenceExpr>,
    into module: inout Module
  ) -> Operand {
    emit(.sink, foldedSequence: expr.foldedSequenceExprs!, into: &module)
  }
```

Sources/Compiler/IR/Emitter.swift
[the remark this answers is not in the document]
lucteo: Fixed

Sources/Compiler/TypedNode.swift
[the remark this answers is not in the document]
dabrahams: It _is_ the whole program, but in this context what it's used for is to supply the `properties` (or `propertyMaps`).  Name things after the abstraction they represent in context, not their types.

lucteo:
> and then can use `FooDecl.Typed`. Something to consider, anyhow.

I love this idea. I would try this as a new PR just dedicated to this. It should be pretty quick.

I would also like the requested code documentation to be addressed in a different PR. I feel that @kyouko-taiga is better suited for this, but if you want me to do it, I'm fine.

Sources/Compiler/TypedNode.swift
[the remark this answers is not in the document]
lucteo: @dabrahams : can I do this as a part of a new PR? Please let me know what you prefer, and I can easily change it. 🙏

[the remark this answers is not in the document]
dabrahams: Oh, absolutely! Always.  I approved it, didn't I?

dabrahams:
I like programSent from my iPhoneOn Dec 11, 2022, at 2:09 PM, Lucian Radu Teodorescu ***@***.***> wrote:﻿
@lucteo commented on this pull request.



In Sources/Compiler/TypedNode.swift:
> @@ -121,6 +121,11 @@ extension TypedNode where ID: DeclID {
   var type: AnyType {
     whole.declTypes[id]!
   }
+
+  /// The implicit captures for the declared entity.
+  var implicitCaptures: [ImplicitCapture]? {
+    whole.implicitCaptures[id]

The main use of this member is to get typed nodes (i.e., children of typed nodes) from their IDs (from a logic perspective). In terms of usage count we use whole 17 times for the logic of getting typed nodes from the whole program, and 11 times to query the properties of the program.
I can add a computed property named properties to be equal to whole, but that sounds less ideal.
My preference is to use wholeProgram (or simplified program), but I'm fine either way. Your call.

Sources/Compiler/IR/Emitter.swift
kyouko-taiga: I have mixed feelings about that. The difference between `nil` and `[]` is sometimes significant.

I think more generally we want the property maps to become total functions at some point.

kyouko-taiga: A solution might be to define `TotalASTProperty` that does the unwrapping internally.

Sources/Compiler/TypedNode.swift
kyouko-taiga: You can think of this type as extending an optional with a third case. In other words, I'm flattening `Optional<Implicit | AnyExprID>`. Perhaps we could make `Name.Domain` conform to `ExpressibleByNilLiteral` to further underline the role of this type as some form of optional.

Sources/Compiler/TypedNode.swift
kyouko-taiga: I don't have a strong opinion. `Domain` works for me, so does `Qualification`.

That said, I think `Domain` is a more common word.

kyouko-taiga:
LGTM. I will address the documentation issues about `NameExpr.Domain` in a separate PR.

Sources/Compiler/TypedNode.swift
dabrahams: Then the comment is wrong.  It should be
```suggestion
  /// The name's qualification.
```

There is no "if."

kyouko-taiga: Could you have a look at https://github.com/val-lang/val/pull/147?

I think the "if" is consistent with the way other "optional" properties have been documented. As mentioned, `NameExpr.Domain` is a kind of optional type, only with two non-nil cases.

If you still think there shouldn't be an "if", then could you suggest this edit in the other PR?

PR #146, opened 2022-12-11
Unwrap lines

dabrahams:
There are two commits here, that should be considered separately.

In the first one, I replaced all breaks after “`(`”, “`[`”, and “`,`” and before “`)`” and “`]`” with spaces and ran the swift-format again.  

In the second commit, I turned off the `respectsExistingLineBreaks` option in the swift-format and let it determine all line breaks.

I don't love _all_ of the results (particularly in long parameter—but not argument—lists), but I believe in some cases it's significantly more readable.  WDYT?

See also #145

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -2,31 +2,15 @@
 import PackageDescription
 
 /// Settings to be passed to swiftc for all targets.
-let allTargetsSwiftSettings: [SwiftSetting] = [
-  .unsafeFlags(["-warnings-as-errors"])
-]
+let allTargetsSwiftSettings: [SwiftSetting] = [.unsafeFlags(["-warnings-as-errors"])]
 
 let package = Package(
-  name: "Val",
-  platforms: [
-    .macOS(.v12)
-  ],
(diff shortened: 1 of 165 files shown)


Package.swift
```diff
@@ -8,15 +8,12 @@ let allTargetsSwiftSettings: [SwiftSetting] = [
 
 let package = Package(
   name: "Val",
-
```
kyouko-taiga, line 11: The line breaks in this file were all intentional. `Package.swift` is a kind of declarative program using the parameter list as its DSL.

kyouko-taiga (commented):
I started reviewing all changes and then I realized that I disagree with most of them. I don't think an automated approach will work well to determine when we need to add line breaks.

I have typically been quite intentional about inserting line breaks and think removing them hurts legibility.

I'd be more inclined to be more lenient when it comes to argument lists. I agree that the tower of arguments fights against the API in some places.

dabrahams:
SGTM, I'll close this.

PR #147, opened 2022-12-12
Improve the docs of 'NameExpr.Domain'

kyouko-taiga:


Sources/Compiler/AST/Expr/NameExpr.swift
```diff
@@ -1,17 +1,25 @@
 /// A name denoting an object.
 public struct NameExpr: Expr {
 
-  public enum Domain: Equatable, Codable {
+  /// The expression of a name's qualification, or absence thereof.
```
dabrahams, line 4:
```suggestion
  /// A name's qualification
```

“none” is a fine qualification.

Sources/Compiler/AST/Expr/NameExpr.swift
```diff
@@ -1,17 +1,25 @@
...
+  ///
+  /// This type is notionaly equivalent to `Optional<D>` where `D` is an enum with the cases
+  /// `implicit` and `expr(AnyExprID)`.
```
dabrahams, line 7: IMO this is explaining far too much; it makes a simple thing complicated.  And incidentally raises the question "why didn't you just use that Optional type?”

Sources/Compiler/AST/Expr/NameExpr.swift
```diff
@@ -1,17 +1,25 @@
...
+  public enum Domain: ExpressibleByNilLiteral, Equatable, Codable {
 
-    /// No domain.
+    /// Domain is absent, as in `bar`.
```
dabrahams, line 10:
```suggestion
    /// Unqualified as in `bar`.
```

Sources/Compiler/AST/Expr/NameExpr.swift
```diff
@@ -1,17 +1,25 @@
...
     case none
 
-    /// Domain is implicit; the expression denotes a type member.
+    /// Domain is implicit, as in `.bar`; the expression denotes a type member.
```
dabrahams, line 13: This one is not clear.  Is `.foo.bar` considered implicit or not?

Sources/Compiler/AST/Expr/NameExpr.swift
```diff
@@ -1,17 +1,25 @@
...
     case implicit
 
-    /// Domain is a value expression or a type identifier.
+    /// Domain is an expression, as in `foo.bar`.
```
dabrahams, line 16: is `foo.bar` supposed to be the domain or the full name in this case?  I can guess but the comment is ambiguous

Sources/Compiler/AST/Expr/NameExpr.swift
dabrahams: @kyouko-taiga inquiring minds want to know!

kyouko-taiga: `.foo.bar` will be represented as two `NameExpr`.
1. `(.foo).bar` where `(.foo)` is a explicit domain, represented by the `.expr(AnyExpr)` case
2. `.foo` where the domain is represented by the `.implicit` case.

More generally, one can think of `.` as a right-associative operator.

How would you suggest I we can document that?

Sources/Compiler/AST/Expr/NameExpr.swift
```diff
@@ -1,17 +1,22 @@
...
-  public enum Domain: Equatable, Codable {
+  /// A name's qualification
+  public enum Domain: ExpressibleByNilLiteral, Equatable, Codable {
 
-    /// No domain.
+    /// Unqualified as in `bar`.
     case none
 
-    /// Domain is implicit; the expression denotes a type member.
+    /// Implicit, as in `.bar`; the expression denotes a type member.
```
dabrahams, line 10:
```suggestion
    /// Implicit as the `.` in `.bar`; the whole name denotes a type member.
```

dabrahams, line 10:
```suggestion
    /// Implicit as the `.` in `.bar`; the whole name denotes a type member.
```

Sources/Compiler/AST/Expr/NameExpr.swift
dabrahams: See the comments I added.

PR #148, opened 2022-12-12
Inner typed

lucteo:
The current syntax of `Typed<AST.Node>` is ugly. We want to make it easier to use.

Instead of that syntax, introduce an inner type alias in each AST node, so that we can write `Node.Typed` to get the typed node from an AST node.


kyouko-taiga (approved):
Very nice!

dabrahams:
I agree it's an improvement.  I think I want s/`Typed`/`Typechecked`/g because typing is too ambiguous, but we can consider that another time.

PR #151, opened 2022-12-13
Make Incidental a property wrapper.

dabrahams:
Unfortunate amount of noise at the declaration site, but uses look better.  What do you think?

### DIFF
--- a/Sources/Compiler/CXX/CXXTypeExpr.swift
+++ b/Sources/Compiler/CXX/CXXTypeExpr.swift
@@ -21,7 +21,7 @@ public struct CXXTypeExpr: CustomStringConvertible {
       if type == ast.coreType(named: "Int") {
         description = "int"
       } else {
-        description = type.name.value
+        description = type.name
       }
 
     case let type as ParameterType:
--- a/Sources/Compiler/IR/Emitter.swift
+++ b/Sources/Compiler/IR/Emitter.swift
@@ -587,7 +587,7 @@ public struct Emitter {
 
     // Determine the bit width of the value.
     let bitWidth: Int
-    switch type.name.value {
+    switch type.name {
     case "Int"    : bitWidth = 64
     case "Int32"  : bitWidth = 32
     default:
--- a/Sources/Compiler/TypeChecking/TypeChecker.swift
+++ b/Sources/Compiler/TypeChecking/TypeChecker.swift
@@ -895,7 +895,7 @@ public struct TypeChecker {
 
                 switch c.base {
                 case let c as AssociatedTypeType:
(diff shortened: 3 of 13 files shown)


kyouko-taiga (commented):
I will cast a _vote blanc_ on this one.

I personally don't like property wrappers. Larger EDSLs aside, I think they only buy marginal benefits at the cost of complexity and "compiler magic". For example, putting myself in the shoes of someone coming from C++ or another language, I think the `@Incidental` wrapper combined with the `private(set)` access level significantly increase the price of the entry ticket to our code base.

Now I need to understand what `@Incidental` does, which is harder and more Swift-specific than when it was a simple generic type. I also need to understand why the property is privately mutable, which I can only figure out if I learn about how property wrappers operate. That's a lot of advanced Swift knowledge for a simple `let` property.

That being said, I agree that the use sites look better, which is also valuable for everyone. So if you think my objections do not outweigh this benefit, you can go ahead.

dabrahams:
> On Dec 13, 2022, at 5:11 AM, Dimi Racordon ***@***.***> wrote:
> 
> I personally don't like property wrappers. Larger EDSLs aside, I think they only buy marginal benefits at the cost of complexity and "compiler magic". For example, putting myself in the shoes of someone coming from C++ or another language, I think the @Incidental wrapper combined with the private(set) access level significantly increase the price of the entry ticket to our code base.
> 
> Now I need to understand what @Incidental does, which is harder and more Swift-specific than when it was a simple generic type. I also need to understand why the property is privately mutable, which I can only figure out if I learn about how property wrappers operate. That's a lot of advanced Swift knowledge for a simple let property.
> 
I agree with your assessment.  The only thing that IMO would justify this change would be if we intended to apply Incidental in many more places than we currently do.  I tried this experiment as a response to your “incidental was making the use sites more complicated” comment.

kyouko-taiga:
> I tried this experiment as a response to your “incidental was making the use sites more complicated” comment.

I thought so. Thanks!

> The only thing that IMO would justify this change would be if we intended to apply Incidental in many more places than we currently do.

Yes, I agree. It could become an idiom in our code base.

I don't think there are that many places where the whole/part relationship doesn't already do the job, though. Looking for all custom equality functions, here are the places where `Incidental` could be added:
- `SourceFile.contents`
- `TypedNode.whole`
- `DoublyLinkedList.address`

We could also use `Incidental` to implement the conformance of `Module` to `Equatable`. But we'll need a custom equality function anyway because the _order_ in which functions appear in `Module.functions` is incidental.

dabrahams:
OK, closing for now.  We always have the experiment for later reference.

dabrahams:
@kyouko-taiga FYI: https://github.com/apple/swift/issues/62543#issuecomment-1347742828

PR #155, opened 2022-12-14
Simplify the API to construct type constraints

kyouko-taiga:


Sources/Compiler/TypeChecking/ConstraintGenerator.swift
```diff
@@ -150,8 +150,7 @@ struct ConstraintGenerator {
       inferredTypes[lhs] = ^TypeVariable(node: lhs.base)
       constraints.append(
         equalityOrSubtypingConstraint(
-          inferredTypes[lhs]!,
```
dabrahams, line 153: It would be better to put trivial things like unwrapping of lines in a separate PR so it's easier to evaluate what's really being done here.

Sources/Compiler/TypeChecking/Constraints/BoundMemberConstraint.swift
```diff
@@ -18,10 +18,10 @@ struct BoundMemberConstraint: Constraint, Hashable {
 
   /// Creates an instance with the given properties.
```
dabrahams, line 19: I know it's not part of this PR, but with such an english-reading API you can almost certainly do better than this comment, FWIW.  It should be possible to describe the abstraction… and then the parameters would probably be better as single-character names and the whole signature described in one line.   There are lots of similar spots.

Sources/Compiler/TypeChecking/Constraints/Constraint.swift
```diff
@@ -35,7 +35,7 @@ extension Constraint where Self: Equatable {
 
 }
 
-/// Creates a subtyping or equality constraint.
+/// Creates a constraint requiring `l` to be either equal to or a subtype of `r`.
 func equalityOrSubtypingConstraint(
```
dabrahams, line 39: These strike me as overwrought explanations, including the name.  X is a (non-strict) subset of X, so it should be possible to say something like that for Y.  I believe this is not an `==` constraint, which the new comment implies it could be, but only a `:` constraint.

Sources/Compiler/TypeChecking/Constraints/Constraint.swift
```diff
@@ -48,3 +48,26 @@ func equalityOrSubtypingConstraint(
     ],
     because: cause)
 }
+
+/// Creates a constraint requiring `subject` to be either equal to `defaultType` or be expressible
+/// by literals associated with `trait`.
```
dabrahams, line 53: That's a pretty weird constraint, as described; maybe it's right but it's not obvious that it should be so complicated.  Why have the "either equal to" part?

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
```diff
@@ -406,32 +405,15 @@ struct ConstraintGenerator {
...
-              ],
-              penalties: 0),
-            .init(
-              constraints: [
-                ConformanceConstraint(^tau, traits: [trait], because: cause)
-              ],
-              penalties: 1),
-          ],
-          because: cause))
+        expressibleByLiteralConstraint(
```
dabrahams, line 409: Seems like a very different constraint now.  Unintended semantic change?

dabrahams (approved):
see comments inline

Sources/Compiler/TypeChecking/Constraints/Constraint.swift
kyouko-taiga: That's linked to the way constraints on literals are resolved.

Because there's no way to enumerate the set of types conforming to a trait in general, it's not possible to solve a constraint `A : P` (i.e., `A` conforms to `P`) without any additional context to further constraint `A`. When that happens, `A : P` will become "stale" during constraint solving and type inference will get stuck, resulting in a type error.

In the case where `P` is `ExpressibleBy***Literal`, the compiler can do a magic trick and try feed `A == DefaultType` into the constraint system if it's about to get stuck.

This process is documented in the solver already. I don't think it should be in the docs of this function.

Sources/Compiler/TypeChecking/Constraints/Constraint.swift
kyouko-taiga: This function does indeed create a disjunction saying either `X == Y` or `X <: Y`. It is motivated by the way subtyping constraints are resolved

In general, it is not possible to enumerate the set of types that are subtype of another type. That's because in an open-world assumption, there's an unbounded number of types that may conform to `any P`. So if all we have is `X <: Y`, we can't guess the type of `X`.

Using the disjunction is a way to try guessing `X == Y` and if that fails, look for other constraints to infer `X` and then make sure `X <: Y` holds.

Sources/Compiler/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: No change. I just refactored constraint into a function.

Sources/Compiler/TypeChecking/Constraints/Constraint.swift
dabrahams: I think you misunderstand my point.
1. The name, and the description as written, makes it sound like—depending on some `init` parameter, for example—an instance is going to either be an equality constraint or a subtyping constraint.  
2. IIUC that's not what this is doing.  Instead, IIUC this is expressing a (non-strict) subtyping constraint, which, IIUC, is the only kind of subtyping constraint in the language.
3. It happens to be implemented as a disjunction, IIUC, for reasons related to how you implement constraint resolution, but that is not part of the abstraction it represents.
4. It ought to be possible to use modifier like “(non-strict)” if you feel it's necessary to clarify what you mean by subtyping.

Sources/Compiler/TypeChecking/Constraints/Constraint.swift
dabrahams: > Because there's no way to enumerate the set of types conforming to a trait in general, it's not possible to solve a constraint A : P (i.e., A conforms to P)

Sorry, I don't even know what it means to “solve a constraint.”  But certainly it is possible for the compiler to deal with a constraint that `A` conforms to `P` without a default or this would be invalid:

```
fun f<A: P>(x: A) {}
```

Obviously I don't know the details here so could be off-base, but instinct says that in the scenario you're describing, I would always want the `defaultType` to also conform to `P`, so _at the level of abstraction of the documentation_, this would become a simple non-disjunction constraint that `A` conforms to `P`, possibly with a default or a hint or something.

Sources/Compiler/TypeChecking/Constraints/Constraint.swift
kyouko-taiga: Ah, I see. I thought the source of confusion was due to the fact that `X <: Y` holds if `X == Y` typically holds in all type systems (including Val!).

TL;DR; I can make the docs more explicit by  saying the function returns a "_disjunction_ constraint requiring ...".

It's not about strictness, because if you give `X == Y ∧ X <: Y` to the solver, it will succeed and conclude that `X` is `Y`. Instead, the abstraction is representing a choice that the solver is allowed to make. Either it can choose to solve the system assuming `X` is exactly `Y`, or it can choose to solve the system assuming `X` is `Y` or any of its subtypes. That's because one cannot enumerate "any of its subtypes" that the latter choice may cause the solver to be stuck.

To understand why that's significant, we have to think about a case where knowing `X` is necessary to infer the type of something else. For example, if we have a call `foo(x, y)`, we may need to know the exact type of `x` to infer the type of `y` and/or the type of `foo`, as the latter may be generic or overloaded.

dabrahams: > X == Y, then X <: Y typically holds in all type systems (including Val!).

I assumed since you documented them as though they were unrelated constraints, that must not be the case. Since that is the case this is **in fact** simply a subtyping constraint.  The disjunction is a detail.

> It's not about strictness

What I meant by "strictness" is given by the use “X is a strict subtype of Y iff X <: Y and X != Y.”  I was assuming, based on what you've written here, that to you "is a subtype" means “is a strict subtype” in that sense.  But now I see that's incorrect.

None of what you say is significant matters to the documentation of *what this thing is*.  It is just a constraint (lower-case "c") that requires `l` to be a subtype of `r`, and it's just an artifact of the way you implemented your solver that every time such a constraint (lower-case "c") is expressed as a `Constraint` it needs to be part of a `Disjunction` with an `EqualityConstraint` or your solver could get stuck.

kyouko-taiga: Not really. There's a type denoting (non-strict) subtyping constraints. So the documentation of the function should mark the difference between creating a constraint with `equalityOrSubtypingConstraint` or with `SubtypingConstraint.init`.

But I'm starting to think a simpler approach would be to define strict subtyping constraints.

PR #160, opened 2022-12-15
Simplify member constraints

kyouko-taiga:


dabrahams (approved):
The use sites are not obviously improved, so I'm not sure why you wan to do this, but won't stand in the way.

PR #168, opened 2022-12-16
Factor out Core module.

dabrahams:


Package.swift
```diff
@@ -48,13 +48,24 @@ let package = Package(
...
         .product(name: "Durian", package: "Durian"),
         .product(name: "BigInt", package: "BigInt"),
       ],
       swiftSettings: allTargetsSwiftSettings),
 
+    .target(
+      name: "Core",
+      dependencies: [
+        "Utils",
+//        .product(name: "Collections", package: "swift-collections"),
```
kyouko-taiga, line 63: Why are these commented out? The dependencies are not needed?

Sources/Core/AST/AST+Diagnostics.swift
```diff
@@ -2,23 +2,23 @@ import Utils
 
 extension Diagnostic {
 
-  static func diagnose(
+  public static func diagnose(
```
kyouko-taiga, line 5: I'm surprised these diagnostics had to become public. I think most of these are parser-related and could be kept internal in the corresponding target.

Others are produced by `AST.validateTypeMember` and could be kept internal in `Code`. It's possible some diagnostics are issued by the parser and the AST validators, but I can't say which on the top of my head.

Sources/Core/Constraints/BoundMemberConstraint.swift
```diff
@@ -1,22 +1,22 @@
 /// A constraint `L.m == R` specifying that `L` has a member of type `R` named `m`.
-struct MemberConstraint: Constraint, Hashable {
+public struct MemberConstraint: Constraint, Hashable {
```
kyouko-taiga, line 2: Instinctively I would have put the type constraints in the front end and kept them internal? Was there a reason it can't be the case?

kyouko-taiga (commented):
I left a handful of inline comments

Package.swift
dabrahams: Yes; I'll eliminate

Sources/Core/AST/AST+Diagnostics.swift
dabrahams: > I'm surprised these diagnostics had to become public.

I'm not sure they had to. I am not entirely smart about where things should live and so may not have chosen the optimal division.  Do you want to suggest I put these in a different module?

Sources/Core/Constraints/BoundMemberConstraint.swift
dabrahams: `ExistentialType` has a `ConstraintSet`, at least.

Sources/Core/AST/AST+Diagnostics.swift
kyouko-taiga: Yes. Here's how I would proceed.
1. Place all of them in `FrontEnd` (where the parser is), with level `internal`.
2. Put back the functions that are requested by the AST validators in `Core`, with level `internal`.
3. Change the level of any function still requested by the parser to `public`.

Sources/Core/Constraints/BoundMemberConstraint.swift
kyouko-taiga: Ah, right! Yes so you were right making them public. My mistake.

Sources/Core/AST/AST+Diagnostics.swift
dabrahams: I think I've [done that now](https://github.com/val-lang/val/pull/168/commits/9891bb8f53a7c4641cb9e637f869a92eafd3dee0). PTAL

Sources/Core/Constraints/BoundMemberConstraint.swift
dabrahams: It's possible I could have left the specific constraint models in the FrontEnd module.  WDYT?

dabrahams: The precondition in `ExistentialType.init` makes that a bit tricky, so I think we should (at least) put this off to a later commit.

kyouko-taiga: Agreed

kyouko-taiga:
Well done!

PR #171, opened 2022-12-17
Improve type inference for integer literals

kyouko-taiga:
The PR allows the type checker to infer the type of an integer literal based on the result of an overload constraint. For example:

```
fun fn(x: Int) -> Int {}
fun fn(x: Double) -> Double {}

public fun main() {
  let _ : Double = fn(x: 1)
}
```

The solver is allowed to infer `1` has type `Double` after overload resolution concludes `fn` has type `[] (x: Double) -> Double`.

### DIFF
--- a/Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
+++ b/Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
@@ -367,7 +367,7 @@ struct ConstraintGenerator {
     let trait = checker.program.ast.coreTrait(named: "ExpressibleByIntegerLiteral")!
     let cause = ConstraintCause(.literal, at: checker.program.ast[id].origin)
 
-    // Constrain the type of the literal to be conforming to `ExpressibleByIntegerLiteral` or,
+    // Constrain the type of the literal to conform to `ExpressibleByIntegerLiteral` or,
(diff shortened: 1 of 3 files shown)


Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -367,25 +367,18 @@ struct ConstraintGenerator {
     let trait = checker.program.ast.coreTrait(named: "ExpressibleByIntegerLiteral")!
     let cause = ConstraintCause(.literal, at: checker.program.ast[id].origin)
 
-    switch expectedTypes[id]?.base {
-    case let tau as TypeVariable:
-      // The type of the expression is a variable, possibly constrained elsewhere; constrain it to
-      // either be `Int` or conform to `ExpressibleByIntegerLiteral`.
+    // Constrain the type of the literal to be conforming to `ExpressibleByIntegerLiteral` or,
```
dabrahams, line 370:
```suggestion
    // Constrain the type of the literal to conform to `ExpressibleByIntegerLiteral` or,
```

dabrahams (approved):
What do I know.  It got smaller, so it must be OK ;-)

PR #177, opened 2022-12-20
Refactor constraint generation

kyouko-taiga:


Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -14,8 +14,8 @@ struct ConstraintGenerator {
     /// The set of type constraints being generated.
     let constraints: [Constraint]
 
-    /// The diagnostics of the errors the generator encountered.
-    let diagnostics: [Diagnostic]
+    /// Indicates that the solver encountered one or more errors.
+    let didFoundError: Bool
```
dabrahams, line 18:
```suggestion
    let foundError: Bool
```
Although the above name is slightly ambiguous; it could be the error that was found.

```suggestion
    let errorFound: Bool
```

doesn't have that problem.

This is the old Cocoa style which seems pointless to me.
```suggestion
    let didFindError: Bool
```

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -14,8 +14,8 @@ struct ConstraintGenerator {
     /// The set of type constraints being generated.
     let constraints: [Constraint]
 
-    /// The diagnostics of the errors the generator encountered.
-    let diagnostics: [Diagnostic]
+    /// Indicates that the solver encountered one or more errors.
```
dabrahams, line 17: And how does it indicate that? By being present?  By being `false`? ;-)

```suggestion
    /// True iff the solver encountered an error.
```

or if you must,

```suggestion
    /// True iff the solver encountered one or more errors.
```

or if you *truly* must

```suggestion
    /// Whether the solver encountered one or more errors.
```

dabrahams: You don't tell me what "error" means in this context.  Without qualification, in programming, "error" usually means "postcondition failure" but I'm guessing this error is something more like a contradiction.  Why not use a more specific word?

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -193,53 +170,43 @@ struct ConstraintGenerator {
...
-    // Nothing to do here.
-  }
-
-  private mutating func visit(
-    floatLiteral id: NodeID<FloatLiteralExpr>,
-    using checker: inout TypeChecker
-  ) {
-    fatalError("not implemented")
+  ) -> AnyType {
+    assumeIsError(id)
```
dabrahams, line 200: I realize this is an idiom throughout and not introduced by this PR, but I don't know what "assume" means in this context.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -108,28 +94,21 @@ struct ConstraintGenerator {
...
-    assume(typeOf: id, equals: boolType, at: checker.program.ast[id].origin)
-  }
-
-  private mutating func visit(
-    bufferLiteral id: NodeID<BufferLiteralExpr>,
-    using checker: inout TypeChecker
-  ) {
-    fatalError("not implemented")
+  ) -> AnyType {
+    assume(
```
dabrahams, line 98: `assume` is undocumented so I can't evaluate its use sites.  However, I find it "weird to be saying assume <some AST node> is <some type>."  I know we use `x is T` at the language level, but I'd rather read `hasType` here (and maybe in the language too) because `is` can mean equality

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -78,28 +78,14 @@ struct ConstraintGenerator {
       return visit(integerLiteral: NodeID(rawValue: expr.rawValue), using: &checker)
     case LambdaExpr.self:
       return visit(lambda: NodeID(rawValue: expr.rawValue), using: &checker)
-    case MapLiteralExpr.self:
```
dabrahams, line 81: Just curious as to what caused all these deletions.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -108,28 +94,21 @@ struct ConstraintGenerator {
...
-
-  private mutating func visit(
-    bufferLiteral id: NodeID<BufferLiteralExpr>,
-    using checker: inout TypeChecker
-  ) {
-    fatalError("not implemented")
+  ) -> AnyType {
+    assume(
+      id, is: checker.program.ast.coreType(named: "Bool")!,
+      at: checker.program.ast[id].origin)
```
dabrahams, line 100: Couldn't assume get the origin from its first parameter?  Presumably the ConstraintGenerator could store the program directly?

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -253,38 +220,35 @@ struct ConstraintGenerator {
...
-      assignToError(id)
-      return
+    if calleeType.isError {
+      return assumeIsError(id)
     }
 
     // Case 2
-    if inferredTypes[callee]!.base is TypeVariable {
+    if calleeType.base is TypeVariable {
       let parameters = visit(arguments: checker.program.ast[id].arguments, using: &checker)
```
dabrahams, line 229: Are these parameters or arguments?  The code here makes it unclear that you know which you're dealing with.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -253,38 +220,35 @@ struct ConstraintGenerator {
...
       let returnType = expectedTypes[id] ?? ^TypeVariable(node: AnyNodeID(id))
 
       constraints.append(
         FunctionCallConstraint(
-          inferredTypes[callee]!, takes: parameters, andReturns: returnType,
+          calleeType, takes: parameters, andReturns: returnType,
```
dabrahams, line 234: I bet this is a constraint that the calleeType can accept _arguments_ of the given type… ?  I would use `canAccept` or something to indicate that these are not necessarily the types of the parameters (unless I'm confused)

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -108,28 +94,21 @@ struct ConstraintGenerator {
...
     using checker: inout TypeChecker
-  ) {
+  ) -> AnyType {
     let node = checker.program.ast[id]
 
     // Realize the type to which the left operand should be converted.
     guard var target = checker.realize(node.right, inScope: scope)?.instance else {
-      assignToError(id)
-      return
+      return assumeIsError(id)
```
dabrahams, line 111: That function is undocumented so can't evaluate.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -427,73 +380,52 @@ struct ConstraintGenerator {
...
-        visit(expr: domainExpr, using: &checker)
-        parentType = inferredTypes[domainExpr]!
+        parentType = visit(expr: domainExpr, using: &checker)
       } else {
         fatalError("not implemented")
       }
       unresolvedComponents = suffix
 
     case .done(let prefix, let suffix):
+      assert(!prefix.isEmpty)
```
dabrahams, line 426: It's always nice to add some descriptive string to an assert about what expectation is being violated, embedding some information there, so when it fires, the person who has to deal with it has a clue.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -427,73 +380,52 @@ struct ConstraintGenerator {
...
       for p in prefix {
-        constrain(p.component, to: p.candidates, using: &checker)
-        parentType = inferredTypes[p.component]
+        parentType = constrain(p.component, to: p.candidates, using: &checker)
```
dabrahams, line 428: It seems odd that you're willing to overwrite `parentType` each time through the loop. Is this really correct?

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -1712,7 +1719,7 @@ public struct TypeChecker {
 
     /// Name resolution failed. The payload contains the undefined component along with the type
     /// in which it was looked up, or `nil` if the component is the expression's first component.
-    case failed(component: NodeID<NameExpr>, parentType: AnyType?)
+    case failed
```
dabrahams, line 1722: No payload; the comment is out of date.

dabrahams (changes requested):
A few things need to be cleaned up.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -1460,16 +1466,13 @@ public struct TypeChecker {
...
-      return nil
-    }
+    infer(typeOf: expr, expecting: expectedType, inScope: scope).succeeded
+      ? exprTypes[expr]!
+      : nil
   }
 
-  /// Returns a solution describing the best guess to type `expr` and its sub-expressions.
+  /// Returns whether the solver succeeded checking the type constraints implied by `expr` along
+  /// with the solution describing the best guess for its type and that of its sub-expressions.
```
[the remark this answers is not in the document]
kyouko-taiga: No, because if inference failed I still want its best solution; it likely contains diagnostics that I can report to explain the failure.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -1510,7 +1517,7 @@ public struct TypeChecker {
     // Consume the solution's errors.
     diagnostics.formUnion(solution.diagnostics)
 
-    return solution
+    return (succeeded: solution.diagnostics.isEmpty, solution: solution)
   }
```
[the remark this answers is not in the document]
kyouko-taiga: Are you talking about [`TypeChecker.infer(typeOf:expecting:inScope:initialConstraints:)`](https://github.com/val-lang/val/blob/bd6c1a5ba71a7f749980276f0691b7b5ca68a614/Sources/FrontEnd/TypeChecking/TypeChecker.swift#L1483) (L1483)?

The function returns two things (return type is `(succeeded: Bool, solution: Solution)`).
The comment says "Returns whether [...] along with [...]."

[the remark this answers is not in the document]
dabrahams: Not the previous function, the [following one](https://github.com/val-lang/val/blob/bd6c1a5ba71a7f749980276f0691b7b5ca68a614/Sources/FrontEnd/TypeChecking/TypeChecker.swift#L1523-L1530).

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -53,23 +53,23 @@ struct ConstraintGenerator {
 
   /// Applies `self` to generate constraints using `checker` to resolve names and realize types.
   mutating func apply(using checker: inout TypeChecker) -> Result {
```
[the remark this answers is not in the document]
kyouko-taiga: I'm hinting at the fact that `self` is notionally a function (its initializer can be understood as building the capture list). So I used "applies" to mean "applies the notional function represented by `self`". Would that be a better comment? Do you have another suggestion?

FWIW, I have some insider information that methods of the `ConstraintGenerator` will migrate to an extension of `TypeChecker` very soon.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -193,53 +170,43 @@ struct ConstraintGenerator {
     switch checker.program.ast[id].success {
     case .expr(let thenExpr):
       expectedTypes[thenExpr] = expectedTypes[id]
-      visit(expr: thenExpr, using: &checker)
-      inferredType = inferredTypes[thenExpr]
+      inferredType = visit(expr: thenExpr, using: &checker)
 
     case .block(let thenBlock):
-      _ = checker.check(brace: thenBlock)
+      if !checker.check(brace: thenBlock) { didFoundError = true }
```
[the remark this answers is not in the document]
kyouko-taiga: Really?

`brace` is for `NodeID<BraceStmt>`, which is what the "then" block of a conditional statement is. I'm open to suggestions.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: Sounds good.

As for the language level, it's a good discussion to open. Note that Val currently doesn't have an `is` operator.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: There were dead artifacts of the time the constraint generator used to be an AST visitor.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: Yes, that's in the pipes for a future patch.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: All of these variables are properly named. We're guessing the type of the parameters from the bottom-up information that we can gather from the arguments. The docs of `visit(arguments:using:)` is explaining the same thing.

I'm open to suggestions if more is needed at the call site.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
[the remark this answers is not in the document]
dabrahams: I'd just say "Returns the constraints for _______ using `checker` to resolve…”
This notional function business is info the reader doesn't need at this point.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: Yes, this is the parent type of the last visited component.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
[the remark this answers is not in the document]
dabrahams: Terrible label then!  It's not a single brace, but a BraceStmt.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
dabrahams: Part of the problem here is that `visit` is a huge missed opportunity to communicate something about what it does.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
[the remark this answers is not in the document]
kyouko-taiga: Noted ;)

All `check` methods are labeled `blah` for `BlahStmt`, so that just follows the pattern. Let's revisit this choice in another patch.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
dabrahams: OK, if I was confused, then the constraint is misnamed and what you really have here is a `FunctionTypeConstraint`.  And I would replace `takes` with `parameters` in this case, just to avoid confusion about whether you're documenting the types it can accept or its concrete argument types.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -805,7 +691,7 @@ struct ConstraintGenerator {
 
     // Check that the labels inferred from the callee are consistent with that of the call.
     if argumentLabels != parameterLabels {
-      diagnostics.append(
+      checker.addDiagnostic(
```
dabrahams, line 694: You probably can't do anything about it in this PR, but I'd rather be reading

```swift
checker.diagnose(labels: argumentLabels, incompatibleWith: parameterLabels)
```

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -34,8 +34,8 @@ struct ConstraintGenerator {
   /// The set of type constraints being generated.
   private var constraints: [Constraint] = []
 
-  /// The diagnostics of the errors the generator encountered.
-  private var diagnostics: [Diagnostic] = []
+  /// Indicates that the solver encountered one or more errors.
+  private var didFoundError = false
```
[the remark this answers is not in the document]
dabrahams: Still ungrammatical and the documentation style is at least not like what we do everywhere else

dabrahams (approved):
Approving, but it looks like some of my earlier remarks were ignored.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
```diff
@@ -51,25 +51,28 @@ struct ConstraintGenerator {
...
+    _ = visit(expr: subject, using: &checker)
     return Result(
       inferredTypes: inferredTypes,
       constraints: constraints,
-      diagnostics: diagnostics)
+      didFoundError: foundConflict)
   }
 
-  private mutating func visit(expr: AnyExprID, using checker: inout TypeChecker) {
+  /// Returns the variable assigned to `expr` in the generated contraints, using `checker` to
```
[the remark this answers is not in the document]
kyouko-taiga: Sorry, I missed that there were two occurrences of `didFoundError`. I replaced by
```swift
/// True iff a constraint could not be solved.
  private var foundConflict = false
```

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: Sorry, I missed that there were two occurrences of properties with this name. I replaced (now instances) by:
```swift
/// True iff a constraint could not be solved.
let foundConflict: Bool
```

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: The property is no called `foundConflict` and the docs don't mention "errors" anymore.

I have to say, though, that it think it's reasonable to use the term "error" in the source code of a type checker to mean something else than a programming error of the type checker itself. I would not expect a component named "TypeChecker" or "ConstraintSolver" to set a flag when it encounters an error in its own implementation.

Sources/FrontEnd/TypeChecking/ConstraintGenerator.swift
kyouko-taiga: I'm afraid I didn't fully understand your comment. Please open an issue if you want to fix this in another patch.

dabrahams:
> On Dec 21, 2022, at 4:01 PM, Dimi Racordon ***@***.***> wrote:
> 
> 
> @kyouko-taiga commented on this pull request.
> 
> I have to say, though, that it think it's reasonable to use the term "error" in the source code of a type checker to mean something else than a programming error of the type checker itself.
> 
Of course it is, but without qualification it could still mean one of several things, e.g. an inconsistency in a particular constraint resolution path or a hard error that needs to be reported to the user.
> I would not expect a component named "TypeChecker" or "ConstraintSolver" to set a flag when it encounters an error in its own implementation.
> 
That's a “bug,” (usually precondition violation) which I never classify as an “error."  In Swift, an error is a response to a recoverable postcondition failure; it’s the thing that gets thrown.

PR #178, opened 2022-12-20
Remove needless copy initializer

kyouko-taiga:
It seems like `ConcreteNodeID` was requiring a copy initializer. Since all types are copyable in Swift, it should be sufficient to just use copy-initialization with `=`, as shown with the suggested change in `TypedNode.swift`.


dabrahams (approved):
of course!

PR #191, opened 2022-12-27
Format check

lucteo:
Add CI check to verify the formatting for the Swift code.

Add script and instructions on how to run swift-format for development.


dabrahams:
I don't think we can make this a requirement for merge until we have all of Dimitri's editor integrations. Is that done?

kyouko-taiga:
Unfortunately, it seems like the simple approach I knew to integrate formatting tools to Xcode only works with proper Xcode projects, not projects managed with SPM. So AFAIU, our best option would be to create an Xcode extension.

That doesn't seem so hard, but it's work.

kyouko-taiga (approved):
LGTM

lucteo:
@dabrahams : are you also good with this?

dabrahams:
If Dimi is happy so am ISent from my iPhoneOn Dec 30, 2022, at 9:55 AM, Lucian Radu Teodorescu ***@***.***> wrote:﻿
@dabrahams : are you also good with this?

PR #193, opened 2022-12-31
Use typed nodes in the IR code

lucteo:
IR code used in a few occasions untyped nodes. Use typed nodes instead.

This is a continuation of the effort to use typed nodes instead of pairing AST nodes with TypedProgram.

### DIFF
--- a/Sources/Core/TypedNode.swift
+++ b/Sources/Core/TypedNode.swift
@@ -135,15 +135,9 @@ extension TypedNode where ID: DeclID {
   }
 
   /// The type of the declared entity.
-  /// In most cases, the type is valid, so we extract the value from the optional by default.
+  /// If the declaration type is invalid, we return `.error` type.
   public var type: AnyType {
-    program.declTypes[id]!
-  }
-
-  /// The type of the declared entity, as optional
-  /// This is to be used in cases in which the type is not always valid.
-  public var typeOpt: AnyType? {
-    program.declTypes[id]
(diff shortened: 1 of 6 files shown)


Sources/Core/TypedNode.swift
```diff
@@ -135,10 +135,17 @@ extension TypedNode where ID: DeclID {
   }
 
   /// The type of the declared entity.
+  /// In most cases, the type is valid, so we extract the value from the optional by default.
```
kyouko-taiga, line 138: I feel like this comment should be an implementation note in the body of this accessor.

Sources/Core/TypedNode.swift
```diff
@@ -135,10 +135,17 @@ extension TypedNode where ID: DeclID {
...
   public var type: AnyType {
     program.declTypes[id]!
```
kyouko-taiga, line 140: If we do as below, we don't need forced unwrapping anymore and we can represent failures to type check an expression with the error type.

```suggestion
    program.declTypes[id] ?? .error
```

kyouko-taiga (changes requested):
Good work.

I think we can avoid exposing the type of a typed expression node as an optional.

Sources/Core/TypedNode.swift
lucteo: Makes a lot of sense. Will fix.

kyouko-taiga (approved):
LGTM

PR #194, opened 2022-12-31
Finish translation of the factorial example to C++

lucteo:
We need to have at least a simple program translatable to C++. This PR will complete the work to ensure that the factorial example is translated to C++ code.

With this, we emit C++ code for most of the Val expressions (actually for most AST nodes). This only treats the happy path, and there are flows that need to be improved, but we have a starting point.

Testing: No new tests were added here. Just make sure that the FactorialExample.val translates to proper C++. More tests will be added in the future.


Sources/CodeGen/CXX/Expr/CXXConditionalExpr.swift
```diff
@@ -0,0 +1,24 @@
...
+  public let condition: CXXRepresentable
+  /// The expression to be returned if the condition is `true`.
+  public let trueExpr: CXXRepresentable
+  /// The expression to be returned if the condition is `false`.
+  public let falseExpr: CXXRepresentable
+
+  /// The original node in Val AST.
+  let original: CondExpr.Typed
+
+  func writeCode<Target: TextOutputStream>(into target: inout Target) {
```
kyouko-taiga, line 16: Note that a conditional expr in Val may not always be transpiled as a ternary expression in C++.
- The condition might involve optional binding.
- The body of the `then` or `else` blocks might be composed of multiple statement.

Sources/CodeGen/CXX/Expr/CXXThisExpr.swift
```diff
@@ -0,0 +1,13 @@
+import Core
+
+/// A C++ `this` expression.
+struct CXXThisExpr: CXXRepresentable {
```
kyouko-taiga, line 4:
```suggestion
struct CXXReceiverExpr: CXXRepresentable {
```

kyouko-taiga (approved):

Sources/CodeGen/CXX/Expr/CXXConditionalExpr.swift
lucteo: Yes, that is true. I thought of this, and ways of solving this. I wanted to fix this after this PR, when I need to walk all Val AST nodes and ensure we cover all possible alternatives.

(In this case, the assumption is that for a ternary operator in C++, we always generate it from an conditional expression in Val. If that would change in the future, I can change this code).

PR #195, opened 2023-01-02
explicitly include `Utils`

compnerd:
Swift 6 will disallow use of modules without an explicit include in the source file.  Add the missing includes in a few sites that trigger this issue.

### DIFF
--- a/Sources/Core/Types/ExistentialType.swift
+++ b/Sources/Core/Types/ExistentialType.swift
@@ -1,3 +1,5 @@
+import Utils
+
 /// An existential type, optionally bound by traits and constraints on associated types.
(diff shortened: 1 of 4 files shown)


compnerd:
CC: @dabrahams
I was able to build on Windows with a recent toolchain snapshot with this set of change(s).

dabrahams:
Happy new year to you too!!

compnerd:
Happy new year @dabrahams!

dabrahams:
Hey @compnerd would you be willing to help us get Windows CI set up?  Thanks!

compnerd:
#198 takes care of that :)

PR #198, opened 2023-01-03
workflows: add Windows to the matrix

compnerd:
Add Windows to the matrix for the CI validation.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -8,16 +8,28 @@ on:
 
 jobs:
   build:
-    name: Swift ${{ matrix.swift }} on ${{ matrix.os }}
+    name: Swift ${{ matrix.swift.version }} on ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
-        os: [macos-12, ubuntu-22.04]
-        swift: ["5.7"]
+        os: [macos-12, ubuntu-22.04, windows-latest]
+        swift: [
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @dabrahams
As requested, one semi-working Windows CI integration.

The test failure _feels_ like it is related to newline shenanigans.  I cannot easily reproduce it locally as I am running a newer version of Swift:
```
>swift --version                                                                                      compnerd.org Swift version 5.8-dev (LLVM d4258b1fff2b41b, Swift 3d3611356ccf83c)                                        Target: x86_64-unknown-windows-msvc
```

I suspect that the changes may have silently resolved the issue, but it likely is good to make the tests more robust.

compnerd:
Turns out that I have the workaround that I have enlisted here.  The issue is definitely the newline conversions (`core.autocrlf` being set to `input` will workaround it by checking out in Unix line endings rather than WIndows line endings).

dabrahams:
@compnerd is this done, or should you perhaps switch it to draft?

compnerd:
@dabrahams this is done, I think that getting it merged is the right thing to do.  However, note that this is technically hiding some unexpected behaviour - namely `\n` and `\r\n` are treated differently.  Now, this might actually be something that is semantically intended from the language perspective and I must admit that I have not looked at the specification and cannot therefore say if that is intentional.

dabrahams:
Thanks; could you file an issue about the newlines?  We can then fix and write a test to make sure it stays that way.  @kyouko-taiga I'm guessing you must be bypassing `Character`, which accounts for this, and using UTF-8 code units instead.

PR #199, opened 2023-01-04
Simplify TestAnnotationHandler

dabrahams:
It doesn't seem like it's even an abstraction, so it could be removed entirely, but that's a bit
more of a project than I want to take on at this point.

### DIFF
--- a/Tests/ValTests/Utils/CXXAnnotationHandler.swift
+++ b/Tests/ValTests/Utils/CXXAnnotationHandler.swift
@@ -20,8 +20,8 @@ struct CXXAnnotationHandler: TestAnnotationHandler {
   init(
     module: CXXModule?,
     ranToCompletion: Bool,
-    diagnostics: [Diagnostic])
-  {
+    diagnostics: [Diagnostic]
+  ) {
     self.cxxHeader = module?.emitHeader() ?? ""
     self.cxxSource = module?.emitSource() ?? ""
     self.defaultHandler = .init(ranToCompletion: ranToCompletion, diagnostics: diagnostics)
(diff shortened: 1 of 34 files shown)


kyouko-taiga:
What's not an abstraction? The test annotation handler?

Tests/ValTests/Utils/CXXAnnotationHandler.swift
```diff
@@ -4,38 +4,27 @@ import XCTest
...
-
   /// The C++ header.
   private let cxxHeader: String
 
   /// The C++ source.
   private let cxxSource: String
 
   /// The recorded issues.
-  private var issues: [XCTIssue] = []
+  private var issues_: [XCTIssue] = []
```
kyouko-taiga, line 14: IIRC, swift-format doesn't like names ending with an underscore.

Tests/ValTests/Utils/ValTestRunner.swift
```diff
@@ -15,22 +15,17 @@ extension ValTestRunner {
   /// Runs the test cases found in `Self.testCaseDirectoryPath` with `runSteps` and handle the
   /// result of each case with an instance of `handler`.
   func runValTests<Handler: TestAnnotationHandler>(
-    handlingResultsWith hander: Handler.Type,
-    _ runSteps: (_ name: String, _ source: SourceFile) -> Handler.Configuration
+    _ runSteps: (_ xcTestName: String, _ source: SourceFile) throws -> Handler
```
kyouko-taiga, line 18: Nice change!

kyouko-taiga (approved):
Approved, but I think there are some swift-format issues.

Tests/ValTests/Utils/CXXAnnotationHandler.swift
dabrahams: That's very silly; they catch it as not obeying lowerCamelCase, as though it's somehow snake_case to end with an underscore.  Fortunately I can start with one.

dabrahams:
Yes, when all the complexity is unwound `TestAnnotationHandler` seems like it's ultimately just a function taking test annotations and returning XCTest issues.  Maybe that's an abstraction; I'm not entirely sure.

PR #206, opened 2023-01-06
Nix TestFileAnnotationHandlers

dabrahams:
The annotations you can handle (aside from the diagnostic annotations) are directly tied to the processing passed to `checkAnnotatedValFiles`, so they shouldn't be separated.


Tests/ValTests/Utils/AnnotatedValFileTest.swift
```diff
@@ -2,13 +2,53 @@ import Core
...
+        return []
+      }
+    )
+  }
+
+  /// Applies `processAndCheck` to each ".val" file in `sourceDirectory` (a path relative to the
+  /// ValTests/ directory of this project) and all the annotations in that file matching the given
+  /// commands, adding any resulting XCTest failures to those that would be produced by
+  /// `checkAnnotatedValFileDiagnostics`.
+  func checkAnnotatedValFiles(
```
kyouko-taiga, line 44: I have a really hard time understanding the API of this function. What are the roles of the closure's parameters?

Tests/ValTests/Utils/AnnotatedValFileTest.swift
```diff
@@ -21,20 +61,77 @@ extension XCTestCase {
         if url.pathExtension != "val" { return true }
 
         let source = try SourceFile(contentsOf: url)
-        let issues = try XCTContext.runActivity(
+        var annotations = TestAnnotation.parseAll(from: source)
+        let p = annotations.partition(by: { checkedCommands.contains($0.command) })
```
kyouko-taiga, line 65: That code looks plausible, but I would not mind a comment explaining what it's trying to achieve. What are we partitioning? Why are we partitioning it?

kyouko-taiga (changes requested):
I have mixed feelings about this PR.

On one hand, I think it's nice to bring the logic handling test annotations closer to their test cases. On the other, I think the proposed design involves more boilerplate on top of a rather complex control flow.

The design I built is based on my experience with so-called "reducers" in [redux](https://redux.js.org). My goal was for test runners to be able to compose an annotation handler by picking the kind of annotations they want to handle. That way, one could write a test runner following this formula:
1. Write the logic for running the test, completely independent from the testing logic
2. Construct an annotation handler
3. Pass the result of the test run to the handler

The proposed changes require test runners to merge some of the logic dealing with diagnostics (i.e., extending a diagnostic log) and use a more elaborated protocol to report errors (i.e., throwing a `DiagnosedError`). As it stands, it is not a win IMO.

dabrahams:
>  On the other, I think the proposed design involves more boilerplate

I'm baffled to that you would say so.  All the use sites got smaller, aside from where the one-off processing for C++ files was moved.  That even includes the cost of reorganizing control flow to have the semantics of the upcoming "driver" work (throwing on failure).

> on top of a rather complex control flow.

Are you saying control flow got more complex?  If so, how?  It took me all week to figure out the control flow of what you had written, FWIW.

> The design I built is based on my experience with so-called "reducers" in [redux](https://redux.js.org/). 

Sounds like you were seduced by a design pattern.  I struggled with design patterns for a while, finding that applying them always prematurely increased complexity.  Eventually someone really schooled in design patterns told me, you're not supposed to *apply* design patterns; you're just supposed to *recognize* them.  After that I wasn't sure there was much point in even doing that.

> My goal was for test runners to be able to compose an annotation handler by picking the kind of annotations they want to handle.

Yes, and it took me all week to figure out that's what you were doing and exactly what semantics you were implementing.  The details were not at all obvious, and much of the behavior was emergent from the interactions between your reducers.  But I worked hard to reproduce the important parts of the capabilities you enabled.

>  That way, one could write a test runner following this formula:
> Write the logic for running the test, completely independent from the testing logic

That statement doesn't make sense to me because AFAICT you're using the term "test" to mean two different things.  It took me all week to figure that out too.  What I think you mean by "running the test" is "processing the file somehow before checking the annotations."  By "the testing logic" I think you mean "checking the annotations."  However the actual thing under test is the outcome of that processing plus the annotation checking.  Once the driver work is done, the processing logic will shrink to just a few lines, and if in some case it doesn't, the way you separate it from any annotation "testing logic" is by using this design pattern known as "a function call."

The validity of using any given annotation handler is totally dependent on what is done in the test, aside from the common base case of checking diagnostics, which I've factored out.  You can't use a `CXXAnnotationHandler` with just any old test.  Did you notice that that "annotation handler" is actually emitting C++ files?

Tests/ValTests/Utils/AnnotatedValFileTest.swift
dabrahams: Thanks; I'll clarify that.

Tests/ValTests/Utils/AnnotatedValFileTest.swift
dabrahams: I thought it was explained by the next two lines of code, but I'll add a comment.

Tests/ValTests/Utils/AnnotatedValFileTest.swift
dabrahams: I think you were meant to refer to the function above it (referenced in the doc) for more clarity, but I agree the doc should stand on its own.

kyouko-taiga:
> Are you saying control flow got more complex? If so, how? It took me all week to figure out the control flow of what you had written, FWIW.

It took me quite a while to understand how control flows in and out of the methods defined in `XCTestCase`'s extensions. Sure, it didn't take a week but I still found it puzzling to the point I was about to ask how it works. Given that I at least understand my own code, I'm sure one can understand how that would leave someone feeling the code got more complex.

> Sounds like you were seduced by a design pattern.

This comment sounds slightly patronizing. Please don't assume how I to design my code.

I went through many iterations on this. Eventually, after many back and forth, I did _recognize_ that what I was building was looking like a Redux reducer. It's entirely possible that it may not have been the most obvious nor appropriate choice but it certainly wasn't motivated by a blind desire to apply a pattern I know.

Notice that it's not even a direct application of the reducer "pattern", it's just inspired by how one organizes code in Redux.

> Yes, and it took me all week to figure out that's what you were doing and exactly what semantics you were implementing.

I'm sorry you had to spend so much time understanding my code. I thought I had built something far simpler to read.

> That statement doesn't make sense to me because AFAICT you're using the term "test" to mean two different things. It took me all week to figure that out too.

Sorry for my poor choice of words. Yes, I meant to separate the "processing logic" from the "annotation checking".

IMO, the protocol your design proposes to have these two parts interact makes the distinction a littler blurrier than the test annotation handlers. The processing logic needs to write information to an `inout` parameter and use exceptions to signal failures. That also makes the control flow a little harder to understand at first sight. The design pattern of a function call typically doesn't involve 3 different channels.

I agree it's probably a good change nonetheless if we keep in mind we'll eventually have a driver to make the factor out the processing logic.

Sources/FrontEnd/AST+Extensions.swift
```diff
@@ -28,7 +28,7 @@ extension AST {
             return true
           }
         } catch let error as DiagnosedError {
-          fatalError(error.diagnostics.first!.description)
+          fatalError(String(error.diagnostics.lazy.map { "\($0)" }.joined(separator: "\n")))
```
kyouko-taiga, line 31: There's a custom string interpolation in Utils/ to do that:
```suggestion
          fatalError("\(list: error.diagnostics, joinedBy: "\n")")
```

kyouko-taiga (approved):

dabrahams:
> It took me quite a while to understand how control flows in and out of the methods defined in XCTestCase's extensions.

I'm honestly deeply confused by this statement.  The part that is complicated is lifted directly from your code.  There is only one extension, with three methods. `checkAnnotatedValFileDiagnostics` very clearly calls `checkAnnotatedValFiles` and that calls `failuresToReport`.  What can I do to make the flow clearer?

> This comment sounds slightly patronizing. Please don't assume how I to design my code.

Apologies for that.  To be clear, though, I wasn't assuming; merely conjecturing.

> IMO, the protocol your design proposes to have these two parts interact makes the distinction a littler blurrier than the test annotation handlers. 

That phrasing sorta assumes your division into parts is somehow the true one; I divided the problem into different parts that match our use cases better.  The checking of diagnostic annotations is one part.  The other is processing and any processing-specific annotation checking.

> The processing logic needs to write information to an inout parameter and use exceptions to signal failures. 

I'm confused again.  You phrase that like there's something undesirable about that and it's a burden but it's *exactly* the protocol we discussed yesterday and agreed to for the driver system.  It doesn't "need to" be organized that way, but our code will be far, far simpler in the long run.

> That also makes the control flow a little harder to understand at first sight. The design pattern of a function call typically doesn't involve 3 different channels.

Fair, but I thought it was not going to be a surprise after I outlined that the driver would use this pattern.

Sources/FrontEnd/AST+Extensions.swift
dabrahams: Sorta the way you abhor property wrappers, I never understood or liked custom interpolations.  But I'll use it if you think it's an improvement.

kyouko-taiga:
> What can I do to make the flow clearer?

The changes on the docs on `checkAnnotatedValFiles` made it much better. I think it's fine now (I approved the PR).

> That phrasing sorta assumes your division into parts is somehow the true one; 

I'm not saying it's the true division, just the one I went for. That's why, reading your changes, I originally tried to find it back.

> Fair, but I thought it was not going to be a surprise after I outlined that the driver would use this pattern.

I didn't think the design we discussed for the driver would be reused in this way.

I guess the fact that `checkAnnotatedValFiles` applies the design by the means of a higher-order function made it harder for me to understand. It's all good now, though.

PR #210, opened 2023-01-06
Encapsulation

dabrahams:
@kyouko-taiga is this a valid change?  It passes all tests.  I need an answer for the driver work.

### DIFF
--- a/Sources/FrontEnd/Parse/Parser.swift
+++ b/Sources/FrontEnd/Parse/Parser.swift
@@ -24,7 +24,7 @@ public enum Parser {
   public struct SourceParseResult {
 
     /// The identitiy of the parsed source, or `nil` if the parsed failed to parse one.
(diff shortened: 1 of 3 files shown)


Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -24,7 +24,7 @@ public enum Parser {
   public struct SourceParseResult {
 
     /// The identitiy of the parsed source, or `nil` if the parsed failed to parse one.
```
kyouko-taiga, line 26:
```suggestion
    /// The identity of the parsed source, or `nil` if the parsed failed to parse one.
```

kyouko-taiga (approved):
Great change. In fact, it even fixes an insidious bug that would have occurred every time the parser managed to create a source file but encountered a fatal syntax error.

PR #218, opened 2023-01-08
Add and use a first-class `Diagnostics` type.

dabrahams:
Because of the way thrown `Errors` are used, I can't be 1000% confident that when using the actual `valc` driver, errors aren't reported twice to the console.  I hope it will become easier to reason about that stuff as the driver work continues.


dabrahams:
@compnerd any idea what's up with the Windoze failures?  It seems to be giving content-free warnings when some modules are loaded, but aside from that I can't see what might be wrong.

compnerd:
That is really rather interesting.  I think that it might be helpful to use `-j 1` as the streams are interleaving and may be hiding the real information.

dabrahams:
Trying @compnerd's suggestion in #219

dabrahams:
Amazing; it passed. Trying to re-run the Windows job here.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -20,64 +20,37 @@ import Utils
...
-      return SourceParseResult(
-        source: nil, diagnostics: state.diagnostics + error.diagnostics)
-    } catch let error {
-      return SourceParseResult(
-        source: nil, diagnostics: state.diagnostics + [.error(String(describing: error))])
-    }
+    // Temporarily stash the ast and diagnostics in the parser state, avoiding CoW costs
+    var state = ParserState(ast: ast, lexer: Lexer(tokenizing: input), diagnostics: diagnostics)
+    defer { diagnostics = state.diagnostics }
+    defer { ast = state.ast }
```
kyouko-taiga, line 37: `ast` was not modified if the parser failed to parse a source file. Was changing this behavior intentional?

Sources/Core/Diagnostic+LoggingOrder.swift
```diff
@@ -1,10 +1,7 @@
-import Core
-import FrontEnd
-
 extension Diagnostic {
```
kyouko-taiga, line 1: Is it really worth defining this extension in a separate file if `isLoggedBefore` must be public? Why not define this method in the primary declaration of `Diagnostic`?

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -789,19 +761,19 @@ public enum Parser {
...
-      throw DiagnosedError(
+      throw Diagnostics(
         prologue.memberModifiers.map(Diagnostic.diagnose(unexpectedMemberModifier:)))
     }
 
     // Create a new `NamespaceDecl`.
     assert(prologue.accessModifiers.count <= 1)
-    let decl = try state.ast.insert(
-      wellFormed: NamespaceDecl(
+    let decl = state.insert(
```
kyouko-taiga, line 775:
```suggestion
    return state.insert(
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -862,8 +834,8 @@ public enum Parser {
     // Create a new `SubscriptDecl`.
     assert(prologue.accessModifiers.count <= 1)
     assert(prologue.memberModifiers.count <= 1)
-    let decl = try state.ast.insert(
-      wellFormed: SubscriptDecl(
+    let decl = state.insert(
```
kyouko-taiga, line 837:
```suggestion
    return state.insert(
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -897,8 +869,8 @@ public enum Parser {
...
```
kyouko-taiga, line 872:
```suggestion
    return state.insert(
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -1082,8 +1054,8 @@ public enum Parser {
 
     // Create a new `ProductTypeDecl`.
     assert(prologue.accessModifiers.count <= 1)
-    let decl = try state.ast.insert(
-      wellFormed: ProductTypeDecl(
+    let decl = state.insert(
```
kyouko-taiga, line 1057:
```suggestion
    return state.insert(
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -1138,19 +1110,19 @@ public enum Parser {
...
-      throw DiagnosedError(
+      throw Diagnostics(
         prologue.memberModifiers.map(Diagnostic.diagnose(unexpectedMemberModifier:)))
     }
 
     // Create a new `TypeAliasDecl`.
     assert(prologue.accessModifiers.count <= 1)
-    let decl = try state.ast.insert(
-      wellFormed: TypeAliasDecl(
+    let decl = state.insert(
```
kyouko-taiga, line 1124:
```suggestion
    return state.insert(
```

kyouko-taiga (approved):
Great changes. I left a few comments and suggestions; nothing that I considered blocking.

Sources/FrontEnd/Parse/Parser.swift
dabrahams: Yes it was, because we agreed it was OK, but also now that I know modules have their own storage maybe I shouldn't change it.  I'll roll that part back, thanks.

Sources/Core/Diagnostic+LoggingOrder.swift
dabrahams: Just because you already had it that way.  I moved the file to the module where it was needed and that was all.  If you want to organize things differently please feel free to make the change after this is in main.

PR #220, opened 2023-01-09
Factor source file computation

dabrahams:


Sources/Core/SourceFile.swift
```diff
@@ -155,3 +156,27 @@ extension SourceFile: CustomStringConvertible {
...
+/// directories are recursively searched for `.val` files, which are considered Val `sourceFiles`;
+/// all others are treated as non-source files and are ignored.
+public func sourceFiles<S: Collection>(in sourcePaths: S) throws -> [SourceFile]
+where S.Element == URL {
+  let explicitSourcePaths = sourcePaths.filter { !$0.hasDirectoryPath }
+  let sourceDirectoryPaths = sourcePaths.filter { $0.hasDirectoryPath }
+
+  // Recursively search the directory paths, adding .val files to `sourceFiles`
+  var result = try explicitSourcePaths.map(SourceFile.init)
+  for d in sourceDirectoryPaths {
```
kyouko-taiga, line 173: Not that we should care about performance that much, but it looks like this is doing roughly 3 linear passes over `sourcePaths` where only one would have been enough.

kyouko-taiga (approved):
Nice boilerplate reduction.

Sources/Core/SourceFile.swift
dabrahams: Um, yeah, but seriously the cost of that as compared to that of the allocations (to say nothing of talking to the filesystem!!) is surely lost in the noise.  This structure is left over from when I was also reporting the skipped files so I could preserve your warning about non-val files.

PR #223, opened 2023-01-11
Pretty print JSON, add type names.

oskidan:
Hi everyone, :wave:

Hopefully this PR makes the generated JSON a bit more readable.

Resolves: #70

### DIFF
--- a/Sources/Core/AST/AnyNode.swift
+++ b/Sources/Core/AST/AnyNode.swift
@@ -5,7 +5,7 @@ struct AnyNode: Codable {
 
   fileprivate enum CodingKeys: String, CodingKey {
 
-    case kind, data
+    case nodeName, kind, data
 
   }
 
@@ -27,8 +27,9 @@ struct AnyNode: Codable {
   func encode(to encoder: Encoder) throws {
     var container = encoder.container(keyedBy: CodingKeys.self)
 
-    let kind = type(of: node).kind
-    try container.encode(kind, forKey: .kind)
+    let nodeType = type(of: node)
+    try container.encode("\(nodeType)", forKey: .nodeName)
+    try container.encode(nodeType.kind, forKey: .kind)
     try container.encode(node)
   }
 }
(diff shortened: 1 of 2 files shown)


Sources/Core/AST/AnyNode.swift
```diff
@@ -27,8 +27,9 @@ struct AnyNode: Codable {
   func encode(to encoder: Encoder) throws {
     var container = encoder.container(keyedBy: CodingKeys.self)
 
-    let kind = type(of: node).kind
-    try container.encode(kind, forKey: .kind)
+    let nodeType = type(of: node)
+    try container.encode("\(nodeType)", forKey: .nodeName)
```
kyouko-taiga, line 31:
```suggestion
    let nodeType = type(of: node)
    try container.encode("\(type(of: node))", forKey: .nodeName)
```

Sources/Core/AST/AnyNode.swift
```diff
@@ -5,7 +5,7 @@ struct AnyNode: Codable {
 
   fileprivate enum CodingKeys: String, CodingKey {
 
-    case kind, data
+    case nodeName, kind, data
```
kyouko-taiga, line 8:
```suggestion
    case nodeType, kind, data
```

kyouko-taiga, line 8:
```suggestion
    case nodeType, kind, data
```

kyouko-taiga (changes requested):
Thanks for this contribution!

I think `CodingKeys.nodeName` should be renamed `CodingKeys.nodeType`. Except that, I the changes should make json outputs more usable. Good job.

Sources/Core/AST/AnyNode.swift
oskidan: Makes sense. Done.

kyouko-taiga (approved):
LGTM

dabrahams (changes requested):
Sorry, I can't tell what this accomplishes by looking at the code.  I'd like to see a test added that would fail before these changes and passes afterward.  Thanks!

C-BJ:
What's wrong with this PR? @kyouko-taiga

kyouko-taiga:
Dave asked for a test that would highlight what the PR is accomplishing.

kyouko-taiga:
I think this PR has gone stale.

There's a related issue on the topic of visualizing ASTs (see #70).

PR #224, opened 2023-01-11
Temporary disable swift-format to figure out issues with building swift-format on CI

lucteo:


lucteo:
See original issue: https://github.com/val-lang/val/issues/222

lucteo:
hmm... CI seems to behave in completely unexpecred ways today... not sure why swift-format check is still running after the last commit.

PR #226, opened 2023-01-15
Emit classes

lucteo:
The CXX transpiler is lacking the ability to emit C++ classes from Val product types.

Add support for emitting classes. The support, at this point, is just basic: we know to write the attributes of simple product types.
We now have 2 top-level declaration types that can be translated into a C++ module: functions and classes. Both of them need to be able to write the "header part" and the "source part".

Testing:
*  add a simple test to show translation of a Point type

### DIFF
--- a/Sources/CodeGen/CXX/CXXModule.swift
+++ b/Sources/CodeGen/CXX/CXXModule.swift
@@ -118,9 +118,59 @@ public struct CXXModule {
 
     let name = CXXIdentifier(decl.identifier.value)
 
+    // Transpile the class membmers.
+    var cxxMembers: [CXXClassDecl.ClassMember] = []
+    for member in decl.members {
+      switch member.kind {
+      case BindingDecl.self:
+        let bindingDecl = BindingDecl.Typed(member)!
+        // Check if the attribute is static or not.
+        var isStatic = false
+        if bindingDecl.memberModifier != nil {
+          switch bindingDecl.memberModifier!.value {
+          case .static:
+            isStatic = true
+          }
+        }
+        // TODO: visit initializer (bindingDecl.initializer)
+        let cxxInitializer: CXXRepresentable? = nil
+        // TODO: pattern introducer (let, var, sink, inout)
+        // Visit the name patterns.
+        for (_, name) in bindingDecl.pattern.subpattern.names {
+          let varDecl = name.decl
+          let cxxAttribute = CXXClassAttribute(
+            type: CXXTypeExpr(varDecl.type, ast: program.ast)!,
+            name: CXXIdentifier(varDecl.name),
+            initializer: cxxInitializer,
+            isStatic: isStatic,
+            original: varDecl)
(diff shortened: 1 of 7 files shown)


Sources/CodeGen/CXX/Decl/CXXTopLevelDecl.swift
```diff
@@ -0,0 +1,10 @@
+/// A C++ top-level declaration that needs to be written both to headers and source files.
+public protocol CXXTopLevelDecl {
```
kyouko-taiga, line 2: Is this protocol necessary? What distinguishes a `CXXDecl` from a `CXXTopLevelDecl`?

kyouko-taiga: Oh, now I see that we no longer have a `CXXDecl` protocol or type (I'm under the impression we had one before). So you can ignore my question.

kyouko-taiga, line 2: I think this protocol should refine `CXXRepresentable`.

Sources/CodeGen/CXX/CXXModule.swift
```diff
@@ -11,96 +11,17 @@ public struct CXXModule {
...
 
   public init(_ decl: ModuleDecl.Typed, for program: TypedProgram) {
     self.valDecl = decl
     self.program = program
   }
 
-  /// Returns the ID of the C++ function declaration corresponding to `valFunctionDecl`.
-  ///
-  /// - Requires: `valFunctionDecl` must be declared in `self.decl`.
-  public mutating func getOrCreateFunction(
```
kyouko-taiga, line 32: `getOrCreateFunction` was meant to be used as a way to handle cases where in Val a function occurs before its declaration, since Val doesn't require forward declarations. Is the code generator able to handle this program?

```val
public fun main() {
  print(foo())
}

fun foo() -> Int { 42 }
```

kyouko-taiga (commented):
I have concerns about the disappearance of `getOrCreateFunction` (see inline comments). Otherwise, LGTM.

Sources/CodeGen/CXX/CXXModule.swift
lucteo: We are not at that point yet, but yes, this code organization allows expressing this. There are multiple ways in which I can see doing that:
* Emit a `CXXFunctionDecl` with a `nil` body -- probably best option
* Introduce a new node of type `CXXForwardFunctionDecl` that does the same thing

In both cases, I envision adding more support in the `CXXModule` structure to keep track of what's being emitted and what not.

Also, the same problem manifests to classes, and probably to everything that we are putting into header files.

I'll deal with these cases when we reach there with the transpiler.

At this point, I wanted to move the logic for emitting functions/classes into `CXXTransipler`, so that we can reuse some of the functions from there. Looking at the code that emits classes, there is still some duplication that needs to be cleaned up.

PR #227, opened 2023-01-15
Add translation plan to CXX transpiler documentation

lucteo:
Add tables with how the Val nodes would be transpiled into CXX nodes (not yet complete). This forms a rough plan on what CXX nodes we need to build and how different Val structures translate to those nodes.


kyouko-taiga (approved):
LGTM

dabrahams:
No objections but I also can't really evaluate it. It would be a lot more useful if it contained some code examples showing how these constructs might be translated.

lucteo:
This document aims at being a rough guide for me, to know what I should tackle. I doesn't need to be 100%; actually I would expect it to change parts of it, while we are working on the transpiler.

I'll merge it as is, and make changes to it, as we move along with the implementation.

PR #230, opened 2023-01-17
Refine pattern inference

kyouko-taiga:


dabrahams:
I'm concerned about seeing a PR called "refine _something-about-the-language-semantics_" with no tests in the PR.

kyouko-taiga:
I created the PR before pushing the new test, sorry about that.

PR #231, opened 2023-01-17
Measure/maintain code coverage in CI

dabrahams:
Causes CI to fail if our code coverage slips; generates a coverage report with every run.

Currently only checks coverage on ubuntu (the GH action plugin only supports that).  This could certainly be improved by making it work on all OSes and eliminating warnings (and there are other plugin options we can explore) but this will suffice to make sure we don't backslide on testing.  We'll have to update the MINIMUM_COVERAGE as we improve things

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -43,7 +43,7 @@ jobs:
       if: matrix.os == 'ubuntu-22.04'
       with:
         SORT_ORDER: +cov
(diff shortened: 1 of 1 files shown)


kyouko-taiga (approved):
Looks like a great change.

How 74 was decided? Is that our current coverage?

dabrahams:
Yes, if you look at the run output you'll see that it comes in at ~74.3.  The plugin reports fractional percentage but only accepts integers.

PR #233, opened 2023-01-17
Ci coverage expect failure

dabrahams:
Trying to verify that this fails the check on ubuntu by having too little coverage.


dabrahams:
Verified

PR #235, opened 2023-01-17
Emit cxx return

lucteo:
Add possibility to emit CXX return statements from Val `return` statements, with or without returning value.

Add tests to cover cases in which `CXXReturnStmt` needs to be emitted:
- function implicitly returning a value
- explicit `return` statement without any value
- explicit `return` statement with value


kyouko-taiga (approved):
LGTM

PR #237, opened 2023-01-18
Fix XCTest shims

kyouko-taiga:
We cannot change the API of the types in XCTest+Shims.swift. Everything in this file is meant to reproduce XCTest's API on non-Apple platforms.


kyouko-taiga:
We cannot change the API of the types in `XCTest+Shims.swift`. Everything in this file is meant to reproduce XCTest's API on non-Apple platforms.

PR #238, opened 2023-01-18
Emit CXX expression statements

lucteo:
Val's `ExprStmt` would be translated into a `CXXExprStmt`. Also, bindings without any pattern will result in a `CXXExprStmt` (with a `CXXVoidCast`)


kyouko-taiga (approved):
LGTM

Issue #239, opened 2023-01-18
Optional `SourceRange` must go


dabrahams:
I'm convinced that storing `SourceRange?` instead of `SourceRange` creates a vast amount of needless complexity and having it defaulted is error-prone, leading to mistakes like [this](https://github.com/val-lang/val/blob/fd204696c146f7083f1967adccbc5ccced1b0b91/Sources/FrontEnd/TypeChecking/TypeChecker%2BDiagnostics.swift#L177) where it is omitted.

We synthesize URLs for the "files" of synthesized code; we can easily synthesize text for that "file"; probably a comment would be most useful.  But in fact we should use a real file for almost every bit of synthesized code. Things like synthesized `self.` qualification have an obvious source range already: use the empty range at the beginning of the expression they qualify.

I would do this job entirely on my own (and I have a huge patch that contains a lot of the work) but there are lots of places in the parser where `nil` is being used as a source range and it's not obvious what to replace it with.  OK, maybe I have a strategy for proceeding.  This issue will probably turn into "eliminate _some-static-member-of-`SourceRange`_ that is a stand-in for `nil`"

dabrahams:
Most likely `SourceRepresentable` should go as well, and all `Node`s should be endowed with a `SourceRange`.

PR #241, opened 2023-01-19
Make most Source ranges non-optional.

dabrahams:
(IR instructions are a separate issue.)

Addresses #239

There will be some follow-up work for @kyouko-taiga to do that I can't address.  That includes:
- elimination of SourceRange.eliminateFIXME
- serializing actual source file contents rather than reading them off disk.

### DIFF
--- a/Sources/CLI/main.swift
+++ b/Sources/CLI/main.swift
@@ -259,7 +259,7 @@ private struct CLI: ParsableCommand {
 
   /// Logs `diagnostic` to the standard error.
   private func log(diagnostic: Diagnostic, asChild isChild: Bool = false) {
-    // Log the location, if available.
+    // Log the location
     let siteFirst = diagnostic.site.first()
     let path = siteFirst.file.url.relativePath
     let (lineFirst, column) = siteFirst.lineAndColumn()
--- a/Sources/CLI/main.swift
+++ b/Sources/CLI/main.swift
@@ -260,11 +260,10 @@ private struct CLI: ParsableCommand {
   /// Logs `diagnostic` to the standard error.
   private func log(diagnostic: Diagnostic, asChild isChild: Bool = false) {
     // Log the location, if available.
-    if let site = diagnostic.site?.first() {
-      let path = site.file.url.relativePath
-      let (line, column) = site.lineAndColumn()
-      write("\(path):\(line):\(column): ".styled([.bold]))
-    }
+    let siteFirst = diagnostic.site.first()
+    let path = siteFirst.file.url.relativePath
+    let (lineFirst, column) = siteFirst.lineAndColumn()
+    write("\(path):\(lineFirst):\(column): ".styled([.bold]))
 
     // Log the level.
     if isChild {
@@ -283,22 +282,21 @@ private struct CLI: ParsableCommand {
     write("\n")
 
     // Log the window
-    if let site = diagnostic.site {
-      let line = site.first().textOfLine()
-      write(line)
+    let site = diagnostic.site
+    let line = site.first().textOfLine()
+    write(line)
 
-      let padding = line.distance(from: line.startIndex, to: site.start)
-      write(String(repeating: " ", count: padding))
+    let padding = line.distance(from: line.startIndex, to: site.start)
+    write(String(repeating: " ", count: padding))
 
-      let count = line.distance(
-        from: site.start, to: min(site.end, line.endIndex))
-      if count > 1 {
-        write(String(repeating: "~", count: count))
-      } else {
-        write("^")
-      }
-      write("\n")
+    let count = line.distance(
+      from: site.start, to: min(site.end, line.endIndex))
+    if count > 1 {
+      write(String(repeating: "~", count: count))
+    } else {
+      write("^")
     }
+    write("\n")
 
     // Log the notes.
     for child in diagnostic.notes {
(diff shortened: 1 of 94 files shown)


Sources/CLI/main.swift
```diff
@@ -260,11 +260,10 @@ private struct CLI: ParsableCommand {
   /// Logs `diagnostic` to the standard error.
   private func log(diagnostic: Diagnostic, asChild isChild: Bool = false) {
     // Log the location, if available.
```
kyouko-taiga, line 262:
```suggestion
    // Log the location.
```

Sources/Core/AST/Decl/AssociatedTypeDecl.swift
```diff
@@ -1,10 +1,10 @@
 /// An associated type declaration.
 public struct AssociatedTypeDecl: SingleEntityDecl {
 
-  public let origin: SourceRange?
+  public let origin: SourceRange
 
   /// The source range of the declaration's introducer, if any.
```
kyouko-taiga, line 6:
```suggestion
  /// The source range of the declaration's introducer.
```

There's probably a lot of "...., if any" comments in the docs of all nodes.

Sources/Core/AST/Expr/ParameterTypeExpr.swift
```diff
@@ -1,25 +1,27 @@
...
+    synthesized: Bool
   ) {
     self.origin = origin
     self.convention = convention
     self.bareType = bareType
+    self.isConventionSynthesized = synthesized
   }
 
-  /// Indicates whether `convention` was synthesized.
-  public var isConventionSynthesized: Bool { convention.origin == nil }
```
kyouko-taiga, line 23: I think there was something elegant about this property that makes me believe perhaps the fact that something has been synthesized should be encoded in the origin. That might be how to replace your `eliminateFIXME`.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -940,23 +941,16 @@ public enum Parser {
...
-      switch body! {
-      case .expr(let id):
-        origin = state.ast[id].origin!
-      case .block(let id):
-        origin = state.ast[id].origin!
-      }
-    }
+    let startIndex = introducer.origin.start
+    // FIXME: @kyouku-taiga should check this for correctness.
+    origin = state.range(from: startIndex)
```
kyouko-taiga, line 953: It's wrong because introducer.origin might be `eliminateFIXME`(formerly `nil`) if `buildSubscriptImplDecl` was called to parse a subscript implementation without an introducer. For example:

```
property foo: Int { 0 } // let introducer is synthetic
```

In that case, we need to get the first position of the contents of the body.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2847,10 +2842,12 @@ public enum Parser {
       .map({ (state, tree) -> NodeID<ParameterTypeExpr> in
         state.insert(
           ParameterTypeExpr(
-            convention: tree.0 ?? SourceRepresentable(value: .let),
+            convention: tree.0 ?? SourceRepresentable(value: .let, range: .eliminateFIXME),
             bareType: tree.1,
             origin: state.range(
-              from: tree.0?.origin!.start ?? state.ast[tree.1].origin!.start)))
+              from: tree.0?.origin.start ?? state.ast[tree.1].origin.start),
+            synthesized: tree.0 == nil  // FIXME: @kyouku-taiga should check this for correctness.
```
kyouko-taiga, line 2849: That's correct.

kyouko-taiga (changes requested):
There's an invalid change in the parser. I left an inline comment. Let me know if you need my help to fix.

Sources/Core/AST/Expr/ParameterTypeExpr.swift
dabrahams: I disagree.  Whether to add a feature (a.k.a. complexity) should be based on whether the problem demands it, not on whether we think there's something elegant about it.  BUT: if you really want to know if something is synthesized, you only need to check if its sourceRange start == end.  These synthesized bits should all (or nearly all) have locations in real source.

Sources/FrontEnd/Parse/Parser.swift
dabrahams: Fixed?

Sources/FrontEnd/Parse/Parser.swift
dabrahams, line 2856:
```suggestion
            synthesized: tree.0 == nil
```

kyouko-taiga (approved):
LGTM

Sources/FrontEnd/Parse/Parser.swift
kyouko-taiga: Yep, good job.

Sources/Core/AST/Expr/ParameterTypeExpr.swift
dabrahams: Oops; I meant to remove it entirely!  It isn't used anywhere.  That was my point about whether the problem demands it

kyouko-taiga: Ah, "au temps pour moi"!

I operated under the assumption it was used in the type checker.

PR #244, opened 2023-01-21
Encode/Decode text of SourceFile, making it an immutable class.

### LINKED ISSUE #242: Serialization of SourceFiles is somewhat busted
Currently URLs are serialized and source text is read back from disk, but the file might not even exist.  Add to that that it will be really wasteful to serialize the URLs.  `SourceFile`s *should* get collected into a dictionary that maps URL to integer ID values at the beginning of AST serialization, then the inverse-mapping array should get serialized as the first step of serializing the AST, then individual `SourceFile` instances should get serialized as their IDs (or ultimately, ideally, inherited from their parent nodes and never serialized).

I have some scraps of code/research that may be useful for this project, but I don't want to do the work myself.

dabrahams:
Codable conformance moves into the main body of SourceFile because... Swift classes and their weird initialization rules.

Fixes #242.

### DIFF
--- a/Sources/Core/SourceFile.swift
+++ b/Sources/Core/SourceFile.swift
@@ -5,7 +5,7 @@ import Utils
 /// A Val source file.
 ///
 /// - Note: two source files are equal if and only if they have the same path in the filesystem.
-public struct SourceFile {
+public final class SourceFile: Codable {
 
   /// A position in the source text.
   public typealias Index = String.Index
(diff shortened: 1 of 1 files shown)


kyouko-taiga:
Does `Codable` preserve class uniqueness? I seems it doesn't in my quick and dirty experiment.

dabrahams:
OMG if Codable doesn't preserve class uniqueness, it is totally unsuitable for serializing OO instance networks!  Can I see your test?

kyouko-taiga:
```swift
import Foundation

class A: Codable {

  let contents: String

  init(_ contents: String) {
    self.contents = contents
  }

}

struct B: Codable {

  let foo: Int
  let bar: A

}

struct C: Codable {

  let ham: [B]

}

let a = A("Hello, world!")
let c = C(ham: [B(foo: 1, bar: a), B(foo: 1, bar: a)])
print(c.ham[0].bar === c.ham[1].bar) // true, as expected

let encoder = JSONEncoder()
let data = try! encoder.encode(c)

let decoder = JSONDecoder()
let c1 = try! decoder.decode(C.self, from: data)
print(c1.ham[0].bar === c1.ham[1].bar) // false, should be true

```

kyouko-taiga:
Some more information on the topic: https://forums.swift.org/t/codable-with-references/13885/69

dabrahams:
I'll rework this.

dabrahams:
Superseded by #281

PR #253, opened 2023-01-23
Emit CXX code for assign statements, expression statements, and local variables

lucteo:
Support for local variable is at its infancy.

Adding CXX translation test for assign statements

### DIFF
--- a/Sources/CodeGen/CXX/CXXTranspiler.swift
+++ b/Sources/CodeGen/CXX/CXXTranspiler.swift
@@ -241,6 +241,8 @@ public struct CXXTranspiler {
       return emit(declStmt: DeclStmt.Typed(stmt)!)
     case ExprStmt.self:
       return emit(exprStmt: ExprStmt.Typed(stmt)!)
+    case AssignStmt.self:
+      return emit(assignStmt: AssignStmt.Typed(stmt)!)
     case ReturnStmt.self:
       return emit(returnStmt: ReturnStmt.Typed(stmt)!)
     default:
@@ -267,6 +269,15 @@ public struct CXXTranspiler {
 
   private mutating func emit(exprStmt stmt: ExprStmt.Typed) -> CXXRepresentable {
     return CXXExprStmt(expr: emitR(expr: stmt.expr), original: AnyNodeID.TypedNode(stmt))
+  }
+
+  private mutating func emit(assignStmt stmt: AssignStmt.Typed) -> CXXRepresentable {
+    let cxxExpr = CXXInfixExpr(
+      callee: CXXIdentifier("="),
+      lhs: emitL(expr: stmt.left, withCapability: .set),
+      rhs: emitR(expr: stmt.right),
+      original: AnyNodeID.TypedNode(stmt))
+    return CXXExprStmt(expr: cxxExpr, original: AnyNodeID.TypedNode(stmt))
   }
 
   private mutating func emit(returnStmt stmt: ReturnStmt.Typed) -> CXXRepresentable {
@@ -540,7 +551,7 @@ public struct CXXTranspiler {
       "&&=", "**":
       // Expand this as a native infix operator call
       return CXXInfixExpr(
-        callee: CXXIdentifier(name), lhs: lhs, rhs: rhs, original: AnyExprID.TypedNode(expr))
+        callee: CXXIdentifier(name), lhs: lhs, rhs: rhs, original: AnyNodeID.TypedNode(expr))
 
     default:
       // Expand this as a regular function call.
--- a/Sources/CodeGen/CXX/CXXTranspiler.swift
+++ b/Sources/CodeGen/CXX/CXXTranspiler.swift
@@ -175,27 +175,8 @@ public struct CXXTranspiler {
   }
 
   private mutating func emit(localBinding decl: BindingDecl.Typed) -> CXXRepresentable {
-    let pattern = decl.pattern
+    let capability = decl.pattern.introducer.value
 
-    switch pattern.introducer.value {
-    case .var, .sinklet:
-      return emit(storedLocalBinding: decl)
-    case .let:
-      return emit(borrowedLocalBinding: decl, withCapability: .let)
-    case .inout:
-      return emit(borrowedLocalBinding: decl, withCapability: .inout)
-    }
-  }
-
-  private mutating func emit(storedLocalBinding decl: BindingDecl.Typed) -> CXXRepresentable {
-    return CXXComment(comment: "local binding", original: AnyNodeID.TypedNode(decl))
-  }
-
-  /// Emits borrowed bindings.
-  private mutating func emit(
-    borrowedLocalBinding decl: BindingDecl.Typed,
-    withCapability capability: AccessEffect
-  ) -> CXXRepresentable {
     // There's nothing to do if there's no initializer.
     if let initializer: TypedNode<AnyExprID> = decl.initializer {
 
@@ -211,15 +192,21 @@ public struct CXXTranspiler {
       for (path, name) in pattern.subpattern.names {
         // TODO: emit code for the patterns.
         let decl = name.decl
+        let _ = path
         stmts.append(
-          CXXComment(
-            comment: "decl \(name), type: \(decl.type.description); path: \(path)",
-            original: AnyNodeID.TypedNode(name)))
+          CXXLocalVarDecl(
+            type: CXXTypeExpr(decl.type, ast: program.ast)!,
+            name: CXXIdentifier(decl.identifier.value),
+            initializer: cxxInitialzer,
+            original: decl)
+        )
       }
       if stmts.isEmpty {
         // No pattern found; just call the initializer, dropping the result.
         let cxxExpr = CXXVoidCast(baseExpr: cxxInitialzer, original: initializer)
         return CXXExprStmt(expr: cxxExpr, original: AnyNodeID.TypedNode(initializer))
+      } else if stmts.count == 1 {
+        return stmts[0]
       } else {
         return CXXScopedBlock(stmts: stmts, original: AnyNodeID.TypedNode(initializer))
       }
--- a/Sources/CodeGen/CXX/CXXTranspiler.swift
+++ b/Sources/CodeGen/CXX/CXXTranspiler.swift
@@ -178,7 +178,7 @@ public struct CXXTranspiler {
     let capability = decl.pattern.introducer.value
 
     // There's nothing to do if there's no initializer.
-    if let initializer: TypedNode<AnyExprID> = decl.initializer {
+    if let initializer = decl.initializer {
 
       let isLValue =
         (initializer.kind == NameExpr.self) || (initializer.kind == SubscriptCallExpr.self)
--- a/Sources/CodeGen/CXX/Decl/CXXLocalVarDecl.swift
+++ b/Sources/CodeGen/CXX/Decl/CXXLocalVarDecl.swift
@@ -1,0 +1,26 @@
+import Core
+
+/// A C++ local variable declaration.
+public struct CXXLocalVarDecl: CXXRepresentable {
+
+  /// The type of the local variable.
+  public let type: CXXTypeExpr
+  /// The name of the local variable.
+  public let name: CXXIdentifier
+  /// The initializer of the local variable.
+  public let initializer: CXXRepresentable?
+
+  /// The original node in Val AST.
+  let original: VarDecl.Typed
+
+  public func writeCode<Target: TextOutputStream>(into target: inout Target) {
+    target.write("\(type) ")
+    name.writeCode(into: &target)
+    if let value = initializer {
+      target.write(" = ")
+      value.writeCode(into: &target)
+    }
+    target.write(";\n")
+  }
+
+}
--- a/Sources/CodeGen/CXX/Expr/CXXInfixExpr.swift
+++ b/Sources/CodeGen/CXX/Expr/CXXInfixExpr.swift
@@ -14,7 +14,7 @@ struct CXXInfixExpr: CXXRepresentable {
   public let rhs: CXXRepresentable
 
   /// The original node in Val AST.
-  /// Typically an expression, but somtimes this can be AssignStmt
+  /// Typically an expression, but may be an `AssignStmt`.
   let original: AnyNodeID.TypedNode?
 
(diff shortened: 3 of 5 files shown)


Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -175,27 +175,8 @@ public struct CXXTranspiler {
   }
 
   private mutating func emit(localBinding decl: BindingDecl.Typed) -> CXXRepresentable {
-    let pattern = decl.pattern
+    let capability = decl.pattern.introducer.value
```
dabrahams, line 178: `capability` is a bad name, just like `pattern` was, and `decl` is.  The name of a variable or parameter should reflect its _role_ if possible, and if you can't do any better than reflecting something about the type of the variable, simply use a single letter name.  

Is `capability` even used in the revised code?

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -175,27 +175,8 @@ public struct CXXTranspiler {
...
-    return CXXComment(comment: "local binding", original: AnyNodeID.TypedNode(decl))
-  }
-
-  /// Emits borrowed bindings.
-  private mutating func emit(
-    borrowedLocalBinding decl: BindingDecl.Typed,
-    withCapability capability: AccessEffect
-  ) -> CXXRepresentable {
     // There's nothing to do if there's no initializer.
     if let initializer: TypedNode<AnyExprID> = decl.initializer {
```
dabrahams, line 181:
```suggestion
    if let initializer = decl.initializer {
```

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -211,15 +192,21 @@ public struct CXXTranspiler {
       for (path, name) in pattern.subpattern.names {
         // TODO: emit code for the patterns.
         let decl = name.decl
+        let _ = path
```
dabrahams, line 195: If you're not using `path`, just omit it and use `_` in the binding pattern in line 192

dabrahams, line 195: Note (line 183) that being a name does not make an expression an lvalue, e.g. a function or type name.

dabrahams: `stmts` is a bad name for reasons mentioned earlier.

Sources/CodeGen/CXX/Decl/CXXLocalVarDecl.swift
```diff
@@ -0,0 +1,26 @@
...
+
+/// A C++ local variable declaration.
+public struct CXXLocalVarDecl: CXXRepresentable {
+
+  /// The type of the local variable.
+  public let type: CXXTypeExpr
+  /// The name of the local variable.
+  public let name: CXXIdentifier
+  /// The initializer of the local variable.
+  public let initializer: CXXRepresentable?
```
dabrahams, line 11: Why doesn't this have a type like `CXXExpr?`

Sources/CodeGen/CXX/Decl/CXXLocalVarDecl.swift
```diff
@@ -0,0 +1,26 @@
...
+
+  /// The original node in Val AST.
+  let original: VarDecl.Typed
+
+  public func writeCode<Target: TextOutputStream>(into target: inout Target) {
```
dabrahams, line 16: Doc comment?

Sources/CodeGen/CXX/Expr/CXXInfixExpr.swift
```diff
@@ -14,7 +14,8 @@ struct CXXInfixExpr: CXXRepresentable {
   public let rhs: CXXRepresentable
 
   /// The original node in Val AST.
-  let original: AnyExprID.TypedNode?
+  /// Typically an expression, but somtimes this can be AssignStmt
```
kyouko-taiga, line 17:
```suggestion
  /// Typically an expression, but may be an `AssignStmt`.
```

kyouko-taiga: Why should we represent assignments with `CXXInfixExpr`? Assignments are statements and not expressions in Val.

dabrahams (approved):
Various nits, but the bigger problem is that I don't understand the context.  I am not in a position to make judgements about the correctness-or-not of the code in `CXXTranspiler.swift`.  Lots of uncommented bits.  What's `emitR`? _Something_ should be done so that I can read this file and understand what it's doing.  I don't want to block merging this, but I think @lucteo and I should try to figure out how to fix the understandability problems.

kyouko-taiga (approved):
I'll approve so that Lucian can progress, but I think we may need to meet so that we can get in sync on various aspect of the syntax and semantics.

Sources/CodeGen/CXX/Expr/CXXInfixExpr.swift
dabrahams: Probably because they're exprs in C++.  But there's a broader question of what we are trying to represent with our C++ AST.  Is it C++, or really just a C++ translation of Val?

kyouko-taiga: Good question. I would instinctively answer that we should go for the latter, as there are probably some C++ constructs that won't get generated from Val, but I might be wrong about this.

lucteo: In my mind, the CXX AST tries to emulate as much as possible the syntax of C++. Sometimes the CXX constructs map to Vap constructs, but sometimes they do not. 

Conditional expression in Val is probably one of the best examples. In C++, we  have the ternary operator, but everything needs to be an expression, and we have the `if` statement, but that cannot be an expression.

On the other hand, I'm not trying to create CXX nodes for everything that C++ supports. I'm trying to incrementally add them as we need to emit them.

lucteo:
> Various nits, but the bigger problem is that I don't understand the context. I am not in a position to make judgements about the correctness-or-not of the code in `CXXTranspiler.swift`. Lots of uncommented bits. What's `emitR`? _Something_ should be done so that I can read this file and understand what it's doing. I don't want to block merging this, but I think @lucteo and I should try to figure out how to fix the understandability problems."

Yes, that makes sense. Let us have a dedicated discussion on the transpiler.

Short summary: I tried to just follow the emitter example and add the small amount of code that can translate some reasonable C++ code for very small examples (lot of TODOs and missing features). I wanted first to have a quick round to have some basic translation for common constructs; I'm still at this point. After this first phase is complete, I want to have a more systematic approach for CXX translation.

> 
> Also, it's _very_ hard to believe that this change doesn't reduce code coverage, considering the branches I see in the code and the simplicity of the one test case added.

For the most part, everything that I've done in the transpiler is covered by some basic examples. Considering that we didn't have a lot of examples before, I would say that the coverage is slowly increasing. But, I haven't watched the numbers, so I may be wrong.

Sources/CodeGen/CXX/CXXTranspiler.swift
kyouko-taiga: FWIW, after trying my best to apply this advice for the last couple of month, I disagree with the generality of this statement (to be fair, @dabrahams did say "if possible"). Sometimes a type has a name that already describes the role of its instances and it seems counterproductive to just invent another name or fall back on a meaningless letter.

For example, in this particular case, would @dabrahams object to the name `pattern` if the type of this variable was called `ASTNode`? `pattern` is the role of this variable in the context of this function. Perhaps it could be called `visitedPattern` but the qualification makes the name longer and  doesn't add much in this context, IMO.

As a reader of this function, I find the name `pattern` more useful than `x` and, as mentioned, don't think this name deserves any qualification.

Sources/CodeGen/CXX/Decl/CXXLocalVarDecl.swift
lucteo: This is handled in #267

Sources/CodeGen/CXX/Decl/CXXLocalVarDecl.swift
lucteo: I don't know how to handle these cases. If I "override" a method of a protocol, do I need to write the same comment over and over? What is the best practice here?

Anyways, I would handle doc-strings in #270, after the major refactoring. This function will disappear.

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: I would prefer to fix this later, after the current set of PRs, to make sure I don't need a lot of rebases. I've created #270 for this.

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: For `path`, I would keep it like that, as I need to revisit this code anyways. For me, having the name there allows me to see what we are iterating over.

For `stmts`, I would prefer to do it later with #270

lucteo: Good to know. I've taken this code from Emitter, and didn't bother that much.
I know that this code needs revisiting, most probably with heavy changes.

L-value vs r-value needs far more attention that it currently received. That is probably after most of the CXX constructs are in place, and we are fixing all the leftovers.

Sources/CodeGen/CXX/CXXTranspiler.swift
dabrahams: Falling back on a "meaningless letter" is hugely beneficial to readability in many cases, especially when its type is evident from other context.  It can massively simplify expressions and reduce line wrapping.  Every character I have to process to read the code is a burden on my brain.  Each one we use should add value.  If it's just echoing the type name it adds approximately zero value, especially if its useful lifetime is short.

`pattern` is _not_ the role in this context; it is not just any pattern, but the pattern that binds the variables.  If we had a language feature where patterns could participate in an initializer expression then it would not be the only pattern here, so what _role_ does it play?  Well, it's the LHS of the initialization (if any). `lhs` would make a much better name than `pattern` (and that assertion would not be invalidated by delayed initialization if we had it, because _in general_ there is a RHS).  I can think of longer and more complicated names with more information in them, but `lhs` is a good one and probably the best choice given the trade-offs.

Sources/CodeGen/CXX/Decl/CXXLocalVarDecl.swift
dabrahams: No you don't, but in those cases it's good practice to write the method in an extension that creates the conformance to the protocol and implements only the protocol's requirements.  That way I can know which protocol to look at for the doc comment.

Sources/CodeGen/CXX/CXXTranspiler.swift
kyouko-taiga: > Falling back on a "meaningless letter" is hugely beneficial to readability in many cases

I don't disagree but I would rather reserve single-letter identifiers in situations where their uses are very close enough to their definition. In this particular case, I would not replace `capability` by a single-letter binding, for example.

> `pattern` is not the role in this context; it is not just any pattern, but the pattern that binds the variables.

Here, it's obvious that `decl` (admittedly a poor name choice) is the main parameter of the function since it's the only one. So, another way to see the problem is to think of the parts of the construct that we're visiting.

A binding declaration has a pattern and an initializer. When I'm reading a method that visit this construct, I'm not surprised to see local variables named after these parts because I see them as shorthands for `decl.pattern` and `decl.initializer`, respectively. Of course, if it were possible for a binding declaration to have multiple patterns, those parts would be named differently and so would the local variable corresponding to these parts.

I would be surprised by `lhs` because I think the name only makes sense if we're picturing the syntactic form of the construct. So there's arguably one more mental hop to make.

dabrahams: > I would be surprised by lhs because I think the name only makes sense if we're picturing the syntactic form of the construct. So there's arguably one more mental hop to make.

I find this objection hard to take seriously; we are operating on an enhanced abstract **syntax** tree, and `lhs`/`rhs` have been previously used in similar places in the code.

Moreover, the goal of naming is very much **not** to avoid surprising you.  The least surprising code in the world names every variable after its type, and the resulting code slips through one's brain without a hiccup: "yup that's an array, yup that's a person …".  Everything one reads looks fine.  You can find scads of this kind of thing on GitHub. In fact, more expressive names *create* an opportunity for surprise, for example when the additional meaning clashes with the reader's (mis)understandings.  Nobody's surprised when a `SourceRange` is called `site` because, well, it's a site in the code and we've used that name for properties.  That doesn't make it a good name for a local variable.

I think nobody needs a shorthand for `decl.pattern`, and that's especially true if you've chosen to rename `decl` as `d` ;-).  When I see `pattern` by itself, I want to know "what pattern?"

kyouko-taiga: I tried to express why I genuinely think that `pattern` is a helpful name for me, as a coder, but it seems I won't be able to convince you. That's okay. I won't be sad if this variable gets named `lhs` here and I hope you'll indulge the many places where I will likely name the pattern of a binding declaration `pattern`.

My larger point, which I think got lost, was that I don't think a general ban on naming a variable after its type is tenable. Sometime the type has a name that describes the role of its instances, at least in some contexts. That's the case of many properties in the AST. For example, the body of a `FunctionDecl` is named `body` and its type is named `FunctionDecl.Body`. I can't think of better names for these entities.

dabrahams: A less syntactic alternative to `lhs` would be `target`.  

You don't need to make that larger point, or worry about it being lost, because nobody's proposing a general ban on naming a variable after its type.  The point is that it is almost always suboptimal and extremely easy, just like using a pointer in C++. It looks trivially OK to say "now I'm going to make an array; let's call it array." We need to train ourselves to be a little alarmed when it happens so that we make the effort to find a better alternative.  

I agree that `body` is the best name for that property of a functiondecl, because that *is* the most specific description of the role that it plays.  In many languages, it's the type that would be named differently, something like  `Stmt` or `CompoundStmt`.  It's not (necessarily) a problem that the body of `FunctionDecl` is a type that's very specific to this use case and thus matches the property name… but it could evolve to be something different.

PR #255, opened 2023-01-23
Disable immediate diagnostic reporting

kyouko-taiga:


Sources/Core/Dianostics.swift
```diff
@@ -1,34 +1,32 @@
...
-  /// is `true`.
-  ///
-  /// - Note: the default is `false` because we build many more tests than command-line tools.
-  public init(reportingToStderr: Bool = false) {
-    self.reportingToStderr = reportingToStderr
+  /// Creates an instance.
+  public init() {}
+
+  /// Creates an instance reporting `d`.
+  public init(_ d: Diagnostic) {
```
dabrahams, line 16: Why move this code?

dabrahams (approved):

Sources/Core/Dianostics.swift
kyouko-taiga: I have this weird obsession about declaring initializers before any other method/computed property. I also like overloaded methods (which initializers are) to be grouped.

These rules are just personal preferences that I'm totally fine with other people not upholding.

dabrahams: I understand that, but we should really establish policies so that trivial reformatting doesn't occur.  These kinds of rearrangements have a cost in merge conflicts and ease-of-review.

kyouko-taiga: I don't mind starting and maintaining a document with all policies. I'm less motivated to extend swift-format so that it can enforce them automatically.

dabrahams:
If you're going to make a policy about initializer positioning, you need to consider whether to forbid them in extensions, especially when they are protocol requirements.  “Don't make rules that are hard to follow” is a good guideline.

PR #266, opened 2023-01-25
Emit cxx control stmt

lucteo:
Basic code to emit Val statements to CXX.

This does not include `for` and `yield` statements, as they require more complex semantics.

Add basic tests to cover some simple scenarios.


kyouko-taiga (approved):
LGTM

PR #267, opened 2023-01-25
Refactoring to separate the logic of writing CXX code from the AST nodes corresponding to CXX code

lucteo:
Currently, the CXX AST nodes also contain the logic of writing the CXX source code. This has multiple problems:
* coupling between two distinct responsibilities
* makes it harder to implement logic that needs to touch multiple types of nodes (e.g., precedence and associativity)
* makes it hard to implement formatting (e.g., indentation)

This refactoring splits the code generation logic from the ASR representation of CXX structures. All the CXX code is now written in the `CXXCodeWriter` struct. We also have a `CodeFormatter` struct that is able to format the generated code (indents, spaces, etc.)

The tests now will "essentialize" the code before comparing it. The "essentialize" process replaces any consecutive whitespace characters (including newlines and tabs) with a single space.


Sources/Utils/String+Extensions.swift
```diff
@@ -37,4 +37,28 @@ extension StringProtocol {
     return r
   }
 
+  /// Replace any sucession of whitespace characters (including newlines and tabs) with a single white space charater.
+  /// This will actually keep the non-space characters ingoring how these are spaced out.
+  public func essentialize() -> String {
```
kyouko-taiga, line 42: This should probably be renamed `canonicalize`.

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /// Adds new text to the output code.
+  mutating func write(_ text: String) {
+    checkAddIndent()
+    code.write(text)
+  }
+
+  /// Adds new text to the output code, with a new line at the end
+  mutating func writeLn(_ text: String) {
```
kyouko-taiga, line 32:
```suggestion
  mutating func writeLine(_ text: String) {
```

I would stay away from abbreviations when possible. I think we can afford the two extra characters here.

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
...
+    checkAddIndent()
+    code.write(text)
+    addNewLine()
+  }
+
+  /// Adds a space to the code that is currently written.
+  mutating func writeSpace() {
```
kyouko-taiga, line 39: Is this really necessary? I would prefer calling `write(" ")` for an economy of concepts.

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
...
+    checkAddIndent()
+    code.write(" ")
+  }
+
+  /// Write a new line to the current code.
+  /// If the next line contains visible characters, they would be indented.
+  mutating func addNewLine() {
```
kyouko-taiga, line 46:
```suggestion
  mutating func writeNewline() {
```

1. We don't capitalize "line" in "newline" in other parts of the codebase.
2. The function should be named `writeNewline` for consistency
3. As mentioned above, I would prefer `write("\n")` for an economy of concepts

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
...
+    code.write("\n")
+    shouldIntent = true
+  }
+
+  /// Increases the indent level and adds a new line.
+  /// Needs to be paired with a followup `exitIndentScope`.
+  mutating func enterIndentScope() {
```
kyouko-taiga, line 53:
```suggestion
  mutating func indent() {
```

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
...
+    curIndent += 1
+    addNewLine()
+  }
+
+  /// Decreases the indent level.
+  /// This does not add a new line; it is typcally called just after a new line.
+  /// Needs to be paired with a previous `enterIndentScope`.
+  mutating func exitIndentScope() {
```
kyouko-taiga, line 61:
```suggestion
  mutating func dedent() {
```

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
+/// Class responsible for the formatiing of the CXX code, as it is generated.
+///
+/// The callers will tell this class what tokens they want to write to the output CXX code,
+/// and this will ensure that proper formatting will be added.
+///
+/// This class will have ownership of the code that is being written.
+struct CodeFormatter: TextOutputStream {
+
+  /// The amount of indentation we apply to the code.
+  private var indentSize: Int
```
kyouko-taiga, line 10:
```suggestion
  private var indentation: Int
```

For consistency with other parts of the codebase

kyouko-taiga, line 10: I bet this will never ever change.

kyouko-taiga, line 10:
```suggestion
  private static let indentation: String = "  "
```
I bet this will be a constant that never ever changes.

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
...
+
+  /// The code that this object is writing.
+  var code: String = ""
+
+  /// The current level of indentation.
+  private var curIndent: Int = 0
```
kyouko-taiga, line 16:
```suggestion
  private var indentationLevel: Int = 0
```

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
...
+  }
+  /// Close a curly brace and decrease the indent.
+  mutating func endBrace() {
+    exitIndentScope()
+    write("}")
+  }
+
+  /// Called when writing some new code (everything except newlines) to ensure that we
+  /// are writing the indent when we are at the beginning of the line.
+  mutating private func checkAddIndent() {
```
kyouko-taiga, line 82:
```suggestion
  mutating private func writeIndentation() {
```

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
...
+  /// Decreases the indent level.
+  /// This does not add a new line; it is typcally called just after a new line.
+  /// Needs to be paired with a previous `enterIndentScope`.
+  mutating func exitIndentScope() {
+    curIndent -= 1
+    assert(curIndent >= 0)
+  }
+
+  /// Open a curly brace and increase the indent.
+  mutating func beginBrace(withSpaceBefore space: Bool = true) {
```
kyouko-taiga, line 67: I'm not convinced it is worth having a `space` parameter. It makes this function more complex only to avoid adding `write(" ")` at some call sites.

Sources/CodeGen/CXX/README.md
```diff
@@ -5,6 +5,12 @@
 The transpiler will take Val AST and convert into a CXX AST corresponding to C++ code.
 The C++ code is then generated by walking the CXX AST producing the textual representation for the Val module.
 
+The translation between Val AST and CXX AST is done in `CXXTranspiler` class.
```
kyouko-taiga, line 8:
```suggestion
The translation between Val AST and CXX AST is done by `CXXTranspiler.
```

Sources/CodeGen/CXX/README.md
```diff
@@ -5,6 +5,12 @@
...
+At this stage, we are looking at AST transformation and we completely ignore the format of the CXX code.
+
+The `CXXCodeWriter` class takes a CXX AST and produces actual CXX code, in a textual form.
```
kyouko-taiga, line 11:
```suggestion
`CXXCodeWriter` takes a CXX AST and produces actual CXX code, in a textual form.
```

Sources/Utils/String+Extensions.swift
```diff
@@ -37,4 +37,28 @@ extension StringProtocol {
...
+    var remaining = self[..<self.endIndex]
+    var res: String = ""
```
kyouko-taiga, line 44:
```suggestion
    var result: String = ""
```

Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -0,0 +1,334 @@
+/// Class used to write the output CXX code from the given CXX AST.
```
kyouko-taiga, line 1:
```suggestion
/// A type used to write the output CXX code from the given CXX AST.
```
Be careful, a class doesn't have the same semantics as a struct in Swift. Both terms are not interchangeable.

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
+/// Class responsible for the formatiing of the CXX code, as it is generated.
```
kyouko-taiga, line 1:
```suggestion
/// A type responsible for the formatting of the CXX code, as it is generated.
```

Sources/CodeGen/CXX/CodeFormatter.swift
```diff
@@ -0,0 +1,88 @@
...
+///
+/// The callers will tell this class what tokens they want to write to the output CXX code,
+/// and this will ensure that proper formatting will be added.
+///
+/// This class will have ownership of the code that is being written.
```
kyouko-taiga, line 6:
```suggestion
/// A type responsible for formatting CXX code, as it is generated.
///
/// Users of the type specify what tokens they want written to the generated CXX code
/// and delegate the formatting logic. Instances of this type have ownership of the code
/// they generate.
```

kyouko-taiga (changes requested):
See inline comments.

Sources/CodeGen/CXX/CodeFormatter.swift
lucteo: It means that this structure will contain the output code. Probably "contain" is a better word than "ownership"

Sources/CodeGen/CXX/CodeFormatter.swift
lucteo: There are multiple programming languages with `writeLn` function. But, yes, we are just adding an utility...

Sources/CodeGen/CXX/CodeFormatter.swift
kyouko-taiga: Does it matter to the caller? You're calling `checkAddIndent` anyway. It's not the responsibility of the caller to determine whether indentation should actually be written.

At the very least, we should come up with a better than than `checkAddIndent`. The name of the method suggests that something is being checked but there isn't any return value nor any assertion that might cause the program to stop.

Sources/CodeGen/CXX/CodeFormatter.swift
kyouko-taiga: You're not making it easier for the callers, because the API of `beginBrace` is more complex. So at the call site, one has to inspect that API to know how `beginBrace` should be called. That's not helped by the fact that the docs of your function doesn't even tell what `space` is doing.

kyouko-taiga (approved):
I left some additional non-blocking comments. Feel free to ignore.

kyouko-taiga (approved):

PR #268, opened 2023-01-25
Add basic support for generating CXX operator expressions

lucteo:
Adding basic structures  to represent CXX operators.

Add code to handle infix operators. This cannot be fully tested, as, at the point of writing this PR, the compiler didn't properly handle the use of infix operators.

Add support for handling prefix and postfix operators. This doesn't work at all, as, at the point of writing the PR, the compiler and standard library did not have any unary operator.

This is more a foundational patch, with more work needing to be done in this area.


Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -615,15 +616,35 @@ public struct CXXTranspiler {
     }
 
     // Emit infix operators.
-    // TODO: Do we need to check the base types?
+    let orig = AnyNodeID.TypedNode(expr)
     let name = nameOfDecl(calleeDecl)
```
kyouko-taiga, line 620: I don't want to block the PR over this but I think this is not the right strategy to transpile operator calls. First, it's a lot of boilerplate, and second it won't work for operators that don't have the same precedence/associativity in Val.

Sources/CodeGen/CXX/Expr/CXXExpr.swift
```diff
@@ -1,2 +1,14 @@
 /// A CXX expression.
-protocol CXXExpr: CXXNode {}
+protocol CXXExpr: CXXNode {
+
+  /// Gets the precedence of the expression.
+  /// This is used to determine the evaluation order of CXX expressions.
+  /// Expressions with smaller precedence values are evaluated in front of expressions with higher precendence values.
+  /// Following the conventions from https://en.cppreference.com/w/cpp/language/operator_precedence
+  func precedence() -> Int
```
kyouko-taiga, line 8: This should likely be a property rather than a function in Swift.
```suggestion
  var precedence: Int { get }
```

Sources/CodeGen/CXX/Expr/CXXExpr.swift
```diff
@@ -1,2 +1,14 @@
...
+
+  /// Returns `true` if this expression needs to be evaluated from left to right.
+  /// Determines the evaluation order for expressions of the same precedence value.
+  func isLeftToRight() -> Bool
```
kyouko-taiga, line 12: This should likely be a property rather than a function in Swift.
```suggestion
  var isLeftToRight: Bool { get }
```

Sources/CodeGen/CXX/Expr/CXXExpr.swift
kyouko-taiga, line 8: Would it be possible to use Val's [`PrecedenceGroup`](https://github.com/val-lang/val/blob/d7d10aa9b002bd1392a87b15f0a583b51681b1a2/Sources/Core/AST/PrecedenceGroup.swift)?

kyouko-taiga (commented):
There's a lot of boilerplate involved in dealing with C++ operators. I'm not sure that's the right approach. Specifically, keeping in mind that Val can defined custom operators and that the ones that are shared with C++ may not have the same precedence/associativity, I'm not sure we should invest that much on keeping a 1-to-1 translation of infix expressions. In other words, I would not be shocked if we transpired `x + y` to `x.infix_plus(y)`.

I recently implemented `Int.infix-` in the standard library. Perhaps it would be best to add a test case that uses this operator before merging.

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: As discussed in the call, the mapping between Val and C++ can be fixed with parenthesis.

I don't see how any other solution would avoid doing manual mapping between Val operators and C++ operators.

Sources/CodeGen/CXX/Expr/CXXExpr.swift
lucteo: Done.

lucteo: No. Val precedence is different from C++ precedence. We need C++ precedence to properly use C++ operators.

Sources/CodeGen/CXX/Expr/CXXExpr.swift
lucteo: Done.

lucteo:
Also added code to exercise the prefix `+` and `-` operators in the standard lib.
The infix operators don't work yet.
I've added some basic code in the standard lib to emulate postfix operators `++` and `--`, but, for some reason, I cannot properly implement them.

While we need to go back on the operators translation, at least there are some basic tests for them.

kyouko-taiga:
> The infix operators don't work yet.
> I've added some basic code in the standard lib to emulate postfix operators ++ and --, but, for some reason, I cannot properly implement them.

Are there issues with the parser and/or type checker? If so, could you open issues?

Sources/Core/TypedNode.swift
```diff
@@ -186,6 +186,16 @@ extension TypedNode where ID == NodeID<NameExpr> {
...
 
+  /// The domain expression node, if the domain is an expression
+  public var domainExpr: AnyExprID.TypedNode? {
+    switch syntax.domain {
+    case .expr(let expr):
+      return program[expr]
+    default:
+      return nil
+    }
+  }
```
lucteo, line 197: @kyouko-taiga , @dabrahams : this feels more complex than it needs to be. Do you know of way of simplifying it?

Library/Core/Int.val
```diff
@@ -30,6 +30,18 @@ public type Int {
...
+  public fun postfix++ () -> Self {
+    self.copy()
+  }
+
+  /// Decrements the current value, and return the old value.
+  /// TODO: fix me
+  public fun postfix-- () -> Self {
+    self.copy()
+  }
+
```
lucteo, line 44: @kyouko-taiga : should I remove these from here?

lucteo: removed

kyouko-taiga (approved):

PR #269, opened 2023-01-25
Make sure that generated CXX code compiles

lucteo:
Currently for a basic example like Factorial, we did not generate compilable C++ code.

This PR fixes that, by doing two things:
* ensuring the name of the module and the names of the written files match (don't include "Main.h", when we generate "main.h")
* When the program contains a `main` function, add a top-level `main` function in the generated CXX code that just calls the one from the appropriate module.


Examples/factorial.val
```diff
@@ -5,6 +5,6 @@ fun factorial(_ n: Int) -> Int {
   if n < 2 { 1 } else { n * factorial(n - 1) }
 }
 
-public fun main() {
-  let _ = factorial(6)
+public fun main() -> Int {
```
kyouko-taiga, line 8: I should add this to the specification, but that code should be illegal, I think. The problem is that `Int` represent 64-bit integers in Val (at least on 64-bit platforms), so it shouldn't be used to return exit statuses.

Eventually, the specification will say that the entry point of a Val program can be declared as either of the following options:
- `public fun main()`
- `public fun main() -> Int32`
- `public fun main(arguments: Arguments)`

For this PR, I would suggest truncating the result of `factorial` to `Int32`. There's an initializer that does that already in the standard library: https://github.com/val-lang/val/blob/ef26e55061399c0c446935e8562902be59c518f6/Library/Core/Int32.val#L9

Examples/factorial.val
```diff
@@ -5,6 +5,6 @@ fun factorial(_ n: Int) -> Int {
...
+  factorial(6)
```
kyouko-taiga, line 9: In fact, I think we shouldn't return the result of `factorial(6)` as the exit status of a program. Although it's traditionally a 32-bit integer, exit statuses can only range from 0 to 255, whereas 6! = 720.

kyouko-taiga (changes requested):
We can't use the result of `factorial(6)` as the exit status of a program for two reasons:
1. `factorial(_:)` returns a 64-bit integer and exit codes must be represented by `Int32`
2. 6! is too large if I'm not mistaken.

Examples/factorial.val
lucteo: I would argue that we need to be able to return a status code from Val to the `main` function. Indeed, on most platforms this gets truncated to an 8-bit (unsigned?) integer. Also, in C++, the type of `int` can actually be 64 bits wide (although not on any major platforms it's only 32 bits).

Examples/factorial.val
lucteo: I can do the first 2 cases as part of this PR, but I would defer the third one to later.

Examples/factorial.val
kyouko-taiga: I'm not contesting that we should be able to return a status code, only that it shouldn't be Val's `Int`.

As for the 8-bit requirement, I think we should keep in mind that we'll probably need to stick with something that works on most platforms if we want to implement tests down the line.

Examples/factorial.val
kyouko-taiga: That's completely fine. I'll define in the specification that `main() -> X` is ill-formed unless `X` is `Int32` anyway.

As for the third case, we'll have to decide what `Arguments` is exactly before we can do anything. So yes, let's skip it for now.

lucteo:
Looking at this PR, with the current translation of types, the only thing that I would change (more or less) is the example.
To properly fix the example, we need to change how we generate code for basic types. We agreed that this will be done soon after.

dabrahams:
Err, image

Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -58,6 +58,15 @@ public struct CXXCodeWriter {
 
     target.endBrace()
 
+    // If we need to generate the CXX entry point for the mdule, write the `main` function here with the proper body.
```
kyouko-taiga, line 61:
```suggestion
    // Write a CXX `main` function if the module has an entry point.
```

Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -58,6 +58,15 @@ public struct CXXCodeWriter {
...
+    if module.cxxEntryPointBody != nil {
+      target.writeNewline()
+      target.writeNewline()
+      target.write("int main()")
```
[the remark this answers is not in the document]
kyouko-taiga: Also, shall we consider using clang-format (see #295)?

Sources/CodeGen/CXX/CXXModule.swift
```diff
@@ -14,6 +14,9 @@ public struct CXXModule {
   /// The C++ top-level declarations for this module
   private(set) var cxxTopLevelDecls: [CXXTopLevelDecl] = []
 
+  /// If the module contains a `main` function, this will be set to call that function in what will be the CXX program entry point.
```
kyouko-taiga, line 17:
```suggestion
  /// Indicates whether this module contains an entry point (i.e., a `main` function).
```

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -98,6 +93,44 @@ public struct CXXTranspiler {
         parameters: cxxParams,
         body: cxxBody,
         original: decl))
+
+    // If this is the `main` function of the program, generate the body that needs to be the CXX entry point.
```
kyouko-taiga, line 97:
```suggestion
    // If this is the Val's entry point, generate a CXX entry point body that calls it.
```

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -98,6 +93,44 @@ public struct CXXTranspiler {
...
+    if identifier.description == "main" {
+      emit(programEntryPointFor: decl, withOutputType: output, into: &module)
+    }
+  }
+
+  /// Translate the body of the entry-point for the CXX program.
```
kyouko-taiga, line 103:
```suggestion
  /// Writes the body of a CXX entry point calling `decl` to `cxxEntryPointBody`, knowing it output type is
  /// expressed by `output` and writing CXX nodes to `module`.
```

kyouko-taiga: Why not returning the `CXXStmt` computed by the function rather than mutating `self`? If it was written that way, would the function still have to be mutable?

Also, why do you need to take `output` as parameter? Can't it be derived from `decl`? If the problem is that you don't want to repeat L49-57, a probably better solution would be to factor out this code in a function `(FunctionDecl.Typed) -> AnyType` so that you can write:
```swift
let output = CXXTypeExpr(outputType(decl), ast: program.ast, asReturnType: true)!
```

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -98,6 +93,44 @@ public struct CXXTranspiler {
...
+  private mutating func emit(
+    programEntryPointFor decl: FunctionDecl.Typed, withOutputType output: CXXTypeExpr,
+    into module: inout CXXModule
+  ) {
+    let orig = AnyNodeID.TypedNode(decl)
```
kyouko-taiga, line 108:
```suggestion
    let valNode = AnyNodeID.TypedNode(decl)
```

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -98,6 +93,44 @@ public struct CXXTranspiler {
...
+
+    // The expression that makes a call to the module `main` function.
+    let callMainExpr = CXXFunctionCallExpr(
```
kyouko-taiga, line 111:
```suggestion
    let callToMain = CXXFunctionCallExpr(
```

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -98,6 +93,44 @@ public struct CXXTranspiler {
...
+    var stmts: [CXXStmt] = []
+    if output.description == "int" {
+      // The expression returns an `int`; make that the program return code.
+      stmts.append(CXXReturnStmt(expr: callMainExpr, original: orig))
+    } else {
+      // Make a plain function call, discarding the result
+      stmts.append(
+        CXXExprStmt(expr: CXXVoidCast(baseExpr: callMainExpr, original: nil), original: orig))
+      // Add a `return 0` statment
+      stmts.append(
```
kyouko-taiga, line 130: Very minor: if I'm not mistaken, the standard says `main` doesn't have to return anything. You could potentially save this line for conciseness.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -98,6 +93,44 @@ public struct CXXTranspiler {
...
+        CXXReturnStmt(expr: CXXIntegerLiteralExpr(value: "0", original: nil), original: orig))
+    }
+    module.cxxEntryPointBody = CXXScopedBlock(stmts: stmts, original: orig)
```
kyouko-taiga, line 133: We should favor returning values over adding side effects in general.

kyouko-taiga (changes requested):
Great work!

I'm think the API of `emit(programEntryPointFor:withOutputType:into:)` could be reworked a little. I left inline comments. Other suggestions are minor.

Sources/CodeGen/CXX/CXXCodeWriter.swift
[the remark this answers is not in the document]
lucteo: That would break the indentation of `int main()`.
I could technically look for '\n' in the given string and break the string at those points, but that feels hackish.

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: I need to revisit this after #285 gets merged.

One possible problem I see is that we might have multiple `main` functions in the same module/program.

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: will be deleted after #285

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: I did not know that. Thank you! Will change.

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: Agree.
In this case I did not see a simple and good solution, without inventing a monadic behaviour. We need to parse multiple entities, and one of them might result in something that needs to be stored in the top-level module (which can be several levels up in the call hierarchy)

Do you see a simple/good solution?

Sources/CodeGen/CXX/CXXCodeWriter.swift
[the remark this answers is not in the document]
kyouko-taiga: I retract my comment but I really think this formatting stuff should be offloaded to `clang-format`. That's out of scope for this PR though.

Sources/CodeGen/CXX/CXXTranspiler.swift
kyouko-taiga: Multiple `main` functions in C++? I'll make sure it's not possible to have multiple main functions in Val (i.e., specify that a program with more than one `main` function is ill-formed).

Sources/CodeGen/CXX/CXXTranspiler.swift
kyouko-taiga: I'm not sure to follow. `emit(programEntryPointFor:withOutputType:into)` is called specifically to produce the `CXXStmt` that is assigned to `cxxEntryPointBody ` unconditionally. Why can't we have the caller(s) make this assignment?

Specifically, rather than doing:
```swift
if identifier.description == "main" {
  emit(programEntryPointFor: decl, withOutputType: output, into: &module)
}
```
we'd have:
```swift
if identifier.description == "main" {
  module.cxxEntryPointBody = emit(programEntryPointFor: decl, withOutputType: output, into: &module)
}
```

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: in my new PR, took advantage of this, and the fact that the `main` function is always at top-level. This way, I don't have to recurse down too much.

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: I went another way and made another traversal of the top-level decls. I took advantage of the fact that we only have `main` at top level. This way, the logic is simple and doesn't require any mutation (there are just a few levevels that I need to pass the restulting `CXXStmt` up.

lucteo:
Abandoing in the favor of #322

It was easier to start over than to do the merge, and then make the required changes

PR #271, opened 2023-01-25
Remove 'SourceRange.eliminateFIXME'

kyouko-taiga:
This PR removes the `eliminateFIXME` values that had been used to replace optional source ranges.

Note that we will probably need to determine better source sites for IR instructions during lowering in the future, so as to improve the accuracy of diagnostics and step-debugging.

### DIFF
--- a/Sources/Core/AST/Decl/BuiltinDecl.swift
+++ b/Sources/Core/AST/Decl/BuiltinDecl.swift
@@ -4,8 +4,10 @@ public struct BuiltinDecl: Decl {
   public let site: SourceRange
 
   public init() {
-    let f = SourceFile(synthesizedText: "/* built-in declaration */")
-    self.site = f.range(f.text.startIndex ..< f.text.endIndex)
+    self.site = Self.source.range(Self.source.text.startIndex ..< Self.source.text.endIndex)
   }
+
+  /// A synthesized source file to define the site of this type's instances.
+  private static let source = SourceFile(synthesizedText: "/* built-in declaration */")
 
 }
--- a/Sources/Core/AST/Decl/BuiltinDecl.swift
+++ b/Sources/Core/AST/Decl/BuiltinDecl.swift
@@ -1,13 +1,11 @@
 /// An abstract node representing a built-in declaration.
 public struct BuiltinDecl: Decl {
 
-  public let site: SourceRange
+  public init() {}
 
-  public init() {
-    self.site = Self.source.range(Self.source.text.startIndex ..< Self.source.text.endIndex)
-  }
+  public var site: SourceRange { Self.site }
 
-  /// A synthesized source file to define the site of this type's instances.
-  private static let source = SourceFile(synthesizedText: "/* built-in declaration */")
+  /// A synthesized site for all instances.
+  private static let site = SourceFile(synthesizedText: "/* built-in declaration */").wholeRange
 
 }
--- a/Sources/Core/AST/Decl/BuiltinDecl.swift
+++ b/Sources/Core/AST/Decl/BuiltinDecl.swift
@@ -3,9 +3,9 @@ public struct BuiltinDecl: Decl {
 
   public init() {}
 
-  public var site: SourceRange { Self.site }
+  public var site: SourceRange { Self.siteConstant }
 
   /// A synthesized site for all instances.
-private static let siteConstant = SourceFile(synthesizedText: "/* built-in declaration */").wholeRange
+  private static let siteConstant = SourceFile(synthesizedText: "/* built-in */").wholeRange
 
 }
--- a/Sources/Core/AST/Decl/BuiltinDecl.swift
+++ b/Sources/Core/AST/Decl/BuiltinDecl.swift
@@ -1,6 +1,11 @@
 /// An abstract node representing a built-in declaration.
 public struct BuiltinDecl: Decl {
 
-  public var site: SourceRange { .eliminateFIXME }
+  public let site: SourceRange
+
+  public init() {
+    let f = SourceFile(synthesizedText: "/* built-in declaration */")
+    self.site = f.range(f.text.startIndex ..< f.text.endIndex)
+  }
 
 }
--- a/Sources/Core/AST/Decl/SubscriptImplDecl.swift
+++ b/Sources/Core/AST/Decl/SubscriptImplDecl.swift
@@ -11,7 +11,7 @@ public struct SubscriptImplDecl: Decl, LexicalScope {
     case block(NodeID<BraceStmt>)
 
     /// The node wrappedby this instance.
-    public var node: AnyNodeID {
+    public var base: AnyNodeID {
       switch self {
       case .expr(let n):
         return AnyNodeID(n)
--- a/Sources/Core/AST/Decl/SubscriptImplDecl.swift
+++ b/Sources/Core/AST/Decl/SubscriptImplDecl.swift
@@ -10,6 +10,16 @@ public struct SubscriptImplDecl: Decl, LexicalScope {
     /// A block body.
     case block(NodeID<BraceStmt>)
 
+    /// The node wrappedby this instance.
+    public var node: AnyNodeID {
+      switch self {
+      case .expr(let n):
+        return AnyNodeID(n)
+      case .block(let n):
+        return AnyNodeID(n)
+      }
+    }
+
   }
 
   public let site: SourceRange
--- a/Sources/Core/AST/Decl/TopLevelDeclSet.swift
+++ b/Sources/Core/AST/Decl/TopLevelDeclSet.swift
@@ -1,15 +1,16 @@
 /// A set of declarations at the top-level of a source file.
 public struct TopLevelDeclSet: Node, LexicalScope {
 
+  public let site: SourceRange
+
   /// The declarations in the set.
   public private(set) var decls: [AnyDeclID]
 
   /// Creates an instance with the given properties.
-  public init(decls: [AnyDeclID] = []) {
+  public init(decls: [AnyDeclID], site: SourceRange) {
     self.decls = decls
+    self.site = site
   }
-
-  public var site: SourceRange { .eliminateFIXME }
 
   public func validateForm(in ast: AST, into diagnostics: inout Diagnostics) {
     for d in decls {
--- a/Sources/Core/TypedNode.swift
+++ b/Sources/Core/TypedNode.swift
@@ -24,6 +24,11 @@ public struct TypedNode<ID: NodeIDProtocol>: Hashable {
     self.id = id
   }
 
+  /// The site from which self was parsed.
+  public var site: SourceRange {
+    program.ast[id].site
+  }
+
   /// Equality comparison; only check the node ID.
   public static func == (lhs: TypedNode<ID>, rhs: TypedNode<ID>) -> Bool {
     lhs.id == rhs.id
--- a/Sources/Core/TypedNode.swift
+++ b/Sources/Core/TypedNode.swift
@@ -70,7 +70,7 @@ extension TypedNode where ID: ConcreteNodeID {
 
   /// The corresponding AST node.
   private var syntax: ID.Subject {
-    program.ast[NodeID(id)!]
+    program.ast[NodeID<ID.Subject>(rawValue: id.rawValue)]
   }
 
   /// Accesses the given member of the corresponding AST node.
--- a/Sources/Core/TypedNode.swift
+++ b/Sources/Core/TypedNode.swift
@@ -58,10 +58,12 @@ extension NodeIDProtocol {
 }
 
 extension TypedProgram {
+
   /// Bundles `id` together with `self`.
   public subscript<TargetID: NodeIDProtocol>(_ id: TargetID) -> TypedNode<TargetID> {
     TypedNode(id, in: self)
   }
+
 }
 
 extension TypedNode where ID: ConcreteNodeID {
@@ -107,6 +109,7 @@ extension TypedNode where ID: ConcreteNodeID {
     program = s.program
     id = myID
   }
+
 }
 
 extension TypedNode {
@@ -117,6 +120,7 @@ extension TypedNode {
 }
 
 extension TypedNode where ID: ScopeID {
+
   /// The parent scope, if any
   public var parent: AnyScopeID.TypedNode? {
     program.scopeToParent[id].map { .init($0, in: program) }
@@ -126,9 +130,11 @@ extension TypedNode where ID: ScopeID {
   public var decls: LazyMapCollection<[AnyDeclID], AnyDeclID.TypedNode> {
     program.scopeToDecls[id, default: []].lazy.map { .init($0, in: program) }
   }
+
 }
 
 extension TypedNode where ID: DeclID {
+
   /// The scope in which this declaration resides.
   public var scope: AnyScopeID.TypedNode {
     .init(program.declToScope[id]!, in: program)
@@ -144,23 +150,29 @@ extension TypedNode where ID: DeclID {
   public var implicitCaptures: [ImplicitCapture]? {
     program.implicitCaptures[id]
   }
+
 }
 
 extension TypedNode where ID == NodeID<VarDecl> {
+
   /// The binding decl containing this var.
   public var binding: BindingDecl.Typed {
     .init(program.varToBinding[id]!, in: program)
   }
+
 }
 
 extension TypedNode where ID: ExprID {
+
   /// The type of this expression
   public var type: AnyType {
     program.exprTypes[id]!
   }
+
 }
 
 extension TypedNode where ID == NodeID<NameExpr> {
+
   public enum Domain: Equatable {
 
     /// No domain.
@@ -188,6 +200,7 @@ extension TypedNode where ID == NodeID<NameExpr> {
 
   /// A reference to a declaration.
   public enum DeclRef: Hashable {
+
     /// A direct reference.
     case direct(AnyDeclID.TypedNode)
 
@@ -201,6 +214,7 @@ extension TypedNode where ID == NodeID<NameExpr> {
       case .member(let d): return d
       }
     }
+
   }
 
   /// The declaration of this name.
@@ -212,16 +226,20 @@ extension TypedNode where ID == NodeID<NameExpr> {
       return .member(program[decl])
     }
   }
+
 }
 
 extension TypedNode where ID: PatternID {
+
   /// The names associated with this pattern.
   public var names: [(path: [Int], pattern: NamePattern.Typed)] {
     program.ast.names(in: id).map({ (path: $0.path, pattern: program[$0.pattern]) })
   }
+
 }
 
 extension TypedNode where ID == NodeID<ModuleDecl> {
+
   /// Collection of (typed) top-level declarations of the module.
   public typealias TopLevelDecls = LazyMapSequence<
     FlattenSequence<
@@ -237,9 +255,11 @@ extension TypedNode where ID == NodeID<ModuleDecl> {
   public var topLevelDecls: TopLevelDecls {
     program.ast.topLevelDecls(id).map({ program[$0] })
   }
+
 }
 
 extension TypedNode where ID == NodeID<FunctionDecl> {
+
   /// The body of the function, containing typed information
   public enum Body {
 
@@ -264,6 +284,7 @@ extension TypedNode where ID == NodeID<FunctionDecl> {
       return .none
     }
   }
+
 }
 
 extension TypedNode where ID == NodeID<SequenceExpr> {
--- a/Sources/FrontEnd/Parse/Parser.swift
+++ b/Sources/FrontEnd/Parse/Parser.swift
@@ -1072,21 +1072,21 @@ public enum Parser {
       if state.ast[m].introducer.value == .memberwiseInit { return m }
     }
 
-    let site = state.emptyRange(at: startIndex)
+    let startOfTypeDecl = state.emptyRange(at: startIndex)
     let receiver = state.insert(
       synthesized: ParameterDecl(
-        identifier: SourceRepresentable(value: "self", range: site),
-        site: site))
+        identifier: SourceRepresentable(value: "self", range: startOfTypeDecl),
+        site: startOfTypeDecl))
     let m = state.insert(
       synthesized: InitializerDecl(
-        introducer: SourceRepresentable(value: .memberwiseInit, range: site),
+        introducer: SourceRepresentable(value: .memberwiseInit, range: startOfTypeDecl),
         attributes: [],
         accessModifier: nil,
         genericClause: nil,
         parameters: [],
         receiver: receiver,
         body: nil,
-        site: site))
+        site: startOfTypeDecl))
     members.append(AnyDeclID(m))
     return m
   }
--- a/Sources/FrontEnd/Parse/Parser.swift
+++ b/Sources/FrontEnd/Parse/Parser.swift
@@ -890,7 +890,7 @@ public enum Parser {
           in: &state,
           introducedBy: SourceRepresentable(
             value: .let,
-            range: state.emptyRange(at: state.ast[body.node].site.start)),
+            range: state.emptyRange(at: state.ast[body.base].site.start)),
           body: body,
           asNonStaticMember: isNonStaticMember)
         return [impl]
@@ -961,7 +961,7 @@ public enum Parser {
     }
 
     let site: SourceRange
-    if let n = body?.node {
+    if let n = body?.base {
       site = introducer.site.extended(toCover: state.ast[n].site)
     } else {
       site = introducer.site
--- a/Sources/FrontEnd/Parse/Parser.swift
+++ b/Sources/FrontEnd/Parse/Parser.swift
@@ -934,7 +934,8 @@ public enum Parser {
     }
   }
 
-  /// Inserts into `state.ast` a subscript implementation having the given `introducer` and `body` returning its ID.
+  /// Inserts a subscript having the given `introducer` and `body` into `state.ast` and returns its
+  /// ID.
   ///
   /// - Parameters:
   ///   - introducer: The introducer of the declaration, or `nil` if it is implicit. In that case,
--- a/Sources/FrontEnd/Parse/Parser.swift
+++ b/Sources/FrontEnd/Parse/Parser.swift
@@ -1042,7 +1042,7 @@ public enum Parser {
 
     // Retrieve or synthesize the type's memberwise initializer.
     var members = parts.1
-    let memberwiseInit = getOrCreateMemberwiseInitializer(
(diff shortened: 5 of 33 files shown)


Sources/Core/AST/Decl/BuiltinDecl.swift
```diff
@@ -1,6 +1,11 @@
 /// An abstract node representing a built-in declaration.
 public struct BuiltinDecl: Decl {
 
-  public var site: SourceRange { .eliminateFIXME }
+  public let site: SourceRange
+
+  public init() {
+    let f = SourceFile(synthesizedText: "/* built-in declaration */")
```
dabrahams, line 7: At lease this could be made a static constant, right?

Sources/Core/AST/Decl/TopLevelDeclSet.swift
```diff
@@ -1,16 +1,17 @@
 /// A set of declarations at the top-level of a source file.
 public struct TopLevelDeclSet: Node, LexicalScope {
 
+  public let site: SourceRange
```
dabrahams, line 4: It's not obvious how/why this change is necessary to serve the purpose of this PR, which makes me wonder if it's necessary.

Sources/Core/AST/Decl/SubscriptImplDecl.swift
```diff
@@ -10,6 +10,16 @@ public struct SubscriptImplDecl: Decl, LexicalScope {
...
+    /// The node wrappedby this instance.
+    public var node: AnyNodeID {
+      switch self {
+      case .expr(let n):
+        return AnyNodeID(n)
+      case .block(let n):
+        return AnyNodeID(n)
+      }
+    }
+
```
dabrahams, line 22: It's not obvious to me how this change serves the purpose of the PR, which makes me wonder if it's necessary.

Sources/Core/AST/Decl/SubscriptImplDecl.swift
```diff
@@ -10,6 +10,16 @@ public struct SubscriptImplDecl: Decl, LexicalScope {
     /// A block body.
     case block(NodeID<BraceStmt>)
 
+    /// The node wrappedby this instance.
```
dabrahams, line 13:
```suggestion
    /// The node wrapped by this instance.
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -30,33 +30,20 @@ public enum Parser {
...
-
-    state.ast[module].addSourceFile(translation)
-    ast = state.ast
-
-    return translation
-  }
-
-  /// Parses an instance of `TopLevelDeclSet`.
-  static func parseSourceFile(in state: inout ParserState) -> NodeID<TopLevelDeclSet> {
+    // Parse the input.
```
dabrahams, line 38: This change took a function I could digest at a glance, and turned it into a longer function.  `parseSourceFile` may not have been well-named, but are you sure this is an improvement?

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -566,8 +567,8 @@ public enum Parser {
     if state.isAtTypeScope && !prologue.isStatic {
       receiver = state.insert(
         synthesized: ParameterDecl(
-          identifier: SourceRepresentable(value: "self", range: .eliminateFIXME),
-          site: .eliminateFIXME))
+          identifier: SourceRepresentable(value: "self", range: head.introducerSite),
```
dabrahams, line 570: Seeing stuff like this makes me wonder if the site of ParameterDecl shouldn't be computed from the sites of its stored properties, rather than being stored.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -710,9 +711,10 @@ public enum Parser {
     withPrologue prologue: DeclPrologue,
     in state: inout ParserState
   ) throws -> NodeID<InitializerDecl>? {
-    // Parse the parts of the declaration.
-    let parser = (take(nameTokenWithValue: "memberwise").and(take(.`init`)))
-    guard let parts = try parser.parse(&state) else { return nil }
+    // Parse the introducer.
+    guard let a = state.take(nameTokenWithValue: "memberwise") else { return nil }
```
dabrahams, line 715: Are there changes in here that don't have to do with the point of this PR?  I don't know how to analyze this one.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -886,7 +888,9 @@ public enum Parser {
       if let body = try subscriptImplDeclBody.parse(&state) {
         let impl = try buildSubscriptImplDecl(
           in: &state,
-          withIntroducer: SourceRepresentable(value: .let, range: .eliminateFIXME),
+          introducedBy: SourceRepresentable(
```
dabrahams, line 891: The change of label is a good one, but seems irrelevant to the purpose of the PR.  I'd appreciate it if renaming steps could be made separately; ones like this don't even require review.  All in all, this is a very large PR and it would be easier to analyze if it only contained relevant changes.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -930,9 +934,18 @@ public enum Parser {
     }
   }
 
+  /// Inserts in `state.ast` a subscript implementation declaration introduced with the given
+  /// introducer and body and returns its ID.
```
dabrahams, line 938: “the given x”, where x is not in `code voice`, is always a poor tradeoff in a doc comment.  There are basically two good ways to handle cases like this:

```suggestion
  /// Inserts into `state.ast` a subscript implementation having the given `introducer` and `body` returning its ID.
  ```
  
  ```suggestion
  /// Inserts into `state.ast` a subscript implementation introduced by `intro`,  with body `b`, returning its ID.
  ```

File: Sources/Core/AST/Decl/SubscriptImplDecl.swift [hunk 10]
```diff
@@ -10,6 +10,16 @@ public struct SubscriptImplDecl: Decl, LexicalScope {
...
+    public var node: AnyNodeID {
```
dabrahams, line 14:
```suggestion
    public var base: AnyNodeID {
```

Sources/Core/AST/Decl/SubscriptImplDecl.swift
dabrahams, line 13: Separately, I'm wondering if `ImplDecl` is not redundant.  Isn't every `Impl` also a `Decl`?

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -930,9 +934,18 @@ public enum Parser {
...
+  ///
+  /// - Parameters:
+  ///   - introducer: The introducer of the declaration, or `nil` if it is implicit. In that case,
```
dabrahams, line 941: It's good to think about how to avoid these clauses when possible.  You could have omitted this description if in the summary fragment you'd said something like “explicitly introduced by `intro`.”

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -930,9 +934,18 @@ public enum Parser {
...
+  ///     it is synthesized as `let`.
+  ///   - body: The body of the declaration, or `nil` if that body should must be synthesized or
+  ///     if the declaration denotes a trait requirement.
+  /// - Requires: `body` is not `nil` if `introducer` is.
```
dabrahams, line 945:
```suggestion
  /// - Requires: if `introducer` is `nil`, body is non-`nil`.
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -1034,7 +1042,10 @@ public enum Parser {
 
     // Retrieve or synthesize the type's memberwise initializer.
     var members = parts.1
-    let memberwiseInit = findOrSynthesizeMemberwiseInitDecl(in: &members, updating: &state)
+    let memberwiseInit = getOrCreateMemberwiseInitializer(
```
dabrahams, line 1045: I'm not sure the memoization action of this function is useful information to the reader. What about just naming it `memberwiseInitializer`?

Otherwise, I'd delete the comment!

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -1051,29 +1062,31 @@ public enum Parser {
 
   /// Returns the first memberwise initializer declaration in `members` or synthesizes an implicit
   /// one, appends it into `members`, and returns it.
-  private static func findOrSynthesizeMemberwiseInitDecl(
-    in members: inout [AnyDeclID],
+  private static func getOrCreateMemberwiseInitializer(
```
dabrahams, line 1065: This again feels like a change that could have been separated from the PR.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -1051,29 +1062,31 @@ public enum Parser {
...
+    ofDeclStartingAt startIndex: SourceFile.Index,
+    among members: inout [AnyDeclID],
     updating state: inout ParserState
   ) -> NodeID<InitializerDecl> {
     for member in members where member.kind == InitializerDecl.self {
       let m = NodeID<InitializerDecl>(rawValue: member.rawValue)
       if state.ast[m].introducer.value == .memberwiseInit { return m }
     }
 
+    let site = state.emptyRange(at: startIndex)
```
dabrahams, line 1075: You can do better.  Maybe

```suggestion
    let startOfTypeDecl = state.emptyRange(at: startIndex)
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2841,12 +2853,14 @@ public enum Parser {
     (maybe(passingConvention)
       .andCollapsingSoftFailures(expr)
       .map({ (state, tree) -> NodeID<ParameterTypeExpr> in
-        state.insert(
+        let site = state.range(from: tree.0?.site.start ?? state.ast[tree.1].site.start)
```
dabrahams, line 2856: Can we do better?  The site of _what_?  It's not just any old site.

Sources/FrontEnd/Parse/ParserState.swift
```diff
@@ -117,6 +117,11 @@ struct ParserState {
     lexer.sourceCode.range(startIndex ..< currentIndex)
   }
 
+  /// Returns an empty site starting and ending at `index`.
```
dabrahams, line 120:
```suggestion
  /// Returns a range starting and ending at `index`.
```

It's hard to explain, but the use of the word site here comes off a bit weird.  A site is usally _of_ or _for_ something.  This one is disembodied.  Also "empty site" is not really a thing, English-wise.

Tests/ValTests/ASTTests.swift
```diff
@@ -18,17 +18,19 @@ final class ASTTests: XCTestCase {
     var ast = AST()
 
     // Create a module declarations.
+    let input = SourceFile(synthesizedText: "")
+    let site = input.range(input.text.startIndex ..< input.text.endIndex)
```
dabrahams, line 22: May as well use your new `emptyRange` function, right?
Can't we do better than `site` here?

Tests/ValTests/ASTTests.swift
```diff
@@ -39,17 +41,21 @@ final class ASTTests: XCTestCase {
     var ast = AST()
 
     // Create a module.
+    let input = SourceFile(synthesizedText: "")
+    let site = input.range(input.text.startIndex ..< input.text.endIndex)
```
dabrahams, line 45: ditto to the above comments.  But this is a repeated code pattern; please factor it out!

dabrahams (commented):
These are just comments for the first commit in the PR.  More to review.

Sources/Core/AST/Decl/TopLevelDeclSet.swift
kyouko-taiga: I had to remove the computed property, which was:
```var
public var site: SourceRange { .eliminateFIXME }
```

Sources/Core/AST/Decl/BuiltinDecl.swift
kyouko-taiga: Yep.

Sources/Core/AST/Decl/SubscriptImplDecl.swift
kyouko-taiga: This change eliminates boilerplate in the parser to retrieve the site of a subscript's body.

Sources/FrontEnd/Parse/Parser.swift
kyouko-taiga: It is not obvious (or even possible) to compute the site of a node from its properties if those are nodes referred by their ID. That's because we need the AST to read the site of an expression represented by `AnyExprID`.

Sources/FrontEnd/Parse/Parser.swift
kyouko-taiga: The changes are related. They serve to identify the start index of the introducer's site.

Sources/FrontEnd/Parse/Parser.swift
kyouko-taiga: Noted.

Sources/Core/AST/Decl/SubscriptImplDecl.swift
kyouko-taiga: That's a good point. I'll make a separate PR.

Sources/FrontEnd/Parse/Parser.swift
kyouko-taiga: Agreed. In this particular case, I went back and forth with the docs until I decided the clause was necessary to clarify what happens if `introducer` is `nil`. I felt like something saying "explicitly introduced by `introducer`" was missing the fact that the introducer is synthesized as `let` if kept implicit.

Sources/FrontEnd/Parse/Parser.swift
kyouko-taiga: I hesitated. The fact that you're asking the same question leads me to think it's fine to just use `memberwiseInitializer `.

Sources/FrontEnd/Parse/Parser.swift
kyouko-taiga: I'll use just `s`

Tests/ValTests/ASTTests.swift
kyouko-taiga: Sure, but factoring out the code pattern is probably not something that should be part of this PR.

Tests/ValTests/ASTTests.swift
kyouko-taiga: `emptyRange` is a method of `ParserState`. It's never used anywhere else because the other synthetic sites we create (for `builtinDecl` and `moduleDecl`) are not empty so I'm note sure we should make it a method of `SourceFile`.

Perhaps it would be more relevant to define a property `SourceFile.wholeRange`.

dabrahams: `emptyRange` doesn't make sense to me as a method of `ParserState`, though.  What is an empty range of a parser state?  Could be anything.  `state.source.emptyRange` makes sense.

Sources/Core/AST/Decl/SubscriptImplDecl.swift
dabrahams: I eventually realized that, and thought I had deleted the comment.  Sorry!

Sources/Core/AST/Decl/TopLevelDeclSet.swift
dabrahams: Oh, the github diff didn't make the corresponding removal obvious.

Tests/ValTests/ASTTests.swift
kyouko-taiga: That's just a helper for a longer expression. The parser state has many helper methods that operate on the lever and/or the input being parsed.

Sources/FrontEnd/Parse/Parser.swift
kyouko-taiga: Yes. I originally kept `parseSourceFile` only to realize it was requiring too many input/outputs to resembles an actual independent unit of computation.

kyouko-taiga: FWIW, `parse(_:into:in:diagnostics:)` is barely longer.

Tests/ValTests/ASTTests.swift
dabrahams: I don't think that's a good idea, FWIW.  It prevents the parser state from representing a clear abstraction; instead it becomes something like the superposition of its parts.

Sources/IR/Operands/Instruction/CondBranchInstruction.swift
```diff
@@ -14,7 +14,14 @@ public struct CondBranchInstruction: Instruction {
   /// The target of the branch if `condition` is false.
   public let targetIfFalse: Block.ID
 
-  public let range: SourceRange?
+  public let site: SourceRange
+
+  init(condition: Operand, targetIfTrue: Block.ID, targetIfFalse: Block.ID, site: SourceRange) {
```
dabrahams, line 19: Curious as to the reason you added this (or omitted it before).

dabrahams (commented):
Comments on 2nd commit

dabrahams (approved):

dabrahams:
**Sorry: Approval was premature; not done reviewing yet**

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
```diff
@@ -910,7 +910,9 @@ extension TypeChecker {
         else { return .incomparable }
 
         // Rank the candidates.
-        switch (refines(lhs, rhs, scope: scope), refines(rhs, lhs, scope: scope)) {
+        let lRefinesR = refines(lhs, rhs, scope: scope, site: program.ast[n].site)
```
dabrahams, line 913: This call reads a bit weirdly.  `site` is a poor label here (arguably `scope` too); what is its role? 

Maybe something like:
```swift
does(lhs, refine: rhs, in: scope, diagnosingConstraintErrorsAt: program.ast[n].site)
```
?

Sources/Core/TypedNode.swift
```diff
@@ -58,10 +58,12 @@ extension NodeIDProtocol {
 }
 
 extension TypedProgram {
+
```
dabrahams, line 66: This commit shoulda been a separate PR.

Sources/Core/TypedNode.swift
```diff
@@ -70,7 +70,7 @@ extension TypedNode where ID: ConcreteNodeID {
 
   /// The corresponding AST node.
   private var syntax: ID.Subject {
-    program.ast[NodeID(id)!]
```
dabrahams, line 71: This commit shoulda been a separate PR.

Can you explain why this change is an improvement?  To me it looks like premature optimization that makes the code harder to read.

Sources/Core/TypedNode.swift
```diff
@@ -24,6 +24,11 @@ public struct TypedNode<ID: NodeIDProtocol>: Hashable {
     self.id = id
   }
 
+  /// The site from which self was parsed.
+  public var site: SourceRange {
```
dabrahams, line 28: isn't this property automatically reflected by the `@dynamicMemberLookup` stuff?

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -934,8 +934,7 @@ public enum Parser {
     }
   }
 
-  /// Inserts in `state.ast` a subscript implementation declaration introduced with the given
-  /// introducer and body and returns its ID.
```
dabrahams: Yeah, that phrasing is pretty awkward, sorry.  Let me try again:

```suggestion
/// Inserts a subscript having the given `introducer` and `body` into `state.ast`, returning its ID.
```

Sources/Core/AST/Decl/BuiltinDecl.swift
```diff
@@ -1,13 +1,11 @@
...
 
-  public init() {
-    self.site = Self.source.range(Self.source.text.startIndex ..< Self.source.text.endIndex)
-  }
+  public var site: SourceRange { Self.site }
 
-  /// A synthesized source file to define the site of this type's instances.
-  private static let source = SourceFile(synthesizedText: "/* built-in declaration */")
+  /// A synthesized site for all instances.
+  private static let site = SourceFile(synthesizedText: "/* built-in declaration */").wholeRange
```
dabrahams, line 9:
```suggestion
private static let siteConstant = SourceFile(synthesizedText: "/* built-in declaration */").wholeRange
```
While it's possible to have static and non-static variables with the same name, it can be confusing.

dabrahams (approved):
OK, review complete. Good job; I had no idea this bit would require so much work!

Sources/IR/Analysis/DefiniteInitializationPass.swift
```diff
@@ -118,10 +118,11 @@ public struct DefiniteInitializationPass: TransformPass {
               // at the end of the predecessor.
               if lhs != rhs {
                 // Deinitialize the object at the end of the predecessor.
+                let terminator = module.terminator(
+                  of: Block.ID(function: functionID, address: predecessor))!
                 module.insert(
```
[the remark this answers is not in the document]
kyouko-taiga: That's a possibility, but note that not all instructions will be created from an AST node. Many will be synthesized from IR passes so the initializer that accepts a SourceRange will always be necessary.

One thing that worries me a little is that it will be difficult to maintain consistent site information once we implement aggressive optimizations that can move instructions around. But that's for future me to address.

Sources/IR/Operands/Instruction/StoreInstruction.swift
```diff
@@ -9,12 +9,12 @@ public struct StoreInstruction: Instruction {
...
 
-  public let range: SourceRange?
+  public let site: SourceRange
 
-  init(_ object: Operand, to target: Operand, range: SourceRange? = nil) {
+  init(_ object: Operand, to target: Operand, site: SourceRange) {
     self.object = object
     self.target = target
-    self.range = range
+    self.site = site
```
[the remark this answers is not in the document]
kyouko-taiga: Okay, sounds reasonable.

Sources/IR/Analysis/ImplicitReturnInsertionPass.swift
```diff
@@ -17,22 +17,22 @@ public struct ImplicitReturnInsertionPass: TransformPass {
...
+      // Nothing to do if there's a terminator instruction.
+      if last.isTerminator { continue }
+
+      if expectedReturnType == .void {
         // Insert missing return instruction.
         module.insert(
-          ReturnInstruction(),
-          at: module.globalEndIndex(of: Block.ID(function: functionID, address: i.address)))
+          ReturnInstruction(site: last.site),
+          at: module.globalEndIndex(of: Block.ID(function: functionID, address: block.address)))
```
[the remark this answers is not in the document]
kyouko-taiga: Yes, that's because they are indices of the doubly linked list, which have an address property denoting their stable position.

Sources/IR/Operands/Instruction/CondBranchInstruction.swift
kyouko-taiga: It was missing before. I added it for consistency. I should probably have done so in another PR. Sorry.

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
kyouko-taiga: Labels looks good but is the function name intentional?

Sources/Core/TypedNode.swift
kyouko-taiga: It's for consistency. I use this initializer every time I know that the conversion will succeed. That makes refactoring simpler if one day we want to change that initializer or how it's used.

Sources/Core/TypedNode.swift
kyouko-taiga: No, it wasn't! Swift was complaining `AnyExprID` was not conforming to `ConcreteNodeID`. I think it's because `site` is a requirement on `Node`, making it part of the existential API.

Sources/FrontEnd/Parse/Parser.swift
dabrahams: Yeah, the finding best representation for the AST and balancing that with other factors continues to be a challenge.

Sources/IR/Analysis/DefiniteInitializationPass.swift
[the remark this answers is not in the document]
dabrahams: I'm not sure that storing source ranges in instructions will make sense in the long run.  Another possiblity is that the IR data structure maintains a mapping between indices or addresses and source locations.  When synthesizing instructions they adopt the source ranges of their "anchor" instructions, i.e. the ones they are inserted `before` or `after`.

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
dabrahams: The base name? I went back and forth about it, and hated most of the things I came up with a little bit.  Probably this one most of all.  I want some standard naming convention for context-requiring functions like this…  How about

```
inFact(lhs, refines: rhs, in: scope, diagnosingConstraintErrorsAt: program.ast[n].site)
```

The need for the context distorts the API a bit because in principle it should be a method of `lhs`.  Could do something like

```
self.constraintSource = program.ast[n].site
lhs.refines(rhs, in: scope, context: &self)
```

🤷 it's hard.  This is the sort of problem `TypedNode` was designed to solve more elegantly.

Sources/Core/TypedNode.swift
dabrahams: I don't buy it.  `T(x)!` is the canonical form for “the conversion could fail dynamically but I know it will succeed.”  It communicates exactly that to the reader, whereas what you wrote is a head-scratcher.

Sources/FrontEnd/Parse/Parser.swift
dabrahams: I guess I was assuming the reader would know that information from the language semantics.

Sources/Core/TypedNode.swift
kyouko-taiga: Okay, you may have a point here, but it remains that I consistently used `NodeID(rawValue:)` everywhere else. So I stand by my point: it will be easier to refactor all those common call sites.

dabrahams: Maybe so, but readability is way more important than ease of refactoring.

kyouko-taiga: Let's open a separate issue if you want to revisit this initializer. Maybe the solution is to remove it altogether. In any case, that's unrelated to this PR.

PR #272, opened 2023-01-25
Consolidate diagnostic constructors

kyouko-taiga:


Sources/FrontEnd/Parse/Parser+Diagnostics.swift
```diff
@@ -85,48 +85,8 @@ extension Diagnostic {
     .error("initializer declaration is not allowed here", at: d.introducer.site)
   }
 
-  static func error(unexpectedMemberModifier m: SourceRepresentable<MemberModifier>) -> Diagnostic {
```
dabrahams, line 88: I don't understand what happened to all of these errors you removed.  If they were unused, then please tighten our minimum code coverage bound.

dabrahams (approved):
Just one comment inline

Sources/FrontEnd/Parse/Parser+Diagnostics.swift
kyouko-taiga: They were duplicated. The exact same error constructors existed in `Parser+Diagnostics.swift` and `AST+Diagnostics.swift`. If you look at the changes in the latter, you'll a bunch of methods becoming public.

I don't think coverage changed. If test were running a constructor in `Parser+Diagnostics.swift`, now they run its equivalent in `AST+Diagnostics.swift`. The only thing that changed should be the number of times the latter is called.

dabrahams: You can validate whether coverage changed by looking at the execution log of the ubuntu runner.  I recently tightened the minimum by 2 points after doing that.  Wow, we're up again:  https://github.com/val-lang/val/actions/runs/4038457794/jobs/6942468136#step:10:12

PR #273, opened 2023-01-25
Code style policy

kyouko-taiga:


Docs/CodeFormatting.md
```diff
@@ -1,51 +1,140 @@
...
+We currently use the version `swift-format` tagged with the commit [0bc2f03](https://github.com/apple/swift-format/commit/0bc2f0381c72d66a949254af22208a81377cf717).
+The best way to get `swift-format` is to obtain it from its [official repository](https://github.com/apple/swift-format) and compile it manually.
+
+**Warning:** please make sure to get the version specified above.
+Different versions of `swift-format` may produce different results.
+
+### Running `swift-format`
+
+The script `Tools/run-swift-format.sh` can be used to check and format code.
+It tool will search for `swift-format` in your path `PATH` or at `/usr/local/bin/swift-format`.
```
dabrahams, line 22:
```suggestion
It will search for `swift-format` in your path `PATH` or at `/usr/local/bin/swift-format`.
```

Docs/CodeFormatting.md
```diff
@@ -1,51 +1,140 @@
...
 ${WORKSPACE_PATH}/../../../Tools/run-swift-format.sh fix
 ```
+
+## Source files
+
+### File encoding
+
+All source files are encoded in UTF-8.
+
+### File names
```
dabrahams, line 65: This isn't exactly `CodeFormatting`, FWIW.

Docs/CodeFormatting.md
```diff
@@ -1,51 +1,140 @@
...
+
+### Column limit
+
+The column limit is 100 characters.
+Any line that would exceed this limit must be line-wrapped.
+
+## Formatting type declarations
+
+Exactly one empty line is added before the first member of a type, protocol, or trait declaration.
+Excatly one empty line is added after each member of a type, protocol, or trait declaration.
```
dabrahams, line 99:
```suggestion
Exactly one empty line is added after each member of a type, protocol, or trait declaration.
```

Docs/CodeFormatting.md
```diff
@@ -1,51 +1,140 @@
...
+For example:
+
+```swift
+struct Foo {
+
+  let bar: Int
```
dabrahams, line 105: IMO showing declarations without doc comments sets a misleading example; we'll then go back and tell people they need to document everything.

Docs/CodeFormatting.md
```diff
@@ -1,51 +1,140 @@
...
+  
+  let ham: Int
+
+}
+```
+
+Nested types are declared first, then static members, then non-static members.
```
dabrahams, line 112: I don't like this ordering much.  The way I see it, there are two ways to read a type declaration, one as an implementor, and the other as a user.  For the user, the public API is paramount and should logically come first, but we're not taking that tack, which is fine with me.

For an implementor, the stored properties (or enum cases) and core initializers are the most important thing for setting context and should be grouped together at the beginning.  Having to declare nested types at the beginning of the the body of the type forces me to emphasize things that, often, should be de-emphasized.  For example, IMO putting [this](https://github.com/val-lang/val/blob/66e51f7d0fb9139cf266dcee47813f75e78c4490/Sources/Core/SourceFile.swift#L230) at the beginning of `SourceFile` would significantly harm readability.  Static members are often the least essential to the nature of a type and I'd usually like to see them last.

dabrahams (commented):
We should talk these things over a bit I think.

Docs/CodeFormatting.md
kyouko-taiga: Good point.

Docs/CodeFormatting.md
kyouko-taiga: Okay, sounds perfectly reasonable. I think we've been sticking pretty close to the "implementor way" in existing code.

kyouko-taiga:
> We should talk these things over a bit I think.

Yes. This document is just a WIP.

Docs/CodeFormatting.md
```diff
@@ -1,51 +1,140 @@
...
+A file that extends an existing type with a protocol (Swift) or trait (Val) conformance is named with a combination of the type name and the protocol or trait name, joined with a plus (`+`) sign.
+For more complex situations, exercise your best judgment.
+
+For example:
+- A Swift file defining a type named `MyType` is named `MyType.swift`.
+- A Swift file defining how `MyType` conforms to `Equatable` is named `MyType+Equatable.swift`.
+
+Avoid defining multiple types, protocols, or traits in the same file unless they are scoped by a main entity or meant to be used only in that file.
+
+### Whitespace characters and indentation
```
lucteo, line 79: I would expect `swift-format` to handle these things. I would not put here rules that are already handled by `swift-format`.

My principle: all styling rules should be automated, so that the programmer can focus on the semantics.

Docs/CodeFormatting.md
lucteo: I would rename this document to "CodingStyle" or something.

Docs/CodeFormatting.md
```diff
@@ -1,51 +1,140 @@
...
+- Tab characters are not used for indentation.
+
+Indentation is set at two horizontal spaces.
+
+## General formatting
+
+### Column limit
+
+The column limit is 100 characters.
+Any line that would exceed this limit must be line-wrapped.
```
lucteo, line 94: This should also be handled by `swift-format`. But I have the feeling that `swift-format` doesn't wrap all the lines with more than 100 characters (I think comments are not following this limit)

dabrahams: Yes it doesn't wrap comments currently.  They have an open issue about that.

dabrahams:
We should probably try to resolve any outstanding issues and land this…

kyouko-taiga:
Okay, I'll put myself to it.

Docs/CodingStyle.md
```diff
@@ -0,0 +1,246 @@
+# Code Formatting
+
+This document describes some guidelines to write code for this project.
+The purpose of these conventions is to keep the style of the codebase consistent.
+Contributors are kindly asked to follow these guidelines and format their code accordingly before submitting new pull requests.
+
+As all set of conventions, many of the rules described in this document are arbitrary and based on personal preferences or simply follow common wisdom.
+For code written in Swift, most of the syntactic conventions are simply inherited from the style currently applied by `swift-format` configured with `.swift-format.json` at the root of this repository.
+Other rules are heavily inspired by [Google's Swift Style Guide](https://google.github.io/swift/).
```
dabrahams, line 9: IMO we should cite the Swift API guidelines as a baseline.

Docs/CodingStyle.md
```diff
@@ -0,0 +1,246 @@
...
+In Swift, use parentheses to declare closure parameters and only use `$0` for simple closures that fit a single line.
+For example:
+
+```swift
+/// Returns `polygon` translated by `d`.
+func translate(polygon: [Vector2], by d: Vector2) -> [Vector2] {
+  polygon.map({ $0 + d })
+}
+
+/// Returns the edges of `polygon` clipped in `area`.
```
dabrahams, line 235:
```suggestion
/// Returns the edges of `polygon` clipped to `area`.
```

dabrahams (approved):
Good starting point!

Docs/CodingStyle.md
```diff
@@ -0,0 +1,246 @@
...
+  /// The length of the vector along the x-axis.
+  public var x: Double = 0.0
+  
+  /// The length of the vector along the y-axis.
+  public var y: Double = 0.0
+
+}
+```
+
+Prefer declaring core type aliases first, followed by stored properties or enum cases, followed by core initializers, followed other members.
```
lucteo, line 111: (minor) I'm not a fan of specifying order of members. I think we should follow whatever order makes the most sense for the type declaration in question. For example, for some types it makes sense to focus on public members, then list private members; but this is not applicable for all types.

kyouko-taiga: I didn't say anything about access levels and also picked the word "prefer" intentionally. The type checker doesn't follow this pattern for justifiable reasons for example.

This guideline only serves to quickly identify the value of a type, which is useful to reason about the way conformance to `Equatable` and `Hashable` may be synthesized.

Docs/CodingStyle.md
```diff
@@ -0,0 +1,246 @@
...
+Tools/run-swift-format.sh lint
+```
+
+To automatically format the code, run the following command:
+
+```bash
+Tools/run-swift-format.sh fix
+```
+
+Please note that it's possible some of your code may still require some hand formatting after running the above command.
```
nickpdemarco, line 36: It might be useful to provide an example of what you mean by this - all I get from this is an ominous feeling that even if I use the tooling, I'm going to be wrong.

kyouko-taiga: Unfortunately, your feeling is accurate. Unless we want to invest a lot of effort into swift-format, I don't think all the rules can be automatically enforced.

PR #276, opened 2023-01-26
Refactor pattern constraints

kyouko-taiga:
This PR refactors how constraint generation operates on patterns for consistency with expressions.

### DIFF
--- a/Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
+++ b/Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
@@ -8,7 +8,7 @@ extension TypeChecker {
   struct InferenceFacts {
 
     /// A map from visited expression to its inferred type.
-    private(set) var inferredTypes = ExprProperty<AnyType>()
+    private(set) var inferredExprTypes = ExprProperty<AnyType>()
 
     /// The set of type constraints between the types involved in the visited expressions.
     private(set) var constraints: [Constraint] = []
@@ -28,13 +28,13 @@ extension TypeChecker {
     fileprivate mutating func assignErrorType<ID: ExprID>(to subject: ID) -> AnyType {
       foundConflict = true
       let ty = AnyType.error
-      inferredTypes[subject] = ty
+      inferredExprTypes[subject] = ty
       return ty
     }
 
     /// Assigns `type` to `subject`.
     fileprivate mutating func assign<ID: ExprID>(_ type: AnyType, to subject: ID) {
-      inferredTypes[subject] = type
+      inferredExprTypes[subject] = type
     }
 
     /// Constrains `subject` to have type `inferredType` and returns either `inferredType` or the
@@ -47,7 +47,7 @@ extension TypeChecker {
       in ast: AST,
       toHaveType inferredType: T
     ) -> AnyType {
-      if let ty = inferredTypes[subject] {
+      if let ty = inferredExprTypes[subject] {
         if ty != inferredType {
           constraints.append(
             EqualityConstraint(
@@ -57,7 +57,7 @@ extension TypeChecker {
         return ty
       } else {
         let ty = ^inferredType
-        inferredTypes[subject] = ty
+        inferredExprTypes[subject] = ty
         return ty
       }
     }
@@ -87,7 +87,7 @@ extension TypeChecker {
     expecting expectedType: AnyType?,
     updating facts: inout InferenceFacts
   ) -> AnyType {
-    defer { assert(facts.inferredTypes[subject] != nil) }
+    defer { assert(facts.inferredExprTypes[subject] != nil) }
 
     switch subject.kind {
     case BooleanLiteralExpr.self:
@@ -351,7 +351,7 @@ extension TypeChecker {
     // Case 3c
     addDiagnostic(
       .error(
-        nonCallableType: facts.inferredTypes[syntax.callee]!,
+        nonCallableType: facts.inferredExprTypes[syntax.callee]!,
         at: program.ast[syntax.callee].site))
     return facts.assignErrorType(to: subject)
   }
@@ -601,7 +601,7 @@ extension TypeChecker {
     //      type and use it at the callee's type.
 
     // Case 1
-    if facts.inferredTypes[syntax.callee]!.isError {
+    if facts.inferredExprTypes[syntax.callee]!.isError {
       return facts.assignErrorType(to: subject)
     }
 
@@ -626,7 +626,7 @@ extension TypeChecker {
     }
 
     // Case 3a
-    if let callable = SubscriptType(facts.inferredTypes[syntax.callee]!) {
+    if let callable = SubscriptType(facts.inferredExprTypes[syntax.callee]!) {
       if parametersMatching(
         arguments: syntax.arguments, of: syntax.callee, inScope: scope,
         expecting: callable.inputs, updating: &facts)
@@ -655,12 +655,13 @@ extension TypeChecker {
     }
 
     // Case 3c
-    let candidates = lookup("[]", memberOf: facts.inferredTypes[syntax.callee]!, inScope: scope)
+    let candidates = lookup(
+      "[]", memberOf: facts.inferredExprTypes[syntax.callee]!, inScope: scope)
     switch candidates.count {
     case 0:
       addDiagnostic(
         .error(
-          noUnnamedSubscriptsIn: facts.inferredTypes[syntax.callee]!,
+          noUnnamedSubscriptsIn: facts.inferredExprTypes[syntax.callee]!,
           at: program.ast[syntax.callee].site))
       return facts.assignErrorType(to: subject)
 
--- a/Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
+++ b/Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
@@ -16,9 +16,9 @@ extension TypeChecker {
     /// True iff a constraint could not be solved.
     private(set) var foundConflict = false
 
-    /// Creates empty base of facts assigning `type` to `subject` if `type` is not `nil`, or an
-    /// empty base otherwise.
-    init<ID: ExprID>(assigning type: AnyType?, to subject: ID) {
+    /// Creates a base of facts assigning `type` to `subject` if `type` is not `nil`, or an empty
+    /// base otherwise.
+    fileprivate init<ID: ExprID>(assigning type: AnyType?, to subject: ID) {
       if let t = type {
         assign(t, to: subject)
       }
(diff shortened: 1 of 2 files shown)


Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -8,33 +8,39 @@ extension TypeChecker {
   struct InferenceFacts {
```
dabrahams, line 8:
```suggestion
  struct Inferences {
```

or maybe:

```suggestion
  struct InferenceSet {
```

IIUC, these "facts" are not necessarily true facts as we'll explore inferences that turn out not to work out in the end.  Aren't they just inferences?

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -8,33 +8,39 @@ extension TypeChecker {
...
 
     /// A map from visited expression to its inferred type.
-    private(set) var inferredTypes = ExprProperty<AnyType>()
+    private(set) var inferredExprTypes = ExprProperty<AnyType>()
```
dabrahams, line 11: I don't think this name change is an improvement. The type already tells me that it's a property of `Expr`s.  Just look at the use-sites and I think it becomes pretty obvious.  Also I think based on the use sites maybe it should not be plural.  That is generally the case with property maps, which act like functions.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -8,33 +8,39 @@ extension TypeChecker {
...
+
+    /// The list of variable declarations visited during constraint generation.
+    private(set) var visitedVarDecls: [NodeID<VarDecl>] = []
```
dabrahams, line 14:
```suggestion
    private(set) var visited: [NodeID<VarDecl>] = []
```

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -8,33 +8,39 @@ extension TypeChecker {
...
     private(set) var constraints: [Constraint] = []
 
     /// True iff a constraint could not be solved.
     private(set) var foundConflict = false
 
-    /// Creates empty base of facts assigning `type` to `subject` if `type` is not `nil`, or an
-    /// empty base otherwise.
-    init<ID: ExprID>(assigning type: AnyType?, to subject: ID) {
+    /// Creates a base of facts assigning `type` to `subject` if `type` is not `nil`, or an empty
+    /// base otherwise.
```
dabrahams, line 23: I'm not sure this initializer represents an abstraction, as its documentation is essentially pseudocode.
IMO it might be better as:

```
    fileprivate init<ID: ExprID>(assigning type: AnyType, to subject: ID) {
      assign(t, to: subject)
    }
```

with use sites explicitly dealing with `nil` as necessary.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -740,6 +766,170 @@ extension TypeChecker {
     return facts.constrain(subject, in: program.ast, toHaveType: TupleType(elementTypes))
   }
 
+  // MARK: Patterns
+
+  mutating func infer(
+    typeOf subject: AnyPatternID,
+    inScope scope: AnyScopeID,
```
dabrahams, line 773:
```suggestion
    in scope: AnyScopeID,
```

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -740,6 +766,170 @@ extension TypeChecker {
...
+    expecting expectedType: AnyType?
+  ) -> (type: AnyType, facts: InferenceFacts) {
+    var facts = InferenceFacts()
+    let inferredType = infer(
+      typeOf: subject, inScope: AnyScopeID(scope), expecting: expectedType, updating: &facts)
+    return (inferredType, facts)
+  }
+
+  /// Returns the type of `subject`, writing facts about its sub-expressions in `facts`.
```
dabrahams, line 782: Needs to say what `scope` and `expecting` are.  You might look at Magoo for how I phrased doc for similar functions in the type checker.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -740,6 +766,170 @@ extension TypeChecker {
     return facts.constrain(subject, in: program.ast, toHaveType: TupleType(elementTypes))
   }
 
+  // MARK: Patterns
+
```
dabrahams, line 770: Doc comment needed

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -740,6 +766,170 @@ extension TypeChecker {
...
+  mutating func infer(
+    typeOf subject: AnyPatternID,
```
dabrahams, line 772:
```suggestion
  mutating func inferType(
    of subject: AnyPatternID,
```

This is basic Swift API guidelines stuff.
Arguably, the label could be dropped.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -740,6 +766,170 @@ extension TypeChecker {
...
+    return (inferredType, facts)
+  }
+
+  /// Returns the type of `subject`, writing facts about its sub-expressions in `facts`.
+  private mutating func infer(
+    typeOf subject: AnyPatternID,
+    inScope scope: AnyScopeID,
+    expecting expectedType: AnyType?,
+    updating facts: inout InferenceFacts
+  ) -> AnyType {
```
dabrahams, line 788:
```suggestion
  private mutating func inferredType(
    of subject: AnyPatternID,
    in scope: AnyScopeID,
    expecting expectedType: AnyType?,
    cache: inout InferenceFacts  // <======= maybe?
  ) -> AnyType {
```

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -740,6 +766,170 @@ extension TypeChecker {
...
+      unreachable()
+    }
+  }
+
+  private mutating func infer(
+    typeOfBindingPattern subject: NodeID<BindingPattern>,
+    inScope scope: AnyScopeID,
+    expecting expectedType: AnyType?,
+    updating facts: inout InferenceFacts
+  ) -> AnyType {
```
dabrahams, line 818: Needs a doc comment

```suggestion
  private mutating func inferredType(
    ofBindingPattern subject: NodeID<BindingPattern>,
    in scope: AnyScopeID,
    expecting expectedType: AnyType?,
    cache: inout InferenceFacts
  ) -> AnyType {
```

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -268,6 +268,15 @@ public struct TypeChecker {
     return success
   }
 
+  /// Sets the realized type of `d`.
```
dabrahams, line 271: to what?

dabrahams (approved):
Comments inline

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
kyouko-taiga: I picked this name from the world of logic programming. In Prolog, for example, a program is often described as a based of facts (and rules). In a sense, the generator is building small logic programs that are "ran" by the the constraint solver. Rules are hard-coded in our cases; they correspond to the rules of Val's type system. So we're only producing facts.

That being said, I'll concede the final say to the English speaker.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
kyouko-taiga: You're right. That's leftover from when I thought we needed `inferredDeclTypes` as well. I'll revert.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
kyouko-taiga: I get that `VarDecl` is in the type of the property, but I think this changes makes the call sites less clear. Typically, one doesn't understand what's being visited in:
```swift
for d in facts.visited {
  ...
}
```

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
kyouko-taiga: If you care about this change, please one an issue. `inScope` is being used consistently in most of the type checker's method. Renaming it would deserve its own PR.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
kyouko-taiga: Those methods are only the individual cases of a documented function that switches over all different patterns. I have not copy-pasted documentation of the top-level visitor (here and in other places) because I think it would only make documentation harder to keep in sync.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
kyouko-taiga: I planned on making the change, but that should be in a separate PR.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
kyouko-taiga: The inference facts are not a cache. They are the logic program being constructed.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
dabrahams: Except that, IIUC, prolog will immediately reject any facts that cause conflicts so that the experience of the user is that all these things actually are facts, while we speculatively explore "facts" that cause conflicts.   I think that term is an appropriate description for the Prolog user.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
dabrahams: OK

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
dabrahams: Absolutely. #24

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
dabrahams: The doc comment could be "`/// see `functionName`"

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
dabrahams: Why I wrote “maybe” there.  the other suggestions in that hunk I'm much more sure of.

Issue #79, opened 2022-11-10
Use the term "name" more consistently


kyouko-taiga:
A name (`Sources/Compiler/AST/Name.swift`) is a value that uniquely identifies a single entity in a given scope. It consists of

- a stem (or base name),
- an array of labels,
- an operator notation (optional), and
- an implement introducer (optional)

For example, `foo(a:b:).sink` is a name whose stem is `foo`, the labels are `a, b` and the introducer is `sink`. `infix+(_:)` is another name whose stem is `infix`, the labels are two `nil` values and the notation is `infix.`

Unfortunately, the term "name" is used rather inconsistently in the code.

### In declaration nodes:

AST nodes representing declarations parsed from source files often have an `identifier` property that describe the stem of some name. This property should be called `stem` to better describe its role.

### In `SingleEntityDecl`:

Declarations that introduce a single entity conform to `SingleEntityDecl`, which requires a property `name`. That property is typed as `String`, but should be typed by `Name` instead.

_Note: It shouldn't be `SourceRepresentable<Name>` because the name's value is not necessarily source representable. Only the stem is._ 

### In the type checker:

There is a family of `lookup` methods in `TypeChecker` that accept a parameter `name` which should be called `stem` instead, to better describe its role (note: the first phase of name lookup ignores labels, notations and introducers).

The method `TypeChecker.resolve` should accept a name rather than its individual parts.

PR #279, opened 2023-01-26
Use 'name' and 'stem' more consistently

kyouko-taiga:
See #79


Sources/Core/AST/Decl/SingleEntityDecl.swift
```diff
@@ -1,7 +1,7 @@
 /// A declaration that introduces a single entity.
 public protocol SingleEntityDecl: Decl {
 
-  /// The name of the entity introduced by the declaration.
-  var name: String { get }
+  /// The stem identifier of the entity introduced by the declaration.
```
dabrahams, line 4: Problem is, I don't know what a "stem identifier" is supposed to be.  This would be a good place to define it with a slightly more verbose comment.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -2108,22 +2109,23 @@ public struct TypeChecker {
...
     for module in program.ast.modules where module != root {
-      matches.formUnion(names(introducedIn: module)[name, default: []])
+      matches.formUnion(names(introducedIn: module)[baseName, default: []])
     }
 
     return matches
   }
 
-  /// Returns the declarations that introduce `name` in the declaration space of `lookupContext`.
+  /// Returns the declarations that introduce a name whose stem is `baseName` in the declaration
```
dabrahams, line 2125: Again, I don't know what "stem" is.  Seems like we should just settle on one of the two terms "base" or "stem" and define it.

dabrahams (approved):
comments inline

Sources/Core/AST/Decl/SingleEntityDecl.swift
kyouko-taiga: The description should belong to `Name`. But know that "stem identifier" is defined in the specification.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
kyouko-taiga: Knowing that the parts of a name are defined in the specification (see General concepts > Preamble), what do you think we should do?

dabrahams: Not sure; my understanding was that the spec was completely outdated.  Going to try to look; the link [here](https://www.val-lang.dev/) sends me to the repo, which makes finding sections hard.  Do we not have a formatted copy online somewhere?  OK, looking for `stem` there, even the spec doesn't even say in `a.b` which name is the stem.  Should we just say "the last component" of the name?  Otherwise, I guess I'd call `b` the “leaf name.”  That at least connotes which end of the name we're talking about.

kyouko-taiga: A stem has nothing to do with a qualified name. `a.b` has two stems, `a` and `b`. A name (not necessarily qualified) has a stem and optional labels, a notation, and an introducer.

Frankly I tried several times to document that the best I could. It seems I have failed but I don't know what to do. Please take the lead on this issue if you know how to address it.

dabrahams: I don't, yet, because I still don't know what stem is.  Is it the part before the open parenthesis of a function name?

kyouko-taiga: Yes.
- `a` => stem is `a`
- `a(foo:_:)` => stem is `a`
- `a.let` => stem is `a`
- `infix+` => stem is `a`
- `Int.infix+` => stem of `Int` is `Int` and stem of `infix+` is `+`

dabrahams: > `infix+` => stem is `a`

How's that work?

kyouko-taiga: Sorry, stem is `+`. I fixed it.

dabrahams: So would it be fair to say that the stem is the key used in name lookup?

kyouko-taiga: Well, name lookup will also consider all other parts of the name. That's because `let f = Int.infix-` is unambiguous while `let g = Int.-` is (bc both `prefix-` and `infix-` are valid candidates). So I'm not sure it's fair to say the stem is the key.

To be precise, name lookup will use as much information as it can. Obviously no one wants to write things like `&pointer.advance(by:).inout(by: 10)` so it has to deal with a possibly incomplete name and typically only get a stem.

PR #280, opened 2023-01-27
Fix operator binding

kyouko-taiga:
This PR fixes a bug that would cause the type checker to pick wrong candidates when solving member constraints, due to operator notation being ignored.

Concretely, the changes make this program to be properly marked ill-typed:
```swift
public fun main() {
  let _ = 99+
}
```


Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
```diff
@@ -424,23 +424,11 @@ struct ConstraintSolver {
...
-    })
-
-    // Catch uses of static members on instances.
-    if nonStaticMatches.isEmpty && !allMatches.isEmpty {
-      diagnostics.append(
-        .error(
-          illegalUseOfStaticMember: goal.memberName, onInstanceOf: goal.subject,
-          at: goal.cause.site))
-    }
+    // Search for members with the specified name.
```
dabrahams, line 427: Is this comment adding anything beyond what the docs for `lookup` say?

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -3646,9 +3627,41 @@ public struct TypeChecker {
     return r
   }
 
+  /// Returns the declarations in `decls` whose names match `n`.
```
dabrahams, line 3630:
```suggestion
  /// Returns the members of `decls` whose names match `n`.
```

But what does it mean to "match" a name?  Is this

```suggestion
  /// Returns the members of `decls` named `n`.
```

?  If not, where is the meaning of "match" defined?

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -3646,9 +3627,41 @@ public struct TypeChecker {
...
+  ///
+  /// - Requires: The base name of all declarations in `decls` is equal to `n-stem`
```
dabrahams, line 3632: Unclear; what's `n-stem`?

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -3646,9 +3627,41 @@ public struct TypeChecker {
...
+  mutating func filter(decls: DeclSet, matching n: Name) -> [AnyDeclID] {
```
dabrahams, line 3633:
```suggestion
mutating func filter(_ decls: DeclSet, named n: Name) -> [AnyDeclID] {
```

should this be a method of `DeclSet`?  If `DeclSet` is (or has) a `Collection`, you could use the existing `filter` with a suitable predicate.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -3646,9 +3627,41 @@ public struct TypeChecker {
...
+      })
+    }
+
+    return matches
+  }
+
   /// Filters the function, method, and subscript declarations in `decls` whose argument labels
   /// match `labels`.
-  mutating func filter(decls: inout [AnyDeclID], withLabels labels: [String?]) {
+  private mutating func filter(decls: inout [AnyDeclID], withLabels labels: [String?]) {
```
dabrahams, line 3664:
```suggestion
  private mutating func remove(elementsNotLabeled desiredLabels: [String?], from decls: inout [AnyDeclID]) {
```

Of course, I am obliged to ask if this should be a method of `[AnyDeclID]`.  Presumably you could use `remove(where:)` and have the typechecker supply a suitable predicate?

dabrahams (approved):
Comments inline

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
kyouko-taiga: It makes it easier to understand what these 3 lines are about without having to go read the documentation of `lookup`. I usually try to add comments so that we can quickly parse the body of a function and get a rough pseudo-code like description of its implementation.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
kyouko-taiga: Was supposed to be `n.stem`

Sources/FrontEnd/TypeChecking/TypeChecker.swift
kyouko-taiga: I think it would make the API worse because we need `self` mutating.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
kyouko-taiga: The reason is that `self` (i.e. the `TypeChecker`) has to be projected mutably in the function and that this filter is only relevant in the context of the type checker. It feels like a lot of boilerplate to declare a conditional extension of `[AnyDecl]` as opposed to simply declaring the method here. I'm not opposed to it though, tradeoffs I guess.

Perhaps `remove(where:)` sidesteps the issue because `self` gets projected in the closure implicitly. The issue is that the predicate is a bit long for an inline closure, and declaring a method just to be a predicate of a single call to `remove(where:)` seems more convoluted than the current solution.

Let me know what you think.

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
dabrahams: Sure, when it's 3 lines, but now it's one line; that's what I was asking about.

kyouko-taiga: Well, it's a rather long line involving two calls to non-trivial methods. I think the comment still help but I can remove it if you think it's needless.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
dabrahams: One of the problems with the name `filter` is that we talk about “filtering out” things that are removed, and yet the closure identifies things that will be kept.  That's what made me start on this, but not what I think is the most important issue: it's that we should use the existing algorithms where they are appropriate.  They are valuable abstractions, and that we use them is part of the Better Code philosophy.  It's very odd in that context to see something removing elements from a collection but not using the standard algorithm.  I think it's worth trying the `remove(where:)` formulation.

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
dabrahams: I'm asking you to consider it.  If you saw that line by itself, would you know what it's doing right away?

Also, in general, please avoid "the specified."  It's a waste of verbiage, and in this case, it seems really inappropriate, because the caller didn't specify the name AFAICT.  

Also, this is srsly nitpicky, but "search for" is a little indirect because it sounds like it's just cranking through the options but not doing anything with them.  If I was going to write the comment as an active thing, I'd use "gather."  But I'd be more likely to use that line to document the meaning of `matches`, which removes the verb.

dabrahams:
Weird; where are the CI checks?

kyouko-taiga:
No checks because it's not merging in main.

dabrahams:
We should probably run CI for everything, no?

kyouko-taiga:
> We should probably run CI for everything, no?

Done.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -3627,87 +3625,75 @@ public struct TypeChecker {
...
+  /// - Requires: The base name of `d` is equal to `n.stem`
+  mutating func decl(_ d: AnyDeclID, named n: Name) -> AnyDeclID? {
+    if !n.labels.isEmpty && !decl(d, isLabeledWith: n.labels) {
+      return nil
     }
 
-    // Filter out candidates whose operator notation does not match.
-    if let notation = n.notation {
-      filter(decls: &matches, withNotation: notation)
+    if let notation = n.notation, !decl(d, hasNotation: notation) {
```
dabrahams, line 3636:
```suggestion
    if let x = n.notation, operatorNotation(d) != x {
```
But I bet you don't even need to unwrap:

```suggestion
    if n.notation != operatorNotation(d) {
```

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
```diff
@@ -425,9 +425,8 @@ struct ConstraintSolver {
     }
 
     // Search for members with the specified name.
-    let matches = checker.filter(
-      decls: checker.lookup(goal.memberName.stem, memberOf: goal.subject, inScope: scope),
-      named: goal.memberName)
+    let matches = checker.lookup(goal.memberName.stem, memberOf: goal.subject, inScope: scope)
+      .compactMap({ checker.decl($0, named: goal.memberName) })
```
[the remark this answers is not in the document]
dabrahams: It's no longer clear to me that the comment is unneeded, but I think if you took the `ifNamed` suggestion and renamed `lookup` to something that describes what it actually returns.  Maybe
```
checker.members(of: goal.subject, named: goal.memberName.stem, in: scope)
```

or, if more accurate,

```
checker.members(of: goal.subject, nameStem: goal.memberName.stem, in: scope)
```

dabrahams (approved):
Making progress!

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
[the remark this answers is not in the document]
kyouko-taiga: > I think it would be cleared up if the label were not `named:` but `ifNamed:`

I think `ifNamed` isn't a perfect fit. `checker.decl(d, named: n)` is returning:
- `d` if it's name is `n`; or
- `m` if `d` is a method declaration and `m` is an implementation/clause/variant whose name is `n`; or
- `nil` otherwise.

Perhaps the name of the method itself should be changed.

> renamed lookup to something that describes what it actually returns.

I am torn here. I think there's value of having all lookup methods share a common prefix, because it helps identifying that they are related. I'm trying to build a kind of vocabulary of the kind of queries that are done during type checking. "Lookup" is one of them.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -3627,87 +3625,75 @@ public struct TypeChecker {
...
-        guard let type = LambdaType(realize(initializerDecl: NodeID(rawValue: d.rawValue))) else {
-          return false
-        }
-        return labels == type.inputs.map({ (p) in p.label })
+  /// Returns `true` iff `d` is a function, method, and subscript declarations whose argument
+  /// labels are `l`.
+  private mutating func decl(_ d: AnyDeclID, isLabeledWith l: [String?]) -> Bool {
+    switch d.kind {
+    case FunctionDecl.self:
+      let decl = program.ast[NodeID<FunctionDecl>(rawValue: d.rawValue)]
```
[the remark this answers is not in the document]
kyouko-taiga: > rawValue init is an obsolete idiom now innit?

Yes, it's been changed in another merged PR already.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -3627,87 +3625,75 @@ public struct TypeChecker {
...
+      return l == decl.parameters.map({ (p) in program.ast[p].label?.value })
 
-      case MethodDecl.self:
-        let decl = program.ast[NodeID<MethodDecl>(rawValue: d.rawValue)]
-        return labels == decl.parameters.map({ (p) in program.ast[p].label?.value })
+    case InitializerDecl.self:
+      guard let type = LambdaType(realize(initializerDecl: NodeID(rawValue: d.rawValue))) else {
```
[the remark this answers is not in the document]
kyouko-taiga: Problem is that without it you'll have `LambdaType(realize(NodeID<InitializerDecl>(d)!))`.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
kyouko-taiga: Unfortunately I still need to unwrap because if `d` has a notation and `n` hasn't, then the method should return `d` anyway. That's why it was called "match" before: it's not a strict comparison.

kyouko-taiga:
Perhaps `TypeChecker.decl` and/or its labels could be renamed to improve the use sites. I propose to push this change to another patch as it's getting unrelated to purpose of this specific PR.

dabrahams:
n/p, can you file the issue?

PR #281, opened 2023-01-28
Properly serialize/deserialize source files

dabrahams:
`SourceFile` contents are now serialized separately from the AST, with the AST just recording an integer ID for each distinct source file.  The lifetime of contents are now maintained by a threadsafe global dictionary.
  
I apologize in advance for the fact that the constituent commits are not particularly meaningful.  I will try to reorganize if any reviewer wants me to; otherwise it should probably be squashed.

### DIFF
--- a/Sources/Utils/FactoryInitializable.swift
+++ b/Sources/Utils/FactoryInitializable.swift
@@ -1,0 +1,35 @@
+/// Classes whose initializers actually create derived classes
+public protocol FactoryInitializable {
+  /// The type of the least-derived class declared to be FactoryInitializable.
+  ///
+  /// - Warning: Do not define this in your FactoryInitializable type!
+  associatedtype FactoryBase: AnyObject, FactoryInitializable = Self
+  // This associatedtype is a trick that captures `Self` at the point where
+  // `FactoryInitializable` enters a class hierarchy; in other contexts, `Self`
+  // refers to the most-derived type.
+}
+
+extension FactoryInitializable where Self: AnyObject {
+  /// Optimally “creates” an instance that is just another reference to `me`.
+  ///
+  /// - Requires: `me is Self`.
+  ///
+  /// Taking `FactoryBase` as a parameter prevents, at compile-time, the
+  /// category of bugs where `me` is not derived from the least-derived ancestor
+  /// of `Self` conforming to `FactoryInitializable`.
+  ///
+  /// However, there are still ways `me` might not be derived from `Self`.  If
+  /// you have factory initializers at more than one level of your class
+  /// hierarchy and you can't control exactly what is passed here, use
+  /// `init(aliasing:)` instead.
+  public init(unsafelyAliasing me: FactoryBase) {
+    self = unsafeDowncast(me, to: Self.self)
+  }
+
+  /// Safely “creates” an instance that is just another reference to `me`.
+  ///
+  /// - Requires: `me is Self`.
+  public init(aliasing me: FactoryBase) {
+    self = me as! Self
+  }
+}
--- a/Sources/Core/AST/AST.swift
+++ b/Sources/Core/AST/AST.swift
@@ -1,19 +1,49 @@
+import Foundation
 import Utils
 import ValModule
 
 /// An abstract syntax tree.
-public struct AST: Codable {
+public struct AST {
+
+  /// The stored representation of an AST; distinguished for encoding/decoding purposes.
+  private struct Storage: Codable {
+    /// The nodes in `self`.
+    public var nodes: [AnyNode] = [AnyNode(BuiltinDecl())]
+
+    /// The indices of the modules.
+    ///
+    /// - Invariant: All referred modules have a different name.
+    public var modules: [NodeID<ModuleDecl>] = []
+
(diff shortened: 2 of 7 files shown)


Sources/Utils/FactoryInitializable.swift
```diff
@@ -0,0 +1,35 @@
+/// Classes whose initializers actually create derived classes
+public protocol FactoryInitializable {
```
kyouko-taiga, line 2: Docs mentions classes but the protocol is not bound by `AnyObject`.

Sources/Utils/FactoryInitializable.swift
```diff
@@ -0,0 +1,35 @@
...
+  ///
+  /// Taking `FactoryBase` as a parameter prevents, at compile-time, the
+  /// category of bugs where `me` is not derived from the least-derived ancestor
+  /// of `Self` conforming to `FactoryInitializable`.
+  ///
+  /// However, there are still ways `me` might not be derived from `Self`.  If
+  /// you have factory initializers at more than one level of your class
+  /// hierarchy and you can't control exactly what is passed here, use
+  /// `init(aliasing:)` instead.
+  public init(unsafelyAliasing me: FactoryBase) {
```
kyouko-taiga, line 25: The purpose of this initializer and `init(aliasing:)` is really not obvious to me. Given some value `t` of type `T: FactoryInitializable`, what's the difference between `let s = t` and `let s = T(unsafelyAliasing: t)`?

kyouko-taiga: Would it be possible to avoid all those shenanigans if the classes that conform to `FactoryInitializable` are required to be final?

kyouko-taiga (commented):
Awesome work!

I left a couple of inline comments about the `FactoriyInitializable` protocol. I also have another design proposal for the way encoders/decoders are extended to support the ad-hoc state:

I'm not a fan of having to make encoders/decoders conform to `Stateful[Encoder|Decoder]` and wonder if we could instead define wrappers parameterized by encoders/decoders. These wrappers could be defined solely for the AST and would address cases where one inadvertently uses an encoder that hasn't been properly prepared by `forAST`.

Another orthogonal but compatible thought is that perhaps we shouldn't make `AST` (publicly) conform to `Codable`. Since the only thing that actually has to be `Codable` is the AST's storage, which is private, I think we could simply provide encoding/decoding methods on the AST that take care of preparing the encoder/decoder:

```swift
extension AST {
  encode<E: TopLevelEncoder>(_ e: E) -> E.Output { ... }
}
```

Sources/Utils/FactoryInitializable.swift
dabrahams: Is that a problem?

Sources/Utils/FactoryInitializable.swift
dabrahams: > what's the difference between let s = t and let s = T(unsafelyAliasing: t)?

Nothing, but you can't use either `s` as the result of a class initialization.

> Would it be possible to avoid all those shenanigans if the classes that conform to FactoryInitializable are required to be final?

I don't know, but that would undermine one of the main use cases for this component (which I wrote long ago, not for this project): `SomeBase(argumentThatDeterminesTheInitializedDerivedClass)`

dabrahams:
> I'm not a fan of having to make encoders/decoders conform to Stateful[Encoder|Decoder] 

Me neither, but it's the best I was able to do after many days working on it.

> and wonder if we could instead define wrappers parameterized by encoders/decoders. 

I don't know what you have in mind specifically, but I did try some designs that used generic wrappers.  They didn't work because the `Encoder`/`Decoder` passed to a `Codable` type's methods is not the same one used by the outer code (`JSONEncoder` in our examples).  In fact `JSONEncoder` is-not-a `Encoder`!

> These wrappers could be defined solely for the AST and would address cases where one inadvertently uses an encoder that hasn't been properly prepared by forAST.

You're welcome to try your own experiment to prove out this approach, but I tried hard to avoid all the drawbacks of this one and couldn't.

> Another orthogonal but compatible thought is that perhaps we shouldn't make AST (publicly) conform to Codable. Since the only thing that actually has to be Codable is the AST's storage, which is private, I think we could simply provide encoding/decoding methods on the AST that take care of preparing the encoder/decoder:

I don't think it works.  Try it if you must, but I warn you that you may waste a bunch of time going down a horrendous rabbit hole.  🐟`able` is maddeningly limited in what it will let you do.  You have to encode both the AST's storage, **and the source files**.  A TopLevelCoder doesn't let you encode two things, because it doesn't update an archive; it produces a finished one.

Sources/Utils/FactoryInitializable.swift
dabrahams: Is there even a way to express the requirement that conforming classes be `final`?

kyouko-taiga (approved):

Sources/Utils/FactoryInitializable.swift
kyouko-taiga: Ugh! Classes really make Swift much uglier.

Issue #270, opened 2023-01-25
Address code review issues for #253


lucteo:
Address the comments from @dabrahams from #253 :
* the use of `capability` variable name
* the use of `stmts` variable name
* use of doc-comments

PR #285, opened 2023-01-29
Rearrange CXX transpiler code (#270)

lucteo:
Rearrange the CXX transpiler code for better readability


lucteo:
@dabrahams : there are still some things that may be improved, but, in general, I think we are better, in the spirit of our discussion.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -61,106 +72,54 @@ public struct CXXTranspiler {
...
+      members: cxx(classMembers: src.members))
+  }
 
-    // Transpile the class membmers.
+  /// Transpiles product type members to C++.
+  private func cxx<C: Collection>(classMembers src: C) -> [CXXClassDecl.ClassMember]
+  where C.Element == AnyDeclID.TypedNode {
     var cxxMembers: [CXXClassDecl.ClassMember] = []
-    for member in decl.members {
+    for member in src {
```
dabrahams, line 115: You could probably use `flatmap` to avoid the raw loop.  LMK if you need guidance.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -61,106 +72,54 @@ public struct CXXTranspiler {
...
-          let cxxAttribute = CXXClassAttribute(
-            type: CXXTypeExpr(varDecl.type, ast: program.ast)!,
-            name: CXXIdentifier(varDecl.name),
-            initializer: cxxInitializer,
-            isStatic: isStatic,
-            original: varDecl)
-          cxxMembers.append(.attribute(cxxAttribute))
-        }
+        // One binding can expand into multiple class attributes
+        cxx(productTypeBinding: BindingDecl.Typed(member)!, into: &cxxMembers)
```
dabrahams, line 119: This use site doesn't make sense, which indicates that `cxx` is the wrong name for this function; it's a noun, which indicates its primary purpose is to return something, but this modifies `cxxMembers`.  `transpile` would be a suitable name here, and the use site would read like a sentence.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -169,597 +128,470 @@ public struct CXXTranspiler {
...
-    module.addTopLevelDecl(CXXClassDecl(name: name, members: cxxMembers, original: decl))
+    return cxxMembers
+  }
+
+  /// Translate a binding decl contained in a product type into one or more C++ class attributes.
+  /// One class attribute will be generated for each name in the binding decl.
+  /// The resulting class members are appended to `res` (the array of class members).
+  private func cxx(
+    productTypeBinding src: BindingDecl.Typed, into res: inout [CXXClassDecl.ClassMember]
+  ) {
```
dabrahams, line 139: - [Don't use abbrevs](https://www.swift.org/documentation/api-design-guidelines/#avoid-abbreviations). There's nothing wrong with a single-character name if alternatives are too long. Otherwise, `source` and `target` would be suitable.  
- Don't write documentation words that simply restate the types in the signature (“the array of class members”)

```suggestion
  /// Appends one C++ class attribute to `target` for each name bound in `source`.
  private func transpile(
    productTypeBinding source: BindingDecl.Typed, into target: inout [CXXClassDecl.ClassMember]
  ) {
```

I hope this gives some idea of what we're going for.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -169,597 +128,470 @@ public struct CXXTranspiler {
...
+    // TODO: pattern introducer (let, var, sink, inout)
+    for (_, name) in src.pattern.subpattern.names {
```
dabrahams, line 141: No raw loops, please.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -169,597 +128,470 @@ public struct CXXTranspiler {
...
-        return CXXExprStmt(expr: cxxExpr, original: AnyNodeID.TypedNode(initializer))
-      } else if stmts.count == 1 {
-        return stmts[0]
-      } else {
-        return CXXScopedBlock(stmts: stmts, original: AnyNodeID.TypedNode(initializer))
+    if numNames > 1 {
+      // We would output multiple variable declaration statements, gtoupped into a scoped block
+      // TODO: scoped block is not good here, because of lifetime issues
+      var varStmts: [CXXStmt] = []
+      for (_, name) in src.pattern.subpattern.names {
```
dabrahams, line 175: No raw loops please.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -169,597 +128,470 @@ public struct CXXTranspiler {
...
-      if stmts.isEmpty {
-        // No pattern found; just call the initializer, dropping the result.
-        let cxxExpr = CXXVoidCast(baseExpr: cxxInitialzer, original: initializer)
-        return CXXExprStmt(expr: cxxExpr, original: AnyNodeID.TypedNode(initializer))
-      } else if stmts.count == 1 {
-        return stmts[0]
-      } else {
-        return CXXScopedBlock(stmts: stmts, original: AnyNodeID.TypedNode(initializer))
+    if numNames > 1 {
+      // We would output multiple variable declaration statements, gtoupped into a scoped block
```
dabrahams, line 172:
```suggestion
      // We would output multiple variable declaration statements, grouped into a scoped block
```

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -4,55 +4,66 @@ import FrontEnd
...
-  mutating func emit(topLevel decl: AnyDeclID.TypedNode, into module: inout CXXModule) {
-    switch decl.kind {
+  // MARK: Declarations
+
+  /// Transpiles to C++ a top-level declaration.
+  func cxx(topLevel src: AnyDeclID.TypedNode) -> CXXTopLevelDecl {
+    switch src.kind {
     case FunctionDecl.self:
-      emit(function: FunctionDecl.Typed(decl)!, into: &module)
+      return cxx(function: FunctionDecl.Typed(src)!)
```
dabrahams, line 37: Shouldn't these functions be methods of the corresponding `TypedNode`s, so we can just write:

```swift
return FunctionDecl.Typed(src)!.cxx()
```
?

dabrahams (commented):
I didn't finish with all the code, but left many comments, of which a large fraction should be applicable to the code I haven't read yet.  I want to give you a chance to absorb those before I finish.

Sources/CodeGen/CXX/CXXTranspiler.swift
dabrahams: Alternately, you could solve the naming problem this way:

```suggestion
        cxxMembers += cxx(productTypeBinding: BindingDecl.Typed(member)!)
```

We're nowhere near being at the point of needing to avoid the cost of the intermediate allocations.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -61,106 +72,54 @@ public struct CXXTranspiler {
...
-
-    // Create the C++ function object.
-    module.addTopLevelDecl(
-      CXXFunctionDecl(
-        identifier: identifier,
-        output: output,
-        parameters: cxxParams,
-        body: cxxBody,
-        original: decl))
+    return cxxParams
```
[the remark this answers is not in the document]
kyouko-taiga: Probably you can simply write:

```swift
return zip(src.parameters, paramTypes).map({ (p, t) in
  .init(CXXIdentifier(p.name), cxx(typeExpr: t))
})
```

@dabrahams  note that one day I'll come for all these uses of trailing closures ;)

Sources/CodeGen/CXX/CXXTranspiler.swift
kyouko-taiga: Hmm, I'm not sure.
1. I feel uneasy about augmenting the API of a bundle.
2. We're likely to need additional information (i.e., not only the `FunctionDecl`). IIUC, that information is stored in `self` here.  In my experience, the quantity of information you need to visit the AST tend to grow, adding more and more arguments to your visitors. For example, I think we'll eventually property maps generated by the borrow checker. It's better to encapsulate all this information in the receiver.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -4,55 +4,66 @@ import FrontEnd
...
-  // MARK: API
-
-  /// Emits the C++ module corresponding to the Val module identified by `decl`.
-  public mutating func emit(module decl: ModuleDecl.Typed) -> CXXModule {
-    var module = CXXModule(decl, for: program)
-    for member in decl.topLevelDecls {
-      emit(topLevel: member, into: &module)
+  /// Transpiles Val Module into a C++ module object.
+  public func transpile(_ source: ModuleDecl.Typed) -> CXXModule {
+    var cxxModule = CXXModule(source)
```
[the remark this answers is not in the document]
lucteo: done

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -4,55 +4,66 @@ import FrontEnd
...
   }
 
-  // MARK: API
-
-  /// Emits the C++ module corresponding to the Val module identified by `decl`.
-  public mutating func emit(module decl: ModuleDecl.Typed) -> CXXModule {
-    var module = CXXModule(decl, for: program)
-    for member in decl.topLevelDecls {
-      emit(topLevel: member, into: &module)
+  /// Transpiles Val Module into a C++ module object.
```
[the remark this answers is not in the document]
lucteo: done

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -4,55 +4,66 @@ import FrontEnd
...
-    return module
+    return cxxModule
   }
 
-  /// Emits the given top-level declaration into `module`.
-  mutating func emit(topLevel decl: AnyDeclID.TypedNode, into module: inout CXXModule) {
-    switch decl.kind {
+  // MARK: Declarations
+
+  /// Transpiles to C++ a top-level declaration.
```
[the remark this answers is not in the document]
lucteo: Isn't this description applicable for 90% of the functions in this file?

And again, isn't this comment deducible from function name and parameters?

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: I also contemplated about placing (some of) these methods in the appropriate AST nodes, but not necessarily liked the implications.

Besides the arguments provided by @kyouko-taiga , I see some other issues:
* there are different ways to transpile the same node, depending on the code who needs it (e.g., BindingDecl)
* some arguments of some of these function might make sense only in the context of the transpiler; it would be strange to add them as AST extensions
* high cohesion -- entities that work together should be placed together. Adding this logic through extensions seems to separate out the logic more than needed.

Sources/CodeGen/CXX/CXXTranspiler.swift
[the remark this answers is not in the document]
lucteo: feeling ashamed :(
fixed

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: done

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: done

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: done

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: will fix in all places. Sorry :(

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -4,55 +4,66 @@ import FrontEnd
...
+/// This takes Val AST (with type information) and transforms it into C++ AST.
+/// It handles one module at a time.
 public struct CXXTranspiler {
 
-  /// The program being transpiled.
-  let program: TypedProgram
+  /// The Val typed nodes.
+  /// Used in a few cases in which we need to obtain the typed nodes from the corresponding Node IDs,
+  /// and to make general queries in the Val AST.
+  let wholeValProgram: TypedProgram
```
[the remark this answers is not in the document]
lucteo: We had a similar discussion when creating `TypedNode`, and we agreed that the `program` should not be a publicly-accessible property of `TypedNode`. It feels strange to access the parent object through an abstraction that should encapsulate only the node. Like saying "give me the car engine from this wheel".

Indeed, the whole Transpiler, with or without state feels like a function `ModuleDecl.Typed x TypedProgram -> CXXModule`. The associated struct doesn't add value.

Is there a better/idiomatic way to group these private functions without dumping them into a node extension?

lucteo:
@dabrahams : I tried to address most of the comments. The one thing that I'm a bit confused about is comments. Good function names and good signatures make most of the comment useless.

Sources/CodeGen/CXX/CXXModule.swift
```diff
@@ -6,22 +6,14 @@ import Utils
...
-  init(_ decl: ModuleDecl.Typed, for program: TypedProgram) {
-    self.valDecl = decl
-    self.program = program
-  }
+  let topLevelDecls: [CXXTopLevelDecl]
 
-  /// Add a top-level C++ declaration to this module.
-  mutating func addTopLevelDecl(_ decl: CXXTopLevelDecl) {
-    cxxTopLevelDecls.append(decl)
+  /// The module name
```
kyouko-taiga, line 14:
```suggestion
  /// The module's name.
```

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -3,56 +3,62 @@ import Core
 import FrontEnd
 import Utils
 
-/// A Val to C++ transpiler.
+/// The conversion of typed Val module AST into corresponding C++ AST.
```
kyouko-taiga, line 6: Is it the conversion or a type that _does_ the conversion?

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -3,56 +3,62 @@ import Core
...
 public struct CXXTranspiler {
 
-  /// The program being transpiled.
-  let program: TypedProgram
+  /// The Val typed nodes.
+  ///
+  /// Used in a few cases in which we need to obtain the typed nodes from the corresponding Node IDs,
+  /// and to make general queries in the Val AST.
```
kyouko-taiga, line 12:
```suggestion
  /// This property is used when we need access to the contents of a node from its ID.
```
But perhaps you don' need this additional precision at all. Programs are consistently used to access the contents of the AST everywhere in the compiler. It's obvious that, if the property exists, it is there for "the cases in which we need to obtain the typed nodes".

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: Writing doc comments is hard :) This is actually Dave's suggestion.

The way I read is that, this type *represents* the conversion. A type typically doesn't do anything, but represents something. The methods in the type will *do* something.

But, truth be told, it's hard for me to follow this logic, even if I know the reasoning behind it for many years. :(

kyouko-taiga: Okay, I guess I can get behind. I think "transpiler" isn't the correct name for the type then. If I get an object of type `CXXTranspiler`, I expect that object to actively do something.

Perhaps the type should be called `CXXTranspilation`? I think I need @dabrahams's input here. I am notoriously bad at naming things _and_ Dave is quite opinionated about names in general.

> A type typically doesn't do anything

Well, I think that's debatable. I'll admit that most likely the abstraction that something like a transpiler represent is a function, not a value. The type actually defines the internal state of this notional function. There's a larger question about the way we want to document these kinds of abstractions.

dabrahams: I don't think it matters too much whether it's called `Transpiler` or `Transpilation`, especially since it shouldn't be exposed outside this file/its module.  And since the former is a more accessible name that commonly refers to a notional function anyway, I would choose that.

Please remember that the name and the doc comment are not supposed to perfectly reflect one another; if they do there is no information in the comment.  The comment tells you how to understand what the name refers to.  A transpiler is just an encapsulation of the process of transpiling things.   The doc comment for a type is supposed to tell you what the thing **is**, not what it **does**.  So what _is_ a CXXTranspiler other than the conversion process?    I'm not attached to the words I suggested.  You could say, "A notional funciton that converts…” which to me just seems like extra verbiage, but if that makes y'all more comfortable I'm OK with it.  As long as the comment describes what the thing is, I'm satisfied.

Sources/CodeGen/CXX/CXXTranspiler.swift
dabrahams: I'm going to take your collective word for this and not think about it anymore for the time being.  It might come up again, in which case feel free to point me here before I waste any more of your time.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -4,55 +4,66 @@ import FrontEnd
...
-  let program: TypedProgram
+  /// The Val typed nodes.
+  /// Used in a few cases in which we need to obtain the typed nodes from the corresponding Node IDs,
+  /// and to make general queries in the Val AST.
+  let wholeValProgram: TypedProgram
 
-  /// Creates a C++ transpiler with a well-typed AST.
-  public init(program: TypedProgram) {
-    self.program = program
+  /// Creates a C++ transpiler from the whole Val program.
```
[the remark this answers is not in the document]
dabrahams: @kyouko-taiga is exactly right.  If there are needless words in your doc comment please omit them.   It's worth writing the doc comment that minimally confirms the signature, only to show that you considered it and you're not omitting anything (e.g. some side-effect of the initializer).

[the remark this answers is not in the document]
lucteo: Ok. Will add minimal comments.
Will take another pass of the file and check comments (and other things that I might hav missed)

Sources/CodeGen/CXX/CXXTranspiler.swift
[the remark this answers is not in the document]
dabrahams: It may be applicable to 90% of the functions.  That kind of thing seems to happen in algorithms that traverse the AST.  That's not terrible.

“Say what it returns” is simply about form, as specified in the API guidelines.

kyouko-taiga (approved):

PR #287, opened 2023-01-30
Refactor the match parser

kyouko-taiga:
This PR brings a couple of improvements on the match expression parser.


dabrahams (approved):
If you're into imperative parser descriptions, I guess this is fine ;-)

kyouko-taiga:
Durian was kind of my last ditch attempt at making a declarative parser DSL in Swift. So until we can play with something nicer, I guess the parser will iteratively become more and more imperative.

dabrahams:
I am 100% ready to play with something nicer :-)

Declarative parsing code with backtracking still feels imperative to me; at least it doesn't seem amenable to local reasoning.  I have to trace control flow to understand what's actually being recognized.

PR #292, opened 2023-01-30
More improvements on the inference API

kyouko-taiga:


Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
```diff
@@ -631,16 +631,11 @@ struct ConstraintSolver {
 
     // Slow path: inspect how the solution compares with the ones we have.
     var shouldInsert = false
-    let lhs = newResult.solution.typeAssumptions.reify(
-      comparator,
-      withVariables: .substituteByError)
+    let lhs = newResult.solution.typeAssumptions.reify(comparator)
```
dabrahams, line 634: Are we reifying a comparator here?

If so, I guess I'd like to understand what that means (not a comment on the code).  If not, maybe the API should change.

Sources/FrontEnd/TypeChecking/SubstitutionMap.swift
```diff
@@ -65,7 +65,14 @@ struct SubstitutionMap {
 
   /// Subtitutes each type variable occuring in `type` by its corresponding substitution in `self`,
   /// apply `substitutionPolicy` to deal with free variables.
-  func reify(_ type: AnyType, withVariables substitutionPolicy: SubstitutionPolicy) -> AnyType {
+  ///
+  /// The default substitution policy is `substituteByError` because we typically use `reify` after
+  /// having build a complete solution and therefore don't expect its result to still contain open
```
dabrahams, line 70:
```suggestion
  /// having built a complete solution and therefore don't expect its result to still contain open
```

Sources/FrontEnd/TypeChecking/SubstitutionMap.swift
```diff
@@ -65,7 +65,14 @@ struct SubstitutionMap {
 
   /// Subtitutes each type variable occuring in `type` by its corresponding substitution in `self`,
   /// apply `substitutionPolicy` to deal with free variables.
-  func reify(_ type: AnyType, withVariables substitutionPolicy: SubstitutionPolicy) -> AnyType {
+  ///
+  /// The default substitution policy is `substituteByError` because we typically use `reify` after
```
dabrahams, line 69: The contents of this comment are not a doc comment for this function.  It looks like a design rationale.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -495,7 +494,7 @@ public struct TypeChecker {
 
       // Otherwise, it's expected to have the realized return type.
       let type = declTypes[id]!.base as! LambdaType
-      let inferredType = deduce(typeOf: expr, expecting: type.output.skolemized, in: id)
+      let inferredType = deduceType(of: expr, expecting: type.output.skolemized, in: id)
```
dabrahams, line 497:
```suggestion
      let inferredType = deducedType(of: expr, expecting: type.output.skolemized, in: id)
```

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -1585,7 +1585,9 @@ public struct TypeChecker {
       : nil
   }
 
-  /// Returns a solution describing the best guess to type `subject` and its sub-expressions.
+  /// Returns the best solution satisfying `initialConstraints` and describing how to assign types
+  /// to `subject` and its sub-expressions and knowing it occurs in `scope` and is expected to have
+  /// a type compatible with `expectedType`.
```
dabrahams, line 1590:
```suggestion
  /// Returns the best assignment of types
  /// to `subject` and its sub-expressions in `scope` satisfying `initialConstraints` and that is expected to have
  /// a type compatible with `expectedType`.
```

My suggestion improves on the wording a lot, but this "expected" thing is completely fuzzy.  Expected by whom?  What in the code causes such an "expectation?"  I'm guessing there's a better word.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
dabrahams, line 497: The name `inferredType` is just another way of saying `deducedType` isn't it?  I don't see the value in using two different words for the same thing here.  More than that, `inferredType` isn't just any old inferred type.  Type of what?  Is the fact that it was inferred even relevant anymore? Is this like saying `angleComputedByArcTan`?  Consider providing useful information in the name or using a single letter.

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
kyouko-taiga: The comparator is the type being used to decide if a solution refines another. It's typically picked as the type being inferred. The rationale is that if we're inferring the type of `e` and get two competing solutions `s1` and `s2`, `s1` refines `s2` if the type it assigns to `e` is subtype of the one `s2` assigns to `e`. This trick is used to skip the much more costly ranking algorithm that compares function signatures à la C++.

The comparator is being reified here to substitute the type variables it may contain.

Sources/FrontEnd/TypeChecking/SubstitutionMap.swift
kyouko-taiga: That's true. The comment was inspired by one you wrote on `Diagnostics` that read something like "the default is true because we build many more libraries than command line tools".

I can remove it but isn't the rationale useful to users of the method?

Sources/FrontEnd/TypeChecking/TypeChecker.swift
kyouko-taiga: Do you feel strongly about this suggestion? Most methods in the type checker are imperative verbs (e.g., `inferType`, `realize`, `check`, ...). I don't see a reason to make an exception here.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
kyouko-taiga: > Expected by whom? 

Expected by the caller? ;)

I have the feeling I'll commit 15 variants of this comment until it gets clear enough so I'll just explain how things work the best I can and shamelessly wait for your wisdom.

`expectedType` is used to propagate top-down type information to the sub-expressions of `subject`. For example, if we have `let x: Double = 42`, the checker will call `deduceType` for `42` with `Double` as the expected type.

Many different constructs may create expectations, not only type annotations. For example, if we have `f(x)`, it's possible the checker can expect a type `T` for `x` if it successfully resolved `f` as a function `(T) -> U`.

The comment says "compatible with" but perhaps it should say "subtype of".

Sources/FrontEnd/TypeChecking/SubstitutionMap.swift
dabrahams: Possibly.  But you can't leave out the summary that says what the thing does.  Oh, now that I look at the diff I see you're leaving that in place, so, nevermind, sorry!

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
dabrahams: If you're doing overload resolution I'd be surprised if the "more costly ranking algorithm" can be skipped in all cases without producing an undesirable user experience, but we'll see.

Most of the first paragraph would make a good inline comment in the implementation.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
dabrahams: All calls whose primary role from the caller's POV is to return a value should be consistent.  By making the suggestion I'm bringing up the possibility that all such methods should change.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
dabrahams: Then it's the _required_ type, or a supertype constraint.

Yes, "subtype" is more specific.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
dabrahams: The overall point is: name APIs to make their use-sites make sense.  When I see `inferredType = deduceType(…)` I actually wonder if you mean something different by deduce and infer; perhaps `deduceType` returns some other type inferred during the process.  When I see `inferredType = deducedType(…)`  I can't be confused.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
kyouko-taiga: Okay, I'll make a separate PR.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
kyouko-taiga: I'll take your word for it but I'm sorry, this kind of arguments don't really hit home. I genuinely don't see why `inferredType = deduceType(...)` is less clear than `inferredType = deducedType(...)`. Perhaps that's because it's because I'm not a native English speaker.

Sources/FrontEnd/TypeChecking/ConstraintSolver.swift
kyouko-taiga: Done

Sources/FrontEnd/TypeChecking/TypeChecker.swift
dabrahams: There is some amount of inherent ambiguity in the first that is not present in the second.  If you don't see that, then let's talk about it until you do understand it.  I don't believe it has anything to do with English.  We could rewrite it in French and the same thing would happen.

PR #293, opened 2023-01-30
Use 'Diagnostics' rather than '[Diagnostic]'

kyouko-taiga:


Sources/Core/Dianostics.swift
```diff
@@ -34,13 +34,27 @@ public struct Diagnostics {
     for d in batch { report(d) }
   }
 
+  /// Merges `self` with `other`.
+  public mutating func merge(_ other: Self) {
```
dabrahams, line 38: Should be called `formUnion`, `insert`, `incorporate` or something that indicates what's being mutated. `x.merge(y)` doesn't say that x is being updated.  (In Val `&x.merge(y)` is fine)

Sources/Core/Dianostics.swift
```diff
@@ -34,13 +34,27 @@ public struct Diagnostics {
...
+  }
+
   /// Throws `self` if any errors were reported.
   public func throwOnError() throws {
     if errorReported { throw self }
   }
 
 }
 
+extension Diagnostics: ExpressibleByArrayLiteral {
```
dabrahams, line 50: Why?  I don't see it being used in this PR.  Shouldn't there at least be a test?

Sources/Core/Dianostics.swift
dabrahams: Please also add a test.

dabrahams (approved):
The change described in the PR title is great, but there are a couple of other seemingly random things thrown in that I have questions about inline.

Sources/Core/Dianostics.swift
kyouko-taiga: It's used by `Solution.init`: https://github.com/val-lang/val/blob/015a4194d02ad2ddb2e40a05ea0cf220f51918e7/Sources/FrontEnd/TypeChecking/Solution.swift#L40
There's no change in the diff because it used to be an array.

Sources/Core/Dianostics.swift
kyouko-taiga: Ah?

I don't mind using `formUnion` but `Swift.Dictionary` has a mutating `merge` method that "[m]erges the given dictionary into this dictionary, using a combining closure to determine the value for any duplicate keys."

kyouko-taiga: Also, do you want a specific unit test for this? Seems a little overkill to do it for this one specific method given that we don't have unit tests on any other individual type, except the AST. The call to `merge` in the `ConstraintSolver` is covered by our current tests.

Sources/Core/Dianostics.swift
dabrahams: To support this one use case, I'd rather just use `.init()`.  But if you write that test I was asking for, you could use literals and there's more justification for the conformance.

Sources/Core/Dianostics.swift
dabrahams: W.r.t. `formUnion`, Diagnostics is a notional set, innit?

In principle, we should have unit tests for everything and should probably get in the habit. it should only take a moment to write.  But if you disagree, let's have a Discussion.

kyouko-taiga: I don't necessarily disagree but that's something we haven't done for any other component (yet), so I don't think it should be in scope for this specific PR, which is about adding a couple of use sites to `Diagnostics`.

Sources/Core/Dianostics.swift
kyouko-taiga: Let's keep the conformance then.

Issue #288, opened 2023-01-30
Revisit the use of this initializer


dabrahams:
See [this example](https://github.com/val-lang/val/pull/271/files/228d928347d87d2e544a26b2d6582483db30a931..c5adc74354f8fcc9bca928d2adc8a0df135cc9f8#diff-fb82a2ae0690aa14551a1d2dd0bf9958c20e0a0420d64614dfd94c14f1307f8cR73) and the subsequent argument.

The question is, should we make this replacement:

```diff
-    program.ast[NodeID(id)!]
+    program.ast[NodeID<ID.Subject>(rawValue: id.rawValue)] 
```

kyouko-taiga:
What do you suggest we do?

I would be highly dissatisfied by a solution that just consist of removing the initializer that optimally converts a type-erased ID to a typed one. That's not only because of optimization, but also because it would be sad if we didn't have a way to provide an API that doesn't require a type check to be done twice every time.

Specifically, consider the following, very frequent pattern:

```swift
func foo(_ d: AnyDeclID) {
  switch d.kind {
  case FunctionDecl.self:
    foo(functionDecl: NodeID(rawVaue: d.rawValue))
  case MethodDecl.self:
    foo(methodDecl: NodeID(rawVaue: d.rawValue))
  default:
     unreachable()
  }
}

func foo(functionDecl d: NodeID<FunctionDecl>) {...}
func foo(methodDecl d: NodeID<MethodDecl>) {...}
```

In this example, I wish we would have an initializer that unconditionally converts `d` to an instance of `NodeID<FunctionDecl>` after we've already checked that `d.kind`  is indeed `FunctionDecl.self` so as to mimic what happens when we use Swift's pattern matching.

```swift
func bar(_ x: Any) {
  switch x {
    case let y as Int:
      print(y + y) // No need to re-cast `y as! Int` here.
  ...
  }
}
```
In other words, I would be disappointed if the solution was to write `NodeID(d)!`.

`NodeID.init(rawValue:)` came to fit this role by accident. It used to be an internal initializer used by the AST when we had only the raw value of an index. That use case doesn't exist anymore, so I bet one solution would be to replace `init(rawValue:)` by a new initializer `init(strawmanLabel:)` that accepts any kind of ID:

```swift
extension NodeID {
  init<T: NodeIDProtocol>(strawmanLabel other: T) {
    precondition(other.kind == Subject.kind) // or perhaps `assert`
    self.rawValue = other.rawValue
  }
}
```

[...]

PR #294, opened 2023-01-30
Eliminate 'NodeID.init(rawValue:)'

kyouko-taiga:
See #288


kyouko-taiga:
This PR removes all uses of `NodeID.init(rawValue)` except the one that creates a new ID when a node is inserted and the one that implements `NodeID.init(_:)`.

This change may have introduced marginal performance regressions in places where the code can assume the type of type-erased ID from AST invariants. Profiling will tell us if we need to handle these cases manually when we'll care about performance.

dabrahams (approved):
Very nice; thank you!

PR #296, opened 2023-01-31
Visit the match expression patterns during type inference

kyouko-taiga:
This PR implements the first step of inferring the type of match expression: visiting the patterns of the match cases and make sure they correspond to the type of the subject.

The changes in the PR do not cover constraint generation for the body of each individual case yet, which should be submitted as a separate patch.

### DIFF
--- a/Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
+++ b/Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
@@ -147,6 +147,10 @@ extension TypeChecker {
       return inferType(
         ofLambdaExpr: NodeID(rawValue: subject.rawValue), in: scope,
         expecting: expectedType, updating: &facts)
+    case MatchExpr.self:
+      return inferType(
+        ofMatchExpr: NodeID(rawValue: subject.rawValue), in: scope,
+        expecting: expectedType, updating: &facts)
     case NameExpr.self:
       return inferType(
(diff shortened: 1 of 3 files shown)


Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -469,6 +473,34 @@ extension TypeChecker {
...
+    in scope: AnyScopeID,
+    expecting expectedType: AnyType?,
+    updating facts: inout InferenceFacts
+  ) -> AnyType {
+    let syntax = program.ast[subject]
+
+    // Visit the subject of the match.
+    let subjectType = inferType(of: syntax.subject, in: scope, expecting: nil, updating: &facts)
+    if subjectType.isError {
+      return facts.assignErrorType(to: subject)
```
dabrahams, line 487: It is really odd to see this check and explicit error assignment here.  Couldn't it be incorporated into `inferType`?

dabrahams (approved):

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
kyouko-taiga: Why is it odd?

The check is a short-circuit in case the subject of the match expression is ill-typed. When that happens, it's likely best to skip checking the cases because we'll get cascading errors when trying to type their patterns.

It's not the role of `inferType` to assign the error type to `subject`. It's inferring the type of another node.

PR #298, opened 2023-02-01
Refactor deferred type check requests

kyouko-taiga:
This PR is an experiment to implement a more principled approach to deferred type checking requests.

The checker sometimes needs to defer the checking of some nodes until the type of the declarations on which it depends have been determined. Usually, this process is handled by realizing the type of all declarations on which an expression depends on demand.

However, this approach cannot apply when the type of a declaration depends on the result of type inference for an expression in which it is contained. For example, consider the following declaration:
```swift
let x = (fun (x) { x })(x: 42)
```
Here, the type of the anonymous function declaration cannot be type checked until its return type has been realized. Because the declaration has no return annotation, however, this type must be inferred from the context. Hence, the checker cannot eagerly realize the declaration. It must pick a type variable for the return type and infer its value from the call. Once it's done, the checker can finally check the function declaration.

Other nodes may require a similar treatment. The goal of this PR is to provide a principled approach to handle them.

In the current code (before the patch), the type checker has to visit a collection of declarations visited during constraint generation after constraint solving and run the conceptual deferred queries. Besides, it must do it at each call site of `TypeChecker.inferType(...)` For example: https://github.com/val-lang/val/blob/33d096cb89732b31aaae59accc52853fde82df52/Sources/FrontEnd/TypeChecking/TypeChecker.swift#L419-L426

For lambdas, constraint generation must register all lambda expressions whose type checking must be deferred (see `TypeChecker. deferTypeChecking(_:)`). These lambdas are then type-checked before returning from `TypeChecker.check(module:)`: https://github.com/val-lang/val/blob/33d096cb89732b31aaae59accc52853fde82df52/Sources/FrontEnd/TypeChecking/TypeChecker.swift#L294

This process is unprincipled, not uniform for all kind of nodes, and error-prone. After the patch, constraint generation can create deferred queries on any node without requiring the caller to implement their semantics. For example: https://github.com/val-lang/val/blob/01eb7bcbf94ecb84b590df72e8b4b0b8a22370e0/Sources/FrontEnd/TypeChecking/TypeChecker%2BConstraintGeneration.swift#L896-L901


Sources/FrontEnd/TypeChecking/AnyDeferredQuery.swift
```diff
@@ -0,0 +1,33 @@
+/// A box wrapping a deferred type checking query.
+private protocol DeferredQueryBox {
+
+  /// Executes this query using `checker` and reading inference results from `solution`.
```
dabrahams, line 4: I don't know what it means to "execute" a query.  If your documentation depends on the reader understanding words that are in the name of the thing you're documenting, that's a sign that you're not conveying anything useful.

Sources/FrontEnd/TypeChecking/AnyDeferredQuery.swift
```diff
@@ -0,0 +1,33 @@
...
+  func _execute(_ checker: inout TypeChecker, _ solution: Solution) -> Bool
```
dabrahams, line 5: Why the leading underscore?

Sources/FrontEnd/TypeChecking/AnyDeferredQuery.swift
```diff
@@ -0,0 +1,33 @@
+/// A box wrapping a deferred type checking query.
```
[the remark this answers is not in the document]
dabrahams: It _looks_ like this is just a function and if you made it so you wouldn't need type erasure shenanigans.

Sources/FrontEnd/TypeChecking/DeferredQuery.swift
```diff
@@ -0,0 +1,29 @@
+import Core
+import Utils
+
+/// A deferred type checking query that should be executed by applying `execute` on `subject` the
+/// type of its constituent parts have been inferred.
```
dabrahams, line 5: I can't parse this sentence as composed.  Semicolon after `subject`?

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
```diff
@@ -75,98 +72,98 @@ extension TypeChecker {
...
-
     /// Indicates that a conflict has been found.
     fileprivate mutating func setConflictFound() {
       foundConflict = true
     }
 
   }
 
+  /// The common state of all `inferTypes(...)` methods as they recursively visit the AST.
+  private typealias VisitationState = (facts: InferenceFacts, deferred: [AnyDeferredQuery])
```
dabrahams, line 83: Should this just be `State`?  Is the fact that there's a kind of visitation pattern in play actually important?

dabrahams (approved):
Seems mostly OK, but could you add something to the PR comment that explains the primary benefit of this refactoring, maybe by showing an example?  To me most of the code reads about the same as before the change.

Sources/FrontEnd/TypeChecking/AnyDeferredQuery.swift
kyouko-taiga: Trying to apply your guidelines, I often feel extremely constrained in my ability to document things adequately. It's hard to have 1-liner docs _and_ build concept descriptions.

Do you honestly don't know what "executing a query" mean or you just worried there may be an ambiguity? The docs on `DeferredQuery` says: "A deferred type checking query that should be executed by applying `execute` on `subject` the type of its constituent parts have been inferred." It's hard to seriously think someone working on a type checker can read this comment and be not understand what the abstraction and accompanying vocabulary is about.

Sources/FrontEnd/TypeChecking/AnyDeferredQuery.swift
kyouko-taiga: Name clash

Sources/FrontEnd/TypeChecking/DeferredQuery.swift
kyouko-taiga: There's a missing word, sorry.

Sources/FrontEnd/TypeChecking/AnyDeferredQuery.swift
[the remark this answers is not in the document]
kyouko-taiga: Good point.

Sources/FrontEnd/TypeChecking/TypeChecker+ConstraintGeneration.swift
kyouko-taiga: Good point.

kyouko-taiga:
> could you add something to the PR comment that explains the primary benefit of this refactoring, maybe by showing an example?

Of course! It's done.

Sources/FrontEnd/TypeChecking/AnyDeferredQuery.swift
dabrahams: I honestly don't know.  Is this just a predicate?  Does a query come back with a list of matching things, like a database query?  

I did read the docs for `DeferredQuery` very carefully.  Yes it says that you execute a query by calling the execute function, but what semantics is that function supposed to have?  If I don't know that, I don't know how to write a valid query, nor do I know how to use one.  "It executes the query" is, obviously, a circular answer to that question.

1-liners are not always adequate; sometimes you need a supporting descriptive paragraph.  

I'm sorry you feel constrained; when you get stuck on these things, feel free to call on me and we can dig into it together.  There's a mutually-supporting process of discovering the nature of underlying abstractions and writing the documentation, and it's not easy.  It eventually becomes natural to know when you haven't quite got it, but it can still be a lot of work in some cases.

Sources/FrontEnd/TypeChecking/AnyDeferredQuery.swift
```diff
@@ -1,33 +1,13 @@
...
-  init<I>(_ q: DeferredQuery<I>) {
-    self.wrapped = q
-  }
+extension DeferredQuery {
 
-  /// Executes this query using `checker` and reading inference results from `solution`.
-  func execute(_ checker: inout TypeChecker, _ solution: Solution) -> Bool {
-    wrapped._execute(&checker, solution)
+  /// Returns a partial application of `q.execute`, capturing `q.subject`.
+  static prefix func ^ (_ q: Self) -> AnyDeferredQuery {
```
dabrahams, line 9: This is a novel meaning for an operator we've used for something very different AFAICT.  Not sure that's a good idea.

dabrahams:
Thanks for the examples.  It's mostly clear now.  I realize now I inadvertently made you write lots more explanation than needed for the review, sorry.  My fault for not being precise enough. All I _really_ needed was to be pointed at the parts of the diff that are essential improvements of this patch as distinguished from trivial things like renames.  You can head off this sort of question by looking at what's in the diff a reviewer will have to process, and seeing whether the "real" change in the patch is obvious.  If it isn't, the options are to break up the PR, to, un-rearrange things you've changed the order of so the diffs can easily be seen, or to write some explanation.  Usually it's a good idea to leave the explanation in a commit message too, not just the PR.  Actually, the content of your explanation makes an excellent commit message, so maybe all your writing isn't wasted.

Sources/FrontEnd/TypeChecking/AnyDeferredQuery.swift
kyouko-taiga: I'm not too attached to it, but I think the operator makes sense in the way that `AnyDeferredQuery` is the type-erased form of `DeferredQuery`.

But, now that I think about it, `DeferredQuery` only serves to declare explicitly the environment of a closure while `AnyDeferredQuery` erases that environment... Perhaps we can simply define `DeferredQuery` as `(inout TypeChecker, Solution) -> Bool`.

dabrahams:
> On Feb 2, 2023, at 11:19 AM, Dimi Racordon ***@***.***> wrote:
> 
> 
> @kyouko-taiga commented on this pull request.
> 
> In Sources/FrontEnd/TypeChecking/AnyDeferredQuery.swift <https://github.com/val-lang/val/pull/298#discussion_r1094975395>:
> 
> >  
> -  /// Executes this query using `checker` and reading inference results from `solution`.
> -  func execute(_ checker: inout TypeChecker, _ solution: Solution) -> Bool {
> -    wrapped._execute(&checker, solution)
> +  /// Returns a partial application of `q.execute`, capturing `q.subject`.
> +  static prefix func ^ (_ q: Self) -> AnyDeferredQuery {
> I'm not too attached to it, but I think the operator makes sense in the way that AnyDeferredQuery is the type-erased form of DeferredQuery.
> 
If the doc comment said that’s what it was doing, it would have made sense to me, but… it doesn’t say that.
> But, now that I think about it, DeferredQuery only serves to declare explicitly the environment of a closure while AnyDeferredQuery erases that environment... Perhaps we can simply define DeferredQuery as (inout TypeChecker, Solution) -> Bool.
> 

I think that’s exactly what I was suggesting earlier.

PR #303, opened 2023-02-02
Better reporting of "unexpected node" error messages

lucteo:
Previously, the code would report things like `unexpected class member` without telling the user what node kind it found and at which location in the Val source file.

With this change we improve the error reporting to:
```
.../CXXTranspiler.swift:114: Fatal error: unexpected class member, found FunctionDecl at location: .../PostfixOperators.val:12:3
```

### DIFF
--- a/Sources/CodeGen/CXX/CXXTranspiler.swift
+++ b/Sources/CodeGen/CXX/CXXTranspiler.swift
@@ -558,12 +558,7 @@ public struct CXXTranspiler {
       return CXXTypeExpr(isReturnType ? "void" : "std::monostate")
 
     case let type as ProductType:
-      // TODO: we should translate this to an "int" struct
-      if type == wholeValProgram.ast.coreType(named: "Int") {
-        return CXXTypeExpr("int")
-      } else {
-        return CXXTypeExpr(type.name.value)
-      }
+      return CXXTypeExpr(type.name.value)
 
     case let type as ParameterType:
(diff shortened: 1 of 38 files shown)


kyouko-taiga (approved):
Nice!

lucteo:
@kyouko-taiga : I merged the PR sitting on top of this, and it merged it into this branch. Sorry for extra changes in the PR.
I also fixed compilation after merging from main.

Probably the easiest way to re-review this is to look at the last commits.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -64,7 +64,7 @@ public struct CXXTranspiler {
         fatalError("not implemented")
 
       default:
-        unreachable()
+        unreachable("unexpected type")
```
kyouko-taiga, line 67: Don't you want to use your `unexpected` function here?

kyouko-taiga (approved):

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: My abstraction doesn't work for types (don't have the location).

I can implement an abstraction, but I would do this in a different PR.

lucteo: Created #335

PR #305, opened 2023-02-02
Stop translating Val `Int` into C++ `int`.

lucteo:
The code will blindly transform all `Int` types into `Int`, even there is no C++ definition for `Int.
This will later be addressed when we translate Val standard library to C++.


kyouko-taiga (approved):
Seems like a good first step.

lucteo:
Hmmm.... It seems that I merged this into `unexpected-node-util` instead of main branch.
I think this happens when your base branch is not yet merged... lessons learned

PR #322, opened 2023-02-03
Make generated C++ code compile

lucteo:
(at least the factorial.val example)

reimplement changes from compile-cxx branch

This PR fixes that, by doing two things:
* ensuring the name of the module and the names of the written files match (don't include "Main.h", when we generate "main.h")
* When the program contains a main function, add a top-level main function in the generated CXX code that just calls the one from the appropriate module.


Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -603,4 +600,49 @@ public struct CXXTranspiler {
...
+  func entryPointBody(topLevel source: AnyDeclID.TypedNode, in module: ModuleDecl.Typed)
+    -> CXXScopedBlock?
+  {
+    let function = FunctionDecl.Typed(source)
+    if function != nil && function!.identifier?.value == "main" {
+      return entryPointBody(mainFunction: FunctionDecl.Typed(source)!, in: module)
+    } else {
+      return nil
+    }
+  }
```
kyouko-taiga, line 620: This pair of function seems overly complex. I guess we're trying to avoid raw loops, but in this case it's impeding readability. Also, I don't think you need any loop in the first place, IIUC:
```suggestion
  func entryPointBody(module source: ModuleDecl.Typed) -> CXXScopedBlock? {
    if let f = decl.topLevelDecls.first(where: { (d) in
      FunctionDecl.Typed(d)?.identifier?.value == "main"
    }) {
      return entryPointBody(mainFunction: FunctionDecl.Typed(f)!, in: module)
    } else {
      return nil
    }
  }
```
It's a bit unfortunate that we have to convert the ID of the main function to `FunctionDecl.Typed` twice. Perhaps we could extend `Sequence` with a method `first(transformedBy:)` to make that simpler.

```swift
extension Sequence {

  public func first<T>(transformedBy transform: (Element) throws -> T?) rethrows -> T? {
    for x in self {
      if let y = try transform(x) { return y }
    }
    return nil
  }

}
```

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -603,4 +600,49 @@ public struct CXXTranspiler {
...
+  private func entryPointBody(mainFunction source: FunctionDecl.Typed, in module: ModuleDecl.Typed)
```
kyouko-taiga, line 621: This function (and the one before) deserve some docs.

kyouko-taiga (commented):
I think we can simplify `entryPointBody` quite a bit. I left inline comments.

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: This was the abstraction I was looking for. Thank you for pointing this out!

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: done

lucteo:
> I think we can simplify `entryPointBody` quite a bit. I left inline comments.

Applied the suggestion for `first(transformedBy:)`. I still kept two functions for better readability. I feel that putting too much code in the lambda seems to make the code harder to read.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -603,4 +600,63 @@ public struct CXXTranspiler {
...
   }
 
+  // MARK: entry point function body
+
+  /// Returns the transpiled body of the entry point function in `source`.
+  func entryPointBody(module source: ModuleDecl.Typed) -> CXXScopedBlock? {
+    let elements = Array(source.topLevelDecls)
+    return elements.first(transformedBy: { (d) in
+      entryPointBody(topLevel: d, in: source)
+    })
```
kyouko-taiga, line 610:
```suggestion
    source.topLevelDecls.first(transformedBy: { (d) in
      entryPointBody(topLevel: d, in: source)
    })
```
One of the big advantage of extending `Sequence` is that you shouldn't have to transform other sequences/collections to `Array` in order to apply generic algorithms.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -603,4 +600,63 @@ public struct CXXTranspiler {
...
+  }
+  /// Returns the transpiled body of the entry point function that may be calling `source`.
+  func entryPointBody(topLevel source: AnyDeclID.TypedNode, in module: ModuleDecl.Typed)
```
kyouko-taiga, line 613: I agree that large closures impede readability, but so does control flow. To read `entryPointBody(module:)` I need to follow control flow to `entryPointBody(topLevel:in:)`, keep in mind what its return type is, and then switch back to `entryPointBody(module:)` to fully understand what's going on.

This whole function only saves 5 lines in the closure of the one on top. It's never reused anywhere else (AFAIK), so really it's only exporting code and causing mental gymnastic.

The docs of the function doesn't help either.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -603,4 +600,63 @@ public struct CXXTranspiler {
...
+  // MARK: entry point function body
+
+  /// Returns the transpiled body of the entry point function in `source`.
+  func entryPointBody(module source: ModuleDecl.Typed) -> CXXScopedBlock? {
+    let elements = Array(source.topLevelDecls)
+    return elements.first(transformedBy: { (d) in
+      entryPointBody(topLevel: d, in: source)
+    })
+  }
+  /// Returns the transpiled body of the entry point function that may be calling `source`.
```
kyouko-taiga, line 612:
```suggestion
  /// Return the body of a function calling `source` if `source` is the entry point of `module`.
  /// Otherwise, returns `nil`.
```

The current docs suggests this function is unconditionally returning scoped block.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -603,4 +600,63 @@ public struct CXXTranspiler {
...
+    }
+    return CXXScopedBlock(stmts: bodyContent)
+  }
+
+}
+
+extension Sequence {
+
+  /// The result of the first scucessful transformation applied to elements in `self`.
+  public func first<T>(transformedBy transform: (Element) throws -> T?) rethrows -> T? {
```
kyouko-taiga, line 655: This extension should go in `Utils/Sequence+Extensions.swift`.

lucteo: this is exactly what I did, and, for some reason it doesn't work. Couldn't find the reason why.

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: sorry. this was a leftover from when I was trying to debug why `first(transformedBy:)` doesn't properly work. Fixed

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: not applicable anymore if I get rid of this function

kyouko-taiga (approved):

PR #324, opened 2023-02-03
Remove builtin symbols

kyouko-taiga:
This PR completes the work on replacing `BuiltinSymbols` by removing this type and updating the use of built-in symbols in the standard library.


Library/Core/Int.val
```diff
@@ -3,26 +3,33 @@ public type Int {
 
   var value: Builtin.i64
 
+  memberwise init
+
+  /// Initializes an instance with the value of `0`.
```
dabrahams, line 8:
```suggestion
  /// Creates an instance with value `0`.
```

Library/Core/Bool.val
```diff
@@ -3,12 +3,19 @@ public type Bool {
 
   var value: Builtin.i1
 
+  memberwise init
+
+  /// Initializes a instance with the value of `0`.
```
dabrahams, line 8:
```suggestion
  /// Creates an instance with value `false`.
```

Library/Core/Double.val
```diff
@@ -3,14 +3,21 @@ public type Double {
 
   var value: Builtin.double
 
+  memberwise init
+
+  /// Creates an instance with the value `0.0`.
```
dabrahams, line 8:
```suggestion
  /// Creates an instance with value `0.0`.
```

Library/Core/Int32.val
```diff
@@ -3,11 +3,16 @@ public type Int32 {
 
   var value: Builtin.i32
 
-  /// Initializes a 32-bit signed integer with the value of `other`.
+  /// Initializes an instance with the value of `0`.
```
dabrahams, line 6:
```suggestion
  /// Creates an instance with value `0`.
```

Library/Core/Int32.val
```diff
@@ -3,11 +3,16 @@ public type Int32 {
...
+  public init() {
+    self.value = Builtin.zeroinitializer_i32()
+  }
+
+  /// Initializes an instance with the value of `other`.
```
dabrahams, line 11:
```suggestion
  /// Creates a copy of `other`.
```

Sources/IR/Operands/Constant/BuiltinFunctionRef.swift
```diff
@@ -3,27 +3,32 @@ import Core
 /// A Val IR reference to a built-in function.
 public struct BuiltinFunctionRef: ConstantProtocol, Hashable {
```
dabrahams, line 4: I find it odd that this exists at all, distinct from `BuiltinFunction`, and I don't know what the word "reference" means in the doc comment.

dabrahams (approved):
🎉

Sources/IR/Operands/Constant/BuiltinFunctionRef.swift
kyouko-taiga: The only difference is that `BuiltinFunctionRef` has a lowered type.

It's a reference because functions are not first-class values in the IR (lambdas are, but lambdas are distinct from functions at the IR level). I would not worry too much about this type because it's most likely to go away very soon.

PR #330, opened 2023-02-04
Remove 'testCode(_:)'

kyouko-taiga:
_Re-open #329 with corrected branch name_

Using `testCode(_:)` instead of `SourceFile.init(synthesized:)` doesn't seem to improve call sites so much, yet it costs an additional free function that can only be used in the `ValTests` target. It is also not obvious what the result of this function is; it doesn't actually test anything, it only creates a source file with the given text.

The changes make `SourceFile` conform to `ExpressibleByStringLiteral` so that we can write:
```swift
let s: SourceFile = """
  import Foo
  fun main() {}
  """
```

### DIFF
--- a/Sources/Core/SourceFile.swift
+++ b/Sources/Core/SourceFile.swift
@@ -33,6 +33,12 @@ public struct SourceFile {
     self.storage = storage
   }
 
+  /// Creates a source file with the specified contents and a unique random `url`.
+  public init(synthesizedText text: String) {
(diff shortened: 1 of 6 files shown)


dabrahams (approved):
Nice idea!  I'm a little nervous about having the conformance outside of the testing code for some reason, but I'm not dead-set against it.

kyouko-taiga:
We could have the conformance declared in a module shared by all test targets (interesting use case BTW, as this wouldn't be possible in Val!)

PR #334, opened 2023-02-04
Improve the API of the IR emitter

kyouko-taiga:


Sources/IR/Emitter.swift
```diff
@@ -329,24 +340,24 @@ public struct Emitter {
...
-      return emitR(sequence: SequenceExpr.Typed(expr)!, into: &module)
+      return emitR(sequenceExpr: SequenceExpr.Typed(expr)!, into: &module)
     default:
       unreachable("unexpected expression")
     }
   }
 
   private mutating func emitR(
-    booleanLiteral expr: BooleanLiteralExpr.Typed,
+    booleanLiteralExpr expr: BooleanLiteralExpr.Typed,
```
dabrahams, line 360: I'm not sure this extra bit of explicitness is actually helping, frankly, _especially_ if you do #340, but even if you don't.  Rvalues are always expressions.

Sources/IR/Emitter.swift
```diff
@@ -750,8 +761,8 @@ public struct Emitter {
     }
   }
 
-  /// Emits `expr` as an operand suitable to be the callee of a `CallInstruction`, updating `conventions`
-  /// and `arguments` if `expr` refers to a bound member function.
+  /// Emits the IR of `expr` as an operand suitable to be the callee of a `CallInstruction` into
+  /// `module`, updating `conventions` and `arguments` if `expr` refers to a bound member function.
```
dabrahams, line 765:
```suggestion
  /// Inserts the IR of lambda expression `expr` into `module`, updating `conventions` and `arguments` if `expr` refers to a bound member function.
```

But you can't just say "updating."  What specific changes are made?

Sources/IR/Emitter.swift
```diff
@@ -856,16 +867,16 @@ public struct Emitter {
 
   // MARK: l-values
 
-  /// Emits `expr` as a l-value with the specified capability into `module` at the current
-  /// insertion point.
+  /// Emits the IR of `expr` treated as a l-value with given `capability` into `module`, inserting
+  /// instructions at the end of `self.insertionBlock`.
```
dabrahams, line 871:
```suggestion
  /// Inserts the IR of the lvalue `expr`, having the given `capability, into `module` 
  /// at the end of the current insertion block.
```

Maybe not ideal, and I'm not 100% sure I'm conveying the intended meaning (my wording assumes `expr` already has the capability)? We always have to be careful with the word "with" because it has so many interpretations. Is the capability getting inserted into the module along with the IR for `expr`?

It's better to avoid "with" altogether when possible, and consider removing it from the label.

Sources/IR/Emitter.swift
```diff
@@ -28,24 +28,27 @@ public struct Emitter {
...
     case TraitDecl.self:
       break
     default:
       unreachable("unexpected declaration")
     }
   }
 
-  /// Emits the given function declaration into `module`.
-  public mutating func emit(function decl: FunctionDecl.Typed, into module: inout Module) {
+  /// Emits the IR of `decl` into `module`.
```
dabrahams, line 50:
```suggestion
  /// Inserts the IR of `decl` into `module`.
```

Sources/IR/Emitter.swift
```diff
@@ -108,53 +111,55 @@ public struct Emitter {
...
-    assert(stack.frames.isEmpty)
+    assert(stack.isEmpty)
   }
 
-  /// Emits the given subscript declaration into `module`.
-  public mutating func emit(
-    subscript decl: SubscriptDecl.Typed,
-    into module: inout Module
-  ) {
+  /// Emits the IR of `decl` into `module`.
```
dabrahams, line 117:
```suggestion
  /// Inserts the IR of `decl` into `module`.
```

Sources/IR/Emitter.swift
```diff
@@ -108,53 +111,55 @@ public struct Emitter {
...
+  private mutating func emit(subscriptDecl decl: SubscriptDecl.Typed, into module: inout Module) {
     fatalError("not implemented")
   }
 
-  /// Emits the product type declaration into `module`.
-  private mutating func emit(product decl: ProductTypeDecl.Typed, into module: inout Module) {
+  /// Emits the IR of `decl` into `module`.
```
dabrahams, line 122:
```suggestion
  /// Inserts the IR of `decl` into `module`.
```

Sources/IR/Emitter.swift
```diff
@@ -108,53 +111,55 @@ public struct Emitter {
...
     case .let:
-      emit(borrowedLocalBinding: decl, withCapability: .let, into: &module)
+      emit(borrowedLocalBindingDecl: decl, withCapability: .let, into: &module)
     case .inout:
-      emit(borrowedLocalBinding: decl, withCapability: .inout, into: &module)
+      emit(borrowedLocalBindingDecl: decl, withCapability: .inout, into: &module)
     }
   }
 
+  /// Emits the IR of `decl` into `module`, assuming it's a local `var` or `sink let` binding.
```
dabrahams, line 155:
```suggestion
  /// Inserts the IR for the local `var` or `sink let` binding `decl` into `module`.
```

Sources/IR/Emitter.swift
```diff
@@ -208,12 +213,15 @@ public struct Emitter {
     }
   }
 
-  /// Emits borrowed bindings.
+  /// Emits the IR of `decl` into `module`, assuming it's a local `let`, `inout` or `set` binding.
```
dabrahams, line 216:
```suggestion
  /// Inserts the IR of the local `let`, `inout` or `set` binding `decl` into `module`.
```

Sources/IR/Emitter.swift
```diff
@@ -329,24 +340,24 @@ public struct Emitter {
...
+      return emitR(nameExpr: NameExpr.Typed(expr)!, into: &module)
     case SequenceExpr.self:
-      return emitR(sequence: SequenceExpr.Typed(expr)!, into: &module)
+      return emitR(sequenceExpr: SequenceExpr.Typed(expr)!, into: &module)
     default:
       unreachable("unexpected expression")
     }
   }
 
   private mutating func emitR(
```
dabrahams, line 359: Might as well document this one, too?

Sources/IR/Emitter.swift
```diff
@@ -891,7 +902,7 @@ public struct Emitter {
   }
 
   private mutating func emitL(
-    name expr: NameExpr.Typed,
+    nameExpr expr: NameExpr.Typed,
```
dabrahams, line 905: Same remark as above.  All lvalues are expressions.

Sources/IR/Emitter.swift
```diff
@@ -954,6 +965,8 @@ public struct Emitter {
 
   // MARK: Helpers
 
+  /// Pops the top frame of `self.stack` and emits a deallocation instruction for each allocation
+  /// in that frame.
```
dabrahams, line 969: Precondition? What if the stack is empty?

Sources/IR/Emitter.swift
```diff
@@ -964,52 +977,49 @@ public struct Emitter {
...
+    /// The allocations in the frame, in FILO order.
     var allocs: [Operand] = []
 
   }
 
-  /// A type describing the state of the call stack during lowering.
+  /// A stack of frames.
   fileprivate struct Stack {
 
     /// The frames in the stack, in FILO order.
```
dabrahams, line 994: FYI: That's seems like a  strange way to use the word FILO, which usually describes the _behavior_ of a queue.  I would generally say “ordered from bottom to top” ersump'n.

Sources/IR/Emitter.swift
```diff
@@ -964,52 +977,49 @@ public struct Emitter {
...
-    var frames: [Frame] = []
+    private var frames: [Frame] = []
+
+    /// True iff the stack is empty.
+    var isEmpty: Bool { frames.isEmpty }
 
     /// Accesses the top frame, assuming the stack is not empty.
```
dabrahams, line 1000: Assuming.

Sources/IR/Emitter.swift
```diff
@@ -964,52 +977,49 @@ public struct Emitter {
...
-      _modify {
-        yield &frames[frames.count - 1]
-      }
+      get { frames[frames.count - 1] }
+      _modify { yield &frames[frames.count - 1] }
     }
 
-    /// Accesses the operand assigned `decl`, assuming the stack is not empty.
-    subscript<ID: DeclID>(decl: ID.TypedNode) -> Operand? {
+    /// Accesses the IR corresponding to `d`, assuming the stack is not empty.
```
dabrahams, line 1006: Assuming.

dabrahams, line 1006: It's very odd to see a subscript without O(1) complexity.  IMO this fact should be documented at the least.

Sources/IR/Emitter.swift
```diff
@@ -964,52 +977,49 @@ public struct Emitter {
...
 
-    /// Pushes a new frame.
+    /// Pushes `newFrame` on the stack.
     mutating func push(_ newFrame: Frame = Frame()) {
       frames.append(newFrame)
     }
 
-    /// Pops a frame, assuming the stack is not empty.
+    /// Pops the top frame, assuming the stack is not empty.
     mutating func pop() {
```
dabrahams, line 1023: Consider a `@discardable` Frame result, which would be consistent with other Swift `pop()` functions.

Sources/Utils/MVS.swift
```diff
@@ -1,3 +1,8 @@
+/// Returns the result of calling `action` with an immutable projection of `value`.
+public func reading<T, U>(_ value: T, _ action: (T) throws -> U) rethrows -> U {
```
dabrahams, line 2: Yeah, here's that "reading" function, right next to the `modifying` function that's been bothering me for the same reason.  I was thinking `modify` and `modifying` should be collapsed, but if not, overloading is better than this use of `ing`.

dabrahams (approved):
Comments inline

Sources/IR/Emitter.swift
```diff
@@ -15,10 +15,10 @@ public struct Emitter {
   /// The basic block in which new instructions are currently inserted.
   private var insertionBlock: Block.ID?
 
-  /// The state of the call stack.
+  /// A stack of frames describing the variables and allocations of each traversed lexical scope.
```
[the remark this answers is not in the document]
kyouko-taiga: Yep, SGTM.

Sources/Utils/MVS.swift
kyouko-taiga: `modify`/`modifying` should be a bundle!

Not at all attached to the `ing` suffix. I'll make the change in another PR.

Sources/IR/Emitter.swift
```diff
@@ -316,9 +324,12 @@ public struct Emitter {
 
   // MARK: r-values
 
-  /// Emits `expr` as a r-value into `module` at the current insertion point.
-  private mutating func emitR<ID: ExprID>(expr: ID.TypedNode, into module: inout Module) -> Operand
-  {
+  /// Emits the IR of `expr` treated as a r-value into `module`, inserting instructions at the end
+  /// of `self.insertionBlock`.
```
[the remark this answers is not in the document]
kyouko-taiga: I've consistently used _r-value_ and _l-value_ so far, but I don't think there's any "official" answer to this question.

Hmm... the C++ standard doesn't use the dash. I guess we shouldn't either.

Sources/IR/Emitter.swift
kyouko-taiga: I'd prefer to remain consistent in the way we name "visitors", but don't feel so strongly about it.

In general, we can't get rid of the suffixes because we have things that share a common prefix like `BindingDecl` and `BindingPattern`. Originally I added the suffix only for those cases, but eventually to all of them for the sake of having uniform APIs.

dabrahams:
> On Feb 5, 2023, at 5:41 PM, Dimi Racordon ***@***.***> wrote:
> 
> In general, we can't get rid of the suffixes because we have things that share a common prefix like BindingDecl and BindingPattern. Originally I added the suffix only for those cases, but eventually to all of them for the sake of having uniform APIs.
> 
IMO uniformity is good insofar as it increases comprehensibility, but is not an end in itself.  Things that explicitly visit lvalues and rvalues don’t have to be like all other things that visit.

Issue #306, opened 2023-02-02
Start translating Val standard library to C++


lucteo:
Ensure that we generate C++ code vor Val standard library.

In the first phase, the code doesn't need to compile, just that we can generate it. Making sure the code compiles is the subject of a follow-up issue/PR

PR #337, opened 2023-02-05
Start translating Val standard library to C++

lucteo:
Implements #306

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,9 +15,3 @@
 /Sources/Parser/**
 !/Sources/Parser/Val.g4
 !/Sources/Parser/ValParser+Extensions.swift
-
-# Generated C++ files
-Main.h
-Main.cpp
-ValStdLib.h
-ValStdLib.cpp
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,9 @@
 /Sources/Parser/**
 !/Sources/Parser/Val.g4
 !/Sources/Parser/ValParser+Extensions.swift
+
+# Generated C++ files
+Main.h
+Main.cpp
+ValStdLib.h
+ValStdLib.cpp
--- a/Sources/CodeGen/CXX/CXXCodeWriter.swift
+++ b/Sources/CodeGen/CXX/CXXCodeWriter.swift
@@ -8,11 +8,14 @@ public struct CXXCodeWriter {
 
   /// Returns the C++ code for a translation unit.
   public func cxxCode(_ source: CXXModule) -> TranslationUnitCode {
-    TranslationUnitCode(headerCode: emitHeaderCode(source), sourceCode: emitSourceCode(source))
+    TranslationUnitCode(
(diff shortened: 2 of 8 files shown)


Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -37,33 +45,33 @@ public struct CXXCodeWriter {
     return target.code
   }
 
-  /// Write the CXX source content for the given module to the given text stream.
-  public func emitSourceCode(_ module: CXXModule) -> String {
+  /// The C++ code for `module` that needs to be present in the source file.
```
kyouko-taiga, line 48:
```suggestion
  /// Returns the C++ code for `module` that needs to be present in the source file.
```

Sources/CodeGen/CXX/TranslationUnitCode.swift
```diff
@@ -0,0 +1,12 @@
+/// The C++ code for a .h/.cpp file pair.
+///
+/// By convention, for each C++ translation unit we generate, we create a header file containing
+/// the interface of the translation unit declarations.
+public struct TranslationUnitCode {
+
+  /// The code that belongs to a C++ header file.
+  public let headerCode: String
+  /// The code that belongs to a C++ source file.
```
kyouko-taiga, line 9:
```suggestion

  /// The code that belongs to a C++ source file.
```

I know the rule about this extra isn't enforced by swift-format but it should.

Sources/Core/TypedProgram.swift
```diff
@@ -26,6 +26,11 @@ public struct TypedProgram: Program {
   /// A map from sequence expressions to their evaluation order.
   public let foldedSequenceExprs: [NodeID<SequenceExpr>: FoldedSequenceExpr]
 
+  /// The Val standard library module.
+  public var corelib: ModuleDecl.Typed? {
+    ast.corelib != nil ? self[ast.corelib!] : nil
```
kyouko-taiga, line 31:
```suggestion
    ast.corelib == nil : nil ? self[ast.corelib!]
```
to avoid testing a negative. Or perhaps even better:
```suggestion
    ast.corelib.map({ self[$0] })
```

.gitignore
```diff
@@ -15,3 +15,9 @@
 /Sources/Parser/**
 !/Sources/Parser/Val.g4
 !/Sources/Parser/ValParser+Extensions.swift
+
+# Generated C++ files
+Main.h
+Main.cpp
+ValStdLib.h
+ValStdLib.cpp
```
kyouko-taiga, line 23: This looks a little weird to have these files generated in the main repository. Is that for convenience when doing local tests? If so let's pick a directory (e.g., `.build`). Otherwise, we should generate a temporary directory as it's already done in `ValCommandTests`.

kyouko-taiga (changes requested):
Great work!

I'm a little worried about the additions to `.gitignore`. I left a comment as well as other minor remarks.

kyouko-taiga (approved):

PR #338, opened 2023-02-05
Remove 'BuiltinDecl'

kyouko-taiga:
This PR removes the `BuiltinDecl` node, which was used as a kind of hack to implement name binding.

The problem is that built-in symbols don't have any AST declaration, so we can't form an instance of `DeclRef` (before this patch) to update `TypeChecker.referredDecls`. `BuiltinDecl` was used to address this issue. The AST would contain a unique instance that could be used as a placeholder every time we had to create a `DeclRef`.

However, this approach has a couple of flaws:
- Because `BuiltinDecl` doesn't carry any information identifying a specific declaration, we always need to rely on additional (and arguably incidental) information to determine the declaration being referred. This problem is visible in IR lowering and transpilation, where we use the name stored in a `NameExpr` to determine the referred declaration.
- We need to synthesize a fake node that doesn't correspond to any actual declaration.

The patch removes `BuiltinDecl` and adds two cases to `DeclRef` to represent references to built-in declarations. One unfortunate consequence of this change is that it makes many use sites of a `DeclRef` slightly more complex, as they can no longer just assume it stores an existing AST declaration.

Also note that while writing this patch, I realized that `DeclRef` I realized it should probably renamed (perhaps `NameBinding`?). However, such a change should belong to another PR.


dabrahams:
> The patch removes BuiltinDecl and adds two cases to DeclRef to represent references to built-in declarations. One unfortunate consequence of this change is that it makes many use sites of a DeclRef slightly more complex, as they can no longer just assume it stores an existing AST declaration.

Are you sure it wouldn't be better to simply inject Decl AST nodes for the builtins into the builtin module?

dabrahams (approved):
This looks totally fine! Given your PR text I was expecting much more complexity bloat.

PR #344, opened 2023-02-06
Make sure generated C++ code for Val standard lib compiles.

lucteo:
Implements #336


Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -4,11 +4,15 @@ public struct CXXCodeWriter {
   /// Initializes the current object.
   public init() {}
 
+  /// Indicates if we are currently writing the standard library module.
+  private var isStdLib: Bool = false
+
```
lucteo, line 9: I'm not quite liking how this class become. I intend to have a PR (or more) soon with refactoring similar to CXXTranspiler, which will hopefully address the fact that we instantiate an empty class just for possibly some state.

Sources/ValCommand/ValCommand.swift
```diff
@@ -214,20 +214,19 @@ public struct ValCommand: ParsableCommand {
 
     // Initialize the transpiler & code writer.
     let transpiler = CXXTranspiler(typedProgram)
-    let codeWriter = CXXCodeWriter()
+    var codeWriter = CXXCodeWriter()
```
lucteo, line 217: this is one of the reasons for refactoring. I would want this to avoid declaring an empty class here.

Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -26,19 +30,40 @@ public struct CXXCodeWriter {
...
 
+    // Do we need to generate extra code for the standard library?
+    if source.isStdLib {
+      // TODO: revsit this
+      target.writeNewline()
+      target.writeLine("inline bool operator <(Int l, Int r) { return l.value < r.value; }")
+      target.writeLine("inline Int operator +(Int l, Int r) { return l.value + r.value; }")
+      target.writeLine("inline Int operator -(Int l, Int r) { return l.value - r.value; }")
+      target.writeLine("inline Int operator *(Int l, Int r) { return l.value * r.value; }")
+      target.writeLine("inline void abort() { ::abort(); }")
```
kyouko-taiga, line 64: You already left a TODO note, but I'd like to add that this should be defined in a separate header that just gets included. No need to maintain Swift code that outputs fully static C++.

Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -148,10 +175,40 @@ public struct CXXCodeWriter {
         target.writeLine("// constructor")
       }
     }
+    // Special code for stdlib types
+    if isStdLib {
+      // For standard value types generate implict conversion constructors from C++ literal types.
+      if ["Int", "Int32", "Double", "Bool"].contains(decl.name.description) {
+        write(conversionCtor: decl, into: &target)
+      }
+    }
```
kyouko-taiga, line 184: This bit is worrisome. The standard library will grow significantly. We should avoid hardcoding about the set of types that it defines in the transpiler.

If you want custom implicit conversions from C++ types to standard library types, then we should define those in a separate C++ file.

Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -148,10 +175,40 @@ public struct CXXCodeWriter {
...
     target.endBrace()
     target.writeLine(";")
   }
 
+  /// Writes to `target` the implicit conversion constructor for `source`, coverting from inner
+  /// attribute type.
+  private func write(conversionCtor source: CXXClassDecl, into target: inout CodeFormatter) {
```
kyouko-taiga, line 191: I think we shouldn't need this.

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -564,19 +573,47 @@ public struct CXXTranspiler {
...
+    case .i(let bitWidth) where bitWidth == 8:
+      return CXXTypeExpr("int8_t")
+    case .i(let bitWidth) where bitWidth == 16:
+      return CXXTypeExpr("int16_t")
+    case .i(let bitWidth) where bitWidth == 32:
+      return CXXTypeExpr("int32_t")
+    case .i(let bitWidth) where bitWidth == 64:
+      return CXXTypeExpr("int64_t")
+    case .half:
+      return CXXTypeExpr("float") // not properly supported; just keep the compiler happy
```
kyouko-taiga, line 604: If it's not supported, why not simply reject it like you did for `i128` for example?

Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -564,19 +573,47 @@ public struct CXXTranspiler {
...
+    case .float:
+      return CXXTypeExpr("float")
+    case .double:
+      return CXXTypeExpr("double")
+    case .fp128:
+      return CXXTypeExpr("long double")
+    case .ptr:
+      return CXXTypeExpr("long")
```
kyouko-taiga, line 612: Is `long` really how raw pointers translate to C++, not `void*`?

kyouko-taiga (approved):
Nice changes!

I think we shouldn't use the transpiler to generate hardcoded C++. It'll be far more convenient and less error-prone to have separate C++ files that we can keep in sync with every change of the standard library.

You can see how to bundle non-Swift files into a product if you look at the `ValModule` target in `Package.swift.

Sources/CodeGen/CXX/CXXCodeWriter.swift
lucteo: I don' think I can do this, as these need to go inside the class declaration.
But I think we can avoid the enumeration here, just by looking if the class has only one attribute, and the attribute has a native type.

Sources/CodeGen/CXX/CXXCodeWriter.swift
lucteo: as discussed in the meeting, I'd like to keep this as a small workaround. It's not ideal, but I don't know at this point a better way to do this.

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: ok, will do

Sources/CodeGen/CXX/CXXTranspiler.swift
lucteo: this wasn't supposed to be like this. I think there was an error from my merge patch.

Sources/CodeGen/CXX/CXXCodeWriter.swift
lucteo: Agree. I'll move this to a dedicated C++ file.

kyouko-taiga (approved):

PR #353, opened 2023-02-07
Redesign the API of 'AbstractTypeLayout'

kyouko-taiga:


Sources/IR/AbstractTypeLayout.swift
```diff
@@ -1,71 +1,67 @@
 import Core
 
-/// A helper object that provides information the abstract layout of a type.
+/// The astract layout of a type, describing the relative offsets of its stored properties.
```
dabrahams, line 3: This is a HUGE improvement over what you had, but I think we can go further:
```suggestion
/// The byte offsets of a type's stored properties.
```

…Unless I'm misunderstanding meaning of the type.  Does it also have stride and size information?

PR #355, opened 2023-02-08
Remove 'TransformPass'

kyouko-taiga:
This patch removes `TransformPass` and implements a more convenient way to abstract over IR pass pipelines.


Sources/IR/Analysis/DefiniteInitializationPass.swift
```diff
@@ -6,18 +6,19 @@ import Utils
...
-  public private(set) var diagnostics: [Diagnostic] = []
-
+  /// Creates an instance that analyzes IR lowered from `program`.
   public init(program: TypedProgram) {
     self.program = program
   }
 
-  public mutating func run(function: Function.ID, module: inout Module) -> Bool {
+  /// Reports any use-before-initialization errors in `function` into `diagnostics`, 
+  /// where `function` is in `module` and `module` is in `self.program`.
```
dabrahams, line 20: You should make a similar change for all `run` methods in the other "passes"

Sources/IR/Analysis/PassPipeline.swift
```diff
@@ -0,0 +1,47 @@
+import Core
+
+/// A collection of IR passes to be applied on a module sequentially.
+public struct PassPipeline {
```
dabrahams, line 4: OK, seeing this leaves me really surprised, after seeing that you're removing `TransformPass`.  I thought we agreed there was no abstraction here. If that's the case, you should just have a function that calls the sequence of other functions, open-coded.  I see no need for all this complication.  Won't something like this work?

```swift
extension TypedProgram {
  mutating func finalize(_ f: Function.ID, diagnostics: inout Diagnostics) {
   insertImplicitReturns(f, diagnostics: &diagnostics)
   checkLifetimes(f, diagnostics: &diagnostics)
   checkDefiniteInitialization(f, diagnostics: &diagnostics)
    ...
  }    
}
```

dabrahams (commented):
I think there's a much simpler way to do this, but LMK if I've missed something.

Sources/IR/Analysis/PassPipeline.swift
kyouko-taiga: Just need a place to write the loop that goes over every function in a module.

An extension of TypedProgram might be the right call.

Sources/IR/Analysis/DefiniteInitializationPass.swift
```diff
@@ -6,18 +6,19 @@ import Utils
...
-  public private(set) var diagnostics: [Diagnostic] = []
-
+  /// Creates an instance that analyzes IR lowered from `program`.
   public init(program: TypedProgram) {
     self.program = program
   }
 
-  public mutating func run(function: Function.ID, module: inout Module) -> Bool {
+  /// Runs this pass represented by this instance on `function` in `module`, reporting warnings
+  /// and errors into `diagnostics`.
```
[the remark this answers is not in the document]
kyouko-taiga: This property map is `module` ...

But we could have module keep a copy of its typed program.

Sources/IR/Analysis/PassPipeline.swift
kyouko-taiga: Actually, we need a way to get to read the module before the passes run for the `--emit raw-ir` option, which I use for debugging.

What part of the pass pipeline are complicated? Perhaps we can just remove `init(_:)` from the public API?

kyouko-taiga:
I simplified the API of the passes so that we no longer need to instantiate them with a typed program. We no longer need these passes to be structs, but I don't want to turn them into functions yet because it will create many conflicts with another PR.

Sources/IR/Analysis/PassPipeline.swift
dabrahams: It's not “a thing.”  If I ask you to tell me what a pass pipeline is and I boil the words down, it just amounts to a function that mutates the TypedProgram and generates diagnostics.  We don't need a collection of steps.  We don't need to type-erase the steps.  We don't need the steps to be  represented as types. It's over-engineered; I don't know how to explain it better.

kyouko-taiga: It will get smaller but I don't want to create too many conflicts by turning the passes into functions yet.

PR #359, opened 2023-02-08
Factor type-checking out of the command pipeline.

dabrahams:
There's a slight impedance mismatch with the way visibility of the builtin module is being handled.  We should discuss that at some point.

### DIFF
--- a/Sources/FrontEnd/AST+Extensions.swift
+++ b/Sources/FrontEnd/AST+Extensions.swift
@@ -4,32 +4,23 @@ import ValModule
 
 extension AST {
 
-  /// Incorporates the core library into `self`.
-  ///
-  /// - Requires: The Core library must not have been already imported.
-  public mutating func importCoreModule() {
-    precondition(!isCoreModuleLoaded, "Core library is already loaded")
-
+  /// An instance that includes the core module.
+  public static var coreModule: AST = {
+    var r = AST()
     do {
       var diagnostics = Diagnostics()
-      corelib = try makeModule(
+      r.corelib = try r.makeModule(
         "Val",
         sourceCode: sourceFiles(in: [ValModule.core!]),
         builtinModuleAccess: true,
         diagnostics: &diagnostics)
 
-      assert(isCoreModuleLoaded)
+      assert(r.isCoreModuleLoaded)
     } catch let error {
       fatalError("Error parsing the core module:\n\(error.localizedDescription)")
     }
-  }
 
-  /// Returns `self`, with syntax for the core module included if it is not there already.
-  public func importingCoreModule() -> Self {
-    if isCoreModuleLoaded { return self }
-    var r = self
-    r.importCoreModule()
     return r
-  }
+  }()
 
 }
--- a/Sources/FrontEnd/AST+Extensions.swift
+++ b/Sources/FrontEnd/AST+Extensions.swift
@@ -4,7 +4,7 @@ import ValModule
 
 extension AST {
 
-  /// Incorporates syntax for the core library into `self`.
+  /// Incorporates the core library into `self`.
   ///
   /// - Requires: The Core library must not have been already imported.
   public mutating func importCoreModule() {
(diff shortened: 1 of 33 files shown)


Sources/FrontEnd/AST+Extensions.swift
```diff
@@ -21,4 +21,42 @@ extension AST {
     }
   }
 
+  /// Returns `self`, fully type-checked, with the core module imported.
+  ///
+  /// - Parameters
+  ///   - standardLibrary: The module, if any, to be checked with the Builtin module visible.
+  ///   - inferenceTracingRange: A range of source code in which to trace type inference, or `nil` 
+  ///     to disable tracing.
+  ///   - diagnostics: channel for errors and warnings.
```
kyouko-taiga, line 30:
```suggestion
  ///   - diagnostics: A channel for errors and warnings.
```

Sources/FrontEnd/AST+Extensions.swift
```diff
@@ -21,4 +21,42 @@ extension AST {
     }
   }
 
+  /// Returns `self`, fully type-checked, with the core module imported.
+  ///
+  /// - Parameters
+  ///   - standardLibrary: The module, if any, to be checked with the Builtin module visible.
```
kyouko-taiga, line 27: Argument is required to be in `self`, no?

Sources/FrontEnd/AST+Extensions.swift
```diff
@@ -21,4 +21,42 @@ extension AST {
...
+  ///   - inferenceTracingRange: A range of source code in which to trace type inference, or `nil` 
+  ///     to disable tracing.
+  ///   - diagnostics: channel for errors and warnings.
+  public func typeChecked(
```
kyouko-taiga, line 31: Wondering if it wouldn't make sense for this method to be an initializer of `TypedProgram`.

kyouko-taiga (approved):
Very nice code reduction!

Sources/FrontEnd/AST+Extensions.swift
dabrahams: Does self know which module is the standard library?  If so, I can drop this.  It's not the same as the core library… unless perhaps you want to say the standard library can't use `Builtin`, in which case I can also drop it or replace it with a flag that corresponds to your command-line argument.

Sources/FrontEnd/AST+Extensions.swift
dabrahams: It will force the AST parameter to be nested in parens at the call site, but we could do that.

Sources/FrontEnd/AST+Extensions.swift
kyouko-taiga: > Does self know which module is the standard library?

`self` doesn't know, but it could. We could add a flag to `AST.makeModule`:
```swift
/// Imports and returns a new module with the given `name` from `sourceCode`, writing diagnostics to
/// `diagnostics`, and sets it as `self.standardLibrary` is `isStandardLibrary` is `true`.
public mutating func makeModule<S: Sequence>(
  _ name: String, sourceCode: S, isStandardLibrary: Bool = false,
  diagnostics: inout Diagnostics
) throws -> NodeID<ModuleDecl>
```

>  It's not the same as the core library

The standard library and the core library are indeed different modules.

>  perhaps you want to say the standard library can't use `Builtin`

I don't mind if it can't see the `Builtin` module by default and we have to pass `--import-builtin` to compile it.

Sources/FrontEnd/AST+Extensions.swift
kyouko-taiga: I think it is better. The general approach to discover how an object is created is to look at its initializers. I bet the most common way we'll build typed programs will be via this function. It might even make sense to make the member wise initializer private.

Sources/FrontEnd/AST+Extensions.swift
dabrahams: > self doesn't know, but it could. We could add a flag to AST.makeModule:

+1; I can add that to the PR.

`--import-builtin` would tend to imply it's imported into all the source files, without an explicit `import` statement.  Is that the semantics you're envisioning?

Sources/FrontEnd/AST+Extensions.swift
dabrahams: OK, will do

dabrahams: Can't make the memberwise initializer private because TypedProgram is in Core where Frontend components like TypeChecker are not accessible :(.

dabrahams:
@kyouko-taiga we lost the code reduction but I think the result is a big improvement.

Issue #345, opened 2023-02-06
Add tests to execute resulting C++ code


lucteo:
Add tests that will take Val code, compile it, translate it to C++, compile the C++ code and run the output executable.

The tests shall not fail if the C++ compiler cannot be found on the system.
The tests shall fail if the C++ compilation fails.

kyouko-taiga:
Part of this work was started in #260

PR #361, opened 2023-02-08
Add tests that execute generated executable

lucteo:
Implements #345.

At this point, we add just one test to prove the flow: Val type checking -> C++ transpilation -> C++ compilation -> execution of the executable.

Contains refactoring to move test annotations functionality in a dedicated testing-only library.


Tests/TestUtils/XCTest+Shims.swift
```diff
@@ -44,7 +44,7 @@ import XCTest
 
   /// An object that contains a file URL and line number that represents a distinct location in
   /// source code.
-  final class XCTSourceCodeLocation: Hashable {
+  public final class XCTSourceCodeLocation: Hashable {
```
kyouko-taiga, line 47: Initializer probably ought to be public too. Do the tests pass on non macOS platforms?

kyouko-taiga: Looks like they do.

kyouko-taiga (changes requested):
It goes in the direction of what we discussed, but I think there's another way to organize this stuff.

I would not have `ExecutionTests` under the `ValCommand` target. These tests do not directly relate to `ValCommand`. They are testing the transpiler.

You could drive the compilation the same way `CXXTests` does and have your `run(_:withArguments:)` method be defined as a member of the `XCTestCase` subclass handling your execution tests. If you do it that way, there's no need to promote `Utils` to a target and it can stay under `ValTests`.

kyouko-taiga:
I am still not convinced that you need to use the annotation system to run your end-to-end tests. The patch is bringing _a lot_ of complexity. It's creating a new target, breaking encapsulation of several components, and defining a new test annotation. All of that to write `//! run` at the top of every test case in `EndToEndTests`.

It's far simpler to just compile a program, execute it, and look at its exit status. I bet all test cases will do just that anyway for the foreseeable future.

Once you can show me sufficiently complex test cases where we _really_ need to inspect the standard output, we can re-open this idea. Until then, I think we should just stick to a simpler approach. I can create a PR to show you what I have in mind if you want.

lucteo:
@kyouko-taiga : it is true that I don't necessarily need the annotation system for this PR. But there are two things that we should be adding to end-to-end tests in the near-medium term:
* check the output of the program (we can generate programs that write to the console with very few changes)
* ability for the generated programs to capture the command line arguments

I would hate to remove the common testing code now just to add it soon after.
Moreover, we start to have shared testing abstractions between our three testing targets.

kyouko-taiga:
> check the output of the program (we can generate programs that write to the console with very few changes)

Yes and no. We can hack a solution quickly, but implementing a proper print will take longer. Regardless, I highly doubt we absolutely need to print anything on the console for the time being. As I mentioned, we can start with e2e tests that just look at the exit status.

> ability for the generated programs to capture the command line arguments

This is definitely not a high priority for now.

> I would hate to remove the common testing code now just to add it soon after.

I'm saying that we won't know how much sharing we actually need until we'll have actual use cases for the shared features.

lucteo:
@kyouko-taiga : I really need in the near future to have tests that produce output, and also be parameterized. Otherwise, the end-to-end tests loose most of what they need to do. I wanted to put up a followup PR this weekend to show that this is relatively easily achieved, but did not get to it.

kyouko-taiga:
We need to discuss visions, because I think there's a mismatch. I'm sorry I put a roadblock on your work here, but this change doesn't go in the direction I'd like.

dabrahams:
Might need to be today if you want me involved. I’m on vacation next weekSent from my iPhoneOn Feb 19, 2023, at 3:01 AM, Dimi Racordon ***@***.***> wrote:﻿
We need to discuss visions, because I think there's a mismatch. I'm sorry I put a roadblock on your work here, but this change doesn't go in the direction I'd like.

dabrahams:
For perspective, "just look at the exit status" may sound on its face like it is too limiting, but I think it works if the test programs themselves include assertions of expected intermediate results, rather than trying to compare printed output.  Does that help, @lucteo?

Tests/TestSupport/AnnotatedValFileTest.swift
```diff
@@ -31,15 +31,15 @@ extension XCTestCase {
   /// the effects of processing don't match the file's diagnostic annotation commands ("diagnostic",
   /// "expect-failure", and "expect-success").
   ///
-  /// - Parameter `sourceDirectory`: a path relative to the ValTests/ directory of this project.
+  /// - Parameter `testCaseDirectory`: where the test files are searched.
```
dabrahams, line 34: Does it have to be absolute?  If not, relative to what is it interpreted?

You changed the name of the parameter but didn't change the place where it appears in the summary fragment.  You removed important information about how to understand it, in favor of redundantly repeating its high-level meaning.

Tests/TestSupport/AnnotatedValFileTest.swift
```diff
@@ -48,19 +48,19 @@ extension XCTestCase {
...
 
-  /// Applies `processAndCheck` to each ".val" file in `sourceDirectory` along with the subset of
+  /// Applies `processAndCheck` to each ".val" file in `testCaseDirectory` along with the subset of
   /// that file's annotations whose commands match `checkedCommands`, and reports resulting XCTest
   /// failures, along with any additional failures where the effects of processing don't match the
   /// file's diagnostic annotation commands ("diagnostic", "expect-failure", and "expect-success").
   ///
   /// - Parameters:
-  ///   - `sourceDirectory`: a path relative to the ValTests/ directory of this project.
+  ///   - `testCaseDirectory`: where the test files are searched.
```
dabrahams, line 57: Replaced important information with redundant info.

Tests/ValTests/CXXTests.swift
```diff
@@ -1,14 +1,16 @@
...
+import TestSupport
 import Utils
 import XCTest
 
 final class CXXTests: XCTestCase {
 
   func testTranspiler() throws {
     try checkAnnotatedValFiles(
-      in: "TestCases/CXX", checkingAnnotationCommands: ["cpp", "h"],
+      in: Bundle.module.url(forResource: "TestCases/CXX", withExtension: nil)!,
```
dabrahams, line 12: It is inot an improvement that now I need to read this mysterious `Bundle` incantation every time `checkAnnotatedValFiles` is called.

dabrahams (changes requested):
See comments

lucteo:
Fixed with #496

PR #369, opened 2023-02-11
Diagnostics => DiagnosticSet

dabrahams:
Diagnostics is just a notional set, so let's make its usage, API, and documentation more consistent with Set.


Sources/Core/DiagnosticSet.swift
```diff
@@ -0,0 +1,65 @@
+import Foundation
```
kyouko-taiga, line 1: I'm surprised to see Foundation here.

dabrahams: Me TOO!

dabrahams: I know why it's there: it used to be writing to stderr.

kyouko-taiga (approved):

Sources/Core/DiagnosticSet.swift
kyouko-taiga: Thought so. Good opportunity to get rid of this import. Everything else looks good.

dabrahams: removed.

PR #374, opened 2023-02-11
Handle diagnostics reported at the end index

kyouko-taiga:
Fixes a bug that would cause off-by-one errors when using `lineAndColumn(_:)` on the last index.

@dabrahams I didn't understand what was the purpose on the invariant on `SourceFile.Storage.lineStarts`:
> /// - Invariant: always starts with `contents.startIndex` and ends with `contents.endIndex`, even
>  ///   if there's no final newline.

All tests pass without this needed this invariant, but I may have missed something,


dabrahams:
There is code that indexes this array with the result of a partitioningpoint call. The invariant ensures that the index is always in bounds. If you are getting wrong results with that element there, there's probably something wrong with your query

dabrahams:
I take it back. I can take a look at this in 15 minutes

kyouko-taiga:
> I take it back. I can take a look at this in 15 minutes

👍

For context, the problem is that `lineAndColumn` returns the wrong line index if there isn't any newline at the end of the file. So if you use that to fetch the text of a specific line, you find yourself out of bounds.

PR #388, opened 2023-02-13
Add possibility to format generated C++ code with clang-format.

lucteo:
If `clang-format` is found in the path, allow the CXXCodeWriter to use it to format the output C++ code.


Sources/CodeGen/CXX/CodeTransform.swift
```diff
@@ -0,0 +1,7 @@
+/// A transformation that can be run on code.
+public typealias CodeTransform = (String) -> String
+
+/// Returns an identity code transform.
+public func identity() -> CodeTransform {
```
kyouko-taiga, line 5: I'm a bit uncomfortable having `identity` be introduced in the global namespace, as a public symbol. I think there's a better way to initialize the code writer (see other comment).

Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -4,7 +4,12 @@ import ValModule
 public struct CXXCodeWriter {
 
   /// Initializes the current object.
-  public init() {}
+  public init(formatter: @escaping CodeTransform = identity()) {
```
kyouko-taiga, line 7:
```suggestion
  public init(formatter: CodeTransform? = nil) {
```

IIUC, that's what your `identity` function is representing: the absence of any transformation.

Sources/CodeGen/CXX/CodeTransform.swift
```diff
@@ -0,0 +1,7 @@
+/// A transformation that can be run on code.
+public typealias CodeTransform = (String) -> String
```
kyouko-taiga, line 2: This should be in the namespace of `CXXCodeWriter`

Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -110,7 +115,7 @@ public struct CXXCodeWriter {
       target.writeNewline()
     }
 
-    return target.code
+    return formatter(target.code)
```
kyouko-taiga, line 118:
```suggestion
    return formatter?(target.code)
```

Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -69,7 +74,7 @@ public struct CXXCodeWriter {
       }
     }
 
-    return target.code
+    return formatter(target.code)
```
kyouko-taiga, line 77:
```suggestion
    return formatter?(target.code)
```

Sources/CodeGen/CXX/CXXCodeWriter.swift
```diff
@@ -4,7 +4,12 @@ import ValModule
 public struct CXXCodeWriter {
 
   /// Initializes the current object.
-  public init() {}
+  public init(formatter: @escaping CodeTransform = identity()) {
+    self.formatter = formatter
+  }
+
+  /// Object used to format output C++ code.
+  private let formatter: CodeTransform
```
kyouko-taiga, line 12:
```suggestion
  private let formatter: CodeTransform?
```

Sources/CodeGen/CXX/ClangFormatTransformation.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+
+/// Returns a C++ code formatter that invokes external `clang-format` to perform the job.
+public func clangFormatTransformation(_ executable: URL, style: String = "llvm") -> (
```
kyouko-taiga, line 4:
```suggestion
public func makeClangFormatter(_ executable: URL, style: String = "llvm") -> (
```

Sources/ValCommand/ValCommand.swift
```diff
@@ -187,9 +187,15 @@ public struct ValCommand: ParsableCommand {
 
     // *** C++ Transpiling ***
 
+    let clangFormat = try? find("clang-format")
+    let codeFormatter: CodeTransform =
+      clangFormat != nil
+      ? clangFormatTransformation(URL(fileURLWithPath: clangFormat!))
+      : identity()
+
```
kyouko-taiga, line 195:
```suggestion
    let codeFormatter: CodeTransform? = (try? find("clang-format")).map({ clangFormatTransformation(URL(fileURLWithPath: $0!)) })
```

Can be improved if `clangFormatTransformation` does the URL construction from a path.

kyouko-taiga (changes requested):
Great work. I'm very excited about this change!

I think there's a way to avoid introducing an `identity` function in the global namespace. See my inline comments.

Sources/CodeGen/CXX/CodeTransform.swift
lucteo: @kyouko-taiga : If `CXXCodeWriter` disappears in the near future, where do you think it's better to put `CodeTransform`?

For me, keeping it here doesn't seem that bad (after removing `identity()` function)

Sources/CodeGen/CXX/ClangFormatter.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+
+/// Returns a C++ code formatter that invokes external `clang-format` to perform the job.
+public func clangFormatter(_ executable: URL, style: String = "llvm") -> (
+  (String) -> String
+) {
+  return { code in
```
kyouko-taiga, line 7:
```suggestion
  { (code) in
```

Sources/ValCommand/ValCommand.swift
```diff
@@ -203,8 +203,18 @@ public struct ValCommand: ParsableCommand {
...
 
+    let codeFormatter: CodeTransform? = (try? find("clang-format")).map({
+      clangFormatter(URL(fileURLWithPath: $0))
+    })
+
+    //       core: program.cxx(program.corelib!, withFormatter: codeFormatter),
+    //       source: program.cxx(program[sourceModule], withFormatter: codeFormatter)
+    // =======
+    //       core: program.cxx(program.coreLibrary!), source: program.cxx(program[sourceModule])
+    // >>>>>>> upstream/main
```
kyouko-taiga, line 214: What's this?

kyouko-taiga (commented):
There seems to be leftover comments in the PR. Apart from that and a couple of minor suggestions, LGTM.

Sources/ValCommand/ValCommand.swift
lucteo: :( apparently I didn't properly save the file.

Issue #399, opened 2023-02-15
Transpiler refers to the core library as the standard library


dabrahams:
This thing we call the core library plays the role of supplying what Rust considers built-in types; it's roughly the stuff you get with [no_std](https://docs.rust-embedded.org/book/intro/no-std.html).

PR #405, opened 2023-02-15
More pipeline cleanups

dabrahams:
I apologize for the mess of these commits.


Sources/CodeGen/CXX/CXXTranspiler.swift
```diff
@@ -6,6 +6,8 @@ import Utils
 /// The conversion of typed Val module AST into corresponding C++ AST.
 public struct CXXTranspiler {
 
+  public static let coreLibModuleName = "ValStdLib"
```
kyouko-taiga, line 9: I'm surprised to see this after #399 was opened. Just want to make sure this line is intentional.

Sources/ValCommand/ValCommand.swift
```diff
@@ -135,103 +135,65 @@ public struct ValCommand: ParsableCommand {
...
     if outputType == .rawAST {
-      let url = outputURL ?? URL(fileURLWithPath: productName + ".ast.json")
-      let encoder = JSONEncoder().forAST
-      try encoder.encode(ast).write(to: url, options: .atomic)
+      try write(ast, to: astFile(productName))
       return
     }
 
-    let typedProgram = try TypedProgram(
+    let p = try TypedProgram(
```
kyouko-taiga, line 150: I feel the scope of this variable is a bit too large for it to be a single-letter identifier.

I can't think of a much better name than `typedProgram`. That's the role this variable has here: the program we're compiling, typed.

Sources/ValCommand/ValCommand.swift
```diff
@@ -341,4 +302,49 @@ public struct ValCommand: ParsableCommand {
   /// A map from executable name to path of the named binary.
   private static var executableLocationCache: [String: String] = [:]
 
+  /// Wrotes a textual descriptioni of `input` to the given `output` file.
+  func write(_ input: AST, to output: URL) throws {
+    let encoder = JSONEncoder().forAST
+    try encoder.encode(input).write(to: output, options: .atomic)
+  }
+
+  /// Given the product name (whatever that means), returns the file to write when "raw-ast" is
```
kyouko-taiga, line 311: > (whatever that means)

Hahahaha.
Well, I think we should discuss this.

Sources/ValCommand/ValCommand.swift
```diff
@@ -341,4 +302,49 @@ public struct ValCommand: ParsableCommand {
   /// A map from executable name to path of the named binary.
   private static var executableLocationCache: [String: String] = [:]
 
+  /// Wrotes a textual descriptioni of `input` to the given `output` file.
```
kyouko-taiga, line 305:
```suggestion
  /// Writes a textual descriptioni of `input` to the given `output` file.
```

Sources/ValCommand/ValCommand.swift
```diff
@@ -341,4 +302,49 @@ public struct ValCommand: ParsableCommand {
...
+    outputURL?.deletingPathExtension() ?? URL(fileURLWithPath: productName)
+  }
+
+}
+
+extension TypedProgram {
+
+  typealias CXX = (syntax: CXXModule, text: TranslationUnitCode)
+
+  func cxx(_ m: ModuleDecl.Typed) -> CXX {
```
kyouko-taiga, line 344: No docs?

Sources/ValCommand/ValCommand.swift
```diff
@@ -135,103 +135,65 @@ public struct ValCommand: ParsableCommand {
...
       return
     }
 
-    // *** C++ Transpiling ***
+    // C++
 
-    // Initialize the transpiler & code writer.
-    let transpiler = CXXTranspiler(typedProgram)
-    var codeWriter = CXXCodeWriter()
+    let cxxModules = (core: p.cxx(p.corelib!), source: p.cxx(p[sourceModule]))
```
lucteo, line 167: neat!

Sources/ValCommand/ValCommand.swift
```diff
@@ -341,4 +302,49 @@ public struct ValCommand: ParsableCommand {
...
+
+}
+
+extension TypedProgram {
+
+  typealias CXX = (syntax: CXXModule, text: TranslationUnitCode)
+
+  func cxx(_ m: ModuleDecl.Typed) -> CXX {
+    let x = CXXTranspiler(self).cxx(m)
+    var w = CXXCodeWriter()
```
lucteo, line 346: Note: I'm also refactoring this, eliminating the need for CXXCodeWriter variable. See https://github.com/val-lang/val/commit/62d4ebba881d04c9a518d5cf346d61e9c9826b7e#diff-afa50b06bcd7e4dfe5601a33c9aa656568e9784f1f0699a4ef9fe2c683705d13

I can put my changes on top of this.

I'm starting to contemplate now if we should put this extension in CXXTranspiler itself, and remove the need to publicly use the class. (similar to what I've started to do with CXXCodeWriter)

Sources/ValCommand/ValCommand.swift
```diff
@@ -341,4 +302,49 @@ public struct ValCommand: ParsableCommand {
...
+  /// "cpp" is selected as the output type.
+  private func sourceModuleCXXOutputBase(_ productName: String) -> URL {
+    outputURL?.deletingPathExtension() ?? URL(fileURLWithPath: productName)
+  }
+
+}
+
+extension TypedProgram {
+
+  typealias CXX = (syntax: CXXModule, text: TranslationUnitCode)
```
lucteo, line 342: I don' think we use `syntax` anymore. If that's the case, then `CXX` == `TranslationUnitCode`.

Also, I would avoid using both `CXX` and `cxx` in the same context, as they can be easily misread (same shape, just the size of the letters are different)

lucteo (commented):
LGTM

Sources/CodeGen/CXX/CXXTranspiler.swift
dabrahams: Very.  #399 is about the identifiers used.  Other stuff is a separate issue.

Sources/ValCommand/ValCommand.swift
dabrahams: I think you should define it, but I'll be happy to talk to you about that if you want.

Sources/ValCommand/ValCommand.swift
dabrahams: `func index(after i: Index) -> Index` is a problem?

Sources/ValCommand/ValCommand.swift
dabrahams: Yes, the `struct` is not a publicly-useful abstraction.  If it should exist at all, it should be `fileprivate`.

Sources/ValCommand/ValCommand.swift
dabrahams: I renamed it `TypedProgram.CXXModule`

Sources/ValCommand/ValCommand.swift
dabrahams: fixed

Sources/ValCommand/ValCommand.swift
dabrahams: "The typed program" is not a role.  A role would be something more like “the source of the IR”

PR #407, opened 2023-02-15
Remove the pass pipeline

kyouko-taiga:


dabrahams (changes requested):
You don't need the array or the loop or the typealias.  This is just a function that calls some other functions. It's not a foregone conclusion that these functions will always want the same set of parameters; the uniformity is a coincidence and at best you can't take advantage of it.  Please, for the love o' Mike, just call each one.  Also, please see if you can find a name for the function that doesn't talk about "passes" but says something about the final state.  Maybe something like `prepareForExecution`?

kyouko-taiga:
I'm guessing you want:

```swift
for f in functions {
  pass1()
  pass2()
  ...
}
```

This will be an issue in the long run.

The order in which the passes apply will be significant once we implement whole-module transformations, like inlining. Further, whole-module passes will also likely be stateful and can't be re-initialized at each loop iteration. So we need the outer loop to iterate over passes, not functions, hence the array.

dabrahams:
When the order of operations is significant, we can control that by reordering them in source code.  This is normal, innit?

```swift
a.sort()
a.removeLast()
```

What's wrong with 
```
func prepareForExecution() {
  pass1(functions)
  pass2(functions)
  ...
}
```

?  If these passes have in common that they process every function once, *maybe* that's worth capturing somehow.

kyouko-taiga:
> What's wrong with
> ```swift
> func prepareForExecution() {
>   pass1(functions)
>   pass2(functions)
>   ...
> }
> ```
It pushes the loop over functions into every pass. I think that's unnecessary boilerplate that it's currently factored out.

> If these passes have in common that they process every function once, maybe that's worth capturing somehow.

That's what I've been doing the whole time.

dabrahams (approved):
I don't buy it yet, but since you're convinced, I won't stand in the way

Issue #252, opened 2023-01-23
Windows CI still flaky (again)


dabrahams:
https://github.com/val-lang/val/actions/runs/3990232735/attempts/1

compnerd:
I was unable to reproduce this locally over 25 runs.  I wonder if there is some sort of parallelism issue here.

C-BJ:
I think this Issue may be closed

C-BJ:
@kyouko-taiga

kyouko-taiga:
Fingers crossed

kyouko-taiga:
Seems Windows CI is still flaky :(

kyouko-taiga:
Seems I can't get a link to the failed CI jobs I re-ran. I'll post a link here next time there's a Windows failure.

C-BJ:
![image](https://user-images.githubusercontent.com/62678643/221355064-b75fc5c5-8dd1-4712-af75-4e32e5c03174.png)
@kyouko-taiga

kyouko-taiga:
I've been looking at that history but couldn't find the failure. I don't remember exactly which branch it was, I commit a lot of stuff these days.

The failure is about 3 days ago, feel free to look for yourself.

C-BJ:
https://github.com/val-lang/val/actions/runs/4242066814/jobs/7373093947
Got it @kyouko-taiga

kyouko-taiga:
Here's another one: https://github.com/val-lang/val/actions/runs/4272221910/attempts/1

C-BJ:
https://github.com/val-lang/val/actions/runs/4335819972/jobs/7570671734

https://github.com/val-lang/val/actions/runs/4335676491/jobs/7570429339

kyouko-taiga:
Sadly, the flakiness is getting worse and starts seriously impeding development. So I decided to disable Windows CI until we can figure out a solution (see https://github.com/val-lang/val/pull/495).

C-BJ:
I noticed that the things happens when using Swift locally same as CI.
Swift also crashes locally.
@kyouko-taiga

C-BJ:
I'm thinking this isn't a problem we can solve either, so what should we do?

C-BJ:
I have an idea

C-BJ:
Assign this issue to me?

compnerd:
@C-BJ would you be able to get a stack trace or minidump from the failure?  That would help understand what is going on (hitting an assert/invalid memory access).

[...]

PR #418, opened 2023-02-18
Update `windows-tag` of Swift in CI

C-BJ:
Windows CI fails too often. Perhaps try updating Swift for Windows again?


C-BJ:
If all goes well after this PR, consider turning #252 off

kyouko-taiga:
Let's try for a while.

PR #425, opened 2023-02-19
Fix printed format of Diagnostic so it is recognized by IDE(mac)s.

### LINKED ISSUE #326: Diagnostics in test failures don't have the right format.
For example, running the tests, I got:

/Users/dave/src/val/Sources/Core/Dianostics.swift:45: error: -[ValCommandTests.ValCommandTests testTypeCheckFailure] : failed: caught error: "file:///Users/dave/src/val/.build/arm64-apple-macosx/debug/Val_ValCommandTests.bundle/Inputs/Failure.val:2:11: error: undefined name 'foo' in this scope"

`/Users/dave/src/val/Sources/Core/Dianostics.swift:45: error:` is parsed fine by my IDE<i>macs</i>, but we need a newline just after the open quotation mark and we need to convert the URL back into a file path.

dabrahams:
Fixes #326.  Please don't fight me on this; capturing these so they show up nicely in XCode is a separate step and must be done explicitly.

### DIFF
--- a/Sources/Core/Diagnostic.swift
+++ b/Sources/Core/Diagnostic.swift
@@ -61,7 +61,7 @@ extension Diagnostic: CustomStringConvertible {
     let prefix: String
     let l = site.first()
     let (line, column) = l.lineAndColumn
-    prefix = "\(l.file.url):\(line):\(column): "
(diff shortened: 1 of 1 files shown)


kyouko-taiga (approved):
OK

PR #426, opened 2023-02-19
Diagnosable sourcefile literals for testing

dabrahams:
During testing, allows unexpected diagnostic reports from Val source code in Swift multiline string literals to point back at the exact place in the original String literal where the problem occurred.

If you insert errors into the source that's now constructed with `SourceFile.diagnosableLiteral` you can see the results for yourself.

### DIFF
--- a/Sources/Core/SourceFile.swift
+++ b/Sources/Core/SourceFile.swift
@@ -33,18 +33,16 @@ public struct SourceFile {
     self.storage = storage
   }
 
-  /// Creates an instance for the given `text` embedded in a multiline string literal in the given
-  /// `swiftFile` at `swiftLine`.
+  /// Creates an instance for the `text` given by a multiline string literal in the given
+  /// `swiftFile`, the literal's textual content (the line after the opening quotes) being
+  /// startLine.
   ///
-  /// The actual text processed will be what's read from the file, regardless of what's in `text`.
-  /// That means if you're going to use any special characters in the literal, the literal should be
-  /// a raw string literal to avoid confusion.
-  fileprivate init(
-    diagnosableLiteral text: String, swiftFile: String = #filePath, swiftLine: Int = #line
-  )
+  /// N.B. The text of the instance will literally be what's in the Swift file, including its
+  /// indentation and any embedded special characters, even if the literal itself is not a raw
+  /// literal or has had indentation stripped by the Swift compiler.
+  fileprivate init(diagnosableLiteral text: String, swiftFile: String, startLine: Int)
     throws
   {
-    let startLine = swiftLine + 2  // Account for typical source code formatting
     let wholeFile = try SourceFile(contentsOf: URL(fileURLWithPath: swiftFile))
     let endLine = startLine + text.lazy.filter(\.isNewline).count
     let fragment = URL(string: "\(wholeFile.url.absoluteString)#L\(startLine)-L\(endLine)")!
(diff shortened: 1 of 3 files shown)


Sources/Core/SourceFile.swift
```diff
@@ -16,7 +14,9 @@ public struct SourceFile {
   public typealias Index = String.Index
 
   /// The contents of the source file.
-  public var text: String { storage.text }
+  ///
+  /// - Note: except for
```
kyouko-taiga, line 18: Except for what?

Sources/Core/SourceFile.swift
```diff
@@ -26,7 +26,29 @@ public struct SourceFile {
...
-    let storage = try Storage(filePath) { try String(contentsOf: filePath) }
+    let storage = try Storage(filePath) { try String(contentsOf: filePath)[...] }
+    self.storage = storage
+  }
+
+  /// Creates an instance for the `text` given by a multiline string literal in the given
+  /// `swiftFile`, the literal's textual content (the line after the opening quotes) being
+  /// startLine.
+  ///
+  /// N.B. The text of the instance will literally be what's in the Swift file, including its
```
kyouko-taiga, line 37:
```suggestion
  /// The text of the instance will literally be what's in the Swift file, including its
```

Very minor, but let's try to avoid introducing different way to write additional notes in the docs.

Sources/Core/SourceFile.swift
```diff
@@ -135,6 +157,31 @@ public struct SourceFile {
 
 }
 
+extension SourceFile {
+
+  /// Returns a SourceFile containing the given text of a multiline string literal, such that
+  /// diagnostics produced in processing that file will point back to the original Swift source.
+  ///
+  /// N.B. The text of the result will literally be what's in the Swift file, including its
```
kyouko-taiga, line 165:
```suggestion
  /// The text of the result will literally be what's in the Swift file, including its
```

Sources/Core/SourceFile.swift
```diff
@@ -135,6 +157,31 @@ public struct SourceFile {
...
+  /// indentation and any embedded special characters, even if the literal itself is not a raw
+  /// literal or has had indentation stripped by the Swift compiler.  It is assumed that the first
```
kyouko-taiga, line 167:
```suggestion
  /// literal or has had indentation stripped by the Swift compiler. It is assumed that the first
```

Sources/Core/SourceFile.swift
```diff
@@ -26,7 +26,29 @@ public struct SourceFile {
...
+  /// Creates an instance for the `text` given by a multiline string literal in the given
+  /// `swiftFile`, the literal's textual content (the line after the opening quotes) being
+  /// startLine.
+  ///
+  /// N.B. The text of the instance will literally be what's in the Swift file, including its
+  /// indentation and any embedded special characters, even if the literal itself is not a raw
+  /// literal or has had indentation stripped by the Swift compiler.
+  fileprivate init(diagnosableLiteral text: String, swiftFile: String, startLine: Int)
+    throws
+  {
```
kyouko-taiga, line 42:
```suggestion
  fileprivate init(
    diagnosableLiteral text: String, swiftFile: String, startLine: Int
  ) throws {
```

You've used this style below, which I prefer.

kyouko-taiga (approved):
It's a quite nice technology but I'm not sure it's really needed. We have very few tests using Swift literals and I don't think it will change. I very much prefer writing test cases in separate files to get a better IDE experience (I have minimal syntax highlighting for VSCode).

But I won't stand in the way if you want this feature.

Sources/Core/SourceFile.swift
dabrahams, line 18:
```suggestion
```

Sources/Core/SourceFile.swift
dabrahams: I let swift-format do what it wants, and I don't think we should be asking for this kind of formatting change in code reviews, but I'll accept your suggestion.

dabrahams:
Yes I want this; I'll be writing interpreter tests this way.

Sources/Core/SourceFile.swift
```diff
@@ -302,12 +350,17 @@ extension SourceFile {
...
+      self.init(e.url) { e.text[...] }
     }
 
     public func encode(to encoder: Encoder) throws {
       var container = encoder.singleValueContainer()
-      try container.encode(Encoding(url: url, text: text))
+      // This could be supported, with caveats, but it's not necessarily a good idea.
+      precondition(
+        text.startIndex == text.base.startIndex && text.endIndex == text.base.endIndex,
+        "Serialization of Val source embedded in Swift string literals is not supported."
```
dabrahams, line 361:
```suggestion
        "Serialization of SourceFile.diagnosableLiteral results is not supported."
```

kyouko-taiga:
> Yes I want this; I'll be writing interpreter tests this way.

I think we should discuss this.

We agreed we'd have separate Val files for tests and generally that's the direction I want to push. There are many advantages to this approach, including the fact that we'll also be able to reuse test cases to compare output across different compilation pipelines.

dabrahams:
The tests I'm talking about are minimal and wouldn't be particularly useful for that.

dabrahams:
I don't think that reuse is going to work well even for separate-file tests, whose annotations are going to have to describe interpreter state.

Issue #431, opened 2023-02-19
Support the MSVC and GCC


C-BJ:
I want to assigned this issue to myself

C-BJ:
I will finish the task next week

PR #450, opened 2023-02-23
Support the MSVC and GCC & add `--cc-flags` command line argument & Update `swift-argument-parser` to version `1.2.0`

C-BJ:
Closes: #431

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,4 +1,4 @@
-name: CI
+name: Build and test
 
 on:
   push:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -20,7 +20,7 @@ on:
 
 jobs:
   build:
-    name: Swift 5 on ${{ matrix.host.os }}-Clang backend
+    name: Swift 5 on ${{ matrix.host.os }} with Clang backend
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,4 +1,4 @@
-name: Build and test
+name: CI
 
 on:
   push:
@@ -20,7 +20,7 @@ on:
 
 jobs:
   build:
-    name: Swift 5 on ${{ matrix.host.os }}
+    name: Swift 5 on ${{ matrix.host.os }}-Clang backend
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -20,7 +20,7 @@ on:
 
 jobs:
   build:
-    name: Swift 5 on ${{ matrix.host.os }} with Clang backend
+    name: Swift 5 on ${{ matrix.host.os }}
     strategy:
       fail-fast: false
       matrix:
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ You can select how deep the compiler should go through the pipeline with the fol
 - `--emit cpp`: Produce a C++ source file.
 - `--emit binary` (default): Produce an executable 
   - Note: by default, C++ files will be compiled with `Clang`. Use `--cc {CXX compiler}` to use another compiler.
-  - Note: You can specify parameters for the CXX compiler to use. like `--cc-flags O3`.
+  - Note: You can specify parameters for the CXX compiler to use (e.g., `--cc-flags O3`).
     - Note: Don't add an extraneous `-`, please use `O3` instead of `-O3`. 
     - Note to MSVC users: Don't add an extraneous `/`, please use `O1` instead of `/O1`. 
     - You can also add more than one such parameter.
--- a/README.md
+++ b/README.md
@@ -38,7 +38,10 @@ You can select how deep the compiler should go through the pipeline with the fol
 - `--emit raw-ir`: Lower the typed AST into Val IR and output the result in a file.
 - `--emit ir`: Run mandatory IR passes and output the result in a file.
 - `--emit cpp`: Produce a C++ source file.
-- `--emit binary` (default): Produce an executable (Clang is required for now).
+- `--emit binary` (default): Produce an executable (Clang / GCC / MSVC is required for now).
+  - The default is `Clang` followed by `GCC` followed by `MSVC`
+  - If you want to customize the C++ compiler, you can set the ValCXXCompiler environment variable to `clang`, `gcc`, `msvc`
+  - If you use MSVC, be sure to use Developer Command Prompt or Developer PowerShell
 
 For example, `valc --emit raw-ast -o main.json main.val` will parse `main.val`, write the untyped AST in `main.json`, and exit the pipeline.
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ You can select how deep the compiler should go through the pipeline with the fol
 - `--emit cpp`: Produce a C++ source file.
 - `--emit binary` (default): Produce an executable (Clang / MSVC / GCC is required for now).
   - The default is `Clang` followed by `MSVC` followed by `GCC`
-  - If you want to customize the C++ compiler, you can set the ValCXXCompiler environment variable to `clang`, `msvc`, `gcc`
+  - If you want to customize the C++ compiler, you can set the `ValCXXCompiler` environment variable to `clang`, `msvc`, `gcc`
   - If you use MSVC, be sure to use Developer Command Prompt or Developer PowerShell
 
 For example, `valc --emit raw-ast -o main.json main.val` will parse `main.val`, write the untyped AST in `main.json`, and exit the pipeline.
--- a/README.md
+++ b/README.md
@@ -40,7 +40,10 @@ You can select how deep the compiler should go through the pipeline with the fol
 - `--emit cpp`: Produce a C++ source file.
 - `--emit binary` (default): Produce an executable 
   - Note: by default, C++ files will be compiled with `Clang`. Use `--cc {CXX compiler}` to use another compiler.
-  - Note: You can specify parameters for the CXX compiler to use. like `--cc-flags -O3`.
+  - Note: You can specify parameters for the CXX compiler to use. like `--cc-flags O3`.
+    - Note: Don't add an extraneous `-`, please use `O3` instead of `-O3`. 
+    - You can also add more than one such parameter.
+    - If you want to specify output file name or something like that, please use the Val command line argument instead of `--cc-flags`.
   - Note to MSVC users: be sure to use Visual Studio Developer Command Prompt or Visual Studio Developer PowerShell.
 
 For example, `valc --emit raw-ast -o main.json main.val` will parse `main.val`, write the untyped AST in `main.json`, and exit the pipeline.
--- a/README.md
+++ b/README.md
@@ -39,10 +39,10 @@ You can select how deep the compiler should go through the pipeline with the fol
 - `--emit ir`: Run mandatory IR passes and output the result in a file.
 - `--emit cpp`: Produce a C++ source file.
 - `--emit binary` (default): Produce an executable 
-  - `Clang` / `MSVC` / `GCC` is required for now. But only Clang has received the most formal support.
+  - `Clang` / `MSVC` / `GCC` is required for now. But only `Clang` has received the formal support.
   - The default CXXCompiler is `Clang`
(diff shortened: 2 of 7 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and test
+name: CI
```
kyouko-taiga, line 1: Why is the length of this name an issue?

"CI" is less descriptive than "Build and test". We might want to add other workflows to CI in the future.

README.md
```diff
@@ -38,7 +38,10 @@ You can select how deep the compiler should go through the pipeline with the fol
 - `--emit raw-ir`: Lower the typed AST into Val IR and output the result in a file.
 - `--emit ir`: Run mandatory IR passes and output the result in a file.
 - `--emit cpp`: Produce a C++ source file.
-- `--emit binary` (default): Produce an executable (Clang is required for now).
+- `--emit binary` (default): Produce an executable (Clang / MSVC / GCC is required for now).
```
kyouko-taiga, line 41: I'd rather have the command line accept an explicit parameter to select the C++ compiler to use and default to clang if the flag isn't set. Val will only grow closer to clang as time goes, so it makes sense to be opinionated here.

Sources/ValCommand/ValCommand.swift
```diff
@@ -212,21 +212,53 @@ public struct ValCommand: ParsableCommand {
     try write(
       cxxModules.source, to: buildDirectory.appendingPathComponent(productName),
       loggingTo: &errorLog)
-
-    let clang = try find("clang++")
-
+      
+    let whichCompiler = ProcessInfo.processInfo.environment["ValCXXCompiler"] ?? ""
```
kyouko-taiga, line 216: This value should be read from a command line argument.

kyouko-taiga (changes requested):
Thanks a lot for this work!

I'd like to keep MVSC and GCC as nice-to-have features rather than committing to support those compilers. One way to alleviate our commitment is to have our compiler allow the user to select the compiler they want to use as a command line argument. Then anyone will be free to opt-out clang to use MVSC, GCC, or anything else.

That also means we don't need to hardcode an arbitrary order in our compiler.

C-BJ:
I finished this PR, btw
image
Does the description of the repository need to be changed
And why aren't Github discussions in the main repository?

kyouko-taiga:
> I finished this PR, btw

Thanks, I'll look at the changes later today.

> Does the description of the repository need to be changed

Why would the description have to change?

> why aren't Github discussions in the main repository? Wouldn't that make easier to access(jump)?

We thought it would be best to keep all discussions not related to the compiler implementation in a separate repository. You can easily access them by hoping through the organization's page.

C-BJ:
I simply thought we would have the Library in this repository, so I just mention it

.github/workflows/build-and-test.yml
```diff
@@ -20,7 +20,7 @@ on:
 
 jobs:
   build:
-    name: Swift 5 on ${{ matrix.host.os }}
+    name: Swift 5 on ${{ matrix.host.os }} with Clang backend
```
kyouko-taiga, line 23: This change is unnecessary. We'll always go for the clang backend, at least for the foreseeable future.

README.md
```diff
@@ -38,7 +38,11 @@ You can select how deep the compiler should go through the pipeline with the fol
 - `--emit raw-ir`: Lower the typed AST into Val IR and output the result in a file.
 - `--emit ir`: Run mandatory IR passes and output the result in a file.
 - `--emit cpp`: Produce a C++ source file.
-- `--emit binary` (default): Produce an executable (Clang is required for now).
+- `--emit binary` (default): Produce an executable 
+  - `Clang` / `MSVC` / `GCC` is required for now. But only `Clang` has received the formal support.
```
kyouko-taiga, line 42: After discussing with other contributors, we've decided that it's better to not even mention support for other compilers. We'll target clang and the ability to use another compiler if you want, without any guarantee.
```suggestion
```

README.md
```diff
@@ -38,7 +38,11 @@ You can select how deep the compiler should go through the pipeline with the fol
...
+  - The default CXXCompiler is `Clang`
+  - If you want to use a compiler other than `Clang`, please use `--CXXCompiler gcc` or `--CXXCompiler msvc`
+  - If you want to use MSVC, be sure to use Visual Studio Developer Command Prompt or Visual Studio Developer PowerShell
```
kyouko-taiga, line 45:
```suggestion
  - Note: by default, C++ files will be compiled with clang. Use `--cc /some/c++/compiler` to use another compiler.
  - Note to MSVC users: be sure to use Visual Studio Developer Command Prompt or Visual Studio Developer PowerShell.
```

kyouko-taiga (changes requested):
After discussing with other contributors, we've decided that we should focus on one C++ compiler rather than hinting about more support.

The current position is that supporting three major C++ compilers is too much work at this stage of the project. Even if support is not "formal", users will rightfully expect most features to just work if there's a gcc or mvsc option built-in the command line.

Our priority is to get a working version of Val and use transpilation to explain the language's semantics in terms of C++. It is possible that we'll work toward supporting the three major C++ compilers in the future, once we shift our focus on interoperability.

In the context of the current PR, I suggest you modify the Val compiler command line so that it can accept the path/name of a custom C++ compiler along with another option to forward flags to that C++ compiler. The idea is to be able to execute:
```bash
valc --cc gcc --cc-flags -O3 factorial.val
```

C-BJ:
After we got the LLVM back end, wasn't it difficult to support interoperability support for multiple versions of CXXCompiler?
As far as I know, Carbon only plans to interoperate with Cpp when using the Clang compiler
Or maybe you're not talking about the LLVM back end?
@kyouko-taiga

kyouko-taiga:
> After we got the LLVM back end, wasn't it difficult to support interoperability support for multiple versions of CXXCompiler?

Yes and no.

Currently, the transpilation story is completely orthogonal to LLVM. That's because the transpiler just walks Val ASTs and generate C++ text. We can continue to operate that way as long as we can afford a second-class interoperability from C++ to Val, i.e., use a C++ API inside a Val program.

In that picture, LLVM is just another backend and generating LLVM IR is no different than generating Javascript sources.

Once we want C++ to Val interoperability, however, we'll need a way to read C++ sources to generate the corresponding Val APIs and necessary glue code. I plan on using clang to do that, meaning there's a connection to LLVM at this point. I have almost no idea how we'll deal with ABI, but I suspect that's the point where we'll diverge more and more from other compilers.

Sources/ValCommand/ValCommand.swift
```diff
@@ -84,6 +105,20 @@ public struct ValCommand: ParsableCommand {
...
+      "Customize the CXX compiler used by the Val backend. From: clang, gcc, msvc(Windows only)",
+      valueName: "CXXCompiler"))
+  private var customizeCXXCompiler: CXXCompiler = .clang
+
+  @Option(
+    name: [.customLong("cc-flags")],
+    help: ArgumentHelp(
+      "Specify flags for the CXX compiler to use",
+      valueName: "CXXCompilerFlags"))
+  private var ccFlags: [String] = ["Empty"]
```
kyouko-taiga, line 120: Why the need for a particular value and not simple:
```suggestion
  private var ccFlags: [String] = []
```

Sources/ValCommand/ValCommand.swift
```diff
@@ -84,6 +105,20 @@ public struct ValCommand: ParsableCommand {
       valueName: "output-type"))
   private var outputType: OutputType = .binary
 
+  @Option(
+    name: [.customLong("cc")],
+    help: ArgumentHelp(
+      "Customize the CXX compiler used by the Val backend. From: clang, gcc, msvc(Windows only)",
+      valueName: "CXXCompiler"))
+  private var customizeCXXCompiler: CXXCompiler = .clang
```
kyouko-taiga, line 113:
```suggestion
  private var cxxCompiler: CXXCompiler = .clang
```
It's not necessarily custom; we'll always use this value IIUC.

Sources/ValCommand/ValCommand.swift
```diff
@@ -212,21 +247,49 @@ public struct ValCommand: ParsableCommand {
...
+    
+    let binaryPath = executablePath(outputURL, productName)
 
-    let clang = try find("clang++")
+    #if os(Windows)
+      if customizeCXXCompiler == .msvc {
+        var arguments =  [ buildDirectory.appendingPathComponent(productName + ".cpp").path,
+          "/link", "/out:" + binaryPath]
+        
+        if ccFlags != ["Empty"] {
```
kyouko-taiga:
```suggestion
        if !ccFlags.isEmpty {
```
Assuming we can declare `ccFlags: [String] = []` as I suggested.

kyouko-taiga (commented):
Starting to look very good. I left a few inline comments.

Sources/ValCommand/ValCommand.swift
C-BJ: Sorry, my fault

C-BJ:
@kyouko-taiga

Sources/ValCommand/ValCommand.swift
```diff
@@ -212,21 +247,47 @@ public struct ValCommand: ParsableCommand {
...
+    let binaryPath = executablePath(outputURL, productName)
 
-    let clang = try find("clang++")
+    #if os(Windows)
+      if cxxCompiler == .msvc {
+        var arguments: [String] =  []
+        
+        for i in (0 ..< ccFlags.count) {
+          arguments.append("/" +  ccFlags[i])
+        }
```
kyouko-taiga, line 266:
```suggestion
        var arguments = ccFlags.map({ "/\($0)" })
        
        for i in (0 ..< ccFlags.count) {
          arguments.append("/" +  ccFlags[i])
        } 
```

Sources/ValCommand/ValCommand.swift
```diff
@@ -212,21 +247,47 @@ public struct ValCommand: ParsableCommand {
...
+      } 
+    #endif
 
-    let binaryPath = executablePath(outputURL, productName)
+    var arguments = ["-o", binaryPath, "-I", buildDirectory.path, 
+      buildDirectory.appendingPathComponent(productName + ".cpp").path]
+    
+    for i in (0 ..< ccFlags.count) {
+      arguments.append("-" +  ccFlags[i])
+    }
```
kyouko-taiga, line 284:
```suggestion
    arguments += ccFlags.map({ "-\($0)" })
```

README.md
```diff
@@ -38,7 +38,14 @@ You can select how deep the compiler should go through the pipeline with the fol
 - `--emit raw-ir`: Lower the typed AST into Val IR and output the result in a file.
 - `--emit ir`: Run mandatory IR passes and output the result in a file.
 - `--emit cpp`: Produce a C++ source file.
-- `--emit binary` (default): Produce an executable (Clang is required for now).
+- `--emit binary` (default): Produce an executable 
+  - Note: by default, C++ files will be compiled with `Clang`. Use `--cc {CXX compiler}` to use another compiler.
+  - Note: You can specify parameters for the CXX compiler to use. like `--cc-flags O3`.
```
kyouko-taiga, line 43:
```suggestion
  - Note: You can specify parameters for the CXX compiler to use (e.g., `--cc-flags O3`).
```

README.md
```diff
@@ -38,7 +38,14 @@ You can select how deep the compiler should go through the pipeline with the fol
...
+    - Note: Don't add an extraneous `-`, please use `O3` instead of `-O3`. 
+    - Note to MSVC users: Don't add an extraneous `/`, please use `O1` instead of `/O1`. 
+    - You can also add more than one such parameter.
```
kyouko-taiga, line 46:
```suggestion
    - You can also add more than one parameter (e.g., `--cc-flags O1 g`)
```
Or maybe I misunderstood your point here.

Another question I have, what happens if we want to pass a flag with an argument?
That can be addressed in a separate PR if this one doesn't support it.

C-BJ: > Another question I have, what happens if we want to pass a flag with an argument?

Let's do that later in another PR

C-BJ: Actually we need `--cc-flags O1 --cc-flags g` instead `--cc-flags O1 g` I'm going to make this clear

kyouko-taiga (approved):
Really nice!

Thanks for the hard work.

C-BJ:
I think this PR can be merge @kyouko-taiga

C-BJ:
I think that's all done @kyouko-taiga

PR #452, opened 2023-02-23
Support Trailing Commas in Literals

### LINKED ISSUE #129: Support trailing separators in compound literals
The parse rejects `[a, b,]` and `[a:1, b:2,]`.

nickpdemarco:
Resolves #129. Following @kyouko-taiga's gradual migration away from Durian, I've rewritten the parsers for map and array literals in a more imperative form, and added tests to verify that trailing commas parse successfully.


Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2481,6 +2454,56 @@ public enum Parser {
...
+
+  private static let nonemptyMapLiteral = DelimitedCommaSeparatedList(
+    openerKind: .lBrack,
+    closerKind: .rBrack,
+    closerDescription: "]",
+    elementParser: Apply(parseMapComponent(in:)))
+
+  private static func parseMapComponent(in state: inout ParserState) throws -> MapLiteralExpr.Element? {
+
+    let backup = state.backup()
```
nickpdemarco, line 2471: @kyouko-taiga Can you provide a general guideline for when to backup/restore state in these routines?

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2481,6 +2454,56 @@ public enum Parser {
...
+    closerDescription: "]",
+    elementParser: Apply(parseExpr(in:)))
+
+  private static let nonemptyMapLiteral = DelimitedCommaSeparatedList(
+    openerKind: .lBrack,
+    closerKind: .rBrack,
+    closerDescription: "]",
+    elementParser: Apply(parseMapComponent(in:)))
+
+  private static func parseMapComponent(in state: inout ParserState) throws -> MapLiteralExpr.Element? {
```
nickpdemarco, line 2469: @kyouko-taiga would you support renaming this to `parseMapElement`? I'm not sure if component is a term of art, but it seems odd to use a different term than the return type.

kyouko-taiga: SGTM

Sources/FrontEnd/Parse/Parser.swift
kyouko-taiga: You need to backup if you want to try another alternative after you have consumed tokens. Everything that can be written with only 1-token lookahead should never have to use `backup`.

For example, if your routine parses this rule:
```ebnf
rule1 ::=
  foo bar
  ham
```
Then you don't need `backup` because with 1-token lookahead you can identify if you'll commit to the first or second alternative, or just turn `nil` (soft failure). If you fail to parse `bar` after `foo`, you should throw (hard failure).

But if you parse this rule:
```ebnf
rule2 ::=
  foo bar* biz
  foo ham biz
```
You'll have to backup after `foo` in case you parsed some `bar` but then fail to parse a `biz`. That's because you'll want to try the second alternative then.
```swift
func parseRule2(in state: inout ParserState) -> Rule2 {
  // Consume `bar` or return a soft failure.
  guard let bar = parseBar(in: &state) else { return nil }

  // Parse the remainder of the first alternative
  let backup = state.backup()  
  var bars: [Ham] = []
  while let b = try parseBar(in: &state) { bars.apend(b) }
  if let biz = parseBiz(in: &state) { return (bar, bars, biz) }

  // Backtrack until after `foo` and parse the remainder of the second alternative.
  state.restore(backup)
  let ham = try state.expect("ham", parseHam(in:))
  let biz = try state.expect("biz", parseBiz(in:))
  return (foo, ham, biz)
}
```

In general, parsers expect not to be called in vain so they should commit as soon as possible and throw (hard failure) when they fail. This approach lets you control backtracking very finely. If a parser wants to try several alternatives that might commit too early, it should catch their hard failures. So another way to implement `rule2` is as follows (notice how we get less control over backtracking ,though):
```swift
func parseRule2(in state: inout ParserState) -> Rule2 {
  let backup = state.backup()
  do {
    // try `foo bar* biz`, soft failures are ignored, hard failures are caught
    if let x = try rule2_first() { return x }
  } catch {
    // recover from a hard failure
    state.restore(backup)
  }
  // try `foo ham biz`
  return try rule_2second()
}
```
The drawback of this approach is that we'll always get the hard failures of the second alternative if none of them worked. That's unfortunate because it's more intuitive and often necessary to parse the preferred alternative first. One workaround is to keep your hard failure around and rethrow it if the second alternative also failed.
```swift
func parseRule2(in state: inout ParserState) -> Rule2 {
  let backup = state.backup()
  let firstAlternativeFailure: Error
  do {
    // try `foo bar* biz`, soft failures are ignored, hard failures are caught
    if let x = try rule2_first() { return x }
  } catch let e {
    // recover from a hard failure
    firstAlternativeFailure = e
    state.restore(backup)
  }
  // try `foo ham biz`
  if let x = try rule_2second() { return x }
  throw firstAlternativeFailure
}
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -3297,6 +3320,10 @@ struct DelimitedCommaSeparatedList<E: Combinator>: Combinator where E.Context ==
       }
     }
 
+    if closer == nil {
+      return nil
```
kyouko-taiga, line 3324: This `nil` return breaks the rules followed by all parser routines. `nil` should only be returned if the parser didn't consume anything from the stream. The invariant we want is:
```swift
func f(s: inout ParserState) throws {
  let before = s.backup()
  if try parseSomething(in: &s) == nil {
    assert(s == before)
  }
}
```

kyouko-taiga:
`DelimitedCommaSeparatedList` already tells you if it parsed the closer delimiter and a trailing comma as part of its return value. I suppose any logic predicated upon the look of the end of a list can use this information. `parseList(in:with:)` shows an example:
```swift
func parseList(...) throws -> [Element] {
    guard let result = try parser.parse(&state) else { return nil }
    if let s = result.trailingSeparator, result.closer != nil {
      state.diagnostics.insert(.error(unexpectedToken: s))
    }
    return result.elements
  }
```
`result` is the return value of a `DelimitedCommaSeparatedList` whose `trailingSeparator` property is the comma token at the end of the list, or `nil` if there isn't any. The same goes for `closer`.

Sources/FrontEnd/Parse/Parser.swift
nickpdemarco: Well explained. I understand. For future readers (and perhaps to test my understanding), I think the following corrects a few typos in the pseudo-implementation of `parseRule2` above, based on this EBNF:

```ebnf
rule2 ::=
  foo bar* biz
  foo ham biz
```

```swift
func parseRule2(in state: inout ParserState) -> Rule2 {
  // Consume `foo` or return a soft failure.
  guard let foo = parseFoo(in: &state) else { return nil }

  // Parse the remainder of the first alternative
  let backup = state.backup()  
  var bars: [Bar] = []
  while let b = try parseBar(in: &state) { bars.append(b) }
  if let biz = parseBiz(in: &state) { return (bar, bars, biz) }

  // Backtrack until after `foo` and parse the remainder of the second alternative.
  state.restore(backup)
  let ham = try state.expect("ham", parseHam(in: &state))
  let biz = try state.expect("biz", parseBiz(in: &state))
  return (foo, ham, biz)
}
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2315,51 +2315,23 @@ public enum Parser {
...
-          MapLiteralExpr(
-            elements: tree.0.1,
-            site: tree.0.0.site.extended(upTo: state.currentIndex)))
-      }))
-
-  private static let mapComponentEmptyContents =
-    (take(.colon)
-      .map({ (_, _) -> [MapLiteralExpr.Element] in [] }))
+    if let buffer = try bufferLiteral.parse(&state) {
+      let expr = state.insert(BufferLiteralExpr(
```
kyouko-taiga, line 2326: You should run swift-format before we merge. This line, for example, will be rearranged.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2315,51 +2315,23 @@ public enum Parser {
   private static func parseCompoundLiteral(in state: inout ParserState) throws -> AnyExprID? {
     let backup = state.backup()
 
-    // Attempt to parse a map literal.
-    do {
-      if let expr = try mapLiteral.parse(&state) { return AnyExprID(expr) }
-    } catch {}
+    if let map = try parseMapLiteral(in: &state) {
```
kyouko-taiga, line 2318: I think this change is incorrect.

The behavior before allowed to try parsing `[1, 2, 3]` as a map literal, fail because we don't see `:` after `1`, bracktrack and try parsing the expression again as a buffer literal. IIUC, the new version will commit to a hard failure after `[1`.

If tests do not cover this, we need to add one that uses `parseExpr` to parse `[1, 2, 3]` and `[a: 1, b: 2, c: 3]`.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2315,51 +2315,23 @@ public enum Parser {
...
+      let expr = state.insert(map)
+      return AnyExprID(expr)
+    }
 
-    // Backtrack and parse a buffer literal.
     state.restore(from: backup)
```
kyouko-taiga, line 2323: There's nothing to restore in your current code. If the conditional binding above doesn't go through, `state` is guaranteed to be unchanged here.

Sources/FrontEnd/Parse/ParserState.swift
```diff
@@ -178,6 +178,30 @@ struct ParserState {
     }
   }
 
+  /// Adds up to `count` tokens from the lexer into the lookahead buffer and returns the number of added tokens.
+  @discardableResult private mutating func advanceLookahead(_ count: Int) -> Int {
```
kyouko-taiga, line 182: Isn't there a `peek(_:)` function that takes a number?

Also, it's a bit strange this one unconditionally fills the lookahead buffer `count` times. It's possible the buffer already contains enough elements when you your variadic `take(_:)`.

Sources/FrontEnd/Parse/ParserState.swift
```diff
@@ -178,6 +178,30 @@ struct ParserState {
...
+    }
+
+    return count
+  }
+
+  /// Consumes and returns the first `kinds.count` tokens if they have the specified kinds.
+  mutating func take(_ kinds: Token.Kind...) -> [Token]? {
+    advanceLookahead(kinds.count)
+
+    if (lookahead.starts(with: kinds, by: {(token, kind) in token.kind == kind })) {
```
kyouko-taiga, line 195: I think this would be better expressed as:
```suggestion
    if peek(kinds.count).elementsEqual(kinds).by({ (a, b) in a.kind == b }) {
```
Then you don't need your `advanceLookahead` function.

It seems `ParserState.peek(_:)` was removed, but it should look like that:
```swift
mutating func peek(_ n: Int) -> Deque<Token>.SubSequence {
  while lookahead.count < n {
    guard let t = lexer.next() else { break }
    lookahead.append(t)
  }
  return lookahead.prefix(upTo: min(n, lookahead.count))
}
```

Sources/FrontEnd/Parse/Parser.swift
kyouko-taiga: I see now that I was wrong.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2481,6 +2453,46 @@ public enum Parser {
...
+    state.restore(from: backup)
+    return nil
+  }
+
+  private static func parseMapLiteral(in state: inout ParserState) throws -> MapLiteralExpr? {
+    if let mapLiteral = state.take(.lBrack, .colon, .rBrack) {
+      return MapLiteralExpr(elements: [], site: mapLiteral.first!.site)
+    }
+
+    if let expr = try nonemptyMapLiteral.parse(&state), expr.closer != nil {
```
kyouko-taiga, line 2488: We can't return `nil` if `expr.closer == `nil`. Parser state will have been mutated, since `nonemptyMapLiteral` will have consumed at least the opening delimiter.

Sources/FrontEnd/Parse/Parser.swift
nickpdemarco: Done 👍

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2313,53 +2313,22 @@ public enum Parser {
...
-    (take(.lBrack).and(mapComponentListContents.or(mapComponentEmptyContents)).and(take(.rBrack))
-      .map({ (state, tree) -> MapLiteralExpr.ID in
-        state.insert(
-          MapLiteralExpr(
-            elements: tree.0.1,
-            site: tree.0.0.site.extended(upTo: state.currentIndex)))
-      }))
+    if let map = try parseMapLiteral(in: &state) {
+      let expr = state.insert(map)
+      return AnyExprID(expr)
```
kyouko-taiga, line 2318:
```suggestion
      return AnyExprID(state.insert(map))
```

kyouko-taiga (approved):
Beautiful!

I think your variadic `take` is a wonderful idea; I'm sure we'll be able to refactor a bunch of existing code with it.

PR #458, opened 2023-02-25
Refactor conformance enumeration

kyouko-taiga:
Before this PR, `TypeChecker.conformedTraits(of:in:)` and `TypeChecker.realize(conformances:in:)` would return `nil` as soon as they encountered an invalid type expression, compelling all call sites to unwrap the result.

### DIFF
--- a/Sources/FrontEnd/TypeChecking/GenericEnvironment.swift
+++ b/Sources/FrontEnd/TypeChecking/GenericEnvironment.swift
@@ -40,11 +40,8 @@ struct GenericEnvironment {
         registerEquivalence(l: equality.left, r: equality.right)
 
       case let conformance as ConformanceConstraint:
-        var allTraits: Set<TraitType> = []
-        for trait in conformance.traits {
-          guard let bases = checker.conformedTraits(of: ^trait, in: scope)
-          else { return nil }
-          allTraits.formUnion(bases)
+        let allTraits: Set<TraitType> = conformance.traits.reduce(into: []) { (r, t) in
+          r.formUnion(checker.conformedTraits(of: ^t, in: scope))
         }
         registerConformance(l: conformance.subject, traits: allTraits)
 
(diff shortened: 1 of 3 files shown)


Sources/FrontEnd/TypeChecking/GenericEnvironment.swift
```diff
@@ -40,11 +40,8 @@ struct GenericEnvironment {
         registerEquivalence(l: equality.left, r: equality.right)
 
       case let conformance as ConformanceConstraint:
-        var allTraits: Set<TraitType> = []
-        for trait in conformance.traits {
-          guard let bases = checker.conformedTraits(of: ^trait, in: scope)
-          else { return nil }
-          allTraits.formUnion(bases)
+        let allTraits: Set<TraitType> = conformance.traits.reduce(into: []) { (r, t) in
+          r.formUnion(checker.conformedTraits(of: ^t, in: scope))
```
kyouko-taiga, line 44: This change hides errors encountered in `conformedTraits`

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -1320,21 +1308,19 @@ public struct TypeChecker {
...
-          conformances: list,
-          in: program.scopeToParent[AnyScopeID(id)!]!)
-      else { return nil }
+      let rhs = ast[p].conformances
+      let requiredTraits = realize(conformances: rhs, in: program.scopeToParent[AnyScopeID(id)!]!)
 
-      if !traits.isEmpty {
-        let cause = ConstraintOrigin(.annotation, at: ast[list[0]].site)
-        constraints.append(ConformanceConstraint(lhs, conformsTo: traits, origin: cause))
+      if requiredTraits.count != rhs.count {
```
kyouko-taiga, line 1317: This won't work if the conformance list contains duplicate items.

PR #461, opened 2023-02-25
Get rid of the DeclLocator

kyouko-taiga:


Sources/IR/Block.swift
```diff
@@ -15,20 +15,20 @@ public struct Block {
     /// The address of the block in the containing function.
     public var address: Function.Blocks.Address
 
-    /// The ID of the instruction at `instAddress` in the block identified by `self`.
-    public func id(at instAddress: Block.Instructions.Address) -> InstructionID {
-      InstructionID(function, address, instAddress)
+    /// The ID of the instruction at `instructionAddress` in the block identified by `self`.
+    public func appending(_ instructionAddress: Block.Instructions.Address) -> InstructionID {
```
dabrahams, line 19: @kyouko-taiga this is not the way the "ing" suffix works.  You can only use it when it returns a notional modified copy of the receiver:

```swift
let a = [1, 2]
let b = a.appending(3)
assert(b == [1, 2, 3])
```

Sources/IR/Block.swift
```diff
@@ -15,20 +15,20 @@ public struct Block {
...
     }
 
-    /// The operand denoting the `index`-th result of the instruction at `instAddress` in the block
-    /// identified by `self`.
-    public func result(at instAddress: Block.Instructions.Address, index: Int) -> Operand {
-      .result(instruction: id(at: instAddress), index: index)
+    /// The operand denoting the `index`-th result of the instruction at `instructionAddress` in
+    /// the block identified by `self`.
+    public func result(at instructionAddress: Block.Instructions.Address, index: Int) -> Operand {
+      .result(instruction: appending(instructionAddress), index: index)
```
dabrahams, line 31: And at the use-site, such names should not be used with an implicit receiver, because by itself "x = appending z" makes no sense.  It's supposed to be read as "x = (y, appending z)"

dabrahams (commented):
Maybe I need to create a document that makes these kinds of conventions explicit?  I thought the way "x, adding 1" was supposed to be read was obvious, but perhaps not.

Sources/IR/Block.swift
kyouko-taiga: I'm confused. Swift key paths have an `appending` method that is doing this. If you write `let s: KeyPath = \[Int].first!` then `s.appending(path: \.bitWidth)` isn't more a notional copy of `s` than an `InstructionID` is of a `Block.ID`.

Sources/IR/Block.swift
kyouko-taiga: We almost never mention `self` explicitly anywhere. Many methods don't make sense as a plain English sentence when the receiver is implicit. `reduce` is one example.
```
extension Sequence where Element == Int {
  func sum() -> Int { reduce(0, +) } // reduce what?
}
```
I was under the impression that it's generally understood that one should add the receiver when reasoning about the expression.

PR #469, opened 2023-02-28
Trying to imbue more semantics into IR naming/commenting

dabrahams:


Sources/IR/Analysis/Lifetime.swift
```diff
@@ -93,7 +93,7 @@ extension Module {
       { (blocks, use) in blocks.insert(use.user.block) })
 
     // Propagate liveness starting from the blocks in which the operand is being used.
-    let cfg = functions[site.function]!.cfg()
+    let cfg = functions[site.owner]!.cfg()
```
kyouko-taiga, line 96: Seeing "owner" makes me think about Val's ownership model, especially during IR analysis. I think we need another name.

Sources/IR/Function.swift
```diff
@@ -1,33 +1,40 @@
 import Core
-import Utils
 import Foundation
+import Utils
+
+// FIXME: it's a point of confusion for me that we have things named "Function" and "Module" and
+// FIXME: they happen to be in the IR, without seeing a distinguishing prefix like IRFunction.  One
+// FIXME:  possibility is thatwe start qualifying these names at their use sites.
```
kyouko-taiga, line 7: Id'd be little sad if we start baking IR qualifiers to everything IR-related. The type is already qualified by the name of its (Swift) module, i.e., it's already called `IR.Function`.

Sources/IR/Function.swift
```diff
@@ -1,33 +1,40 @@
...
 
 /// A collection of basic blocks representing a lowered function.
 public struct Function {
 
   /// A collection of blocks with stable identities.
   public typealias Blocks = DoublyLinkedList<Block>
 
-  /// The profile of a IR function input.
-  public typealias Input = (convention: AccessEffect, type: LoweredType)
+  /// A parameter's type and convention (`let`, `inout`, `sink`).
```
kyouko-taiga, line 15:
```suggestion
  /// A parameter's type and convention (`let`, `inout`, `sink`, `yielded`).
```

Sources/IR/Function.swift
```diff
@@ -1,33 +1,40 @@
...
 
   /// The mangled name of the function.
   public let name: String
 
   /// The debug name of the function, if any.
   public let debugName: String?
 
+  // DWA FIXME: what does it mean for a function to be "anchored" somewhere?
+
+  // Is this just "site?" Shouldn't we store the whole source range though the closing brace?
```
kyouko-taiga, line 26: No, because not all IR functions are lowered version of a Val function with an actual site in the AST. For example, the synthesized implementations of `Int`'s conformance to `Sinkable` don't exist anywhere in the AST but they must be anchored somewhere because the IR analysis needs a site when it generates diagnostics.

Sources/IR/Operands/Operand.swift
```diff
@@ -10,16 +10,17 @@ public enum Operand: Hashable {
   /// A constant value.
   case constant(Constant)
 
+  // FIXME: this is just used for a precondition check.  Worth defining?
```
kyouko-taiga, line 13: I think so. The precondition catches a pretty common mistake one can make when generating IR. I certainly made this kind of mistakes more often than I can count using LLVM.

Sources/IR/Analysis/Module+InsertImplicitReturns.swift
```diff
@@ -8,15 +8,15 @@ extension Module {
   /// - Requires: `f` is in `self`.
   public mutating func insertImplicitReturns(in f: Function.ID, diagnostics: inout DiagnosticSet) {
     /// The expected return type of the function.
-    let expectedReturnType = self[f].output.astType
+    let expectedReturnType = self[f].returnType.astType
```
kyouko-taiga, line 11: If you want this change, we should also modify `LambdaType`, `MethodType` and `SubscriptType` for consistency.

dabrahams:
Stale

PR #474, opened 2023-02-28
Add/use a first class Function.BlockID

dabrahams:


kyouko-taiga (approved):
I'm a little concerned about is `BlockID` being defined as an alias to `Functions.Blocks.Address`. So there's only a very subtle name difference between `Block.ID` and `BlockID` which might create some confusion. It's also confusing that `BlockID` doesn't denote a tiered key path while `InstructionID` does. What about `TieredBlockID` vs `BlockID`?

The question doesn't come for functions because the values of their tiered key paths and the that of their "IDs in container" coincide, but the problem will surface again with instructions.

I'm not entirely convinced about `blockInFunction` when I look at call sites like `result.define(source.address, predecessorOf: s.target.blockInFunction)`. But thinking about tiered instruction IDs, which have 3 levels, I agree that this choice brings more consistency across the board.

One thing to consider: Sometimes I wonder if it wouldn't be simpler to define `InstructionID` as a pair whose first element is a `Block.ID`(the tiered identifier) and second element is the instruction's ID in its block (i.e., it's "address" in the doubly-linked list). Perhaps that approach solves the naming issue:
```swift
// Or perhaps `x: TieredInstructionID`
func foo(x: InstructionID) {
  print(x.block.function)
  print(x.block.idInFunction)
  print(x.idInBlock)
}
```
We could even have `idInContainer` but that's less specific.

dabrahams:
> I'm a little concerned about is BlockID being defined as an alias to Functions.Blocks.Address.

It's not just `BlockID`, but `Function.BlockID`.  I started with `Block.IDInFunction` but I didn't like the coupling; I can change it though.

> What about TieredBlockID vs BlockID?
We discussed `GlobalBlockID` yesterday.  I'd prefer that to `Tiered` because it is semantic rather than structural.

Your idea about having a cons-list form for these identifiers is kind of neat.  I'll have a think about that.

dabrahams:
Closing due to staleness

PR #475, opened 2023-02-28
Reduce egregious use of reduce

dabrahams:


Sources/Core/SourceFile.swift
```diff
@@ -280,12 +280,8 @@ extension SourceFile: CustomStringConvertible {
 /// directories are recursively searched for `.val` files, which are considered Val source files;
 /// all others are treated as non-source files and are ignored.
 public func sourceFiles<S: Sequence<URL>>(in sourcePaths: S) throws -> [SourceFile] {
-  try sourcePaths.reduce(into: []) { (result, f) in
-    if !f.hasDirectoryPath {
-      try result.append(SourceFile(contentsOf: f))
-    } else {
-      try result.append(contentsOf: sourceFiles(in: f, withExtension: "val"))
-    }
+  try sourcePaths.flatMap { p in
```
kyouko-taiga, line 283:
```suggestion
  try sourcePaths.flatMap { (p) in
```

Sources/IR/AbstractTypeLayout.swift
```diff
@@ -64,18 +64,14 @@ extension TypedProgram {
   fileprivate func properties(of t: AnyType) -> [AbstractTypeLayout.StoredProperty] {
     switch t.base {
     case let p as ProductType:
-      return self[p.decl].members.reduce(into: []) { (result, m) in
-        if let b = BindingDecl.Typed(m) {
-          result.append(
-            contentsOf: b.pattern.names
-              .map({ (_, name) in (name.decl.baseName, name.decl.type) }))
-        }
+      return self[p.decl].members.flatMap { m in
```
kyouko-taiga, line 67:
```suggestion
      return self[p.decl].members.flatMap { (m) in
```

kyouko-taiga (approved):
define: egregious
> extremely bad in a way that is very noticeable

That terrible?

dabrahams:
> That terrible?

Sorry, very poor choice of word; I actually thought it just meant overdone and needless.

Sources/Core/SourceFile.swift
dabrahams: I'll apply these changes because I don't care that much, but I'm generally not in favor of doing things like this.  It feels like the product of a very personal preference that has no good technical reason.  It does not make the Swift code easier to read for anyone who's been programming in Swift for a little while; in general it just adds ceremony and development friction.  People will continue to forget, and Swift programmers not steeped in the project will wonder why we're inserting these extra characters.

kyouko-taiga: > It feels like the product of a very personal preference that has no good technical reason. 

Teaching has strongly convinced me that Swift's syntactic mess surrounding closures hurts readability and learnability. I conceded on trailing closure syntax only to make the use of swift-format more convenient. It doesn't mean I no longer care. At least I think we can strive for consistency. I initially skipped this suggestion and came back to it when I saw that you added (or maybe left) parentheses in all other changes of this PR.

> People will continue to forget, and Swift programmers not steeped in the project will wonder why we're inserting these extra characters.

I know, but I'm ready to make my case as many times as necessary.

dabrahams: IMO it's important to separate your position on how Swift *should have been* designed (some of which I agree with), from the question of what's the most effective way to use Swift *as designed*.  I'm not sure we're doing that.

PR #486, opened 2023-03-06
Factor out the infrastructure shared by abstract interpreters

kyouko-taiga:
Two phases of the IR analysis/transformation pipeline are implemented as abstract IR interpreters, which share a significant part of their logic. In an effort to avoid code duplication, I tried to factor out these common parts.

The main contribution of this PR is `AbstractInterpreter`. This type represents the state of an abstract interpreter and implements the algorithm for selecting the basic blocks to process and forming the abstract evaluation contexts in which their IR is interpreted. It is used as follows: https://github.com/val-lang/val/blob/36b01ba1a0d669b25bb2d2a7e4310199e0b6d7f0/Sources/IR/Analysis/Module%2BOwnership.swift#L12-L36

The closure passed to `AbstractInterpreter.fixedPoint` will be called for each block to process until a fixed point is reached.

### DIFF
--- a/Sources/IR/Analysis/Module+NormalizeObjectStates.swift
+++ b/Sources/IR/Analysis/Module+NormalizeObjectStates.swift
@@ -9,118 +9,11 @@ extension Module {
   ///
   /// - Requires: `f` is in `self`.
   public mutating func normalizeObjectStates(in f: Function.ID, diagnostics: inout DiagnosticSet) {
+    var machine = AbstractInterpreter(analyzing: f, in: self, entryContext: entryContext(of: f))
 
-    /// The control flow graph of the function to analyze.
-    var cfg = self[f].cfg()
-
-    /// The dominator tree of the function to analyze.
-    var dominatorTree = DominatorTree(function: f, cfg: cfg, in: self)
-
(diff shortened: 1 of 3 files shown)


Sources/IR/Analysis/Module+NormalizeObjectStates.swift
```diff
@@ -397,20 +293,19 @@ extension Module {
...
+        machine.removeWork(s.targetIfTrue.address)
 
       default:
         fatalError("not implemented")
       }
 
       // Recompute the control flow graph and dominator tree.
-      cfg = functions[f]!.cfg()
-      dominatorTree = .init(function: f, cfg: cfg, in: self)
+      machine.recomputeControlFlow(self)
```
kyouko-taiga, line 308: I'm not entirely satisfied about the need for this particular call.

Because we removed a basic block (see a few lines above), we need to tell the machine to update its CFG and dominator tree. However, I worry that the need to call a method to do so might be error prone. Further, it compels us to pass the machine as an `inout` parameter all the way down to this function.

I haven't managed to come up with a better idea yet.

PR #496, opened 2023-03-08
Add support for end-to-end tests.

lucteo:
We compile and execute all test .val files. We expect that all the compiled programs to execute successfully.

### DIFF
--- a/Tests/EndToEndTests/ExecutionTests.swift
+++ b/Tests/EndToEndTests/ExecutionTests.swift
@@ -1,5 +1,5 @@
-import Core
 import ArgumentParser
+import Core
 import ValCommand
 import XCTest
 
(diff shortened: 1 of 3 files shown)


Tests/EndToEndTests/ExecutionTests.swift
```diff
@@ -0,0 +1,68 @@
...
+    temporaryDirectory.appendingPathComponent("\(UUID())")
+  }
+
+}
+
+extension String: Log {
+
+  public var hasANSIColorSupport: Bool { false }
+
+}
```
lucteo, line 68: @kyouko-taiga: I don't quite like the fact that these last 2 things are duplicated between test projects.
If that's ok with you, I would prefer you deal with these when you have the tests better organized.

kyouko-taiga: Okay, will do.

kyouko-taiga (approved):
LGTM

PR #498, opened 2023-03-09
Arrange C++ code writer code

lucteo:
Drop the `CXXCodeWriter` class. Add better abstractions to write the code. Simplify code writing.


Sources/CodeGen/CXX/CXXModule+Extensions.swift
```diff
@@ -0,0 +1,748 @@
+import ValModule
+
+extension CXXModule {
+
+  /// Returns the C++ code for `self`.
+  public func code(withFormatter formatter: CodeTransform? = nil)
```
kyouko-taiga, line 6:
```suggestion
  public func code(formatter: CodeTransform? = nil)
```

kyouko-taiga (approved):
Great change!

PR #501, opened 2023-03-10
MVS meets FP

kyouko-taiga:


Sources/IR/Emitter.swift
```diff
@@ -220,7 +220,7 @@ public struct Emitter {
     into module: inout Module
   ) {
     precondition(program.isLocal(decl.id))
-    precondition(reading(decl.pattern.introducer.value, { ($0 == .var) || ($0 == .sinklet) }))
+    precondition(read(decl.pattern.introducer.value, { ($0 == .var) || ($0 == .sinklet) }))
```
dabrahams, line 223: I don't understand why the trailing closure syntax is OK above but not here.

Sources/Core/TypeRelations.swift
```diff
@@ -20,8 +20,8 @@ public struct TypeRelations {
     _ c: Conformance,
     testingContainmentWith p: P
   ) -> (inserted: Bool, conformanceAfterInsert: Conformance) {
-    modifying(&conformances[canonical(c.model), default: [:]]) { (traitToConformance) in
-      modifying(&traitToConformance[c.concept, default: []]) { (allConformances) in
+    modify(&conformances[canonical(c.model), default: [:]]) { (traitToConformance) in
```
dabrahams, line 23: Not related to this PR, but I find the logic of this function hard to follow (it's perhaps too nested and functional) and the contract awkward.  I suggest it may be possible to do better by reassessing how the use-sites are formulated.

Sources/FrontEnd/TypeChecking/ConstraintSystem.swift
```diff
@@ -485,21 +485,24 @@ struct ConstraintSystem {
     }
 
     // Generate the list of candidates.
-    let matches = checker.lookup(goal.memberName.stem, memberOf: goal.subject, exposedTo: scope)
-      .compactMap({ checker.decl(in: $0, named: goal.memberName) })
```
dabrahams, line 489: What was wrong with this first use of `compactMap`?  is `decl` a mutating method?
The advantage of this step  for clarity was that it was an understandable unit of functionality, rather than having everything nested in one call.

Sources/FrontEnd/TypeChecking/ConstraintSystem.swift
```diff
@@ -485,21 +485,24 @@ struct ConstraintSystem {
...
-
-      // TODO: Handle bound generic typess
-
-      return OverloadConstraint.Predicate(
-        reference: .member(match),
-        type: matchType,
-        constraints: [],
-        penalties: checker.program.isRequirement(match) ? 1 : 0)
-    })
+    let candidates: [OverloadConstraint.Predicate] = modified([]) { (candidates) in
```
dabrahams, line 488:
```suggestion
    let candidates = modified([] as [OverloadConstraint.Predicate]) { (candidates) in
```

```suggestion
    let candidates = modified([OverloadConstraint.Predicate]()) { (candidates) in
```

Sources/FrontEnd/TypeChecking/ConstraintSystem.swift
```diff
@@ -762,7 +765,7 @@ struct ConstraintSystem {
 
   /// Inserts `batch` into the fresh set.
   private mutating func insert<S: Sequence<Constraint>>(fresh batch: S) -> [GoalIdentity] {
-    batch.map { (g) in insert(fresh: g) }
+    batch.map(mutating: &self, { (s, g) in s.insert(fresh: g) })
```
dabrahams, line 768: The verbosity increase is unfortunate.

Sources/FrontEnd/TypeChecking/ConstraintSystem.swift
dabrahams: Why not use `compactMap(mutating:)` here?

dabrahams (approved):
I have a few thoughts, having seen all of this in context:

- Maybe `mutating: &x` is totally redundant, and should be shortened to `&x`.  I reached for `with: &x` first but that's redundant too.
- I don't like `modified` very much anymore.  What we're doing there, logically speaking, is declaring a var, doing something imperative with it, and using its value.  I was thinking something like one of these would work:

    ```swift
    let something = varValue([]) { x in ... mutate(&x); mutate1(&x); }
    let something = varValue(initially: []) { x in ... mutate(&x); mutate1(&x); }
    let something = finalVarValue([]) { x in ... mutate(&x); mutate1(&x); }
    ```
    
    along with that, the version that returns something other than the value of the variable (do we have that?):
    
    ```swift
    let something = withVar([]) { x in ... mutate(&x); return y; }
    ```
- More globally, I'm a little concerned that the use of these functions reflect a focus on purity at the expense of other important things.  If the available scope of a binding extends over one more statement than needed, it's not the end of the world, and the cost of "fixing" that may outweigh any benefits.  Remember your zen of python (`python3 -c 'import this'`)

Sources/Core/Constraints/ConformanceConstraint.swift
```diff
@@ -23,7 +23,7 @@ public struct ConformanceConstraint: Constraint, Hashable {
   }
 
   public mutating func modifyTypes(_ transform: (AnyType) -> AnyType) {
-    modify(&subject, with: transform)
+    assign(&subject, to: transform)
```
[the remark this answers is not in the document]
dabrahams: In fairness, English around this can be a little confusing.  A variable is assigned a value, but a value is assigned to a variable.

Sources/IR/Emitter.swift
kyouko-taiga: Above where?

The answer is probably that I will always instinctively favor non-trailing closure syntax, not only because of personal preferences but also because Xcode will add parentheses with I use autocompletion. I only ever use trailing closure syntax to keep swift-format from inserting needless newlines.

Sources/Core/TypeRelations.swift
kyouko-taiga: You'll have to be more specific or propose your own PR, because I'm not sure to understand what's unclear about this use site.

Sources/FrontEnd/TypeChecking/ConstraintSystem.swift
kyouko-taiga: I very much prefer declaring types using annotations. I will not impose this convention on others, though.

> Why not use compactMap(mutating:) here?

I came to the conclusion that writing an imperative algorithm here was clearer.

kyouko-taiga:
> Maybe mutating: `&x` is totally redundant, and should be shortened to `&x`.

I disagree because I think the label draws the attention on the fact that it's not your good old map. I'd vote to keep it.

> I don't like `modified` very much anymore.

It has grown on me; also, I like it better than any of your alternatives because it's more descriptive. `varValue` doesn't convey a lot of intent. In contrast, `let value = modified x by-some-lambda` makes more sense.

> More globally, I'm a little concerned that the use of these functions reflect a focus on purity at the expense of other important things.

I don't think I overused it. I kept mutable locals in a few places, when the added verbosity didn't feel worth its cost.

I think I attach more value to "purity" than you. Just like you think it's important to keep members as private as possible, I also think it's important to keep local values as immutable as possible. When I see `var` I immediately wonder where it's being mutated in its scope.

Sources/Core/Constraints/ConformanceConstraint.swift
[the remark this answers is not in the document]
kyouko-taiga: I'm fine with the last one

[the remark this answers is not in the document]
kyouko-taiga: Ah but now this bloody trailing closure syntax is getting in the way because there's another `modify`. Well, unless you really strongly object to `assign`, I vote we keep it.

Sources/FrontEnd/TypeChecking/ConstraintSystem.swift
kyouko-taiga: Yes, `decl` is mutating.

Sources/Core/Constraints/ConformanceConstraint.swift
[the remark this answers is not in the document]
dabrahams: I strongly object, sorry.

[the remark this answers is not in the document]
dabrahams: I don't see how TCS is getting in the way.  Also did you not read my review, 2nd bullet?

Sources/FrontEnd/TypeChecking/ConstraintSystem.swift
dabrahams: > prefer declaring types using annotations

You're not declaring a type.  Are you saying you very much prefer explicitly declared types on all bindings?  I think my rewrite requires less mental energy to understand, and I think I can explain why, but it's just a suggestion.  I wouldn't insist either, but I don't understand what principle is driving your choice.

Sources/IR/Emitter.swift
dabrahams: > where?

Don't worry about it.  I understand your method of choosing as you explained it.

> Xcode will add parentheses with I use autocompletion

Readability should take precedence over conveniences like this for the writer.  I know you will say it's more readable with more nesting (the inner parens don't help either), and it may be true for you, but I don't believe it's true for Swift programmers at large.  Anyway, I'm going to stop questioning the way you do closures now, so it probably doesn't matter.

Sources/FrontEnd/TypeChecking/ConstraintSystem.swift
kyouko-taiga: >  I don't understand what principle is driving your choice.

This is a regular, everyday, normal binding declaration:

```swift
let x: [T]
```

This is a declaration with a hint to disambiguate:

```swift
let x: SourceFile = """
print("こんにちは")
"""
```

To me, this establish precedent that we add annotations on the binding declaration when the compilers need a hint. So I prefer

```swift
let x: [Int] = []
```

over

```swift
let x = [] as [Int]
let y = Array<Int>()
let z = [Int]() // <- This one hurts readability, IMO
```

Sources/Core/Constraints/ConformanceConstraint.swift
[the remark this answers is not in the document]
kyouko-taiga: We already have this `modify`:
```swift
/// Returns the result of calling `action` with a mutable projection of `value`.
public func modify<T, U>(_ value: inout T, _ action: (inout T) throws -> U) rethrows -> U {
  try action(&value)
}
```

I saw your second bullet and I answered. I want to keep `modified`.

Sources/FrontEnd/TypeChecking/ConstraintSystem.swift
dabrahams: Bindings not initialized at declaration result in awkward code and are usually discouraged.  Therefore `let x: [T]` is an outlier.  In the particular case in question, I'm not coercing the result; I'm actually telling the reader what's getting passed _into_ the `modify` call.  I agree with you about the `z` case.  Anyway, do with this what you will; I've made my points as well as I know how.

kyouko-taiga: > Bindings not initialized at declaration result in awkward code and are usually discouraged.

Not true for stored properties of types.

dabrahams:
>  I think the label draws the attention on the fact that it's not your good old map. I'd vote to keep it.

I don't know if that attention is needed.  My inclination would be to try the code without the label and only add it after looking at the code.  And for the purpose of calling attention, any old label would do so at least we could use something shorter, like `with:`.  "mutating" truly conveys nothing that isn't already there at the call site.

dabrahams:
>I don't think I overused it. I kept mutable locals in a few places, when the added verbosity didn't feel worth its cost.

IMO the uses in the new `map` and `compactMap` are clearly not helping.

kyouko-taiga:
> IMO the uses in the new map and compactMap are clearly not helping.

I agree they aren't a clear improvement.

I still think there's a need those FP algorithms in a world where `inout` exists. The conciseness of `batch.map({ insert($0) })` is a product of the reuse of generic algorithms.

Note that if we handle caching differently in the type checker, it's likely that some of these map and compactMap will become "pure".

kyouko-taiga:
FWIW, I don't think an ought argument like "`map` should not be used with a mutating lambda" trumps the benefits in readability of a line like `batch.map({ insert($0) })`

dabrahams:
I agree about the "ought argument".  I think it begins to apply when the body gets more complicated.  I don't know what the guideline should be.

Sources/Core/Constraints/ConformanceConstraint.swift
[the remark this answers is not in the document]
kyouko-taiga: I propose `update`.

dabrahams:
> I think I attach more value to "purity" than you. 

But then you go on to make an argument that sounds like you're saying I am just as concerned with "purity" as you are?

> Just like you think it's important to keep members as private as possible, I also think it's important to keep local values as immutable as possible. 

These are not at all equivalent because maximizing privacy of members by changing access control level has no cost to readability, and because the technical debt introduced by non-local coupling is high. These are practical, not purity, concerns.  Also, I would never introduce nesting (a readability cost) just to further limit access to something that could be `fileprivate` (e.g. so that nothing could access it except exactly the components that need it).  In other words, it's not “as private as possible;” it's “as private as practical.”  

If you take a scope that should already be small enough to reason about locally (e.g. a function) and carve it up into regions where some variables are explicitly mutable or accessible, the annotation cost of that carving (extra function calls, braces, etc.) can work against readability.  And if these annotations will begin to look like they justify themselves because *when they are in place*, the scope is just complex enough to make the strictness feel necessary.  

If it's actually hard to reason about what's happening with your local bindings, it is probably worth breaking the scope of mutation or lifetime into a separate, named function.  The readability difference between

```swift
let x = thingWithNamedSemantics(y, z)
```

and

```swift
let x = modified(someValue) { state in
  // open-coded logic
}
```

Is significant in favor of the former.

Sources/Core/Constraints/ConformanceConstraint.swift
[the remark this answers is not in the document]
dabrahams: That proposal is fine in principle but leaves some details unresolved. What do the use-sites look like?  That's the ultimate test.

kyouko-taiga:
> But then you go on to make an argument that sounds like you're saying I am just as concerned with "purity" as you are?

I'm saying you're concerned about a different kind of purity.

> [...] changing access control level has no cost to readability

That's not entirely true. There's a cost in building new forwarding abstractions to keep members private. That cost might be justified but it's there.

> If it's actually hard to reason about what's happening with your local bindings, it is probably worth breaking the scope of mutation or lifetime into a separate, named function.

The problem is when you need a lot of information from your current context. There's a high cost to pass arguments around in verbosity; and we loose the benefits of inference. So, sometimes

```swift
let x = thingWithNamedSemantics(y, z)
```

can turn into:

```swift
let y = thingWithNamedSemantics(y, z, temporary1, &temporary2, &temporary3)
...
func thingWithNamedSemantics(_ z: T, _ z: U, t0: inout Super<Duper, Long<Name>>, ...) { ... }
```

**Obviously**, it's better to create named abstractions in general. In some cases, though, the best approach might not be so obvious. I have example of cases where it seems nearly impossible to break the algorithm, at least without significant loss of efficiency.

Anyway, I'm not sure we will fully agree on this topic. I think I understood your points and presented mines to the best of my ability. Regarding this PR specifically, I think we should salvage the relevant name changes and err on the side of the status quo for everything else.

dabrahams:
> I'm saying you're concerned about a different kind of purity.

I don't understand; I don't see it as a purity argument at all.

>> [...] changing access control level has no cost to readability
>
> That's not entirely true. There's a cost in building new forwarding abstractions to keep members private. 

It is.  Either an abstraction is appropriate and the cost of its declaration is more than outweighed by the improvement it creates at use-sites, or it isn't appropriate and you use something like `private(set)` which is practically free.

> The problem is when you need a lot of information from your current context.

**Obviously** cases occasionally come up where practically speaking, it's better to use something like `modified`.  My point is that I'm concerned about these things being reached for solely on the basis that they restrict the scope of mutability, but that restriction should be outweighed by other costs.

Issue #436, opened 2023-02-19
Vision for organizing tests


kyouko-taiga:
I want to open this issue to share my vision about the future of testing for Val compiler.

First, I want to distinguish between 3 categories of tests:
1. Unit tests of Swift components
2. Integration tests of compiler parts
3. End-to-end tests of compiled/interpreted Val programs.

Unit tests should obviously be written in Swift and organized according to the common practice. For integeration and end-to-end tests, I want to favor writing tests in Val rather than having Val fragments scattered across Swift files. I believe inline Val code should be used mostly for the purpose of unit testing components that require an AST.

Writing separate Val files will help us progressively build a corpus of programs that I believe to be valuable for a handful of reasons: 
1. to provide a collection of Val program examples that people can find easily.
2. to provide an easy way for people to contribute new tests without having to know anything about Swift.
3. to facilitate looking for recurring patterns in Val code with classical grep-like approaches.
4. to facilitate reusing test cases.

For the sake of simplicity, the proposed architecture currently assumes a program to be a single Val file. However, we should keep in mind that we'll eventually have to write tests for programs consisting of several files or even modules. That being said, I believe having a great setup for handling individual files is a clear step toward one that can handle multiple modules.

I want to avoid (without being completely opposed to) "hacking" features into Val for the purpose of testing it from Swift. Testing is an important part the developer's experience so we should be sure to develop a satisfying story in Val, and one that stands on its own. I believe having to eat our own dog food will help us making informed decisions.

Although building such a test framework in Val comes is quite ambitious, I believe we can do it progressively. We would start by simply running programs and check their exit status, quickly implement assertions to support more elaborate tests, and then slowly iterate on the design of a library.

Now let's have a look at the proposed arborescence of `Tests/`:

```
Tests/
  |- CompilerTests/
  |  |- CompilerTests.swift
  |  |- TypeChecking.testsuite/
  |  |- CXXCodeGen.testsuite/
  |  |- LLVMCodeGen.testsuite/
  |  |- Parsing.testsuite/
  |  |  |- UnterminatedComment.val
  |  |  |- ...
  |
  |- UnitTests/
  |  |- Core/
  |  |  |- SourceFileTests.swift
  |  |  |- ...
  |  |
  |  |- Utils/
  |     |- ...
  |
  |- Val
  |  |- ValTests.swift
  |  |- Foo.testsuite/
  |  |- Bar.testsuite/
  |  |- Library/
  |  |  |- Core.testsuite/
[...]

PR #508, opened 2023-03-11
Change the way to organizing tests

C-BJ:
Fixes: #436
I cannot guarantee that I fully understand the meaning of #436.
 If there is anything I have misunderstood, please mention it.
@kyouko-taiga


kyouko-taiga:
Sorry for the slow response time. I'm attending a conference this week so I'll take a closer look at your PR over the weekend.

C-BJ:
In case you didn't see it, please check out: https://github.com/val-lang/vscode-val/pull/1 @kyouko-taiga

kyouko-taiga:
The PR gets us closer to what I'd like to implement, but we're not completely there yet. There are some changes we need to make on the tests themselves, not simply move files around. Most significantly, I'd like `CompilerTests/` to contain all (and only) integration tests as Val files that would be processed by a single runner written in Swift.

Do you mind if we put this PR on hold for a little? I think it will be easier to merge it once the LLVM backend lands. Perhaps I could also take it from there, since I have a clear idea of the whole picture.

C-BJ:
ok

Issue #507, opened 2023-03-11
Improve testing inside Val


lucteo:
With the introduction of end-to-end tests, we should be able to use Val programs to create tests.
We need to make sure we can report non-zero exit codes, and we can implement asserts in Val.

lucteo:
Fixed with #509

PR #509, opened 2023-03-11
Improvements on the factorial end-to-end test

lucteo:
See #507 

* Allow tests to return Int32 value
* Allow the use of basic asserts in the tests


Tests/EndToEndTests/TestCases/Factorial.val
```diff
@@ -2,7 +2,14 @@ fun factorial(_ n: Int) -> Int {
   if n < 2 { 1 } else { n * factorial(n - 1) }
 }
 
+fun assert(_ b: Bool) {
+  // TODO: if !b {
```
lucteo, line 6: @kyouko-taiga : for some reason, Bool negation doesn't work. I'll leave it like that for now, and we can go back to it later.

lucteo:
oh... I merged another PR in this one ... this makes it harder for review ... sorry for that

kyouko-taiga (approved):
LGTM.

I feel like the C++ and LLVM backend have entered into a race ;)

PR #511, opened 2023-03-11
Handle precedence and associativity in the generated C++ code

### LINKED ISSUE #308: Handle precedence and associativity for expressions

lucteo:
Make sure we use parenthesis whenever it makes sense to implement precedence and associativity rules for C++.

Fixes #308


lucteo:
Bad base of pr

PR #512, opened 2023-03-11
Handle precedence and associativity in the generated C++ code

### LINKED ISSUE #308: Handle precedence and associativity for expressions

lucteo:
Make sure we use parenthesis whenever it makes sense to implement precedence and associativity rules for C++.

Fixes #308


Sources/CodeGen/CXX/CXXModule+Extensions.swift
```diff
@@ -412,14 +412,22 @@ extension CXXContinueStmt: Writeable {
 
 private struct AnyExpr: Writeable {
 
+  /// The C++ expression to write to the output C++ stream.
   let source: CXXExpr
 
-  init(_ source: CXXExpr) {
+  /// True if we need to write the expression in parentheses.
```
kyouko-taiga, line 418:
```suggestion
  /// `true` iff we need to write the expression in parentheses.
```

Minor: only for consistency with other docs

kyouko-taiga (approved):
Nice!

PR #519, opened 2023-03-13
Translate initializers to C++

lucteo:
Implements #317


Sources/CodeGen/CXX/CXXModule+Extensions.swift
```diff
@@ -35,6 +35,7 @@ private struct HeaderFile: Writeable {
       output << "#include <variant>\n"
       output << "#include <cstdint>\n"
       output << "#include <cstdlib>\n"
+      output << AdditionalFileContent("NativeCodePreamble.h")
```
kyouko-taiga, line 38: Is `AdditionalFileContent` a function? Why does it start with a capital letter?

Tests/ValTests/TestCases/CXXCoreLibrary/GeneratedConstructors.val
```diff
@@ -0,0 +1,14 @@
+//! h Int() { value = native_zeroinitializer_int64_t(); }
+
+//! h Bool() { value = native_zeroinitializer_bool(); }
+
+//! h Double() { value = native_zeroinitializer_double(); }
+
+//! h Float() { value = native_zeroinitializer_float(); }
```
kyouko-taiga, line 7: Just a comment: seems like it might be difficult to avoid unnecessary copies in C++ when we have declarations like that:
```swift
type Vector2 {
  var x: Double; var y: Double
}

type Circle {
  var center: Vector2
  var radius: Double
  init(centered_at center: sink Vector2) {
    self.radius = 1.0
    self.center = center
  }
}
```
Note that in Val, `Circle`'s initializer doesn't create a "zero" value for the stored properties. So `self.center = center` is initialization, not assignment.

lucteo: True.
While we can transforom some of the assignments in initialization, I would prefer to handle these "optimizations" later. For most cases, this is should not be a concern.

kyouko-taiga (approved):
LGTM

Issue #517, opened 2023-03-12
Windows EndToEndTests faild


C-BJ:
Github Action:
https://github.com/C-BJ/val/actions/runs/4397242691/jobs/7700233917

C-BJ:
![image](https://user-imag@lucteo .githubusercontent.com/62678643/224703897-97666305-ae0b-481c-8f02-e7c4e20e4b05.png)
@lucteo

kyouko-taiga:
What am I looking at?

compnerd:
@kyouko-taiga - that appears to be a local reproduction of this elusive crasher.  It does look like a `fatalError`, which should trigger a minidump, which might contain clues as to what is happening.

C-BJ:
full output:
```
> Test run started at 2023/3/14 20:48:40 <

[0/1] Planning build
Test Suite 'Selected tests' started at 2023-03-14 12:48:50.339
Test Suite 'ExecutionTests' started at 2023-03-14 12:48:50.386
Test Case 'ExecutionTests.testExecution' started at 2023-03-14 12:48:50.386
<EXPR>:0: error: ExecutionTests.testExecution : threw error "Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory""
Test Case 'ExecutionTests.testExecution' failed (0.062 seconds)
Test Suite 'ExecutionTests' failed at 2023-03-14 12:48:50.448
         Executed 1 test, with 1 failure (1 unexpected) in 0.062 (0.062) seconds
Test Suite 'Selected tests' failed at 2023-03-14 12:48:50.448
         Executed 1 test, with 1 failure (1 unexpected) in 0.062 (0.062) seconds
```

PR #520, opened 2023-03-14
Update ExecutionTests.swift

C-BJ:
Fixes: #517

### DIFF
--- a/Tests/EndToEndTests/ExecutionTests.swift
+++ b/Tests/EndToEndTests/ExecutionTests.swift
@@ -5,10 +5,15 @@ import XCTest
 
 final class ExecutionTests: XCTestCase {
 
-  /// Compiles and executes all tests in `TestCases` directory, and ensures they return success.
+  /// End to end tests
   func testExecution() throws {
-    let s = Bundle.module.url(forResource: "TestCases", withExtension: nil)!
-    for testFile in try! sourceFiles(in: [s]) {
+    try testVal("./TestCases/Factorial.val")
+  }
+  
+  /// Compiles and executes tests, and ensures they return success.
+  func testVal(_ fileURL: String) throws {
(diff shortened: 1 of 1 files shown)


C-BJ:
@kyouko-taiga

Tests/EndToEndTests/ExecutionTests.swift
```diff
@@ -5,10 +5,15 @@ import XCTest
 
 final class ExecutionTests: XCTestCase {
 
-  /// Compiles and executes all tests in `TestCases` directory, and ensures they return success.
+  /// End to end tests
   func testExecution() throws {
-    let s = Bundle.module.url(forResource: "TestCases", withExtension: nil)!
-    for testFile in try! sourceFiles(in: [s]) {
+    try testVal("./TestCases/Factorial.val")
```
kyouko-taiga, line 10: I don't understand what the PR is bringing? Is it just that you want to avoid the trap when `sourceFiles(in:)` throws? If so you can just drop the `try!` from `testExecution`:

```suggestion
    for testFile in try sourceFiles(in: [s]) {
```

kyouko-taiga (changes requested):
See inline comment.

Tests/EndToEndTests/ExecutionTests.swift
C-BJ: The Bundle.module.url function cannot return the correct value in Windows.

kyouko-taiga: Then at the very least this should be behind a `#if os(Windows)`. I feel uneasy about using relative paths and I'm not even sure this will work consistently across all platforms.

Also, `Bundle.module.url` not returning correct values on Windows sounds like a bug that should be reported to Swift.

C-BJ:
See #521

PR #521, opened 2023-03-15
Fix bugs to support windows

C-BJ:
Fixes: #517


C-BJ:
But, I am not sure whether the build or test's retry steps will work properly (because I cannot test). If there is any problem, please feel free to tell me later.

C-BJ:
> > many project, regardless of programming language, have similar problems
> 
> That would seem to imply that Microsoft, who owns both GitHub and Windows, can't make them work together reliably. Am I reading that right, or is there something more that all these projects have in common?

I think CI flake in Windows has nothing to do with Github Action and everything to do with Windows itself.

C-BJ:
CI errors are expected because https://github.com/val-lang/Swifty-LLVM/pull/9 need to be merged first. @kyouko-taiga

kyouko-taiga:
Before we consider moving forward with the retry trick to handle Windows CI failures, have you looked into this answer @C-BJ https://github.com/orgs/community/discussions/26933 ? Could it be related to our problems?

C-BJ:
https://forums.swift.org/t/stability-of-swift-on-windows/63610/30

kyouko-taiga:
I see. Well then let's go ahead with your approach. I'll merge your PR on Swifty-LLVM and this one asap.

I made some changes on `main` that have surely introduced some issues on Windows. Now `valc` requires access to a support library that is compiled separately. On Unix, it gets installed at `/use/local/lib` be default but I don't know what should be done for Windows.

Here's an overview of how things work on Unix:
- The target `Support` (in `Sources/`) contains C/C++ files that implement runtime and support functions.
- We use SPM to compile this target, resulting in object files in `.build/release/Support.build/src`
- We gather all of these files in a library named `liberal_support.a; that's the support library
- We create a SDK which is just a directory with a `lib/` sub-directory containing that library
- We install that SDK at `/usr/local/lib/val`

You can see all steps in `Tools/install.sh`. I guess we should create a similar script for Windows.

This approach is probably not a long-term solution. But I am really not an expert with C/C++ build tools so I just hacked something that works locally and with CI. If you know a better, more portable way to move forward, I'm all ears!

C-BJ:
I'll look into it as soon as I can, but I've been a little busy

kyouko-taiga:
No worries. Let me know if you need any help.

C-BJ:
There is another problem, I now use the Windows LLVM version in https://github.com/c3lang/win-llvm. Do you think we need to separate maintain a repository? Compile LLVM ourself?

kyouko-taiga:
I don't think so. Why not simply use this repository to get LLVM for the Windows CI integration and continue to rely on the approaches currently in place for macOS and Linux?

C-BJ:
sorry wdym?

C-BJ:
I know it doesn't make any difference, it's just a lot of languages do it, I'm just asking

kyouko-taiga:
CI is configured differently for macOS and linux. On macOS, we download and install LLVM using a GitHub action ([install-llvm-action](https://github.com/KyleMayes/install-llvm-action)). On Linux, we use `apt` (see the Dockerfile in `.devcontainer/`).

I guess that on Windows we can refer to this other version that you're using. If I understand correctly, that's what you've done on Swifty-LLVM.

kyouko-taiga:
> I know it doesn't make any difference, it's just a lot of languages do it, I'm just asking

Ah, I see. Perhaps we should continue with the current setup for the time being. There's a value having the CI confirm that everything works as expected if you're downloading and installing LLVM from sources, as it is likely what most user would do locally on their machine.

C-BJ:
Actually what I'm saying is, should we build LLVM ourselves instead of using something that someone else has already built, actually we don't have any problem using someone else's build, so I think we should do that later only when we have additional needs for LLVM builds

kyouko-taiga:
I have to admit that I don't have a strong opinion on the subject. If you think it's better to use someone else's build (like as you said many languages do), then I'm fine with that.

C-BJ:
sorry, I mean other languages mostly build their own. 
also it's a small thing, when we need it, we build it ourselves

C-BJ:
> I see. Well then let's go ahead with your approach. Thanks a lot for investigating! I'll merge your PR on Swifty-LLVM and this one asap.
> 
> I made some changes on `main` that have surely introduced some issues on Windows. Now `valc` requires access to a support library that is compiled separately. On Unix, it gets installed at `/use/local/lib` be default but I don't know what should be done for Windows.
> 
> Here's an overview of how things work on Unix:
> 
> * The target `Support` (in `Sources/`) contains C/C++ files that implement runtime and support functions.
> * We use SPM to compile this target, resulting in object files in `.build/release/Support.build/src`
> * We gather all of these files in a library named `liberal_support.a; that's the support library
> * We create a SDK which is just a directory with a `lib/` sub-directory containing that library
> * We install that SDK at `/usr/local/lib/val`
> 
> You can see all steps in `Tools/install.sh`. I guess we should create a similar script for Windows.
> 
> This approach is probably not a long-term solution. But I am really not an expert with C/C++ build tools so I just hacked something that works locally and with CI. If you know a better, more portable way to move forward, I'm all ears!

I'm pretty much done. I'll finish it then commit it tomorrow or the day after tomorrow. @kyouko-taiga

Tests/EndToEndTests/ExecutionTests.swift
```diff
@@ -6,21 +6,43 @@ import XCTest
...
-      do {
-        let (status, _) = try run(output)
-        XCTAssertEqual(
-          status, 0,
-          "Execution of binary for test \(testFile.baseName) failed with exit code \(status)")
-      } catch {
-        XCTFail("While testing \(testFile.baseName), cannot execute: \(output)")
+  #if os(Windows)
+    func testExecution() throws {
+      try testValOnWindows("./TestCases/Factorial.val")
```
kyouko-taiga, line 11: I remember we had a discussion about this but I don't remember the conclusion. Why can't we use `Bundle.module.url(forResource: "TestCases", withExtension: nil)!` on Windows?

The problem with your approach is that we'll have to edit this file every time we add a new test will a long list of hard-coded filenames, the number of end-to-end tests is expected to grow significantly.

We're using `Bundle.module.url` in several places. That's how collect the test cases of the parser, the type checker, and the IR emitter. Are these tests broken on Windows too?

kyouko-taiga:
Great job @C-BJ, thanks a lot!

I just have one question I asked inline.

Tests/EndToEndTests/ExecutionTests.swift
C-BJ: I think I can look into it a little more

kyouko-taiga: Got it.

C-BJ:
I don't know why, but there seems to be a test on Windows that goes into an infinite loop. @kyouko-taiga

C-BJ:
maybe testWhileStmtWithMultipleConditions

kyouko-taiga:
> maybe testWhileStmtWithMultipleConditions

That'd be surprising. This test is just running the parser, not running any loop. In fact, there are no execution tests that are running loops AFAIK.

C-BJ:
The local test was successful, so I don't know what the problem is

C-BJ:
https://github.com/val-lang/val/actions/runs/4787251828/jobs/8512255958
https://github.com/val-lang/val/actions/runs/4787300582/jobs/8512371120

C-BJ:
I'd like to ask you to stop these three CI
https://github.com/val-lang/val/actions/runs/4787251828/jobs/8512255958
https://github.com/val-lang/val/actions/runs/4787300582/jobs/8512371120
https://github.com/val-lang/val/actions/runs/4787431398/jobs/8512658800

Tests/EndToEndTests/ExecutionTests.swift
C-BJ: image
This error appears to be due to a path slash error.
Is this a Swift error?
@kyouko-taiga

kyouko-taiga: I think so. AFAIK We're not building any path that would contain a reference to `.build/debug` so I don't think this URL has been built by us.

There's one difference between `ExecutingTests` and other Val tests. To gather test case URLs, the former always uses `Bundle.module.url` while the the latter use this method: https://github.com/val-lang/val/blob/3778b59e738aea6c7a14c24cd0ad24ad1c9ee6dd/Tests/ValTests/Utils/AnnotatedValFileTest.swift#L116-L126

I know that the condition (L120) fails on Linux and so we gather URLs `Bundle.module.url`. On macOS, however, we use the first branch (L121). If the same happens on Windows, then I guess those Val tests (parser, type checker, ...) do not use `Bundle.module.url`. That would explain why `ValTests` pass on Windows but `ExecutionTests` does not.

kyouko-taiga:
I guess everything is possible but I'd be surprised. `testCPP` only generate C++ files and then asks the OS if those files exist. If I had to guess I would expect the problem to occur with one of the tests that are running third-party tools, like `ValCommandTests.testBinary`, `ExecutionTests.testExecution`, or `ExecutionTests.testHelloWorld`. Those are calling the linker in a sub-process.

You can try to disable these tests by renaming those methods so that they have any other prefix (i.e., not start with `test`).

Tests/EndToEndTests/ExecutionTests.swift
C-BJ: I haven't used Swift before contributing to Val, so do you have any solutions?

kyouko-taiga: I suspect the problem is caused by a bug in `Foundation` which we probably can't address directly. So my suggestion would be to use something like the `testSuite` method above to collect test URLs on Windows.

Here's a sketch of the solution:

```swift
final class ExecutionTests: XCTestCase {

  /// The URL of the `EndToEndTests/` directory of this project.
  private static let root = URL(fileURLWithPath: String(#filePath))
    .deletingLastPathComponent()

  /// Returns the Val test cases in the suite at `path` relative to root of the test target
  /// containing this file.
  ///
  /// The suite is sought at `path` relative to `EndToEndTests/`. If no such directory exists, the
  /// suite is sought at `path` relative to the root of the resource bundle associated with the
  /// current Swift module.
  private func testSuite(at path: String) throws -> [SourceFile] { 
     let p = URL(fileURLWithPath: path, relativeTo: ExecutionTests.root) 
     if p.hasDirectoryPath { 
       return try sourceFiles(in: [p]) 
     } else { 
       let s = Bundle.module.url(forResource: path, withExtension: nil)! 
       return try sourceFiles(in: [s]) 
     } 
   } 

  /// Compiles and executes all tests in `TestCases` directory, and ensures they return success.
  func testExecution() throws {
    let testCases = try testSuite(at: "TestCases")
    for testFile in testCases {
      let output = try compile(testFile.url, with: ["--emit", "binary"])
      ...
    }
  }

}
```

If my suspicion is correct, code should go through the first branch of the condition in `testSuite(at:)` on Windows. You can make sure of that with an assertion:
```
#if os(Windows)
assertionFailure()
#endif
```
If I am correct, then Windows would no longer use `Bundle.module.url` to gather URLs.

kyouko-taiga: I can implement that idea if you want but I don't have the means to run Windows myself so you'd have to do the actual testing on your end.

C-BJ: ok

C-BJ:
image
@k
@kyouko-taiga

kyouko-taiga:
Maybe `\n` isn't interpreted the same way on Windows?

C-BJ:
```Test Case 'ParserTests.testParse' started at 2023-04-28 12:04:15.741
Swift/ContiguousArrayBuffer.swift:600: Fatal error: Index out of range
Current stack trace:
0    (null)                             0x00007ffcad84e200 swift_stdlib_reportFatalErrorInFile + 134```
This error seems to occur only locally, not in CI

**Comment by C-BJ** [comment]
```
> Test run started at 2023/4/28 20:19:42 <

Test Suite 'Selected tests' started at 2023-04-28 12:19:42.838
Test Suite 'EmitterTests' started at 2023-04-28 12:19:42.846
Test Case 'EmitterTests.testEmitter' started at 2023-04-28 12:19:42.846
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Unused.val:7: error: EmitterTests.testEmitter : missing expected diagnostic: 'binding 'x' was never used'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Unused.val:11: error: EmitterTests.testEmitter : missing expected diagnostic: 'binding 'z0' was never used'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Unused.val:19: error: EmitterTests.testEmitter : missing expected diagnostic: 'binding 'x10' was never used'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Unused.val:19: error: EmitterTests.testEmitter : missing expected diagnostic: 'binding 'x11' was never used'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Unused.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'binding 'x10' was never used'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Unused.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'binding 'x' was never used'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Unused.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'binding 'x11' was never used'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Unused.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'binding 'z0' was never used'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Ownership.val:76: error: EmitterTests.testEmitter : missing expected diagnostic: 'illegal mutable access'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Ownership.val:89: error: EmitterTests.testEmitter : missing expected diagnostic: 'illegal immutable access'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Ownership.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'illegal mutable access'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Ownership.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'illegal immutable access'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/NumericLiteral.val:13: error: EmitterTests.testEmitter : missing expected diagnostic: 'integer literal '128' overflows when stored into 'Int8''
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/NumericLiteral.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'integer literal '128' overflows when stored into 'Int8''
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/LocalBindings.val:7: error: EmitterTests.testEmitter : missing expected diagnostic: 'initialization of inout binding must be marked for mutation'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/LocalBindings.val:9: error: EmitterTests.testEmitter : missing expected diagnostic: 'declaration of let binding requires an initializer'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/LocalBindings.val:10: error: EmitterTests.testEmitter : missing expected diagnostic: 'declaration of inout binding requires an initializer'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/LocalBindings.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'declaration of let binding requires an initializer'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/LocalBindings.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'initialization of inout binding must be marked for mutation'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/LocalBindings.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'declaration of inout binding requires an initializer'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/ImplicitReturn.val:10: error: EmitterTests.testEmitter : missing expected diagnostic: 'missing return in function expected to return 'Int''
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/ImplicitReturn.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'missing return in function expected to return 'Int''
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/DI.val:16: error: EmitterTests.testEmitter : missing expected diagnostic: 'use of uninitialized object'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/DI.val:18: error: EmitterTests.testEmitter : missing expected diagnostic: 'use of uninitialized object'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/DI.val:24: error: EmitterTests.testEmitter : missing expected diagnostic: 'use of consumed object'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/DI.val:25: error: EmitterTests.testEmitter : missing expected diagnostic: 'use of consumed object'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/DI.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'use of consumed object'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/DI.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'use of consumed object'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/DI.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'use of uninitialized object'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/DI.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'use of uninitialized object'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Assignment.val:6: error: EmitterTests.testEmitter : missing expected diagnostic: 'left-hand side of assignment must be marked for mutation'
C:/Users/19454/Desktop/val/Tests/ValTests/TestCases/Lowering/Assignment.val:1: error: EmitterTests.testEmitter : unexpected diagnostic: 'left-hand side of assignment must be marked for mutation'
Test Case 'EmitterTests.testEmitter' failed (5.825 seconds)
Test Suite 'EmitterTests' failed at 2023-04-28 12:19:48.671
         Executed 1 test, with 32 failures (0 unexpected) in 5.825 (5.825) seconds
Test Suite 'Selected tests' failed at 2023-04-28 12:19:48.671
         Executed 1 test, with 32 failures (0 unexpected) in 5.825 (5.825) seconds
```

**Comment by C-BJ** [comment]
emit cpp will enter an infinite loop under windows. There's no output, there's no end

**Comment by kyouko-taiga** [comment]
I don't know exactly what I can do. It's nearly impossible for me to debug anything on Windows, because I don't have a way to run that OS. If you're lost and don't know how to proceed, I think we should put Windows support on hold until we can energy more effort on it.

I should focus my efforts on overall compiler implementation.

**Comment by C-BJ** [comment]
The clang-format code doesn't work properly on windows, so I'll leave it out of the window

**Comment by C-BJ** [comment]
CI passed @kyouko-taiga

**Comment by C-BJ** [comment]
I think this PR is ready for final review of the merge @kyouko-taiga

File: .github/workflows/build-and-test.yml [hunk 2]
```diff
@@ -97,3 +97,79 @@ jobs:
...
+      run: Add-Content $env:GITHUB_PATH 'C:\llvm-15.0.6-windows-x86-msvc17-libcmt\bin'
+
+    - run: llvm-config --version
+
+    - name: Build support library
+      id: build-support-library
+      continue-on-error: true
+      run: swift build -c release --target Support
+
+    - name: Retry again on failure
```
kyouko-taiga, line 149:
```suggestion
    - name: Retry on failure
```

.github/workflows/build-and-test.yml
```diff
@@ -97,3 +97,79 @@ jobs:
...
+      
+    - name: Install support library
+      run: ./Tools/install_CI.bat
+
+    - name: Build (Release)
+      id: build
+      continue-on-error: true
+      run: swift build -v -c release
+
+    - name: Retry again on failure
```
kyouko-taiga, line 162:
```suggestion
    - name: Retry on failure
```

.github/workflows/build-and-test.yml
```diff
@@ -97,3 +97,79 @@ jobs:
...
+      continue-on-error: false
+      if: steps.build.outcome != 'success'
+      run: swift build -v -c release
+
+    - name: Test (Release)
+      id: test
+      continue-on-error: true
+      run: swift test -c release -Xswiftc -enable-testing ${{matrix.host.test-options}}
+
+    - name: Retry again on failure
```
kyouko-taiga, line 172:
```suggestion
    - name: Retry on failure
```

README.md
```diff
@@ -23,10 +21,13 @@ That's Val compiler!
 
 In order to compile Val programs, the compiler needs to locate Val's SDK.
 By default, `valc` will look for the SDK at `/usr/local/lib/val` on Unix system.
-On Wondows, it will look for a directory named `val` in the working directory.
+On Wondows, it will look for the `.val` directory in the user directory.
```
kyouko-taiga, line 24:
```suggestion
On Windows, it will look for a `.val` directory in the user directory.
```

Sources/ValCommand/ValCommand.swift
```diff
@@ -57,7 +57,16 @@ public struct ValCommand: ParsableCommand {
   /// The default location of Val's SDK.
   private static func defaultValSDK() -> URL {
     #if os(Windows)
-      return URL(fileURLWithPath: FileManager.default.currentDirectoryPath, isDirectory: true)
+      let environment = ProcessInfo.processInfo.environment["USERPROFILE"]!
+      //For CI
+      if environment.contains("runneradmin") {
+        return URL(
+          fileURLWithPath:
+            "D:/a/val/val/.build/x86_64-unknown-windows-msvc/release/Support.build/src/.val")
```
kyouko-taiga, line 65:
```suggestion
            "D:/a/val/val/.build/x86_64-unknown-windows-msvc/release/Support.build/src/.val")
```
There's got to be another way than hardcoding paths in the source. Why not set an environment variable when you configure the CI?

Sources/ValCommand/ValCommand.swift
```diff
@@ -57,7 +57,16 @@ public struct ValCommand: ParsableCommand {
...
+      }
+      //For user
+      //Get user directory
```
kyouko-taiga, line 68:
```suggestion
```

Sources/ValCommand/ValCommand.swift
```diff
@@ -190,10 +199,13 @@ public struct ValCommand: ParsableCommand {
     // C++
 
     if outputType == .cpp {
-      let codeFormatter: CodeTransform? = (try? find("clang-format")).map({
-        clangFormatter(URL(fileURLWithPath: $0))
-      })
-
+      #if os(Windows)
+        let codeFormatter: CodeTransform? = nil
```
kyouko-taiga, line 203:
```suggestion
        // Note: clang-format doesn't work on Windows.
        let codeFormatter: CodeTransform? = nil
```

Sources/ValCommand/ValCommand.swift
```diff
@@ -235,6 +249,7 @@ public struct ValCommand: ParsableCommand {
 
   /// Combines the object files located at `objects` into an executable file at `binaryPath`,
   /// logging diagnostics to `log`.
+  /// Make Val code executable on MacOS(C++ backend)
```
kyouko-taiga, line 252:
```suggestion
```
This comment is just repeating what's already in the name of the function.

Sources/ValCommand/ValCommand.swift
```diff
@@ -257,7 +272,8 @@ public struct ValCommand: ParsableCommand {
 
   /// Combines the object files located at `objects` into an executable file at `binaryPath`,
   /// logging diagnostics to `log`.
-  private func makeLinuxExecutable<L: Log>(
+  /// Make Val code executable on Windows and Linux(C++ backend)
```
kyouko-taiga, line 275:
```suggestion
  /// Make Val code executable on Windows and Linux.
```
We're not using C++ here.

You should just add "Windows and Linux" at the end of the existing comment. Please don't add a new line:

```swift
/// Combines the object files located at `objects` into a Linux or Windows executable file at `binaryPath`,
/// logging diagnostics to `log`.
```

Tests/EndToEndTests/ExecutionTests.swift
```diff
@@ -6,21 +6,44 @@ import XCTest
...
-          status, 0,
-          "Execution of binary for test \(testFile.baseName) failed with exit code \(status)")
-      } catch {
-        XCTFail("While testing \(testFile.baseName), cannot execute: \(output)")
+  #if os(Windows)
+    func testExecution() throws {
+      try testValOnWindows("./TestCases/Destroy.val")
+      try testValOnWindows("./TestCases/Existential.val")
+      try testValOnWindows("./TestCases/Factorial.val")
+      try testValOnWindows("./TestCases/Lambda.val")
```
kyouko-taiga, line 14: We need a way to collect these URLs programmatically. Having to update this list manually every time we add new test cases is not tenable. Since since most other contributors don't even Windows, they are likely to forget because they won't even see whether or not something changed when they run tests locally.

kyouko-taiga: If the approach I suggested doesn't work, then I vote we simply run the "Hello, World" test on Windows and skip those ones until we can figure out a solution.

kyouko-taiga (changes requested):
Great job. I left a few comments.

The main issue is the complexity of `ExecutionTests.testExecution()`. We should avoid having so widely different behaviors. If we can't make Windows find URLs, then let's disable this test until we can figure out a solution.

I left other inline comments.

Tests/EndToEndTests/ExecutionTests.swift
C-BJ: Can you try the method you mentioned earlier?

Sources/ValCommand/ValCommand.swift
C-BJ: The reason we need to do these thing in CI is because that Github Action does not seem to allow direct access to the user's directory.

Tests/EndToEndTests/ExecutionTests.swift
C-BJ: I thought, I'll try again

C-BJ:
@kyouko-taiga

Sources/ValCommand/ValCommand.swift
kyouko-taiga: I understand. I'm suggesting that you read an environment variable that gets set by the CI configuration script before the tests are ran. You can hardcode this path in the environment variable.

I just don't want to have hardcoded paths in the source files.

Tests/EndToEndTests/ExecutionTests.swift
kyouko-taiga: I pushed it on the branch. Can you tell me what Windows sees?

I would try to debug the following:
- See what the URL `p` generated by `ExecutionTests.testSuite` looks like (L13). It should normally be the URL of the directory containing the test cases in the target, i.e., `Tests/EndToEndTests/TestCases` relative to the root of the repository.
- If the URL looks fine, what's the value of `p.hasDirectoryPath` (L14)
- If it's true, then what does `sourceFiles(in:)` returns (L15).
- If any of the above doesn't work, then try to understand why it seems work in `ValTests`.

If none of that work, then disable `testExecution` on Windows and add an issue to fix it when you or someone else has time.

C-BJ: I did it.

C-BJ:
I use a very ingenious way to avoid errors, and we are able to not change the source code when we added a new test file. @kyouko-taiga

C-BJ:
I don't know how to create environment variables in CI for Windows

C-BJ:
I don't know how to create environment variables in CI on Windows

Tests/EndToEndTests/ExecutionTests.swift
```diff
@@ -6,21 +6,48 @@ import XCTest
...
-      do {
-        let (status, _) = try run(output)
-        XCTAssertEqual(
-          status, 0,
-          "Execution of binary for test \(testFile.baseName) failed with exit code \(status)")
-      } catch {
-        XCTFail("While testing \(testFile.baseName), cannot execute: \(output)")
+  #if os(Windows)
+    ///Test all the Val code in the TestCases folder
+    func testExecution() throws {
```
kyouko-taiga, line 11: I might be clever but I don't understand what it is doing and why it has to be so intricate.

Please, let's try not to write large sections of code in `#if os(Windows)`. It is hard to read and will be difficult to maintain. What's the issue on Windows? Are other tests working? If so why?

Here's how things should work:
- `Bundle.module.url(forResource:withExtension:)` gives us the URL of the directory containing the test cases
- `sourceFiles(in:)` gives us a sequence with all the `.val` files in this directory.

If any of these methods are not working as expected on Windows, they should be fixed, not worked around. I don't want to merge hacks that just "happen" to work because we got lucky. We should have one single definition of `testExecution` that is nearly identical for all OSes. Other tests seem to be doing fine, so I don't think there's any reason to overengineer conditional code here.

kyouko-taiga: > I don't know how to create environment variables in CI on Windows

Then wait a bit, because I think I'll be able to remove the need for the Val SDK (at least for now).

C-BJ: > I might be clever but I don't understand what it is doing and why it has to be so intricate.
> 
> Please, let's try not to write large sections of code in `#if os(Windows)`. It is hard to read and will be difficult to maintain. What's the issue on Windows? Are other tests working? If so why?
> 
> Here's how things should work:
> 
> * `Bundle.module.url(forResource:withExtension:)` gives us the URL of the directory containing the test cases
> * `sourceFiles(in:)` gives us a sequence with all the `.val` files in this directory.
> 
> If any of these methods are not working as expected on Windows, they should be fixed, not worked around. I don't want to merge hacks that just "happen" to work because we got lucky. We should have one single definition of `testExecution` that is nearly identical for all OSes. Other tests seem to be doing fine, so I don't think there's any reason to overengineer conditional code here.

You can take a look, and the URL produced seem to be fine. I don't know what the problem is

image

kyouko-taiga: If you can't reproduce your local behavior on CI, then I'm not sure we can realistically hope to get CI working for Windows.

I'm personally out of options here, I'm really sorry. I'll ask around to see if someone has an idea, but until we can find the cause of the issue I don't want to merge hacky solutions just to appease CI.

C-BJ: Tests also fail locally. I just want you to see if these outputs are correct?

C-BJ:
I can now make CI work correctly in my way, and do not need to make many modifications in the future, but you seem to have some requirements that I don't really understand.

Tests/EndToEndTests/ExecutionTests.swift
C-BJ: The test can also run on CI now, I'm just showing you the output of the url

C-BJ:
As long as we can guarantee the accuracy of CI, it's not a big deal if there's a little bit more code for windows, or if there's a specific path in the code I think

C-BJ:
We probably won't be adding platform-specific code like this again, so I don't think it's really a problem to handle Windows CI this way until we find a better solution

C-BJ:
If you're worried, I can explain the above complex code to you and make sure it's correct

C-BJ:
If you need anything, just let me know, and I'll do my best to fix it

C-BJ:
I just think that Windows CI is really more important than these little code issues

C-BJ:
@kyouko-taiga

C-BJ:
I'm sorry if these message makes you feel uncomfortable

Tests/EndToEndTests/ExecutionTests.swift
C-BJ: Can you run the same code in the screenshot on your system and let me take a look at the output

C-BJ:
I'm now trying to see if I can do better

C-BJ:
I found the normal solution

C-BJ:
I think I've done the rest of your request for this PR @kyouko-taiga

Tests/EndToEndTests/ExecutionTests.swift
```diff
@@ -9,7 +9,10 @@ final class ExecutionTests: XCTestCase {
   func testExecution() throws {
     let s = Bundle.module.url(forResource: "TestCases", withExtension: nil)!
     for testFile in try! sourceFiles(in: [s]) {
-      let output = try compile(testFile.url, with: ["--emit", "binary"])
+      let output =
+        try compile(
+          URL(fileURLWithPath: testFile.url.path),
```
kyouko-taiga, line 14: This was the fix? Building a fresh URL from the existing path?

That's fine with me, but if that's the case we should add a comment to explain the workaround. As it stands, I would be tempted to refactor this code and get back to the original version.

kyouko-taiga (changes requested):
Good job!

Conflicts are caused by the removal of the support library and C++ transpilation (we decided to put development of the latter on hold).

There's also a change in Swifty-LLVM that's likely to cause your branch to fail now. See https://github.com/val-lang/Swifty-LLVM/issues/15

C-BJ:
removal of the support library and C++ transpilation?
Why?

C-BJ:
https://github.com/val-lang/Swifty-LLVM/pull/16
@kyouko-taiga

kyouko-taiga:
> removal of the support library and C++ transpilation?
Why?

It is slowing development down. As we shifted focus on the LLVM backend to catch-up with our goal of having a 1.0 this year, keeping the C++ transpiler code around just means creating more opportunities for conflicts as we make changes to other components.

C-BJ:
> [val-lang/Swifty-LLVM#16](https://github.com/val-lang/Swifty-LLVM/pull/16) @kyouko-taiga

@kyouko-taiga

C-BJ:
There seems to be a problem with the linking process

README.md
```diff
@@ -21,6 +19,10 @@ swift build -c release
 That command will create an executable named `valc` in `.build/release`.
 That's Val compiler!
 
+A custom location can be specified with the option `--valc`.
+
+For details about the environment configuration, see CI Settings.
+
```
kyouko-taiga, line 25:
```suggestion
```

This is obsolete, right?

Tests/EndToEndTests/ExecutionTests.swift
```diff
@@ -9,7 +9,11 @@ final class ExecutionTests: XCTestCase {
   func testExecution() throws {
     let s = Bundle.module.url(forResource: "TestCases", withExtension: nil)!
     for testFile in try! sourceFiles(in: [s]) {
-      let output = try compile(testFile.url, with: ["--emit", "binary"])
+      //Testing on Windows requires this not the most concise piece of assignment operation code.
```
kyouko-taiga, line 12:
```suggestion
      // Note: testing on Windows requires explicit call to `URL.init(fileURLWithPath:)`
```

kyouko-taiga (changes requested):
Changes Look good to me. I left two small suggestions.

kyouko-taiga:
How does one link object files on windows?
On macOS we use the linker directly, not clang. Maybe you can do the same on Windows.

kyouko-taiga:
https://stackdiary.com/linux-docs/ld/

I suppose there's a utility on Windows to link object files too. Also, [LLVM has a linker](https://lld.llvm.org).

C-BJ:
```
```

C-BJ:
stdlib.h stdio.h Not supported by the Microsoft C++ standard library

C-BJ:
So I'm thinking about using other C++ standard libraries on Windows

C-BJ:
```lld-link: error: undefined symbol: aligned_alloc
>>> referenced by C:/Users/19454/AppData/Local/Temp/Val.o:(Val.77.aligned_alloc#48)

lld-link: error: undefined symbol: fdopen
>>> referenced by C:/Users/19454/AppData/Local/Temp/Val.o:(Val.77.fdopen#62)```
I can confirm that these two functions do not exist in Microsoft's C/C++ standard library
@kyouko-taiga

**Comment by C-BJ** [comment]
We can use other standard libraries to solve this problem, but it's really cumbersome(I didn't try it locally) Where is the code associated with these links? Just modify it a little bit for windows

**Comment by C-BJ** [comment]
My idea is to find a more generic function and than modify LibC.val to run on all three major systems

**Comment by C-BJ** [comment]
I think we need Val to support conditional compilation on different systems to solve this problem

**Comment by C-BJ** [comment]
I will create two `LibC.val` for Unix and Windows

**Comment by C-BJ** [comment]
There is a problem with `_aligned_malloc` (alternative to `aligned_alloc` in windows) and `aligned_alloc`, the order of inputting parameters is opposite. @kyouko-taiga

**Comment by C-BJ** [comment]
As you can see, now that the CI for Windows has passed, the key is to solve the problem of multi-platform conflicts. @kyouko-taiga

**Comment by C-BJ** [comment]
I have found the solution

**Comment by C-BJ** [comment]
I will create two `LibC.val` for Unix and Windows
But how do I modify `Library.swift` and something else?

**Comment by C-BJ** [comment]
I will create two `LibC.val` for Unix and Windows
But how do I modify `Library.swift` and something else?

**Comment by kyouko-taiga** [comment]
It would be nice if we can come up with a principled solution rather than hacks just to make things work.

Conditional compilation will have to be implemented in Val anyways, so that is not unreasonable to push for it. What things that worries me is that if there is a large number of functions that needs to be interfaced differently on Windows, it will be hard to create a nice system interface in Val.

I suggest you open an discussion for that [over there](https://github.com/val-lang/val-lang.github.io/discussions) so we can perhaps get some input from other members of the community.

**Comment by C-BJ** [comment]
It's not a hack, it's just that Microsoft doesn't have the functions we need to use in the standard library, that's all. I'm not sure what else is possible

**Comment by C-BJ** [comment]
@kyouko-taiga

**Comment by C-BJ** [comment]
There isn't much that needs to be interfacing differently on Windows, just a few things that are different with the C standard library (you can imagine migrating C code from Linux to VC using only the standard library with very little modification).

**Comment by kyouko-taiga** [comment]
Not saying your solution is a hack. But I'd like a unified system layer in Val that doesn't depend on the OS you're using. Providing such a layer will require us to abstract over different OSes. Since the layer must be part of Val's core library, it think it wouldn't hurt to think ahead and see how we can design a convenient way to account for these differences.

**Comment by C-BJ** [comment]
It's worth mentioning that the conflict we ran into was about the C standard library, maybe we shouldn't rely on it?

**Comment by C-BJ** [comment]
There are actually other solutions, such as using other C standard libraries on Windows, but it might be more cumbersome

**Comment by kyouko-taiga** [comment]
It is a very large effort to reimplement system calls directly in Val. It may be an interesting project in the long run, but in the short term I think it is more reasonable to reuse the C library.

**Comment by kyouko-taiga** [comment]
Please keep in mind that, while Windows support would be _**awesome**_ to have, I think it is more important to focus on having a working language implementation that is easy to setup in the short run.

Most of our development is being done on Unix-like (specifically macOS) systems at the moment, so it's a reasonable choice to focus on the development experience on these systems in the short term. If we can afford to take shortcuts on Unix platforms, I'm ready to accept some technical debt. In this specific case, that means that I think Libc is out best bet.

I can see several possibilities to accommodate this choice on Windows. We can write C shims for Libc and link them to Val (similar to what the "support library" was doing) or we can get started with conditional compilation, pick a standard library on Windows, and write an adapter layer on Val. The latter option is probably more work, but it will also lead to a simpler setup for Windows user, probably.

I think this PR is not the best place to discuss these choices. That's why I suggest you open a community discussion.

**Comment by C-BJ** [comment]
https://github.com/val-lang/val-lang.github.io/discussions/64

**Comment by dabrahams** [comment]
One minor nit/note: we're already not using the Github CI matrix feature the way we should; the idea is to factor out common structure of build steps, and this one makes yet another copy of the workflow just for Windows.  I hope that if Windows gets sorted out we can fit it into a single set of matrix jobs.

**Comment by kyouko-taiga** [comment]
> One minor nit/note: we're already not using the Github CI matrix feature the way we should; the idea is to factor out common structure of build steps, and this one makes yet another copy of the workflow just for Windows. I hope that if Windows gets sorted out we can fit it into a single set of matrix jobs.

Seems unlikely from what I understood has been going with our attempts to support Windows. Tooling around Swift and LLVM is different on Windows. The LibC problem also suggests further setup differences. IMO, if we really wanted to streamline a setup for all different OSes, we'd be better off using a tool that's made just for that and just call it from the GH workflow. Sadly, I don't have the time to work on that.

**Comment by dabrahams** [comment]
There are ways to support differences; you can have conditionals on steps.  It's just programming.

**Comment by kyouko-taiga** [comment]
It's not about whether or not you can program the workflow to work. It's about the benefits of doing it. Conditionals shine when there are only slight differences in the middle of relatively similar processes. Once you need to put a condition before most steps, the benefits of using a matrix become questionable.

GH workflows are cumbersome to develop and maintain because the fix/run/debug loop is painfully slow. Further, the efforts we put into them do not contribute to the configuration of the developer's environments. I think it will be more beneficial to factor out this work into a tool that we can reuse for CI as well as users and contributors.

**Comment by C-BJ** [comment]
How long do you think it will take us to decide on that problem? @kyouko-taiga

**Comment by kyouko-taiga** [comment]
A lot of moving pieces here. It is not an easy topic and I can't give it my full attention because I'm quite busy with the implementation of the generics. I will ask what other main contributors think about https://github.com/val-lang/val-lang.github.io/discussions/64 and we can plan something before the end of the week.

**Comment by C-BJ** [comment]
Is there any progress? @kyouko-taiga 
I don't think it's really that hard to decide, you know

**Comment by kyouko-taiga** [comment]
I think @dabrahams wanted to chime in.

**Comment by dabrahams** [comment]
Sorry, what is the question, precisely?

**Comment by kyouko-taiga** [comment]
The question is: what should we do in the short term w.r.t. Windows integration ?

IIRC you said you wanted to take the lead on this PR's review.

The latest problem is that Windows doesn't seem to support libc, which is currently a requirement of Val's standard library. There are ways to address this problem but AFAICT all of them require some non-trivial work, notably including the support of conditional compilation in Val. I expended on this topic [here](val-lang/val-lang.github.io#64).

My own two cents: as I am quite busy with other features, I personally don't have a lot of cycles to spend on this integration. Further, I am worried that the aforementioned problem is only one of a never ending list of Windows-specific issues that will keep slowing down development. Given the current state of the project, I would personally recommand to press pause on Windows and come back once we have a 1.0 release.

I would recommend Windows users to use either Docker or WSL. Thanks to Nick's contribution, the former should be really easy to setup.

**Comment by dabrahams** [comment]
> The question is: what should we do in the short term w.r.t. Windows integration ?
>
> IIRC you said you wanted to take the lead on this PR's review.

I did? I wonder why.

> The latest problem is that Windows doesn't seem to support libc, which is currently a requirement of Val's standard library. There are ways to address this problem but AFAICT all of them require some non-trivial work, notably including the support of conditional compilation in Val. I expended on this topic [here](https://github.com/val-lang/val/pull/val-lang/val-lang.github.io#64).

image

And there are no comments from you [here](https://github.com/val-lang/val-lang.github.io/discussions/64), so…?

> My own two cents: as I am quite busy with other features, I personally don't have a lot of cycles to spend on this integration. Further, I am worried that the aforementioned problem is only one of a never ending list of Windows-specific issues that will keep slowing down development. Given the current state of the project, I would personally recommand to press pause on Windows and come back once we have a 1.0 release.
> 
> I would recommend Windows users to use either Docker or WSL. Thanks to Nick's contribution, the former should be really easy to setup.

Seems to me that if we have an unsolved problem with libc, and we depend on it unconditionally, there's nothing to be done right now.  We don't have a working PR we can merge.  So to me it seems the obvious short term move is that Windows is unsupported until someone actually puts in the time to make our code pass the tests on Windows.  Maybe I'm missing something; the answer to the question seems too simple.  What other choice do we have?

**Comment by C-BJ** [comment]
1. I have successfully passed Windows CI   
2. Windows is not without libc, but if you use Microsoft's libc, there will be very few functions that do not meet the C standard.

**Comment by dabrahams** [comment]
The latest CI results don't show you passing anything, so I don't know what to make of your claim in 1.  If you can update this PR so it passes, I'll be happy to review the changes and see if I feel we can incorporate them.

**Comment by C-BJ** [comment]
It pass the Windows CI at https://github.com/val-lang/val/pull/521/commits/a93c55cf210677c700e150e4cb2884c99d4fee99 and my subsequent commits were trying the conditional compilation method. All in all I've recommitted the code that passes Windows CI @dabrahams

**Comment by C-BJ** [comment]
wait

**Comment by C-BJ** [comment]
@dabrahams @kyouko-taiga

**Comment by kyouko-taiga** [comment]
Unless I'm missing something, your PR is only passing on Windows but regresses on the other OSes. I understand you may be blocked by the inability to use conditional compilation in Val, but if that's the case, then my remarks stand and we should press pause on Windows integration.

I sympathize with the frustration that may be caused by my position, but please understand that any feature we merge is a burden we have to carry as the project progresses. It's one thing to get a green checkmark on Windows once. It's another to continuously support this OS as we're completing Val's implementation. It doesn't mean I don't care about Windows, only that it isn't a priority at this moment. Until it becomes one, there's always the option to run the compile through Docker or WSL.

If you are still dead set on supporting Windows natively, then perhaps you could try to work on the implementation of features blocking this PR, notably including conditional compilation.

**Comment by dabrahams** [comment]
It seems to me that the portability issue could be solved pretty simply by reintroducing a C file in Val's core library that exposes the non-underscored symbols (the ones defined on unixes) to the windows build.  It would be conditionally empty using the preprocessor on all other OSes.

**Comment by C-BJ** [comment]
> It seems to me that the portability issue could be solved pretty simply by reintroducing a C file in Val's core library that exposes the non-underscored symbols (the ones defined on unixes) to the windows build. It would be conditionally empty using the preprocessor on all other OSes.

I agree with this approach

**Comment by C-BJ** [comment]
```error: couldn't build D:\a\val\val\.build\x86_64-unknown-windows-msvc\release\ValTests.build\ASTTests.swift.o because of missing inputs: D:\a\val\val\.build\plugins\outputs\val\ValTests\TestGeneratorPlugin\ValFileTests.swift```
Where is the `ValFileTests.swift`?

**Comment by kyouko-taiga** [comment]
Might be a file produced by our test case generator @dabrahams ?

**Comment by dabrahams** [comment]
> Might be a file produced by our test case generator @dabrahams ?

Correct.  It's possible you're running into [this](https://github.com/apple/swift-package-manager/issues/6595) SPM bug.

**Comment by C-BJ** [comment]
But why does Linux run this test properly, and does it have anything to do with this code?
```
          swift test -c ${{ matrix.configuration }} ${{ matrix.host.test-options }} |
            tee testoutput.txt && (
              (grep -q "[.']EndToEndTests[/. ]test_" testoutput.txt && grep -q "[.']ValTests[/. ]test_" testoutput.txt) ||
              (echo "error: generated tests failed to run; see
              https://github.com/apple/swift-package-manager/issues/6595" && false) )
```

**Comment by dabrahams** [comment]
@C-BJ That code exists to ensure that when the bug occurs, it fails loudly and we notice that the tests haven't run, because the only way we've seen the bug manifest before now was to silently skip tests.  If you read through the referenced bug report, you'll see that why it sometimes works and sometimes doesn't is a complete mystery to us.  Changing seemingly-irrelevant things like the name of the directory containing the project seem to affect it.

**Comment by C-BJ** [comment]
what I can do for this? or try something? or disable this test generation for Windows?

**Comment by dabrahams** [comment]
> Why does Linux run this test properly?
> 
> image
> 
> only on xcode?

Theres no Xcode in linux. Read everything in the issue. Nobody understands this, so nobody can answer your question.

**Comment by C-BJ** [comment]
Since we're again running into (suspected) Swift bugs preventing this PR from going forward, I decided to finish the rest of the code for Windows support in this PR, but not Enable Windows CI.

**Comment by kyouko-taiga** [comment]
Sounds quite reasonable. If you managed to make everything work on your machine we might be able to claim unofficial and experimental Windows support on the README page, with a link to a an issue or community discussion to discuss troubleshooting.

**Comment by C-BJ** [comment]
Yes, But I have a question, how do I make LibC.val use functions from my newly created LibC.h instead of functions from the C standard library?

**Comment by C-BJ** [comment]
> 
It seems to me that the portability issue could be solved pretty simply by reintroducing a C file in Val's core library that exposes the non-underscored symbols (the ones defined on unixes) to the windows build. It would be conditionally empty using the preprocessor on all other OSes.

I can't make it very clear, but based on the methodology described above, I think you should be able to understand it.

**Comment by kyouko-taiga** [comment]
I think you need to define a variant of `makeLinuxExecutable` in `ValCommand.swift` that will pass different flags to the linker.

**Comment by C-BJ** [comment]
thanks, I will try it later

**Comment by C-BJ** [comment]
Done

**Comment by C-BJ** [comment]
I decided to keep CI to build Val on Windows, but not test it

File: .github/workflows/build-and-test.yml [hunk 16]
```diff
@@ -115,3 +115,72 @@ jobs:
...
+
+#    - name: Test (Release)
+#      id: test
+#      continue-on-error: true
+#      run: swift test -c release
+
+#    - name: Retry on failure
+#      continue-on-error: false
+#      if: steps.test.outcome != 'success'
+#      run: swift test -c release
```
kyouko-taiga, line 186:
```suggestion
```
Probably better to avoid keeping commented lines. We already have commits with these if you want to look at how things were done in the future.

Examples/fortytwo.val
```diff
@@ -1,3 +1,3 @@
-public fun main() -> Int32 {
+public fun main() -> Int {
```
kyouko-taiga, line 1: Why changing this?

README.md
```diff
@@ -10,7 +10,7 @@ Please visit our [website](https://val-lang.dev) to get more information about t
 This project is written in [Swift](https://swift.org) and distributed in the form of a package, built with [Swift Package Manager](https://swift.org/package-manager/).
 You will need Swift 5.7 or higher to build the compiler from sources.
 
-*Note to Windows users: although this project is **not** Unix-specific, Windows support is not guaranteed due to the instability of continuous integration (see https://github.com/val-lang/val/issues/252).*
+*Note to Windows users: although this project is **not** Unix-specific, and is expected to run properly on Windows, but due to an unknown bug in SPM, we are unable to ensure Val is running correctly on Windows via CI testing. For details about the environment configuration, see CI build setting.*
```
kyouko-taiga, line 13: We should keep the note as it is. We' can't guarantee support until CI works. We can't claim it is "expected" to run properly until we can reliably test that it's true.

Sources/ValCommand/ValCommand.swift
```diff
@@ -282,6 +285,22 @@ public struct ValCommand: ParsableCommand {
     try runCommandLine(find("clang++"), arguments, diagnostics: &diagnostics)
   }
 
+  /// `binaryPath`, logging diagnostics to `log`.
```
kyouko-taiga, line 288: Missing docs?

Sources/ValCommand/ValCommand.swift
```diff
@@ -282,6 +285,22 @@ public struct ValCommand: ParsableCommand {
...
+    at binaryPath: String,
+    linking objects: [URL],
+    diagnostics: inout DiagnosticSet
+  ) throws {
+    var arguments = [
+      "-defaultlib:ValLibC", "-defaultlib:msvcrt", "-out:" + binaryPath,
+    ]
+    arguments.append(contentsOf: objects.map(\.path))
+
+    // Note: We use "clang" rather than "ld" so that to deal with the entry point of the program.
```
kyouko-taiga, line 299: The comment is wrong. You're using `lld-link`.

.github/workflows/build-and-test.yml
C-BJ: OK

Examples/fortytwo.val
C-BJ: Oh, My bad

README.md
C-BJ: I see

Sources/ValCommand/ValCommand.swift
C-BJ: I see

C-BJ:
I find that when the return value of main is 42 of type Int32, my command line will give a warning of abnormal termination of the program. Is this normal?

Examples/fortytwo.val
C-BJ: I wanted to change the type to test it, but I forgot to change it back

kyouko-taiga:
> I find that when the return value of main is number(except zero) of type Int32, my command line will give a warning of abnormal termination of the program. Is this normal?

Yes, normal return status is 0. Here we want to return another value to confirm that the program is doing something.

C-BJ:
but if the type of the return value is Int, my command line will not give a warning of abnormal termination of the program.

C-BJ:
btw, What else does this PR need to do? Can it be merged now?

kyouko-taiga:
> but if the type of the return value is Int, my command line will not give a warning of abnormal termination of the program.

That's a bug of the compiler.

README.md
```diff
@@ -10,7 +10,7 @@ Please visit our [website](https://val-lang.dev) to get more information about t
 This project is written in [Swift](https://swift.org) and distributed in the form of a package, built with [Swift Package Manager](https://swift.org/package-manager/).
 You will need Swift 5.7 or higher to build the compiler from sources.
 
-*Note to Windows users: although this project is **not** Unix-specific, Windows support is not guaranteed due to the instability of continuous integration (see https://github.com/val-lang/val/issues/252).*
+*Note to Windows users: although this project is **not** Unix-specific, Windows support is not guaranteed due to the unexpected continuous integration failure (see https://github.com/val-lang/val/issues/805).*
```
kyouko-taiga, line 13: Please don't change the message.

If you want to add links to your issue, refer to it in #252. Otherwise it will be harder to retrieve valuable information we already discussed.

kyouko-taiga (changes requested):
Revert your change in the README and I think we should be fine. I'd like another pair of eyes on the PR before we merge it though.

README.md
C-BJ: But the problem we discussed in #252 has been solved

C-BJ: The issue with windows CI's flaky was resolved by retry

kyouko-taiga: Not really. We're just trying a second time hoping the tests pass and assume two failures in a row are noteworthy.  CI is still flaky.

C-BJ: The new issue is that Swift tests don't run properly on Windows

C-BJ: OK

kyouko-taiga: Then link both issues. But please keep the rest of the message as is.

C-BJ:
I suggest you change the title of #252. Because it can't even pass the test right now

C-BJ:
Is there any progress?

Tests/EndToEndTests/ExecutionTests.swift
```diff
@@ -14,7 +14,11 @@ final class ExecutionTests: XCTestCase {
     let output = try compile(f, with: ["--emit", "binary", "-o", "hello"])
     let result = try run(output)
     XCTAssertEqual(result.status, 0, "Exit code is \(result.status)")
-    XCTAssertEqual(result.standardOutput, "Hello, World!\n")
+    #if os(Windows)
+      XCTAssertEqual(result.standardOutput, "Hello, World!\r\n")
+    #else
+      XCTAssertEqual(result.standardOutput, "Hello, World!\n")
+    #endif
```
dabrahams, line 21:
```suggestion
    XCTAssert(result.standardOutput.last?.isNewline ?? false) // Allow for windows newline.
    XCTAssertEqual(result.standardOutput.dropLast(), "Hello, World!")
```

Sources/ValCommand/ValCommand.swift
```diff
@@ -282,6 +286,21 @@ public struct ValCommand: ParsableCommand {
...
+    at binaryPath: String,
+    linking objects: [URL],
+    diagnostics: inout DiagnosticSet
+  ) throws {
+    var arguments = [
+      "-defaultlib:ValLibC", "-defaultlib:msvcrt", "-out:" + binaryPath,
+    ]
+    arguments.append(contentsOf: objects.map(\.path))
+
+    try runCommandLine(find("lld-link"), arguments, diagnostics: &diagnostics)
```
dabrahams, line 301:
```suggestion
    try runCommandLine(
      find("lld-link"), 
      ["-defaultlib:ValLibC", "-defaultlib:msvcrt", "-out:" + binaryPath] + objects.map(\.path),
      diagnostics: &diagnostics)
```

dabrahams (changes requested):
Mostly this looks pretty awesome!  Please respond to my comments and I'm very inclined to merge this.

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -67,6 +66,8 @@ struct GenerateValFileTests: ParsableCommand {
 
 }
 
+GenerateValFileTests.main()
```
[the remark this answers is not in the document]
C-BJ: https://forums.swift.org/t/main-in-a-single-swift-file/63079

.github/workflows/build-and-test.yml
```diff
@@ -117,3 +117,63 @@ jobs:
...
+      run: xcopy ValLibC.lib c:\program" "files\microsoft" "visual" "studio\2022\enterprise\vc\tools\msvc\14.35.32215\lib\x64\
+
+    - name: Build (Release)
+      id: build
+      continue-on-error: true
+      run: swift build -v -c release
+
+    - name: Retry on failure
+      continue-on-error: false
+      if: steps.build.outcome != 'success'
```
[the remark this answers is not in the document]
C-BJ: I have set the id of the `Build (Release)` step to `build`

.github/workflows/build-and-test.yml
```diff
@@ -117,3 +117,63 @@ jobs:
...
+
+    - run: |
+        git config --global core.autocrlf input
+    - uses: actions/checkout@v3
+
+    - name: Swift version
+      run: swift --version
+
+    - name: Set up LLVM 15.0.6
+      run: |
```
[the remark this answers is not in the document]
C-BJ: OK

C-BJ:
Format check has an unexpected error, but it has nothing to do with this PR

C-BJ:
Seems to be a network issue with the Github Action

C-BJ:
Well, it seems to be all right now

.github/workflows/build-and-test.yml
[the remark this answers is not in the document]
C-BJ: Oh, I think I misunderstood what you meant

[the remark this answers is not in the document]
C-BJ: Yes, If the first of the commands fails, the CI run will be shown as failing

dabrahams:
Thanks for all your work, @C-BJ !

PR #522, opened 2023-03-15
Add performance paper to Docs

camio:


Docs/Performance.org
```diff
@@ -0,0 +1,138 @@
...
+
+This document collects some examples that illustrate performance benefits of a language that
+enforces strict mutable value semantics. These examples were inspired from Fedor G Pikus's /The Art
+of Writing Efficient Programs/.
+
+* Branch free optimization
+
+Consider the following two functions, ~f~ and ~g~.
+
+#+begin_src C++
```
dabrahams, line 13: The text should give some context for what language you're about to show the reader.

dabrahams (approved):
Great start; one note inline

Docs/Performance.org
```diff
@@ -0,0 +1,138 @@
...
+  {
+    f( {1, 2, 3}, flag );
+  }
+#+end_src
+
+Since the compiler cannot, in general, see the implementation of ~g~ it needs to assume that
+something like this is possible.
+
+Again, in a strict mutable value semantic language, this isn't possible because only one piece of
+code has modification access which, by the way, implies global variables are not permitted.
```
kyouko-taiga, line 133: That's not entirely correct. You can have global variables but mutable access must be guaranteed unique. Statically enforced MVS will require immutability but dynamically enforced MVS could rely on some ad-hoc mechanism, like reference counting.

Docs/Performance.org
```diff
@@ -0,0 +1,138 @@
+#+TITLE: Performance Benefits of Mutable Value Semantics
+
+* Abstract
+
+This document collects some examples that illustrate performance benefits of a language that
+enforces strict mutable value semantics. These examples were inspired from Fedor G Pikus's /The Art
```
kyouko-taiga, line 6: I think you should also specify whether MVS is assumed to be enforced purely statically, as it has an impact on the conservative assumptions one has to make.

Despite some runtime overhead, dynamically enforced uniqueness can be leveraged for optimization too, just like sound gradual type systems can be leveraged to avoid indirections.

PR #525, opened 2023-03-16
Remove obsolete windows CI steps and reduce duplication with matrix

nickpdemarco:
This PR removes now-obsolete Windows CI configuration, and simplifies what remains by making heavier use of the GH actions `matrix` feature.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -20,55 +20,43 @@ on:
 
 jobs:
   build:
-    name: Swift 5 on ${{ matrix.host.os }}
+    name: "${{ matrix.host.os }} / ${{ matrix.configuration }} / Swift ${{ matrix.swift.version }}"
     strategy:
       fail-fast: false
       matrix:
         host: [
           { type: macos, os: macos-latest },
-          { type: linux, os: ubuntu-latest, test-options: --enable-code-coverage },
+          { type: linux, os: ubuntu-latest, test-options: "-Xswiftc -enable-testing --enable-code-coverage" },
+        ]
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -20,55 +20,43 @@ on:
...
-        tag: ${{ matrix.swift.windows-tag }}
+    - run: git config --global core.autocrlf input
 
-    - run: |
-        git config --global core.autocrlf input
     - uses: actions/checkout@v3
 
-    - name: Swift version
-      run: swift --version
+    - run: swift --version
```
nickpdemarco, line 51: I realize this is a point of style but I found it terribly redundant to have "Swift Version" written out twice.

.github/workflows/build-and-test.yml
```diff
@@ -20,55 +20,43 @@ on:
...
+      run: swift build -v -c ${{ matrix.configuration }}
+
+    - name: Test ${{ matrix.configuration }}
+      run: swift test -v -c ${{ matrix.configuration }} ${{matrix.host.test-options}}
 
-    - name: Build (Release)
-      run: swift build -v -c release
-    - name: Test (Release)
-      run: swift test -c release -Xswiftc -enable-testing ${{matrix.host.test-options}}
-    - name: Check code coverage (Release)
```
nickpdemarco, line 71: There was no difference I could see between (Release) and (Debug) code coverage tests, so I've consolidated to one.

kyouko-taiga: It seems like this change caused the CI to fail.

PR #527, opened 2023-03-18
Add Zig cc support

C-BJ:


kyouko-taiga:
I'm not sure we should merge this PR. Please note that it's not a goal of Val to support as many C++ compilers as possible. The conditional code required for each option is a significant maintenance burden that I don't want to put on ourselves.

I think a better option would be to accept the name of the compiler the user wants as an argument; then they can choose whatever they want and adapt the command-line invocation accordingly.

```bash
valc --emit cpp main.val # defaults to clang
valc --emit cpp --cc zig --cflags xyz main.val
valc --emit cpp --cc icpc --cflags xyz main.val
...
```

kyouko-taiga:
Also note that I'll soon merge changes that will make LLVM the default backend when `--emit` isn't passed. So compilation via C++ will no longer be driven by `valc`.

C-BJ:
ok, I see

PR #532, opened 2023-04-02
Translate methods to C++

lucteo:
Fixes #318

PR #535, opened 2023-04-04
Add a devcontainer spec & associated CI

nickpdemarco:
Like https://github.com/val-lang/Swifty-LLVM/pull/7, this creates a [devcontainer](https://containers.dev/) specification and runs Linux CI based on it. This improves consistency between the local Linux development experience and what's actually run in CI.


kyouko-taiga (changes requested):
It seems like the tests have disappeared from the workflow. There's only the format check. Am I missing something?

nickpdemarco:
Sorry @kyouko-taiga - sneaky typo broke everything without an error. Should be fixed now.

.github/workflows/build-and-test.yml
```diff
@@ -19,58 +19,64 @@ on:
...
+    name: "Devcontainer: ${{ matrix.host.os }}/${{ matrix.configuration }}"
     strategy:
       fail-fast: false
       matrix:
         host: [
-          { type: macos, os: macos-latest },
-          { type: linux, os: ubuntu-latest, test-options: --enable-code-coverage },
+          { type: linux, os: ubuntu-latest, test-options: "-Xswiftc -enable-testing --enable-code-coverage" }
+        ]
+        configuration: [ "debug", "release" ]
```
kyouko-taiga, line 30: I'm not entirely convinced this is a big improvement because it's forcing the CI to download and install dependencies for the debug and release version jobs. I worry that this may increase the time to run the CI jobs considerably.

kyouko-taiga (approved):
A little concerned about the impact of this change on CI execution times, but overall it's a nice PR.

If you think my concerns are not justified, then go ahead and merge!

nickpdemarco:
I understand your concern, but it doesn't appear to make a significant difference. In fact, it might speed up CI.

This PR checked in **6m42s**

Here's a few PR checks I just polled:

- [#532](https://github.com/val-lang/val/actions/runs/4590602808): **7m48s**
- [#533](https://github.com/val-lang/val/actions/runs/4600804594) **9m29s**
- [#531](https://github.com/val-lang/val/actions/runs/4586752285) **7m50s**

You can look at all of the successful runs [here](https://github.com/val-lang/val/actions?query=is%3Asuccess+) and it seems that this branch is not an outlier. My guess is any delta introduced here is dominated by the time we normally wait to get connected with a GitHub runner. 

That said, if we notice a significant change in the future, this is an easy revert.

.github/workflows/build-and-test.yml
nickpdemarco: I understand your concern - I don't know enough about GH runners to explain this away, but it seems like the checks for this PR are not noticeably faster or slower than our existing CI runs. There might be some clever caching at play; I'm not sure.

PR #539, opened 2023-04-10
Disable devcontainers

nickpdemarco:
This is a check to see if the devcontainer CI configuration is the cause of the CI instability we're seeing.


nickpdemarco:
This turned out to be irrelevant to fixing the CI issues we are facing.

PR #595, opened 2023-05-08
work around / fix dynamic LOE violation

dabrahams:
This LOE violation has only been reproduced by me so far and it's not yet clear whether it's spurious, or a real violation that is only caught dynamically.


kyouko-taiga (approved):
LGTM. Nicely done!

PR #604, opened 2023-05-09
CI improvement bundle

dabrahams:
- have build steps build the tests too
- ensure multi-part build steps fail if any part fails (using the bash shell apparently also does this)
- eliminate a broken/illegal import that was causing some tests to fail
- stop enabling code coverage where we don't use it
- create and use `llvm.pc` without `sudo`.


kyouko-taiga (approved):
LGTM

PR #606, opened 2023-05-10
SPM `--parallel` test / GH action interaction demo

dabrahams:
This PR demonstrates how using `swift test --parallel` in a GH action can fail.  As you can see, the previous commit in the chain passed CI, but this one, which only adds the `--parallel` option, fails on both Linux and MacOS.  I believe there's some nondeterminism involved here.


dabrahams:
Never intended to be merged

PR #618, opened 2023-05-13
Delay the textual rendering of Diagnostics as much as possible

dabrahams:
Will allow us to use the test annotation framework for all our tests, including the end-to-end ones.


Sources/ValCommand/DiagnosticLog.swift
```diff
@@ -0,0 +1,157 @@
+import Core
+
+extension DiagnosticSet {
+
+  /// Write a textual representation of `self` into `output` using the given style.
```
kyouko-taiga, line 5:
```suggestion
  /// Writes a textual representation of `self` into `output` using the given style.
```

Sources/ValCommand/StandardErrorLog.swift
```diff
@@ -1,19 +1,25 @@
...
-  /// Appends `text` to the stream.
-  func write(_ text: String) {
-    FileHandle.standardError.write(Data(text.utf8))
+  /// Appends `data` to `self`.
+  public mutating func write(_ data: Data) {
+    base.write(data)
   }
+}
+
+var standardError = OutputFileHandle(.standardError)
```
kyouko-taiga, line 21: Maybe we should define this as a static variable in `OutputFileHandle` ?

Sources/ValCommand/StandardErrorLog.swift
```diff
@@ -1,19 +1,25 @@
 import Foundation
 
-/// A wrapper type implementing `FileHandle.standardError`'s conformance to `Log`.
-struct StandardErrorLog: Log {
+struct OutputFileHandle: TextOutputStream {
```
kyouko-taiga, line 3: File should be renamed after the type for consistency.

kyouko-taiga (approved):
Good work. I left a couple of inline comments but LGTM.

Sources/ValCommand/StandardErrorLog.swift
dabrahams: 🤷 Makes it more inconvenient to use, if more hygienic.  I'm not attached to this design, and am happy to have you change it whenever the mood strikes.

dabrahams: or if you feel strongly, I can change it now.

dabrahams: I started to, but it made code ugly and hard to read.

kyouko-taiga: Let's keep it that way then. I don't feel strongly about it.

PR #645, opened 2023-05-28
Generate XCTests for each Val file processed.

dabrahams:


.github/workflows/build-and-test.yml
```diff
@@ -43,13 +43,17 @@ jobs:
...
+              (grep -q "[.']EndToEndTests[/. ]test_" testoutput.txt && grep -q "[.']ValTests[/. ]test_" testoutput.txt) ||
+              (echo "error: generated tests failed to run; see
+              https://github.com/apple/swift-package-manager/issues/6595" && false) )
 
     - name: Check code coverage
       uses: mattpolzin/swift-codecov-action@0.7.3
       with:
         SORT_ORDER: +cov
-        MINIMUM_COVERAGE: 85
+        MINIMUM_COVERAGE: 84
```
kyouko-taiga, line 56: Looks like a regression.

Sources/Core/SourceFile.swift
```diff
@@ -28,6 +28,11 @@ public struct SourceFile {
     self.storage = storage
   }
 
+  /// Creates an instance representing the file at `filePath`.
+  public init(atPath filePath: String) throws {
```
kyouko-taiga, line 32:
```suggestion
  public init(at filePath: String) throws {
```

The label might contain needless information. Looking at a use site `SourceFile(at: someBinding)`, it seems obvious to me that `someBinding` is a path.

Plugins/TestGeneratorPlugin/TestGeneratorPlugin.swift
```diff
@@ -0,0 +1,23 @@
...
+/// our build process.
+@main
+struct TestGeneratorPlugin: BuildToolPlugin {
+
+  func createBuildCommands(context: PluginContext, target: Target) throws -> [Command] {
+    guard let target = target as? SourceModuleTarget else { return [] }
+    let inputPaths = target.sourceFiles(withSuffix: "val").map(\.path)
+    let outputPath = context.pluginWorkDirectory.appending("ValFileTests.swift")
+
+    let cmd: Command = .buildCommand(
```
kyouko-taiga, line 13:
```suggestion
    let cmd = Command.buildCommand(
```

I really don't feel strongly about it but I think it may be better to avoid using qualification inference when we're just calling a static member directly.

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
...
+
+  @Option(
+    name: [.customShort("n")],
+    help: ArgumentHelp("Name of generated test case.", valueName: "XCTestCase name"))
+  var testCaseName: String
+
+  @Argument(
+    help: "Paths of annotated val source files to be tested.",
+    transform: URL.init(fileURLWithPath:)
+  )
```
kyouko-taiga, line 23:
```suggestion
    transform: URL.init(fileURLWithPath:))
```

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
+import ArgumentParser
+import Foundation
+
+/// A command-line tool that, given a list of .val files, generates XCTest cases for annotated .val
+/// files as part of our build process.
```
kyouko-taiga, line 5: The docs are a little unclear. Does the tool only generates XCTest cases for **annotated** .val files, ignoring non-annotated ones?

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
...
+  var testCaseName: String
+
+  @Argument(
+    help: "Paths of annotated val source files to be tested.",
+    transform: URL.init(fileURLWithPath:)
+  )
+  var valSourceFiles: [URL]
+
+  /// Returns the Swift source of the test function for the Val at `source`, having `firstLine` as
+  /// its first line.
```
kyouko-taiga, line 27:
```suggestion
  /// Returns the Swift source of the test function for the Val file at `source`, having `firstLine` as
  /// its first line.
```

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
...
+  func swiftFunctionTesting(valAt source: URL, withFirstLine firstLine: Substring) throws -> String
```
kyouko-taiga, line 28:
```suggestion
  func swiftFunctionTesting(_ source: URL, withFirstLine firstLine: Substring) throws -> String
```
Not convinced the label is useful.

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
...
+          contentsOf: Data("\(f.path):1: error: \(e.details)\n".utf8))
+        GenerateValFileTests.exit(withError: ExitCode(-1))
+      }
+    }
+
+    output += "\n}\n"
+
+    try output.write(to: outputURL, atomically: true, encoding: .utf8)
+
+  }
```
kyouko-taiga, line 67:
```suggestion
  }

```

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
...
+      import XCTest
+
+      final class \(testCaseName.asSwiftIdentifier): XCTestCase {
+
+      """
+
+    for f in valSourceFiles {
+      let firstLine = try String(contentsOf: f).split(separator: "\n", maxSplits: 1).first ?? ""
+      do {
+        output += try swiftFunctionTesting(valAt: f, withFirstLine: firstLine)
```
kyouko-taiga, line 55:
```suggestion
        output += try swiftFunctionTesting(valAt: f)
```

`firstLine` is computed here only to be passed to `swiftFunctionTesting`, but the first argument is enough to compute this value in the callee. That would make its API simpler, as the call site would not have to ensure it is satisfying the relationship between `source` and `firstLine`.

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
...
+
+  }
+}
+
+extension String {
+
+  /// Returns a valid Swift identifier made by replacing non-identifier characters in `self` with
+  /// underscores, and prefixing it with "X" if it happens to begin with a number.
+  fileprivate var asSwiftIdentifier: String {
+    let r = String(self.lazy.map { $0.isNumber || $0.isLetter ? $0 : "_" })
```
kyouko-taiga, line 75:
```suggestion
    let r = String(self.map { $0.isNumber || $0.isLetter ? $0 : "_" })
```

`lazy.map` is almost always slower than an eager `map` for this kind of small closures. I ran benchmarks. The fastest approach is `reduce`, but it's slightly harder to read and the gain is too small to justify the extra complexity.

```suggestion
    let r = reduce(into: "", { (s, c) in s.append((c.isNumber || c.isLetter) ? c : "_") })
```

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
...
+    return (r.first?.isNumber ?? true) ? "X" + r : r
+  }
+
+}
+
+extension Substring {
```
kyouko-taiga, line 81: The extensions should be defined in `Utils/String+Extensions.swift` if possible.

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
...
+  /// empty, returning the removed substring.
+  fileprivate mutating func removeUntil(firstSatisfies p: (Character) -> Bool) -> Substring {
+    let firstToKeep = firstIndex(where: p) ?? endIndex
+    defer { self = self[firstToKeep...] }
+    return self[..<firstToKeep]
+  }
+
+}
+
+extension StringProtocol where Self.SubSequence == Substring {
```
kyouko-taiga, line 110: I would probably define this extension in another file to reduce the length of this one. The members are not `fileprivate` so moving this code to another file won't reduce encapsulation.

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
...
+
+  /// Interpreting `self` as the first line of an annotated Val test file, returns the embedded test
+  /// method name and expectation of success.
+  func parsedAsFirstLineOfAnnotatedValFileTest() throws -> (
+    methodName: Substring, expectSuccess: Bool
+  ) {
+    var text = self[...]
+    if !text.removeLeading("//- ") {
```
kyouko-taiga, line 118: A particular reason you decided to use `//-` as a prefix for your annotations, as opposed to `//!` which is used for other annotations (see [TestAnnotations.swift](https://github.com/val-lang/val/blob/bcafafd6af3a80f6e271104c44d055377adf0d8d/Tests/ValTests/Utils/TestAnnotation.swift)).

kyouko-taiga:
Review not completed; will continue later.

Tests/EndToEndTests/ExecutionTests.swift
```diff
@@ -32,35 +17,68 @@ final class ExecutionTests: XCTestCase {
...
 
+}
+
+extension XCTestCase {
+
+  /// Compiles and runs the val file at `valFilePath`, `XCTAssert`ing that diagnostics and exit
+  /// codes match annotated expectations.
+  func compileAndRun(_ valFilePath: String, expectSuccess: Bool)
+    throws
+  {
```
kyouko-taiga, line 28:
```suggestion
  func compileAndRun(_ valFilePath: String, expectSuccess: Bool) throws {
```

kyouko-taiga (commented):
I finished reviewing all the changes. Very nice overall. All my my thoughts are in inline comments.

dabrahams:
Great Feedback, Dimi

.github/workflows/build-and-test.yml
dabrahams: Nope, good catch!

dabrahams:
```suggestion
        MINIMUM_COVERAGE: 85
```

Plugins/TestGeneratorPlugin/TestGeneratorPlugin.swift
dabrahams: Absolutely.

Sources/GenerateValFileTests/GenerateValFileTests.swift
dabrahams: Technically that's overspecified, but OK 😉

Sources/GenerateValFileTests/GenerateValFileTests.swift
dabrahams: I am.  This is a case of weak type information.  We could be testing anything.

Sources/GenerateValFileTests/GenerateValFileTests.swift
dabrahams: Good point; this was left over from when I was collecting all the first lines async before generating the code.

Sources/GenerateValFileTests/GenerateValFileTests.swift
dabrahams: It's a crime that an eager map is faster, but thanks.

Sources/GenerateValFileTests/GenerateValFileTests.swift
```diff
@@ -0,0 +1,153 @@
...
+    return self[..<firstToKeep]
+  }
+
+}
+
+extension StringProtocol where Self.SubSequence == Substring {
+
+  /// Interpreting `self` as the first line of an annotated Val test file, returns the embedded test
+  /// method name and expectation of success.
+  func parsedAsFirstLineOfAnnotatedValFileTest() throws -> (
```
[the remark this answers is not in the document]
dabrahams: How would it be more straightforward? That will add a new nominal type that's just a simple aggregate and complicate the use site a bit.

Sources/GenerateValFileTests/GenerateValFileTests.swift
dabrahams: Because it doesn't have the format of a TestAnnotation, would confuse annotation processing, and can't appear at an arbitrary place in the file.

Tests/EndToEndTests/ExecutionTests.swift
```diff
@@ -32,35 +17,68 @@ final class ExecutionTests: XCTestCase {
...
     let cli = try ValCommand.parse(arguments + ["-o", output.relativePath, input.relativePath])
     let (status, diagnostics) = try cli.execute()
+    if !status.isSuccess {
+      throw diagnostics
+    }
 
-    XCTAssert(status.isSuccess, "Compilation of \(input) failed with exit code \(status.rawValue)")
     XCTAssert(
-      diagnostics.isEmpty,
+      !diagnostics.containsError,
```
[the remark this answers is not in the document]
dabrahams: I think so but will double-check

Sources/GenerateValFileTests/GenerateValFileTests.swift
kyouko-taiga: Feel free to ignore if you think my suggestion is unnecessary.

Sources/GenerateValFileTests/GenerateValFileTests.swift
kyouko-taiga: Works for me then.

Sources/GenerateValFileTests/GenerateValFileTests.swift
kyouko-taiga: I'm fine with the status quo.

Sources/GenerateValFileTests/GenerateValFileTests.swift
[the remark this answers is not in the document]
kyouko-taiga: OK, makes sense.

Sources/GenerateValFileTests/GenerateValFileTests.swift
dabrahams: Thanks

Sources/GenerateValFileTests/GenerateValFileTests.swift
kyouko-taiga: Makes sense. You can ignore my comment.

Tests/EndToEndTests/ExecutionTests.swift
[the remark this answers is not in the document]
dabrahams: Yeah, the semantics were wrong.  If that fires, it means that the status code returned was zero but there were errors in the diagnostics.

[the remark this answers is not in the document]
kyouko-taiga: `diagnostics.isEmpty` means that there can be no diagnostic, including warnings. `!diagnostics.containsError` means that we're allowing warnings.

Sources/GenerateValFileTests/GenerateValFileTests.swift
dabrahams: Actually once they get moved to Utils they need to be generalized to Collection :-)

Sources/GenerateValFileTests/GenerateValFileTests.swift
dabrahams: made `fileprivate`

Tests/EndToEndTests/ExecutionTests.swift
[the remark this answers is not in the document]
dabrahams: Yes, I'm well aware, since I wrote those properties.  Think about it again and if you need me to explain, slack me.

Issue #84, opened 2022-11-13
Synthesize default access modifiers


kyouko-taiga:
Add a default access modifier (likely `private`) and synthesize it on all declaration when `public` is not explicitly supplied.

nickpdemarco:
I'm taking a look at this - can you clarify which declarations should have a synthesized `private`? I'm assuming things like import and namespace declarations don't make sense to label as private.

kyouko-taiga:
The idea is that all declarations that currently have an optional access modifier property, like `FunctionDecl`, `SubscriptDecl`, etc., should have a non-optional property instead, replacing the `nil` cases by a synthesized `private` modifier.

PR #651, opened 2023-06-07
`private` Access Modifier

jlscheerer:
This PR adds support for the `private` access modifier (#84), as discussed with @kyouko-taiga.
It is possible to mark functions, method, binding, etc. declaration explicitly as `private`, e.g.,:
```
private fun my_private_function(_ a: let Int, _ b: let Int) -> Int { ... }
```
Furthermore, any declaration without access modifiers is treated implicitly as private.
We may consider a changing the `SourceRange` of the implicitly "generated" private access modifiers.

### DIFF
--- a/Sources/Core/AST/Decl/BindingDecl.swift
+++ b/Sources/Core/AST/Decl/BindingDecl.swift
@@ -7,7 +7,7 @@ public struct BindingDecl: Decl {
   public let attributes: [SourceRepresentable<Attribute>]
 
   /// The access modifier of the declaration, if any.
-  public let accessModifier: SourceRepresentable<AccessModifier>?
+  public let accessModifier: SourceRepresentable<AccessModifier>
 
   /// The member modifier of the declaration.
   public let memberModifier: SourceRepresentable<MemberModifier>?
@@ -29,14 +29,15 @@ public struct BindingDecl: Decl {
   ) {
     self.site = site
     self.attributes = attributes
-    self.accessModifier = accessModifier
+    // implicitly mark the binding as private
+    self.accessModifier = accessModifier ?? SourceRepresentable(value: .private, range: site)
     self.memberModifier = memberModifier
     self.pattern = pattern
     self.initializer = initializer
   }
 
   /// Returns whether the declaration is public.
-  public var isPublic: Bool { accessModifier?.value == .public }
+  public var isPublic: Bool { accessModifier.value == .public }
 
   /// Returns whether the declaration denotes a static member.
   public var isStatic: Bool { memberModifier?.value == .static }
--- a/Sources/Core/AST/Decl/BindingDecl.swift
+++ b/Sources/Core/AST/Decl/BindingDecl.swift
@@ -21,7 +21,7 @@ public struct BindingDecl: Decl {
   /// Creates an instance with the given properties.
   public init(
     attributes: [SourceRepresentable<Attribute>] = [],
-    accessModifier: SourceRepresentable<AccessModifier>? = nil,
+    accessModifier: SourceRepresentable<AccessModifier>,
     memberModifier: SourceRepresentable<MemberModifier>? = nil,
     pattern: BindingPattern.ID,
     initializer: AnyExprID?,
@@ -29,8 +29,7 @@ public struct BindingDecl: Decl {
   ) {
     self.site = site
     self.attributes = attributes
-    // implicitly mark the binding as private
-    self.accessModifier = accessModifier ?? SourceRepresentable(value: .private, range: site)
+    self.accessModifier = accessModifier
     self.memberModifier = memberModifier
     self.pattern = pattern
     self.initializer = initializer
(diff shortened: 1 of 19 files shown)


Sources/Core/AST/Decl/BindingDecl.swift
```diff
@@ -29,14 +29,15 @@ public struct BindingDecl: Decl {
   ) {
     self.site = site
     self.attributes = attributes
-    self.accessModifier = accessModifier
+    // implicitly mark the binding as private
```
kyouko-taiga, line 32:
```suggestion
    // Declarations are private by default.
```

Sources/Core/AST/Decl/BindingDecl.swift
```diff
@@ -29,14 +29,15 @@ public struct BindingDecl: Decl {
...
+    self.accessModifier = accessModifier ?? SourceRepresentable(value: .private, range: site)
```
kyouko-taiga, line 33: The constructor should take a non-optional property and let the parser synthesize it. That is how other properties are synthesized already.

Sources/Core/AST/Decl/ConformanceDecl.swift
```diff
@@ -28,7 +28,8 @@ public struct ConformanceDecl: TypeExtendingDecl {
     site: SourceRange
   ) {
     self.site = site
-    self.accessModifier = accessModifier
+    // implicitly mark conformance(s) as private
+    self.accessModifier = accessModifier ?? SourceRepresentable(value: .private, range: site)
```
kyouko-taiga, line 32: Same remark about the non-optional constructor argument.

Sources/Core/AST/Decl/ExtensionDecl.swift
```diff
@@ -24,7 +24,8 @@ public struct ExtensionDecl: TypeExtendingDecl {
     site: SourceRange
   ) {
     self.site = site
-    self.accessModifier = accessModifier
+    // implicitly mark the extension as private
+    self.accessModifier = accessModifier ?? SourceRepresentable(value: .private, range: site)
```
kyouko-taiga, line 28: Same remark about the non-optional constructor argument.

Sources/Core/AST/Decl/InitializerDecl.swift
```diff
@@ -53,7 +53,8 @@ public struct InitializerDecl: GenericDecl, GenericScope {
     self.site = site
     self.introducer = introducer
     self.attributes = attributes
-    self.accessModifier = accessModifier
+    // implicitly mark the initializer as private
+    self.accessModifier = accessModifier ?? SourceRepresentable(value: .private, range: site)
```
kyouko-taiga, line 57: Same remark about the non-optional constructor argument.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -133,6 +133,14 @@ public enum Parser {
               appearsBeforeAccessModifier: access))
         }
 
+        // Catch incosistent access modifiers.
+        else if let prev = accessModifiers.first, prev != access {
+          state.diagnostics.insert(
+            .error(
+              inconsistentAccessModifiers: access,
+              appearsAfterPreviousAccessModifier: prev))
+        }
```
kyouko-taiga, line 142:
```suggestion
        else if let p = accessModifiers.first, p != access {
          state.diagnostics.insert(
            .error(
              inconsistentAccessModifiers: access,
              appearsAfterPreviousAccessModifier: p))
        }
```

We tend to avoid abbreviations in the code, or use single letter variables if the context allows.

kyouko-taiga (changes requested):
Excellent work!

The only change that should be made is that modifiers should be synthesized by the parser rather than in the declarations' constructors.

Sources/Core/AST/Decl/BindingDecl.swift
jlscheerer: Thank you very much for the feedback! Do you have any reference about how I can synthesize an object that is not an AST node (e.g., the AccessModifier)?

kyouko-taiga: You can look at examples of synthesized declarations, like this [one](https://github.com/val-lang/val/blob/900601c03be0963c5fb21308d778df7c546fbb7e/Sources/FrontEnd/Parse/Parser.swift#L1057-L1086).

You can pick any range that makes sense for the synthesized site. For instance, we often use the first position of the declaration, like [here](https://github.com/val-lang/val/blob/900601c03be0963c5fb21308d778df7c546fbb7e/Sources/FrontEnd/Parse/Parser.swift#LL925C9-L925C88).

jlscheerer:
Thanks again for your review! I hope I have addressed all outstanding comments.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -133,6 +133,14 @@ public enum Parser {
               appearsBeforeAccessModifier: access))
         }
 
+        // Catch incosistent access modifiers.
```
kyouko-taiga, line 136:
```suggestion
        // Catch inconsistent access modifiers.
```

Issue #677, opened 2023-06-20
"is not representable in LLVM" assertion


dabrahams:
With `--import-builtin`, this file makes the compiler assert, with

CodeGenLLVM/LoweredProgram+Extensions.swift:152: Fatal error: 'Pointee' is not representable in LLVM

```
type Pointer<Pointee> {

  var base: Builtin.ptr

  memberwise init

  public property q: Pointee {
    yield base as* (remote let Pointee)
  }
}

fun test(p: Pointer<Int>) {
  var x: Int = p.q // Actual problem: var should be let
}

public fun main() {}
```

kyouko-taiga:
The problem seems to be that `p.q` is inferred as having type `Pointee` instead of `Int`. I'm almost certain the issue comes from `TypeChecker.inferredType(of:shapedBy:updating:)` (the overload for `SubscriptCallExpr`).

PR #680, opened 2023-06-21
Monomorphize the type of subscript references during lowering

kyouko-taiga:
This PR fixes the assertion failure mentioned in #677. It does not fix the issue with `var`.

### DIFF
--- a/Sources/IR/Emitter.swift
+++ b/Sources/IR/Emitter.swift
@@ -1824,7 +1824,7 @@ public struct Emitter {
       return emitComputedProperty(boundTo: receiver, declaredBy: i, parameterizedBy: a, at: site)
     }
 
-    let t = SubscriptType(program.relations.canonical(program[d].type))!
+    let t = SubscriptType(program.canonicalType(of: d, parameterizedBy: a))!
     let r = append(module.makeAccess(t.capabilities, from: receiver, at: site))[0]
 
     var variants: [AccessEffect: Function.ID] = [:]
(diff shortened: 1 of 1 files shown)


dabrahams:
I'm slightly concerned that monomorphization is _required_ for anything to work.  Is that concern misplaced?

kyouko-taiga:
Monomorphizing types isn't the same as monomorphizing a function or subscript. It just means replacing generic parameters, like going from a polymorphic `<T>(T) -> T` to a monomorphic `(Int) -> Int`.

This process is indeed required for any generic program to work but I don't think it should be cause for concern.

dabrahams:
I'm pretty sure I know what it means, and I'm still a bit concerned.  In a resilient generics model, you can't monomorphize all the types in the compiler because some are only discovered at runtime.  

I also don't know what to make of it when you say monomorphizing types is not like monomorphizing a function and then use monomorphizing a function as an example of what it's like, FWIW.  But maybe I'm missing something!

kyouko-taiga:
In a resilient generics model you still have access to type signatures. Say you have this type in a module:
```swift
public type Box<T> {
  public subscript foo: T { ... }
}
```
Another module can import `Box`, not know anything about its implementation, but still be allowed to substitute `T` for `Int` and conclude that `Box<Int>.foo` has type `Int`. That's what monomorphizing a type means.

The example I gave is monomorphizing a function _type_, not a function. Monomorphizing a function requires generating new code, thus requiring access beyond the resilience boundary. Monomorphizing a type only requires reading the information advertised by the resilience boundary.

If you're still worried, I think it's best to take this in-person or we risk piling up misunderstanding.

dabrahams:
Got it; thanks for explaining

PR #696, opened 2023-06-27
Rationalize names of floating types

### LINKED ISSUE #694: Use less confusing floating-point type names
General purpose floating-point numbers should be represented by `Float` instead of `Double`, which would represent a "machine-size" floating-point number. Floating-point numbers with specific bit widths should be represented by `Float16`, `Float32`, `Float64` and `Float128`.

dabrahams:
Fixes #694


kyouko-taiga (approved):
Perfect.

PR #716, opened 2023-07-06
Inline storage view Instruction

dabrahams:


Sources/IR/Analysis/Module+Ownership.swift
```diff
@@ -120,20 +120,22 @@ extension Module {
...
 
       // Skip the instruction if an error occured upstream.
-      guard let base = context.locals[s.base] else {
+      guard let base = context.locals[s.source] else {
         assert(diagnostics.containsError)
         return
       }
 
-      let newLocations = base.unwrapLocations()!.map({ $0.appending(s.elementPath) })
+      if s.targetPath.elementOffset != nil { fatalError("pointer indexing not fully implemented") }
```
kyouko-taiga, line 136: Now I get why you wanted an optional.

I think it would be equivalent to replace `nil` by `IR.IntegerConstant(0, bitWidth: 32)`.

dabrahams:
Withdrawn in favor of having two separate instructions for pointer offsets and lensing into records.

PR #719, opened 2023-07-07
element_addr => subfield_view

dabrahams:
Renaming, commenting, etc. in preparation for separate pointer_offset instruction


Sources/Core/AST/AST.swift
```diff
@@ -316,8 +318,9 @@ public struct AST {
   ///   tuples, then they have equal lengths and labels.
   public func walking(
     pattern: AnyPatternID, expression: AnyExprID,
-    at root: PartPath = [],
-    _ action: (_ path: PartPath, _ subpattern: AnyPatternID, _ subexpression: AnyExprID) -> Void
+    at root: SubfieldID = [],
```
kyouko-taiga, line 321: This line here shows makes me a little unsure about `SubfieldID` being the name of the abstraction. The empty "path" doesn't identify a sub-field, it identifies the record itself. I don't have a much nicer solution on the top of my head though, `PartPath` had the same issue.

Sources/IR/Analysis/AbstractLocation.swift
```diff
@@ -4,7 +4,7 @@ import Core
 enum AbstractLocation: Hashable {
 
   /// A root location.
-  case root(Operand)
+  case recordAddress(Operand)
```
kyouko-taiga, line 7: This renaming doesn't work because a root location does not necessarily identifies a record. It may, for example, be the address of an existential container, which isn't a record.

I think the name "record" should denote "things with fields". It's certainly used that way currently (see `AnyType.hasRecordLayout`).

Sources/IR/Analysis/Module+NormalizeObjectStates.swift
```diff
@@ -686,39 +688,39 @@ extension State: CustomStringConvertible {
...
-  var consumed: [(path: PartPath, consumers: State.Consumers)]
+  var consumed: [(subfield: SubfieldID, consumers: State.Consumers)]
 
 }
 
 extension AbstractObject.Value where Domain == State {
 
   /// If `self` is `.partial`, the paths to `self`'s parts; otherwise, `nil`.
-  fileprivate var paths: PartPaths? {
+  fileprivate var subfieldsByStatus: SubfieldStatus? {
```
kyouko-taiga, line 708: Is the `ByStatus` really necessary?

It confused me at call sites; I only understood once I looked at this definition. The abstraction does make sense, but I don't think the property name does.

Sources/Core/AST/AST.swift
dabrahams: I hear you.

`RecordPath`?  `RecordFieldPath`?

Sources/IR/Analysis/AbstractLocation.swift
dabrahams: 100% agreed.

Sources/IR/Analysis/Module+NormalizeObjectStates.swift
dabrahams: Don't know about _necessary_, but if I got something called `subfields` I'd expect it to be a single collection.  Are you sure you aren't confused by it only because you know the prior code well?

Sources/IR/Analysis/AbstractLocation.swift
dabrahams: This is weird; are you sure I didn't undo this change?  I don't see it in my local code; it's still called `root` there.

Sources/Core/AST/AST.swift
kyouko-taiga: I like `RecordPath`

Sources/IR/Analysis/AbstractLocation.swift
kyouko-taiga: I'm sure it is, unless e0d4fcc3a41996feb609ad755d94addfe7991bc9 isn't the last commit. See here: https://github.com/val-lang/val/blob/e0d4fcc3a41996feb609ad755d94addfe7991bc9/Sources/IR/Analysis/AbstractLocation.swift#L7

Sources/Core/AST/AST.swift
dabrahams: Doesn't bother you that you can apply an empty `RecordPath` to the address of a non-record?

dabrahams: I think that's probably OK.  You?

kyouko-taiga: I'm okay.

Sources/IR/Analysis/Module+NormalizeObjectStates.swift
dabrahams: Then I should rename the type.  I'd offer `fieldsByInitializationState` but I don't want to make you ill.  I will go to `subfields` unless you say otherwise.

kyouko-taiga: I would personally choose `subfields` but am 100% okay with `fieldsByInitializationState` if you think it helps.

dabrahams: Well now this is interesting.  I meant to float `subfieldsByInitializationState` but mistyped.  Are these actually all the subfields (recursively determined), or is the thing just wastefully storing arrays when it could store single indices?  If the latter, I _would_ use `fields`

Sources/Core/AST/AST.swift
dabrahams: You're missing my point.  You have allowed for `field view %1` with an empty path.  The type of `%1` doesn't need to be a record type in that case, and presumably might not be.  I personally wonder why empty paths are allowd.

Sources/IR/Analysis/Module+NormalizeObjectStates.swift
kyouko-taiga: Its computing record paths recursively and returns all part by initialization state at arbitrary depths. A path is expensed iff it identifies an aggregate whose parts have different states.

Sources/Core/AST/AST.swift
kyouko-taiga: The instruction must check if the path is valid in `%1` regardless of its depth. That's because you could have `%1` be a`{any Copyable, Int}` and a path `[0]`, this identifying a non-record. That's no different than using the instruction to get `[]` from `any Copyable`.

> I personally wonder why empty paths are allowed.

They used to be disallowed but that caused code to grow more complex in other places. Eventually I decided that we should not bother and eventually have a pass that would optimize sequences of the form:
```
%x = field_view %0, 1, 2
%y = field_view %x, 3, 4
%z = field_view %x, 5, 6
```

dabrahams: > thus identifying a non-record

This is not about whether the path identifies a record, whose address is the output.  This is about whether the object at the input address (1st argument) is even a record type to begin with.

kyouko-taiga: Now I understand your point. I agree

PR #753, opened 2023-07-09
advanced by strides instruction for pointers

dabrahams:
Most of the 2nd commit of this PR is a WAG.  Please review carefully.

### DIFF
--- a/Sources/IR/Analysis/Module+Depolymorphize.swift
+++ b/Sources/IR/Analysis/Module+Depolymorphize.swift
@@ -146,8 +146,8 @@ extension Module {
       switch sourceModule[i] {
       case is AddressToPointerInstruction:
         rewrite(addressToPointer: i, to: b)
-      case is AdvancedByStridesInstruction:
-        rewrite(advancedByStrides: i, to: b)
+      case is AdvancedByBytesInstruction:
+        rewrite(advancedByBytes: i, to: b)
       case is AllocStackInstruction:
         rewrite(allocStack: i, to: b)
       case is BorrowInstruction:
@@ -207,10 +207,10 @@ extension Module {
     }
 
     /// Rewrites `i`, which is in `r.function`, into `result`, at the end of `b`.
-    func rewrite(advancedByStrides i: InstructionID, to b: Block.ID) {
-      let s = sourceModule[i] as! AdvancedByStridesInstruction
+    func rewrite(advancedByBytes i: InstructionID, to b: Block.ID) {
+      let s = sourceModule[i] as! AdvancedByBytesInstruction
       append(
-        makeAdvancedByStrides(source: rewritten(s.source), offset: rewritten(s.offset), at: s.site),
+        makeAdvancedByBytes(source: rewritten(s.source), offset: rewritten(s.offset), at: s.site),
         to: b)
     }
 
--- a/Sources/IR/Analysis/Module+Depolymorphize.swift
+++ b/Sources/IR/Analysis/Module+Depolymorphize.swift
@@ -210,7 +210,7 @@ extension Module {
     func rewrite(advancedByBytes i: InstructionID, to b: Block.ID) {
       let s = sourceModule[i] as! AdvancedByBytesInstruction
       append(
-        makeAdvancedByBytes(source: rewritten(s.source), offset: rewritten(s.offset), at: s.site),
+        makeAdvancedByBytes(source: rewritten(s.base), offset: rewritten(s.byteOffset), at: s.site),
         to: b)
     }
 
--- a/Sources/IR/Analysis/Module+Depolymorphize.swift
+++ b/Sources/IR/Analysis/Module+Depolymorphize.swift
@@ -146,6 +146,8 @@ extension Module {
       switch sourceModule[i] {
       case is AddressToPointerInstruction:
         rewrite(addressToPointer: i, to: b)
+      case is AdvancedByStridesInstruction:
+        rewrite(advancedByStrides: i, to: b)
       case is AllocStackInstruction:
         rewrite(allocStack: i, to: b)
       case is BorrowInstruction:
@@ -202,6 +204,14 @@ extension Module {
     func rewrite(addressToPointer i: InstructionID, to b: Block.ID) {
       let s = sourceModule[i] as! AddressToPointerInstruction
       append(makeAddressToPointer(rewritten(s.source), at: s.site), to: b)
+    }
+
+    /// Rewrites `i`, which is in `r.function`, into `result`, at the end of `b`.
+    func rewrite(advancedByStrides i: InstructionID, to b: Block.ID) {
+      let s = sourceModule[i] as! AdvancedByStridesInstruction
+      append(
+        makeAdvancedByStrides(source: rewritten(s.source), offset: rewritten(s.offset), at: s.site),
+        to: b)
     }
 
     /// Rewrites `i`, which is in `r.function`, into `result`, at the end of `b`.
--- a/Sources/IR/Operands/Instruction/AdvancedByStridesInstruction.swift
+++ b/Sources/IR/Operands/Instruction/AdvancedByStridesInstruction.swift
@@ -1,0 +1,67 @@
+import Core
+
+/// Computes a `source` address value advanced by `offset` strides of the `source`'s content type
+/// (`source + offset` in C terms).
+public struct AdvancedByStridesInstruction: Instruction {
+
+  /// The pointer value that is advanced to produce the result.
+  public private(set) var source: Operand
+
+  /// An integer number of strides to advance the source value.
+  public private(set) var offset: Operand
+
+  /// The site of the val source from which `self` was generated.
+  public let site: SourceRange
+
+  /// The content type of the `source` (and resulting) address value.
+  private let contentType: LoweredType
+
+  /// Creates an instance with the given properties.
+  fileprivate init(
+    contentType: LoweredType,
+    source: Operand,
+    offset: Operand,
+    site: SourceRange
+  ) {
+    self.contentType = contentType
+    self.source = source
+    self.offset = offset
+    self.site = site
+  }
+
+  public var types: [LoweredType] { [contentType] }
+
+  public var operands: [Operand] { [source, offset] }
+
+  public mutating func replaceOperand(at i: Int, with new: Operand) {
+    precondition(i >= 0 && i < 2)
+    if i == 0 { source = new } else { offset = new }
+  }
+
+}
+
+extension AdvancedByStridesInstruction: CustomStringConvertible {
+
+  public var description: String {
+    "\(source) advanced by \(offset) strides"
+  }
+
+}
+
+extension Module {
+
+  /// Creates an `advanced by strides` instruction anchored at `site` computing the `source` address
+  /// value advanced by `offset` strides of the `source`'s content type.
+  func makeAdvancedByStrides(source: Operand, offset: Operand, at site: SourceRange)
+    -> AdvancedByStridesInstruction
+  {
+    precondition(type(of: source).isAddress)
+    precondition(type(of: offset).isAddress)
+    return .init(
+      contentType: type(of: source),
+      source: source,
+      offset: offset,
+      site: site)
+  }
+
+}
--- a/Sources/CodeGen/LLVM/Transpilation.swift
+++ b/Sources/CodeGen/LLVM/Transpilation.swift
@@ -461,8 +461,8 @@ extension LLVM.Module {
       switch m[i] {
       case is IR.AddressToPointerInstruction:
         insert(addressToPointer: i)
-      case is IR.AdvancedByStridesInstruction:
-        insert(advancedByStrides: i)
+      case is IR.AdvancedByBytesInstruction:
+        insert(advancedByBytes: i)
       case is IR.AllocStackInstruction:
         insert(allocStack: i)
       case is IR.BorrowInstruction:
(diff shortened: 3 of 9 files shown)


Sources/IR/Operands/Instruction/AdvancedByStridesInstruction.swift
```diff
@@ -0,0 +1,67 @@
...
+    "\(source) advanced by \(offset) strides"
+  }
+
+}
+
+extension Module {
+
+  /// Creates an `advanced by strides` instruction anchored at `site` computing the `source` address
+  /// value advanced by `offset` strides of the `source`'s content type.
+  func makeAdvancedByStrides(source: Operand, offset: Operand, at site: SourceRange)
```
kyouko-taiga, line 55: I think there's something wrong with this constructor. Your instruction will apply on `Builtin.ptr`, so asking `self` for the type of `source` will invariably give you a built-in pointer. Whether `source` has to be an address is another question. I'd say probably not.

I'll add more details in the review message.

Sources/IR/Analysis/Module+Depolymorphize.swift
```diff
@@ -200,6 +202,14 @@ extension Module {
       append(makeAddressToPointer(rewritten(s.source), at: s.site), to: b)
     }
 
+    /// Rewrites `i`, which is in `r.function`, into `result`, at the end of `b`.
+    func rewrite(advancedByStrides i: InstructionID, to b: Block.ID) {
+      let s = sourceModule[i] as! AdvancedByStridesInstruction
+      append(
+        makeAdvancedByStrides(source: rewritten(s.source), offset: rewritten(s.offset), at: s.site),
+        to: b)
```
kyouko-taiga, line 210: Here we can see why `s.contentType` is significant.

You're implementing the monomorphization of an offset computation that depends on a type to compute the stride. If that type happens to be generic in this context, clearly it has to be monomorphized.

kyouko-taiga (changes requested):
It's a very nice PR. Good job on navigating the design of IR analyses!

One issue, however, is that your instruction isn't built with the right values. Let's look at an example to explain:

Assume we have this code in Val:

```swift
fun foo(_ x: T[4], i: Int) {
  bar(x[i])
}
```

IIUC, the main purpose of `advanced_by_strides` is to compute the address of the `i`-th element of a buffer. So in the example it will be used to compute the address of `x[i]`.

Assuming perfect inlining, at the IR level, I expect `x[i]` to be lowered as follows:

```
%1: &Builtin.i64 = subfield_view %i, 0
%2: Builtin.i64 = load %1
%3: Builtin.ptr = address_to_pointer %x
%4: Builtin.ptr = advanced %3 by %2 strides
%5: pointer_to_address %4 as (remote let T)
```

You'll notice that something is missing then. How are we supposed to compute the stride by which `%3` is advanced? The answer is `AdvancedByStridesInstructions.contentType`, which we should make a part of the textual representation. So the 4th line becomes:

```
%4: Builtin.ptr = advanced %3 by %2 strides_of T
```

Because the argument of your instruction is a built-in pointer, its type is invariably an opaque `Builtin.ptr`. I also think it should be an object type rather than an address type. We're in one of the rare situations where we have to deal with actual values because of the way LLVM's `getemelementptr` works.

So at construction, we need to pass it `contentType` explicitly. Only the code generator can know the type of `source`'s pointee. We also need to monomorphize this type when we're rewriting `advanced _by_strides` during monomorphization.

dabrahams:
> It's a very nice PR. Good job on navigating the design of IR analyses!

Well, as noted, most of that is a copy/paste WAG.

> IIUC, the main purpose of `advanced_by_strides` is to compute the address of the `i`-th element of a buffer. 

No, it's to do pointer arithmetic.  The buffer functionality will be built on that.

> ```
> %4: Builtin.ptr = advanced %3 by %2 strides
> ```

```
%4: Builtin.ptr =  %3 advanced by %2 strides
```

> You'll notice that something is missing then. How are we supposed to compute the stride by which `%3` is advanced? 

My presumption was that it was encoded in the runtime value of `%3`.  If `Builtin.ptr` is non-generic (i.e. as opaque as an LLVM pointer), of course, it can't be.

The answer is `AdvancedByStridesInstructions.contentType`, which we should make a part of the textual representation. So the 4th line becomes:
> 
> ```
> %4: Builtin.ptr = advanced %3 by %2 strides_of T
> ```

Underscore needed?


```
%4: Builtin.ptr = %3 advanced by %2 strides of T
```

> Because the argument of your instruction is a built-in pointer, its type is invariably an opaque `Builtin.ptr`. I also think it should be an object type rather than an address type. 

I guess I can look at load and store to see how to encode the latter distinction? Anything else I need to know about that?

> We're in one of the rare situations where we have to deal with actual values because of the way LLVM's `getemelementptr` works.
> 
> So at construction, we need to pass it `contentType` explicitly, 

We do, though.  I _think_ what you're really saying is that `contentType` needs to be an operand, because we don't know its monomorphized value when the instruction is generated?

But then the following makes me think maybe I have that wrong:

> as only the code generator can know the type of `source`'s pointee. We also need to monomorphize this type when we're rewriting `advanced _by_strides` during monomorphization.

dabrahams:
If all the builtin types are non-generic, I guess I need to encode the offset type as an operand as well?  Or have I misunderstood?

kyouko-taiga:
> Underscore needed?

No. `%3 advanced by %2 strides of T` is fine.

> I guess I can look at load and store to see how to encode the latter distinction? Anything else I need to know about that?

Not sure to understand the question.

But at this point I'd suggest that you simply expect `source` to have an object type and ignore other details about `load`/`store`.

> We do, though. I think what you're really saying is that contentType needs to be an operand, because we don't know its monomorphized value when the instruction is generated?

As you suspected later, the problem is indeed that built-in _pointer_ types are non-generic. In other words, the module only knows that `source` is a pointer, it doesn't know anything about the pointee's type.

It would be reasonable to ask for more type safety, but since this problem only relates to `Builtin.ptr` and that we can expect the front-end to have type-checked the actual program, I think it bring a lot of complexity for very little gain.

> If all the builtin types are non-generic, I guess I need to encode the offset type as an operand as well? Or have I misunderstood?

I'd answer yes to the question "are all built-in types non-generic" but the remainder of the question suggests we do not operate with the same definition. To me non-generic should be understood as "not parametrizable by another type". We do not have a `Builtin.ptr<T>` type, meaning that a pointer type says nothing about the type of its pointee, but `Builtin.i64` is distinct from `Builtin.fp128`.

So there's no need to encode the type of the offset because it is known by the module.

In the IR fragment example I wrote:

```
%1: &Builtin.i64 = subfield_view %i, 0
%2: Builtin.i64 = load %1
%3: Builtin.ptr = address_to_pointer %x
```

Notice that `%2` has type `Builtin.i64` and `%3` has type `Builtin.ptr`. These correspond exactly to what `module.type(of:)` will return.

dabrahams:
> > I guess I can look at load and store to see how to encode the latter distinction? Anything else I need to know about that?
> 
> Not sure to understand the question.

I have only dealt with address types so far, and I am doing everything by example, so I thought there might be some things I needed to know.

> But at this point I'd suggest that you simply expect `source` to have an object type and ignore other details about `load`/`store`.

> > We do, though. I think what you're really saying is that contentType needs to be an operand, because we don't know its monomorphized value when the instruction is generated?
> 
> As you suspected later, the problem is indeed that built-in _pointer_ types are non-generic. In other words, the module only knows that `source` is a pointer, it doesn't know anything about the pointee's type.
> 
> It would be reasonable to ask for more type safety, but since this problem only relates to `Builtin.ptr` and that we can expect the front-end to have type-checked the actual program, I think it would bring a lot of complexity for very little gain.

It's not about type-safety, I think; it's just a question of whether type information is threaded through instructions explicitly or can be carried along in generic values.  I'm not sure what the tradeoffs are, but if it wasn't considered it might be worth keeping in mind as a possibility.

> > If all the builtin types are non-generic, I guess I need to encode the offset type as an operand as well? Or have I misunderstood?
> 
> To me non-generic should be understood as "not parametrizable by another type". 

That's exactly what I meant.

We do not have a `Builtin.ptr<T>` type, meaning that a pointer type says nothing about the type of its pointee, but `Builtin.i64` is distinct from `Builtin.fp128`.
> 
> So there's no need to encode the type of the offset because it is known by the module.

I'm a little confused here.  As coded, instruction can deal with an offset of any integer type.  That can allow any Val integer type to be added to a pointer, and presumably we could code a generic offset operation.  The module can't know the eventual type of that integer until the code is monomorphized, which—at resilience boundaries—may never happen.

Now you could say that level of genericity is overkill, but at an IR level (AFAIK) I don't have access to a word-sized integer type yet.

kyouko-taiga:
> I'm a little confused here. As coded, instruction can deal with an offset of any integer type. That can allow any Val integer type to be added to a pointer, and presumably we could code a generic offset operation.

Not any Val integer type, any `Builtin.ixx` type, which will be extracted from the corresponding Val type.

> The module can't know the eventual type of that integer until the code is monomorphized, which—at resilience boundaries—may never happen.

I don't think it's an issue. You'll need to tell the compiler about which built-in integer type you're using. Consider this function:

```
public fun advance<T, U: BinaryInteger>(
  _ base: Pointer<T>, by_strides_of strides: U
) -> Pointer<T> {
  let v: Builtin.ptr = base.value
  let w = Builtin.advance(v, by: magic(strides), strides_of: T)
  return .new(value: w)
}
```

Here, we need `magic` to give us a built-in integer type from `U`.

But that makes me wonder if we even need to care about the strides at all in the IR. Is there any downside in offsetting pointers by bytes at this level? In thing we really care about is to implement `advanced(by:strides:)` in Val, which could be:

```swift
public fun advance<T, U: BinaryInteger>(
  _ base: Pointer<T>, by_strides_of strides: U
) -> Pointer<T> {
  let v: Builtin.ptr = base.value
  let w = Builtin.advance(v, by: Int(strides) * Metatype<T>.stride)
  return .new(value: w)
}
```

> Now you could say that level of genericity is overkill, but at an IR level (AFAIK) I don't have access to a word-sized integer type yet.

Actually now there's `Builtin.word` :)

kyouko-taiga:
>  Is there any downside in offsetting pointers by bytes at this level?

Maybe that would make LLVM worse at doing optimization. I will try to investigate.

dabrahams:
> > I'm a little confused here. As coded, instruction can deal with an offset of any integer type. That can allow any Val integer type to be added to a pointer, and presumably we could code a generic offset operation.
> 
> Not any Val integer type, any `Builtin.ixx` type, which will be extracted from the corresponding Val type.

“can allow” meant you can build it, not that it's a direct consequence.

> > The module can't know the eventual type of that integer until the code is monomorphized, which—at resilience boundaries—may never happen.
> 
> I don't think it's an issue. You'll need to tell the compiler about which built-in integer type you're using. Consider this function:
> 
> ```swift
> public fun advance<T, U: BinaryInteger>(
>   _ base: Pointer<T>, by_strides_of strides: U
> ) -> Pointer<T> {
>   let v: Builtin.ptr = base.value
>   let w = Builtin.advance(v, by: magic(strides), strides_of: T)
>   return .new(value: w)
> }
> ```
> 
> Here, we need `magic` to give us a built-in integer type from `U`.

oof.

I was imagining `magic` returning something like an `BuiltinValue` instance rather than a concrete `Builtin.word`, and in that case I think I'd need to pass the builtin type as a parameter because the module wouldn't know it(?)

> But that makes me wonder if we even need to care about the strides at all in the IR. Is there any downside in offsetting pointers by bytes at this level? 

I think the only issue is IR readability.  You could push the same stride multiplication up into Val code, but nobody would want to read the result.  Your choice.

> Actually now there's `Builtin.word` :)

Yes, but I don't know how to access it from the compiler's implementation code (where I can use things like `i32`).

kyouko-taiga:
For our information, it _seems_ like LLVM is equally good at optimizing `getelementptr` using byte offsets. Definite answer would require more investigation.

> oof.
>
> I was imagining magic returning something like an `BuiltinValue` instance rather than a concrete Builtin.word, and in that case I think I'd need to pass the builtin type as a parameter because the module wouldn't know it(?)

It can return `Builtin.word` or any `Builtin.ixx`. Something like a `BuiltinValue` would require support from the run-time. In that case, I would say that it is overkill. But maybe I miss something.

> I think the only issue is IR readability. You could push the same stride multiplication up into Val code, but nobody would want to read the result. Your choice.

That is a fair point. There are serious pros and cons to both approaches.

I care about the readability of the IR, but that of Val has precedence. IMO the second approach is simpler to understand at the Val level, even in a generic context. Maybe that should be the criterion?

I guess if you pressed me for an answer right now, I'd say we should implement the instruction to operate on strides of bytes.

> Yes, but I don't know how to access it from the compiler's implementation code (where I can use things like i32).

- In Val: `Builtin.word` (with `--import-builtin`)
- In code: `BuiltinType.word` or `AnyType.builtin(.word)` (in the `Core` module)

For example:

```swift
extension Module {

  func makeMyInstruction(_ source: Operand, at site: SourceRange) -> MyInstruction {
    precondition(type(of: source) == .object(AnyType.builtin(.word)))
    return .init(source: source, site: site)
  }

}

```

dabrahams:
One other thing, that might be significant: if we do the multiplication in Val there will be overflow checking for safety, which we'll need to bypass(?) or optimize

kyouko-taiga:
> One other thing, that might be significant: if we do the multiplication in Val there will be overflow checking for safety, which we'll need to bypass(?) or optimize

True. I guess we would then have to choose whether we want to make `Pointer<T>.advanced(by:)` an unsafe operation or if we are willing to pay of the overflow checking. I think you are better positioned to answer this question than I am.

kyouko-taiga:
To summarize, we have these options:
1. Have the builtin instruction be aware of the pointee's type
2. Operate on strides of bytes and pay for overflow checking in Val, hoping it gets optimized in general
3. Operate on strides of bytes and add the `unsafe` effect to `Pointer<T>.advanced(by:)`

FWIW, I would personally pick 3-2-1 in this order. My rationale is that using `getelementpr inrange` with a too large offset is UB anyway.

dabrahams:
Pointer advancing could still be safe if we just truncate the multiplication at the address space boundary (like using `*&` in swift).  I'm fine with byte-sized motion in the IR if that's what you want.

dabrahams:
My current take is that your number 3 is fine.

kyouko-taiga (approved):

PR #759, opened 2023-07-13
Initial infrastructure for mangling/demangling

kyouko-taiga:


Sources/Core/Mangling/Base64Digit.swift
dabrahams, line 1: I wonder if Base64 is really the most efficient thing;  how many characters are actually available for use?  Is it just a coincidence that LLVM assembly ids only allow 64 characters?

kyouko-taiga: LLVM allows 2 more. We use one of them as a bit to determine if a string has been sanitized.

Base 64 has nice properties that we could exploit to optimize some numeric operations. So I don't think using base 65 (or even 66) would buy us much.

kyouko-taiga: I tried many different way to deal with strings that contain characters outside of the 66 LLVM allows. Sanitization is one of them. It will remove illegal characters and encode the necessary instructions to put them back. There's a lengthy comment explaining the whole process in String+AssemblySanitized.swift.

PR #761, opened 2023-07-13
Improve assertion diagnostic

dabrahams:


Sources/CodeGen/LLVM/TypeLowering.swift
```diff
@@ -146,6 +146,8 @@ extension LoweredProgram {
 }
 
 /// Traps indicating that `t` is not representable in LLVM.
-private func notLLVMRepresentable<T: TypeProtocol>(_ t: T) -> Never {
-  preconditionFailure("'\(t)' is not representable in LLVM")
+private func notLLVMRepresentable<T: TypeProtocol>(
+  _ t: T, file: StaticString = #filePath, line: UInt = #line
+) -> Never {
+  preconditionFailure("'\(t)' is not representable in LLVM", file: (file), line: line)
```
kyouko-taiga, line 152: nit: a reason `file` is parenthesized?

dabrahams: Swift will warn you if you forward `#filePath` to a thing called `file` otherwise.

I loathe the whole `#file`/`#fileID` compression nonsense.

PR #770, opened 2023-07-17
Function.parameters => Function.genericParameters

dabrahams:
“Function parameters” already means something else.

### DIFF
--- a/Sources/IR/Analysis/Module+Depolymorphize.swift
+++ b/Sources/IR/Analysis/Module+Depolymorphize.swift
@@ -453,7 +453,7 @@ extension Module {
       name: "<\(list: parameterization.values), \(useScope)>(\(source.name))",
       site: source.site,
       linkage: .module,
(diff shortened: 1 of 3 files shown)


kyouko-taiga:
I think we should consider using the terms "static"/"compile-time" and "dynamic"/"run-time" as qualifiers. they are less ambiguous than the terms "generic" and "non-generic".

dabrahams:
I'm not sure I agree.  Is a dispatch through a witness table implementing a resilient generic static or dynamic?  IMO what you have there is statically typed but dynamically dispatched.

"generic parameter" has a perfectly unambiguous meaning as far as I can tell.

As for “compile-time,” that will become very ambiguous when we start evaluating runtime code in compile-time contexts at compile time.

kyouko-taiga:
Okay, I'm happy to go along with this change. We'll see if it deserve to be revisited once we get to non-type arguments.

PR #772, opened 2023-07-18
Transpile => Lower

dabrahams:
Even this renaming suggests a further simplification: let's just talk explicitly about generating LLVM rather than some non-specific "lowering".  I'll push another that begins to do that so we can evaluate.


kyouko-taiga:
The problem with this renaming is that we cannot tell whether a "lowered thing" belongs to Val IR or LLVM. I think we need either a qualifying namespace or a different term.

dabrahams:
Agreed; closing in favor of the refactoring recently discussed that begins with #773

PR #773, opened 2023-07-18
LoweredXXX => IR.XXX

dabrahams:
Note that a Swift bug means we need to say `IR.Type_` rather than `IR.Type`.


kyouko-taiga (approved):
LGTM. The idea of qualifying IR-relates names with `IR` is growing on me.

PR #777, opened 2023-07-20
Parameterized IR.Module

dabrahams:
@kyouko-taiga this is some of the refactoring we discussed.  There are lots of directions to go with it and I'm not sure when to consider my changes a finished unit.  I'd like your feedback on the direction I've taken.


Sources/IR/Module.swift
```diff
@@ -2,17 +2,50 @@ import Core
...
+
+  public var typedProgram: TypedProgram { base }
+
+}
+
+/// An IR Module in a program that has not been fully lowered to IR.
+public typealias ModuleUnderConstruction = Module<Core.TypedProgram>
+
+/// An IR Module in a program that has been fully lowered to IR.
+public typealias FinishedModule = Module<IR.Program>
```
kyouko-taiga, line 36: This isn't fully accurate. We'll still be allowed to mutate modules once we implement optimization passes, but at this point they will no longer be "under construction".

kyouko-taiga (commented):
I am not completely convinced these changes are clear improvements. I left inline comments about `program` => `container` and `Module` => `ModuleUnderConstruction`.

I also have to admit that the idea of using a typestate to track progress on the module sounded a little better than it looks. There's at least one other state we should capture if we want to be accurate (for optimization passes) and so I worry that we're just putting lipstick on a `Module`/`MutableModule` pair of types, which is probably an anti-pattern. A better way to indicate mutation in Swift is using `var`/`inout`.

Sources/IR/Module.swift
dabrahams: I suppose you're saying the name `FinishedModule` is inaccurate?  Given what you've just told me, I'm not sure “under construction” is great either.  `PartialModule` and `CompleteModule`?

Sources/IR/Module.swift
```diff
@@ -2,17 +2,50 @@ import Core
...
 /// designated as its entry point (i.e., the `main` function of a Val program).
-public struct Module {
+public struct Module<Container: IR.ModuleContainer> {
 
   /// The identity of a global defined in a Val IR module.
   public typealias GlobalID = Int
 
   /// The program defining the functions in `self`.
-  public let program: TypedProgram
+  public let container: Container
```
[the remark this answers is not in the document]
dabrahams: I understand and don't completely disagree.  It's part of my attempt to address my struggle with code that looks like `part.biggerThing.biggerThingStill`.  In a world of whole-part value semantics, I think we need some kind of syntactic convention for getting to “the thing that contains `x`” and this was one idea.  I'd be more than happy to discuss alternatives.

Sources/IR/Module.swift
dabrahams: Note that as long as this `Module` contains its whole, we can't mutate the whole through operations on `Module` instances, so I'm not sure “we'll still be allowed to mutate modules…” can be realized.

kyouko-taiga: The whole is a program which I think never has to be mutated (except for the part that the module represents) so I don't think there's an issue.

But maybe it's something we should take online because I worry some kind of misunderstanding might be sneaking in.

Sources/IR/Module.swift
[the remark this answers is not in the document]
kyouko-taiga: Half-baked ideas:

```swift
Program(containg: m).relations
// or
IR.Module.Container(containing: m).relations
// or
container(of: m).relations
```

Sources/IR/Module.swift
dabrahams: I'm pretty sure an `IR.Program` contains modules.  Yep.  Currently "under construction" but that ought to get fixed one way or another if we pursue this.

```swift
  /// A map from module ID to its lowered form.
  public private(set) var modules: [ModuleDecl.ID: IR.ModuleUnderConstruction]
```

So a `IR.Module<IR.Program>` can't really be mutated directly.

Sources/IR/Module.swift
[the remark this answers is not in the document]
dabrahams: Oh, you just want to be able to call it `program` when it's used internally to `IR.Module`?  Only the first of your ideas involves the word "Program".

dabrahams:
> I also have to admit that the idea of using a typestate to track progress on the module sounded a little better than it looks. There's at least one other state we should capture if we want to be accurate (for optimization passes) and so I worry that we're just putting lipstick on a `Module`/`MutableModule` pair of types, which is probably an anti-pattern. A better way to indicate mutation in Swift is using `var`/`inout`.

I'm not thrilled with that either.  But remember, the _real_ goal was to have a way to talk about "a module in the context of an `IR.Program`."  Maybe that's just a wholly different beast than an “`IR.Module`” (or `ModuleUnderConstruction`).  I don't know.

Lots of thoughts running thru my brain stem:

`Core.TypedProgram.IRModule`
`IR.Program.Module`

PR #823, opened 2023-08-02
Use YAML "folded" literals in build-and-test.yml

nickpdemarco:
See [this stack overflow post](https://stackoverflow.com/questions/59954185/github-action-split-long-command-into-multiple-lines). We've been misusing YAML's "literal block scalar" when writing commands.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -41,7 +41,7 @@ jobs:
     - name: Build and Test
       uses: devcontainers/ci@v0.3
       with:
-        runCmd: |
+        runCmd: >
           swift package resolve &&
           .build/checkouts/Swifty-LLVM/Tools/make-pkgconfig.sh llvm.pc &&
           export PKG_CONFIG_PATH=$PWD &&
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -42,6 +42,7 @@ jobs:
       uses: devcontainers/ci@v0.3
       with:
         runCmd: |
+          set -o pipefail &&
```
dabrahams, line 45: Why is this one `set -o pipefail` and the other one `shell: bash`?
If `shell: bash` doesn't work in this context, we should use explicit `set`s everywhere.

Do we not also want `set -e`? Let's just be as explicit as possible.

dabrahams (approved):
Overall, great work discovering the truth about the shells (which information I'd been unable to find).

Please think about my questions.  Also think about whether we want to remove the top-level `&&`s.
Finally, remember I asked that you put documentation about whatever you learned into the `.yml` file?

dabrahams:
@nickpdemarco Do we need to chat to get this moving forward?

nickpdemarco:
@dabrahams sorry, no - I got distracted by the compiler explorer work. I'll address your comments now.

.github/workflows/build-and-test.yml
nickpdemarco: Here, we're leveraging a custom [GitHub action](https://github.com/devcontainers/ci) which does not support a `shell` argument. It probably should, so I [raised an issue](https://github.com/devcontainers/ci/issues/256). For now, I'll be explicit in all steps for consistency, as you ask.

dabrahams:
@nickpdemarco it looks like a test is still disabled in this PR?  Please merge once you've fixed that!

Tests/ValTests/LexerTests.swift
```diff
@@ -82,7 +82,7 @@ final class LexerTests: XCTestCase {
       in: input)
   }
 
-  func testFloatingPointSequences() {
+  func disabledTestFloatingPointSequences() {
```
dabrahams, line 85:
```suggestion
  func testFloatingPointSequences() {
```

PR #827, opened 2023-08-03
Refactor Integer and Float Tokenization/Parsing for Future Parsing Issue Fix

esleche:
This PR adjusts the way integers and floats are tokenized and parsed, fixing the tuple key path parsing issue described in https://github.com/val-lang/val/issues/783
Specifically, floats are no longer tokenized as a single .float token. Instead, they are broken down into parts like .int, .dot, and .exponent. This change allows for more granular control over the tokenization process and directly resolves the aforementioned issue.

**Tokenization Changes:**
- **Before this PR:**
  - The floating-point number `1.12e+123` was tokenized as: `.float "1.12e+123"`

- **After this PR:**
  - The floating-point number `1.12e+123` is now tokenized as:
    - `.int, "1"`
    - `.dot, "."`
    - `.int, "12"`
    - `.exponent, "e+123"`

Fixes https://github.com/val-lang/val/issues/783

### DIFF
--- a/Sources/FrontEnd/Parse/Lexer.swift
+++ b/Sources/FrontEnd/Parse/Lexer.swift
@@ -170,7 +170,7 @@ public struct Lexer: IteratorProtocol, Sequence {
       token.kind = k
       token.site.extend(upTo: index)
 
-      if let nextChar = peek(), (nextChar == "e" || nextChar == "E") {
+      if let nextChar = peek(), nextChar == "e" || nextChar == "E" {
         var exponent = Token(kind: .exponent, site: location ..< location)
         discard()
         exponent.site.extend(upTo: index)
--- a/Sources/FrontEnd/Parse/Lexer.swift
+++ b/Sources/FrontEnd/Parse/Lexer.swift
@@ -17,9 +17,15 @@ public struct Lexer: IteratorProtocol, Sequence {
 
   /// The current location of the lexer in `sourceCode`.
   public var location: SourcePosition { sourceCode.position(index) }
+  private var pendingExponentToken: Token? = nil
 
   /// Advances to the next token and returns it, or returns `nil` if no next token exists.
   public mutating func next() -> Token? {
+    if let exponentToken = pendingExponentToken {
+      pendingExponentToken = nil
+      return exponentToken
+    }
+
     // Skip whitespaces and comments.
     while true {
       if index == sourceCode.text.endIndex { return nil }
@@ -159,10 +165,25 @@ public struct Lexer: IteratorProtocol, Sequence {
       return token
     }
 
-    // Scan numeric literals.
-    if let k = scanNumericLiteral() {
+    // Scan integral literals.
+    if let k = scanIntegralLiteral() {
       token.kind = k
       token.site.extend(upTo: index)
+
+      if let nextChar = peek(), (nextChar == "e" || nextChar == "E") {
+        var exponent = Token(kind: .exponent, site: location ..< location)
+        discard()
+        exponent.site.extend(upTo: index)
+
+        if let _ = scanIntegralLiteral(true) {
+          exponent.site.extend(upTo: index)
+        } else {
+          exponent.kind = .invalid
+        }
+
+        pendingExponentToken = exponent
+      }
+
       return token
     }
 
@@ -322,10 +343,10 @@ public struct Lexer: IteratorProtocol, Sequence {
     return sourceCode.text[start ..< index]
   }
 
-  /// Consumes a numeric literal and returns its kind, or returns `nil` if the stream starts with a
+  /// Consumes an integral literal and returns its kind, or returns `nil` if the stream starts with a
   /// different token.
-  private mutating func scanNumericLiteral() -> Token.Kind? {
-    let i = take("-") ?? index
+  private mutating func scanIntegralLiteral(_ allowPlus: Bool = false) -> Token.Kind? {
+    let i = (allowPlus ? take("+") : nil) ?? take("-") ?? index
     guard let head = peek(), head.isDecDigit else {
       index = i
       return nil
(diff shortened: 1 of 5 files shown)


Sources/FrontEnd/Parse/Lexer.swift
```diff
@@ -159,10 +165,25 @@ public struct Lexer: IteratorProtocol, Sequence {
...
     }
 
-    // Scan numeric literals.
-    if let k = scanNumericLiteral() {
+    // Scan integral literals.
+    if let k = scanIntegralLiteral() {
       token.kind = k
       token.site.extend(upTo: index)
+
+      if let nextChar = peek(), nextChar == "e" || nextChar == "E" {
```
kyouko-taiga, line 173:
```suggestion
      if let next = peek(), next == "e" || next == "E" {
```
There's no need to encode the type of the variable in its name.

Sources/FrontEnd/Parse/Lexer.swift
```diff
@@ -322,10 +343,10 @@ public struct Lexer: IteratorProtocol, Sequence {
     return sourceCode.text[start ..< index]
   }
 
-  /// Consumes a numeric literal and returns its kind, or returns `nil` if the stream starts with a
+  /// Consumes an integral literal and returns its kind, or returns `nil` if the stream starts with a
   /// different token.
-  private mutating func scanNumericLiteral() -> Token.Kind? {
-    let i = take("-") ?? index
+  private mutating func scanIntegralLiteral(_ allowPlus: Bool = false) -> Token.Kind? {
```
kyouko-taiga, line 348: Could you document the additional parameter?

Also I think it would deserve a label. It is not obvious what the Boolean argument means at call sites (e.g., `scanIntegerLiteral(true)` not as clear as `scanIntegerLiteral(allowingPlus: true)`).

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -1714,6 +1711,29 @@ public enum Parser {
     }
   }
 
+  private static func parseFloatLiteralExpr(in state: inout ParserState) throws -> FloatLiteralExpr
+    .ID?
+  {
+    guard let first = state.take(.int) else { return nil }
+
+    if let _ = state.take(if: { $0.kind == .dot }) {
```
kyouko-taiga, line 1719:
```suggestion
    if state.take(if: { $0.kind == .dot }) != nil {
```
This is easier to read.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -1714,6 +1711,29 @@ public enum Parser {
...
+      guard let _ = state.take(if: { $0.kind == .int }) else { return nil }
```
kyouko-taiga, line 1720:
```suggestion
      guard state.take(if: { $0.kind == .int }) != nil else { return nil }
```

kyouko-taiga (changes requested):
Very nice work!

I left a few inline comments. Most are nit pick except the one about the documentation of the additional parameter to `scanIntegerLiteral`.

esleche:
> Very nice work!
> 
> I left a few inline comments. Most are nit picks except the one about the documentation of the additional parameter to `scanIntegerLiteral`.

Thanks, it is my first time with Swift, so sorry about that.
Fixing the issues.

Sources/FrontEnd/Parse/Lexer.swift
esleche: Done

Sources/FrontEnd/Parse/Lexer.swift
```diff
@@ -322,10 +343,11 @@ public struct Lexer: IteratorProtocol, Sequence {
     return sourceCode.text[start ..< index]
   }
 
-  /// Consumes a numeric literal and returns its kind, or returns `nil` if the stream starts with a
-  /// different token.
-  private mutating func scanNumericLiteral() -> Token.Kind? {
-    let i = take("-") ?? index
+  /// Consumes an integral literal and returns its kind, or returns `nil` if it fails to scan a valid integer.
+  /// - Parameter allowingPlus: If set to `true`, allows the integral literal to begin with a "+" sign.
```
kyouko-taiga, line 347:
```suggestion
  ///
  /// - Parameter allowingPlus: If set to `true`, allows the integral literal to begin with a "+" sign.
```
Nice comment! I would not have guessed the behavior if `false`.

Sources/FrontEnd/Parse/Lexer.swift
```diff
@@ -17,9 +17,15 @@ public struct Lexer: IteratorProtocol, Sequence {
 
   /// The current location of the lexer in `sourceCode`.
   public var location: SourcePosition { sourceCode.position(index) }
+  private var pendingExponentToken: Token? = nil
```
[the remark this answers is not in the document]
kyouko-taiga: Ah, the problem is that you have to emit an exponent as a single token!

Hmm, then I think we should not have an `.exponent` token and have the lexer produce `.int, .dot, .int, .name, .operator, .int` for `1.123e+9` and arrange everything in the parser.

The alternative is to somehow preserve the property that `self.sourcePosition` is immediately after the last produced token. The parser relies on this invariant.

Sources/FrontEnd/Parse/Lexer.swift
```diff
@@ -17,6 +17,7 @@ public struct Lexer: IteratorProtocol, Sequence {
 
   /// The current location of the lexer in `sourceCode`.
   public var location: SourcePosition { sourceCode.position(index) }
+  private var previousTokenWasInt = false
```
kyouko-taiga, line 20:
```suggestion

  /// `true` iff the last scanned token was `.int`.
  private var previousTokenWasInt = false
```

kyouko-taiga (approved):
LGTM

PR #828, opened 2023-08-03
Update Swifty-LLVM and Swift Syntax packages

nickpdemarco:
The most salient change is [here](https://github.com/val-lang/val/pull/828/files#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR37-R38); the rest is the result of running `swift package update`. 

This change removes our dependency on our [fork of swift-format](https://github.com/val-lang/swift-format), which was put in place to reduce inconsistency between local dependency versions and those available on CI; namely `swift-format` and `swift-syntax`. The inconsistencies would produce build errors on CI which would not show up on local runs of `Tools/run-swift-format.sh -d lint`.

The forked repo attempted to lock the version of swift-format by [committing a `Package.resolved`](https://github.com/apple/swift-format/compare/main...val-lang:swift-format:main) file that specifies particular Git SHAs for its dependencies.

However, this does not seem to be effective. I'm not a SPM expert, but a quick look at Val's [`Package.resolved`](https://github.com/val-lang/val/blob/main/Package.resolved#L54) shows a different SHA for `swift-syntax` than the one [specified in the fork](https://github.com/apple/swift-format/compare/main...val-lang:swift-format:main#diff-3b22d20b6ae7cdf1e2591c647bcae94a3db9ab955f497aeb9d0715f468e6e6ceR18). If anyone with more SPM experience can explain, I'd be curious to know why this didn't work.

As an alternative, this PR simply version locks swift-format to its latest release (508.0.1) in `Package.swift` instead of pointing to `main`. This should achieve the desired effect of version locking swift-format more simply, ensuring consistency between local and CI builds, without the added complexity of a fork.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -51,7 +51,7 @@
       "location" : "https://github.com/apple/swift-syntax.git",
       "state" : {
         "branch" : "main",
-        "revision" : "8212815c39c1c66d6400f32f72280f06ef94af83"
+        "revision" : "250b431726c7b05309770caaa9989a4166ea6ba7"
       }
     },
     {
@@ -78,7 +78,7 @@
       "location" : "https://github.com/val-lang/Swifty-LLVM",
       "state" : {
         "branch" : "main",
-        "revision" : "6d3d421cdabcfd761bac91b650ccc839c41a6ee6"
+        "revision" : "4ba25d9615b6931033161ffcfc0a73a06809d781"
       }
     }
   ],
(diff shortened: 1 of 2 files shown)


lucteo (approved):
As the swift-format version is past the changes that Dave wanted to have for our repo, I think that just pointing to a fixed version in swift-format will work now (didn't work in the past).

kyouko-taiga (approved):

PR #834, opened 2023-08-07
Emit Intel Assembly

nickpdemarco:
To support getting Val onto Compiler Explorer, we need a way to emit Intel ASM. This PR adds a simple `--emit` option to the pipeline, updates documentation, and adds a smoke test consistent with other output modes.


kyouko-taiga (approved):
LGTM

PR #853, opened 2023-08-10
Fixing generic parameter shadowing

### LINKED ISSUE #847: Unable to build with Swift 5.9
Attempting to build with the following Swift version results in a build failure.

```
swift-driver version: 1.87.1 Apple Swift version 5.9 (swiftlang-5.9.0.128.2 clang-1500.0.40.1)
Target: x86_64-apple-macosx14.0
```

```
val/Sources/FrontEnd/TypeChecking/TypeChecker.swift:908:25: error: generic parameter 'T' shadows generic parameter from outer scope with the same name; this is an error in Swift 6
    func checkSatisfied<T: Decl>(callable requirement: T.ID, named requiredName: Name) {
```

This is caused by a generic method with a generic parameter name T containing nested generic functions that also use a generic parameter T. I can see that a new TypeChecker is in the works, but it appears that is also suffers from this same issue. Using a different generic parameter name for the nested functions should resolve the issue.

gsfreema:
This fixes #847 for the main branch. This change will ultimately be replaced by the new TypeChecker implementation, but it is a simple fix to remove a blocker for anyone wanting to check out the current state of Val using the latest beta Swift compiler.


gsfreema:
It should be noted that if this PR is merged, it will likely cause merge conflicts for the new-typechecker branch. However, it is unclear when that new work will be ready. If it is soon, then it might be best to merge #854 in the new-typechecker branch which will fix the main branch once that work is completed. But if that work will be a while, it might be worth the merge conflict so that the main branch looks more stable for newcomers with the latest Swift compiler.

kyouko-taiga:
Sorry I missed this PR and merged the new type checker already.

kyouko-taiga (approved):
LGTM

PR #859, opened 2023-08-10
Update README to use the new name where possible

Zachiah:
## Notes
- ~~I made the repository links go to `hylo-lang/hylo` as I figured the repo will get renamed very shortly. Let me know if this is incorrect~~
- On line 38 (now 41) It said "That's Val compiler!" this seems grammatically incorrect to me so I changed it to "That's Hylo's compiler!" it could also be "That's the Hylo compiler!" if you prefer.
- I added a warning about broken links and confusion to the top of the README. Let me know if this is OK or if I should remove it
- I just renamed the site in the email in `CODE_OF_CONDUCT.md`. Please verify this is correct

### EDIT
The repo was renamed while I was typing. I'll fix it right quick
- I just also fixed CONTRIBUTING.md and CODE_OF_CONDUCT.md since they seemed straight forward

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,9 +1,9 @@
 # Code of Conduct
 
-The Val community works to foster a positive environment for individuals from all walks of life and commits to treating each and everyone with kindness and respect, regardless of age, gender identity and expression, sexual identity and orientation, ethnicity, race, nationality, religion, socio-economic status, disability or level of experience.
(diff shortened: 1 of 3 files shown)


Zachiah:
Let me know if I should fix `CodeFormatting.md`, `.spi.yml`, `Package.resolved`, `Package.swift`, and/or `SharedMutable.swift`. They all have `val-lang`. I was a bit hesitant to change things that could potentially break things which is why I ask.

kyouko-taiga:
I guess CodeFormatting.md can be changed, but other files shouldn't yet. Thanks!

kyouko-taiga (approved):
Nice, thanks a lot!

Issue #857, opened 2023-08-10
Test optimized builds


kyouko-taiga:
Currently, end-to-end tests only run debug builds produced by the compiler. That means we never test the `-O` option. We should run all end-to-end tests on optimized builds in addition to the debug ones.

PR #864, opened 2023-08-11
Test optimized builds

C-BJ:
Fixes: #857


C-BJ:
The test failed, but I don't know what I did wrong.
Also, do we need to extract the tests of test optimized builds, but if so, I'm not sure how the comments at the beginning of `TestGeneratorPlugin.swift` and the test file are going to be changed.
@kyouko-taiga

C-BJ:
Optimised `factorial.hylo` built locally on Windows will not work properly

C-BJ:
@kyouko-taiga

kyouko-taiga:
Thanks for taking on this issue. It has revealed like some of the attributes we're adding on LLVM parameters cause programs to fail in optimized builds. We should investigate.

As for the generating new tests, perhaps @dabrahams can give pointers.

dabrahams:
> I don't know how to write the new test in a separate function

I don't know what you mean by this exactly.  We have lots of single function tests, like [this one](https://github.com/hylo-lang/hyloc/blob/34ddc9e984a26c4937194bc652b869808fbe1dee/Tests/HyloTests/ParserTests.swift#L43) for example.

C-BJ:
`TestGeneratorPlugin.swift` generates test files for the end to end test. But I don't know how to use it for the new test.
https://github.com/hylo-lang/hyloc/commit/4aebc15047eedbd21a82c3abf82006f3b212f682 should explain what I'm talking about. I tried what I wanted to do in this commit.

C-BJ:
@dabrahams

PR #901, opened 2023-08-18
Fix lexer issue with 'e' following integers across lines

esleche:
Fixes https://github.com/hylo-lang/hyloc/issues/873

### DIFF
--- a/Sources/FrontEnd/Parse/Lexer.swift
+++ b/Sources/FrontEnd/Parse/Lexer.swift
@@ -27,15 +27,9 @@ public struct Lexer: IteratorProtocol, Sequence {
     while true {
       if index == sourceCode.text.endIndex { return nil }
 
-      // Skip newlines.
-      if sourceCode.text[index].isNewline {
-        previousTokenWasInt = false
-        discard()
-        continue
-      }
-
       // Skip whitespaces.
       if sourceCode.text[index].isWhitespace {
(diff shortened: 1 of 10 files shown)


Sources/FrontEnd/Parse/Lexer.swift
```diff
@@ -27,6 +27,13 @@ public struct Lexer: IteratorProtocol, Sequence {
     while true {
       if index == sourceCode.text.endIndex { return nil }
 
+      // Skip newlines.
+      if sourceCode.text[index].isNewline {
+        previousTokenWasInt = false
```
kyouko-taiga, line 32: It's a little suspicious that you can unset `preciousTokenWasInt` here but not after any other whitespace. Why is that?

kyouko-taiga (changes requested):
It's a bit overkill to create end-to-end tests only to check that the lexer is doing its work properly. It would be better to write these tests as parser tests (see `HyloTests/TestCases/Parsing`) or, if type checking is needed, as type checking tests (see `HyloTests/TestCases/TypeChecking`).

Sources/FrontEnd/Parse/Lexer.swift
esleche: Done 👍

kyouko-taiga (approved):
Nice changes, thanks!
LGTM.

PR #919, opened 2023-08-22
Use cacheing for llvm installation action

dabrahams:


dabrahams:
Dead end: cacheing slows things down on a Mac runner, and we do need setup-swift.

Issue #805, opened 2023-07-28
Enable testing on Windows


C-BJ:
We are currently blocked by:

```
error: couldn't build D:\a\val\val\.build\x86_64-unknown-windows-msvc\release\ValTests.build\ASTTests.swift.o 
because of missing inputs: D:\a\val\val\.build\plugins\outputs\val\ValTests\TestGeneratorPlugin\ValFileTests.swift
```

Maybe it has something to do with [this](https://github.com/apple/swift-package-manager/issues/6595) SPM bug.

dabrahams:
It actually is related to [this bug](https://github.com/apple/swift-package-manager/issues/6859).

PR #926, opened 2023-08-23
Solve the "Swift won't show the full path" problem…

dabrahams:
…once and for all.


dabrahams:
This PR is currently blocked by [this SPM bug](https://github.com/apple/swift-package-manager/issues/6859), which is currently also blocking us from fixing #805.

dabrahams:
Superseded by #928

PR #939, opened 2023-08-26
Pointers

dabrahams:
There's a fair amount to discuss here; feedback appreciated.  I did my own mini-review to prompt discussion.


Library/Hylo/Core/Bitcast.hylo
```diff
@@ -1,11 +1,11 @@
 /// Projects `value` with its memory representation reinterpreted as a value of type `U`.
 public subscript unsafe_bitcast<T, U>(_ value: T): U {
   let {
-    sink let p = Pointer<T>.to[value].copy().value
-    yield p as* (remote let U)
+    sink let p: Pointer<U> = Pointer(type_punning: pointer[to: value])
```
dabrahams, line 4: I don't have better ideas for an argument label but am open to suggestions.

Library/Hylo/Core/Bitcast.hylo
```diff
@@ -1,11 +1,11 @@
 /// Projects `value` with its memory representation reinterpreted as a value of type `U`.
```
dabrahams, line 1: It occurred to me that this is missing preconditions, but thinking about them made me question its implementation.
The preconditions implied by this implementation are that the argument is suitably aligned for a value of type `U` and that there is an instance of `U` at the argument's address.  But I don't think that's what we want for `unsafe_bitcast` at all, and I suspect the implementation needs to use IR primitives that don't imply looking at the argument's memory as though it was a `U`,

Library/Hylo/Core/Bitcast.hylo
```diff
@@ -1,11 +1,11 @@
...
 public subscript unsafe_bitcast<T, U>(_ value: T): U {
```
dabrahams, line 2: It occurs to me that we shouldn't use the name `value` for anything; it's certainly information-free in this context.  Not making that change preemptively because it will cause some of my other comments to become outdated.

Library/Hylo/Core/CVoidPointer.hylo
```diff
@@ -0,0 +1,43 @@
+// This type only exists because we don't seem to be able to make generic types like Pointer ForeignConvertible.
```
dabrahams, line 1: This is approximately exactly what you called `RawPointer`, without any extra trappings that make it like the real pointer types.  It's merely there to be a vehicle for FFI.

By the comment I mean to say C `void*` should really map to `PointerToMutable<Never>`, etc., but if I try to use generic types in FFI signatures, I get link errors, presumably because generic functions are not being monomorphized into externally accessible symbols(?)

Library/Hylo/Core/MutablePointer.hylo
dabrahams, line 1: That type's new name is `PointerToMutable`.

Library/Hylo/Core/Pointer.hylo
```diff
@@ -1,43 +1,57 @@
...
+  /// - Requires: the address stores an object of type Pointee and nothing modifies
+  ///   that storage during the lifetime of this projection.
+  public subscript unsafe(): Pointee {
+    yield base as* (remote let Pointee)
   }
 
-  /// Creates a copy of `other`.
-  public init(_ other: MutableRawPointer) {
-    &self.value = other.value
+  /// Creates an instance referring to the same address as `p`.
```
dabrahams, line 17:
```suggestion
  /// Creates an instance representing the same address as `p`.
```

Library/Hylo/Core/Pointer.hylo
```diff
@@ -1,43 +1,57 @@
...
-  var value: Builtin.ptr
+  /// The raw bits of the address.
+  var base: Builtin.ptr
 
   memberwise init
 
-  /// Creates a copy of `other`.
-  public init(_ other: RawPointer) {
-    &self.value = other.value
+  /// The value at the given address.
```
dabrahams, line 9:
```suggestion
  /// The `Pointer` at the address.
```

Library/Hylo/Core/Pointer.hylo
```diff
@@ -1,43 +1,57 @@
...
+  /// `usafe[]` subscript (which see) are followed.
+  public init<Other>(type_punning p: Pointer<Other>) {
+    &base = p.base
   }
 
-  /// Projects a pointer to `pointee`.
-  public static subscript to(_ pointee: Pointee): Self {
-    Pointer(value: Builtin.address(of: pointee))
+  /// Creates an instance that does not address any usable storage.
+  public static fun null() -> Self {
```
dabrahams, line 31: Considering making this a static `let`:
today, we have to call `.null()` with parentheses, but we get an independent value.  If we make it a `let`, initializing a variable to `null` will require a `.copy()`.  Huh, that change would only be worth it if people make lots more comparisons against null than they make initializations to null.  Do they?  If we had some equivalent of Swift's `NilLiteralConvertible` that would be the ultimate sugar.  We could also relieve the pressure of that `copy()` by making `PointerToMutable()` a way to create null. Interesting set of tradeoffs.

Library/Hylo/Core/Pointer.hylo
```diff
@@ -1,43 +1,57 @@
-/// A pointer for accessing typed data.
-public type Pointer<Pointee> {
+/// A typed memory address whose contents can be read.
+public type Pointer<Pointee>: Regular {
 
-  var value: Builtin.ptr
+  /// The raw bits of the address.
+  var base: Builtin.ptr
```
dabrahams, line 5: I don't think `value` was a good name for this property, but I'm not all that thrilled with `base`. 
`self.bits`
`self.storage`
`self.lowered`
all come to mind, with the last one being the most descriptive.  I would probably access it with qualification always to keep it readable if we chose `lowered`.

Library/Hylo/Core/PointerToMutable.hylo
```diff
@@ -0,0 +1,74 @@
...
+
+  /// The raw bits of the address.
+  var base: Builtin.ptr
+
+  memberwise init
+
+  /// The value at the given address.
+  ///
+  /// - Requires: the address stores a mutable object of type Pointee and nothing accesses that
+  ///   storage except through this projection during its lifetime.
```
dabrahams, line 12: This is a bit too strict for the `let` accessor.  I should update it to be clear that only applies to the `inout` accessor and that another requirement applies to `let`.  IMO we should have a documentation convention that supports separate documentation for clauses in addition to overall documentation of the subscript, and will try using that for my update.

Library/Hylo/Core/PointerToMutable.hylo
```diff
@@ -0,0 +1,74 @@
...
+public type PointerToMutable<Pointee>: Regular {
+
+  /// The raw bits of the address.
+  var base: Builtin.ptr
+
+  memberwise init
+
+  /// The value at the given address.
+  ///
+  /// - Requires: the address stores a mutable object of type Pointee and nothing accesses that
```
dabrahams, line 11:
```suggestion
  /// - Requires: the address stores a mutable object of type `Pointee` and nothing accesses that
```

Library/Hylo/Core/PointerToMutable.hylo
```diff
@@ -0,0 +1,74 @@
...
+  /// - Warning: while this initializer is not in itself unsafe, it should be used with caution.
+  public init(adding_mutation_to p: Pointer<Pointee>) {
+    &base = p.base
+  }
+
+  /// Creates an instance referring to the same address as `p`.
+  ///
+  /// Note that dereferencing type punned pointers will cause problems unless the rules for the
+  /// `usafe[]` subscript (which see) are followed.
+  public init<Other>(type_punning p: PointerToMutable<Other>) {
```
dabrahams, line 29: Is "type_punning an address" a thing?  If not, a different label is probably appropriate. `same_address_as`, though, I fear is too innocuous-sounding for the danger surrounding this operation.

Library/Hylo/Core/PointerToMutable.hylo
```diff
@@ -0,0 +1,74 @@
...
+    &base = p.base
+  }
+
+  /// Creates an instance that does not address any usable storage.
```
dabrahams, line 33: Hmm, it occurs to me that there's something to document about the representational-sameness of `Pointer` and `PointerToMutable`.

Library/Hylo/Core/PointerToMutable.hylo
```diff
@@ -0,0 +1,74 @@
...
+  public static fun null() -> Self {
+    .new(base: Builtin.zeroinitializer_ptr())
+  }
+
+  /// Returns the result of applying `initialize` to the uninitialized `Pointee` storage.
+  ///
+  /// - Requires: the `MemoryLayout<Pointee>.size()` bytes starting at the address are uninitialized
+  ///   and suitably aligned for `Pointee`.
+  public fun unsafe_initialize_pointee<E, R>(_ initialize: inout [E](set Pointee) -> R) -> R {
```
dabrahams, line 42: I was surprised to realize that something like this is still needed, even though we have subscripts.  It's possible that's only because we don't have `new_value` yet and so can't implement a `set` subscript, but I think this operation is unusual enough that it probably shouldn't be spelled similarly to pointee access.

Library/Hylo/Core/RawPointer.hylo
dabrahams, line 1: See `CVoidPointer`

Library/Hylo/Core/Regular.hylo
```diff
@@ -0,0 +1,11 @@
...
+/// (roughly per Stepanov).
+///
+/// - Move is value-preserving.
+/// - Destruction has no side-effects.
+public trait SemiRegular: Deinitializable, Movable, Equatable {}
+
+/// Regular types (roughly per Stepanov).
+///
+/// Copies have equal value.
+public trait Regular: Deinitializable, Copyable, Movable, Equatable {}
```
dabrahams, line 11: Having this be a nominal trait rather than an alias for the trait composition (`Deinitializable & Copyable & Movable & Equatable` in Swift) seems wrong.  Can we express the latter today?

Library/Hylo/DynamicBuffer.hylo
```diff
@@ -1,6 +1,6 @@
 /// A buffer of elements stored in a contiguous region of memory, whose size is determined at
 /// instance creation.
-public type DynamicBuffer<Header, Element>: Deinitializable {
+public type DynamicBuffer<Header: Deinitializable, Element>: Deinitializable {
```
dabrahams, line 3: I couldn't get it to compile without this.  And besides, we do want to deinitialize the header, don't we?

Library/Hylo/Pointers+DynamicAllocation.hylo
```diff
@@ -1,29 +1,22 @@
...
 
   /// Allocates memory for `count` bytes at given `alignment`.
-  public static fun allocate(count: Int, aligned_at alignment: Int) -> Self {
-    .new(mutating: aligned_alloc(alignment, count))
-  }
-
-  /// Deallocates the memory previously allocated at `self`.
-  public fun deallocate() {
-    free(RawPointer(self))
+  public static fun allocate_bytes(count: Int, aligned_at alignment: Int) -> Self {
```
dabrahams, line 4: It's very unclear to me that these are best represented as `static` functions and I'd like to think about alternatives a bit.

Library/Hylo/Print.hylo
```diff
@@ -1,11 +1,11 @@
 /// Writes the textual representation of `item` to the standard output.
 public fun print(_ item: String, terminator: String = "\n") {
   let stream = stdout()
-  _ = fwrite(item.utf8, 1, item.size, stream)
-  _ = fwrite(terminator.utf8, 1, 1, stream)
+  _ = fwrite(CVoidPointer(base: item.utf8.base), 1, item.size, CVoidPointer(base: stream.base))
+  _ = fwrite(CVoidPointer(base: terminator.utf8.base), 1, 1, CVoidPointer(base: stream.base))
```
dabrahams, line 5: Could make this a lot less painful by projecting a `CVoidPointer` out of these other pointers via a property.

Library/Hylo/Core/CVoidPointer.hylo
kyouko-taiga: > but if I try to use generic types in FFI signatures, I get link errors, presumably because generic functions are not being monomorphized into externally accessible symbols(?)

Possible. I should investigate for a definitive answer.

Library/Hylo/Core/Comparable.hylo
```diff
@@ -0,0 +1,20 @@
+/// A type whose instances' values have a standard total ordering
+///
+/// Two equal values `a` and `b` are unordered, i.e. `a < b` and `b < a` are both false.
```
kyouko-taiga, line 3: Nit: I think we'll need to add the full properties to be upheld in the trait description.

Library/Hylo/Core/Equatable.hylo
```diff
@@ -0,0 +1,17 @@
...
+public trait Equatable {
+
+  /// Returns `true` iff `other` has an equivalent value.
+  fun infix== (_ other: Self) -> Bool
+
+}
+
+extension Equatable {
+
+  /// Returns `false` iff `other` has an equivalent value.
```
kyouko-taiga, line 14:
```suggestion
  /// Returns `false` iff `other` has not an equivalent value.
```

Library/Hylo/Core/MemoryAddress.hylo
```diff
@@ -0,0 +1,13 @@
+// Can't really use this for anything yet due to https://github.com/hylo-lang/hylo/issues/936
+/// A memory address.
+typealias MemoryAddress = PointerToMutable<Never>
+
+public extension PointerToMutable where Pointee == Never {
+
+  /// Creates an instance pointing to the same address as `p`
```
kyouko-taiga, line 7:
```suggestion
  /// Creates an instance pointing to the same address as `p`.
```

Library/Hylo/Core/MemoryAddress.hylo
```diff
@@ -0,0 +1,13 @@
...
+  public init<T>(_ p: PointerToMutable<T>) { &self.base = p.base }
+
+  /// Creates an instance pointing to the same address as `p`
```
kyouko-taiga, line 10:
```suggestion
  /// Creates an instance pointing to the same address as `p`.
```

Library/Hylo/Core/MemoryAddress.hylo
kyouko-taiga: A question occurred to me. Is a pointer pointing to an address or a value? I could imagine describing a pointer as _being_ an address.

Library/Hylo/Core/MutablePointer.hylo
kyouko-taiga: Which is an awesome name change!

Library/Hylo/Core/Pointer.hylo
kyouko-taiga: I think you wrote `raw` somewhere, which I liked. Otherwise I would go for `lowered`.

Library/Hylo/Core/Pointer.hylo
kyouko-taiga: Is "at the address" still necessary then?

Library/Hylo/Core/Pointer.hylo
```diff
@@ -1,43 +1,57 @@
...
-  }
-
-  /// Returns the result of applying `action` to a projection of the value referenced by `self`.
-  public fun with_pointee<T>(_ action: [](Pointee) -> T) -> T {
-    action(value as* (remote let Pointee))
+  /// Creates an instance referring to the same address as `p`.
+  ///
+  /// Note that dereferencing type punned pointers will cause problems unless the rules for the
+  /// `usafe[]` subscript (which see) are followed.
+  public init<Other>(type_punning p: Pointer<Other>) {
```
kyouko-taiga, line 26:
```suggestion
  public init<Other>(punning p: Pointer<Other>) {
```
Are there other forms of punning we should care about?

Library/Hylo/Core/Pointer.hylo
kyouko-taiga: There is a very interesting question to answer about this kind of pattern. Note that a similar situation will occur with `AdditiveArithmetic.zero`.

We can certainly implement `ExpressibleByNilLiteral` (which I think is the modern name for the protocol you mentioned). The support for `ExpressibleBy***Literal` is already implemented in the type checker, so it should not be an insurmountable task.

Library/Hylo/Core/Pointer.hylo
```diff
@@ -1,43 +1,57 @@
...
+    Pointer(base: base)
+  }
+
+}
+
+public conformance Pointer: Equatable {
+
+  /// Returns `true` iff `other` has an equivalent value.
+  public fun infix== (_ other: Self) -> Bool {
+    Bool(value: Builtin.icmp_eq_word(Builtin.ptrtoint_word(base), Builtin.ptrtoint_word(other.base)))
```
kyouko-taiga, line 49:
```suggestion
  Bool(value: Builtin.icmp_eq_ptr(base, other.base))
```

Library/Hylo/Core/Pointer.hylo
```diff
@@ -1,43 +1,57 @@
...
   }
 
 }
+
+/// The address of `x`.
```
kyouko-taiga, line 54: We should probably mention that using a copy of the projection after the projection ends is UB.

Library/Hylo/Core/PointerToMutable.hylo
kyouko-taiga: What if we document the variants?
```
/// Projects a part of `x`.
public subscript p(x: T): Int {
  let { ... }

  /// The projection is greater than zero when it ends.
  inout { ... }
}
```

Library/Hylo/Core/PointerToMutable.hylo
kyouko-taiga: The problem is that we can't use mutating subscripts on non-mutating values. Using a `set` variant would require mutable access to the pointer itself because in the common case, initializing a part does mutate the whole.

Library/Hylo/Core/Regular.hylo
kyouko-taiga: We can't, at least currently.

Swift has some fancy semantics for aliases of protocols because they can be used either as types or protocols depending on the context. I bet that's because we can use a protocol as a type unless it has associated types. In Hylo, traits are never types. We always need to use existential types (i.e., `any P`). So I think it would be inconsistent to be able to write `typealias R = P & Q`.

Perhaps we should consider introducing `traitalias R = ...` or `trait alias R = ...` (note: I am contemplating the idea of using `type alias` rather than `typealias`).

Library/Hylo/DynamicBuffer.hylo
kyouko-taiga: Oh yes, we do.

Library/Hylo/Core/Bitcast.hylo
dabrahams: @kyouko-taiga I was really hoping to discuss [this](https://github.com/hylo-lang/hylo/pull/939#discussion_r1310959046) with you.

Library/Hylo/Core/Comparable.hylo
dabrahams: Saying `<` is a total ordering in addition to what's there should be sufficient, right?

Library/Hylo/Core/Equatable.hylo
dabrahams: I'm pretty sure that would be wrong.  This is `!=`.

Library/Hylo/Core/MemoryAddress.hylo
dabrahams: Agreed, the pointer is an address.  But it may not point to a value, so I think
```suggestion
  /// Creates an instance representing the same address as `p`.
```

Library/Hylo/Core/Comparable.hylo
dabrahams, line 3:
```suggestion
/// `<` is a total ordering; `a == b` implies that `a < b` and `b < a` are both false.
```

Library/Hylo/Core/MemoryAddress.hylo
dabrahams:
```suggestion
  /// Creates an instance representing the same address as `p`.
```

Library/Hylo/Core/Pointer.hylo
dabrahams: I think so, because `Pointee` by itself in “code voice” only specifies a type and not its relationship to the “pointer”

Library/Hylo/Core/Pointer.hylo
```diff
@@ -1,43 +1,57 @@
...
 
   memberwise init
 
-  /// Creates a copy of `other`.
-  public init(_ other: RawPointer) {
-    &self.value = other.value
+  /// The value at the given address.
+  ///
+  /// - Requires: the address stores an object of type Pointee and nothing modifies
+  ///   that storage during the lifetime of this projection.
```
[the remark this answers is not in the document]
dabrahams: OK, but it's not just storage _for_ an object of type `Pointee`; there actually is an object there.  So I'd say “of” instead.

dabrahams, line 12:
```suggestion
  /// - Requires: `self` is the address of an object of type `Pointee` and its storage
  ///   is accessed only through this projection during the projection's lifetime.
```

Library/Hylo/Core/Pointer.hylo
dabrahams: Not that I know of, but I think it's not a universally understood term.  "Type punning" seems a bit more suggestive, and after all, I don't think we need to make this case too terse.

Library/Hylo/Core/Pointer.hylo
dabrahams: I think I will make this a `let` for now, and start eliminating needless `copy()`s (hopefully with the compiler's help!) once we have the `set` accessor support we discussed.

Library/Hylo/Core/Pointer.hylo
dabrahams: It isn't!  Dereferencing it _might_ be UB, if the lifetime of `x` has ended, but that's covered in the precondition of `subscript unsafe()`.

Library/Hylo/Core/PointerToMutable.hylo
dabrahams: That's exactly the documentation convention I meant.

Library/Hylo/Core/PointerToMutable.hylo
dabrahams: The lack of `nonmutating inout` is not a big problem; we already use `p.copy()` to get around it.

Library/Hylo/Core/Regular.hylo
dabrahams: The only downside I can see with `type alias` and `trait alias` is that the newly declared name will always be followed by `=` while for non-aliases it will never be `=`.  That kinda suggests distinct keywords would be better because `alias` is not just an adjectival modifier for `type` or `trait`, but actually switches us to a very different form of declaration.  These points look pretty subtle and aesthetic to me, so I don't feel strongly either way…

OK, thinking more about that, I think I want single keywords if only because all the other introducers are single keywords.

Library/Hylo/DynamicBuffer.hylo
dabrahams: You might want to ask yourself why the old code compiled but the new code doesn't.  Bug?

Library/Hylo/Core/Pointer.hylo
dabrahams: Looks like #953 and #954 are blocking this change.

Library/Hylo/Core/PointerToMutable.hylo
dabrahams: That can wait, I think.  Don't give away the store.

Library/Hylo/Core/Bitcast.hylo
kyouko-taiga: You're right about the alignment. There's also a size requirement. `T` be at least as large as `U`.

There isn't necessary an instance of `U` at the argument's address. The subscript is just allowing you to reinterpret the bits at this address of the value starting _as though_ there was one. So you should be able to write this once buffers are supported:

```
public fun main() {
  var x = [1, 2, 3]
  inout y: Int = unsafe_bitcast[&x]
  &x += 2
  print(x) // [3, 2, 3]
}
```

So another precondition is that the bit representation of the argument is a valid representation for an instance of type `U`.

> But I don't think that's what we want for `unsafe_bitcast` at all, and I suspect the implementation needs to use IR primitives that don't imply looking at the argument's memory as though it was a `U`.

I didn't understand this comment. What would we want for `unsafe_bitcast` in your opinion?

The implementation is indeed looking at the argument's memory as though it was a `U`. We're creating a pointer to `T`, casting it to a pointer to `U`, and dereferencing the value of that pointer. I _think_ in C++ the semantics would be:

```c++
template<typename T, typename U>
U& unsafe_bitcast(T& value) {
  auto* p = &value;
  auto* q = static_cast<U*>(static_cast<void*>(p));
  return *q;
}
```

The primitives used to achieve that in Hylo IR are `address_to_pointer` (which is no-op in LLVM since the argument is in fact already a pointer) and `pointer_to_address` (also no-op since LLVM pointers are opaque). The copy of the pointer is done with load/store that I'm 99% sure should get optimized away.

Library/Hylo/Core/Equatable.hylo
kyouko-taiga: Oh, I missed that the sentence starts with "Returns `false`".

IMO this is confusing. I think we should rather saying when something returns `true` consistently because this is how most predicates and Boolean properties are typically documented.

Library/Hylo/Core/PointerToMutable.hylo
kyouko-taiga: I'm sorry, I didn't understand your comment and the double negation in your clarification confused me further. What is it that you would like to write ideally?

Library/Hylo/Core/Regular.hylo
kyouko-taiga: `memberwise init`

Library/Hylo/DynamicBuffer.hylo
kyouko-taiga: Not necessarily because the old code never deinitialized the header. Could you write a reproducer showing the compilation error without the trait bound?

Library/Hylo/Pointers+DynamicAllocation.hylo
kyouko-taiga: The only alternative seems to be an initializer. Did you have other solutions in mind?

Library/Hylo/Core/Comparable.hylo
kyouko-taiga: Technically yes, but I'm thinking that these very fundamental traits might want to be a little more educational in their documentation. Being a little verbose about what a total ordering and other similar "well-established" properties might go a long way to make the language feel more inclusive to people with a light or dusty mathematical background.

Library/Hylo/Core/Regular.hylo
kyouko-taiga: But I don't feel strongly either way so if you want single keywords, we'll have single keywords.

Library/Hylo/Core/Comparable.hylo
dabrahams: Oh, absolutely, the docs should contain wikipedia links if nothing else.

Library/Hylo/Pointers+DynamicAllocation.hylo
dabrahams: free functions are also possible, not that I'm advocating for them.

Library/Hylo/DynamicBuffer.hylo
dabrahams: It would just be a snapshot of this branch with the trait bound removed.  You want that?

Library/Hylo/Core/Regular.hylo
dabrahams: Hm, I think of `memberwise` as a modifier here.  It's still just introducing the same kind of initializer, just asking for synthesis.  `trait alias` would not declare a `trait`.

Library/Hylo/Core/PointerToMutable.hylo
dabrahams: I'm not necessarily saying I know what the ideal is.  All I'm saying is, I originally thought I was going to replace all the explicit inversion of control in your pointers with projections, but having worked on the code I no longer think that replacement would be appropriate for this initialization case.

Library/Hylo/DynamicBuffer.hylo
kyouko-taiga: No, I think I found a bug but it is now hidden by another one. I'll take care of opening the right issues.

PR #945, opened 2023-08-31
Spell check GitHub action

WalterSmuts:
This *shouldn't* get in the way of development too much. It's [designed](https://github.com/crate-ci/typos/blob/master/docs/design.md) to work with code and have a low false-positive count. Feel free to drop the `Add github spell check workflow` commit if you recon it'll be more hassle than  worth.

### DIFF
--- a/.github/workflows/spell-check.yml
+++ b/.github/workflows/spell-check.yml
@@ -1,0 +1,13 @@
+name: Spell check
+on: [pull_request, push]
+
+jobs:
+  run:
+    name: Spell Check using typos
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Actions Repository
+      uses: actions/checkout@v3
+
+    - name: Check spelling of file.txt
+      uses: crate-ci/typos@v1.0.4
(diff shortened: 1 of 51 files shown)


dabrahams:
This is awesome.  My only question: what's the procedure for silencing a false positive?

WalterSmuts:
> My only question: what's the procedure for silencing a false positive?

Add it to the `[default.extend-words]` in the `.typos.toml` file. There's already two defined:
```
inout = "inout" # Mutable projection keyword
olt = "olt" # Abbreviation for "ordered less than"
```

The list is defined as a map of valid corrections. TBH not too sure why it's a correction in this case. Likely related to the tool working off of correction lists and not valid lists. See [the docs on false positives](https://github.com/crate-ci/typos#false-positives) for more info.

dabrahams:
If you could add something to [this text](https://github.com/hylo-lang/hylo/blob/main/CONTRIBUTING.md#contributing-code) describing how contributors can spell-check their code locally, I think we can accept this PR with gratitude!

WalterSmuts:
Think that should suffice :+1:

PR #956, opened 2023-09-04
Add conditional to publishing actions

WalterSmuts:
Contributors may want to run actions on their own forks of hylo. The GitHub pages publishing action should always be excluded in such a run.


dabrahams:
Afraid this [broke the build](https://github.com/hylo-lang/hylo/actions/runs/6115035602/job/16597821940).  Care to try again?

dabrahams:
Actually, you didn't break anything; I reapplied your patch and fixed the real problem.

PR #987, opened 2023-09-11
Richer bool interface

nickpdemarco:
Emulating what I can from [swift/core/bool](https://developer.apple.com/documentation/swift/bool).


Library/Hylo/Core/Bool.hylo
```diff
@@ -24,6 +19,40 @@ public type Bool {
...
+
+  /// Performs a logical OR on two boolean values.
+  public fun infix|| <E> (_ rhs: [E]() -> Bool) -> Bool {
+    if self { true } else { rhs() }
+  }
+
+  /// A textual representation of the Boolean value.
+  property description: String {
+    if self { "true" } else { "false" }
+  }
```
kyouko-taiga, line 55: IMO this should be a method.

Library/Hylo/Core/Bool.hylo
```diff
@@ -24,6 +19,40 @@ public type Bool {
...
+
+  public fun infix|| (_ rhs: Bool) -> Bool {
+    Bool(value: Builtin.or_i1(self.value, rhs.value))
+  }
+
+  /// Performs a logical AND on two boolean values.
+  public fun infix&& <E> (_ rhs: [E]() -> Bool) -> Bool {
+    if self { rhs() } else { false }
+  }
```
kyouko-taiga, line 45: We can't do better to implement short-cut behavior for now because we didn't implement `@autoclosure` (or similar), but this signature is implying that one will have to write a lambda to use `infix&&`:

```
public fun main() {
  if foo && fun() { bar() } {
    ...
  }
}
```

Perhaps we should renounce to short-cut behavior until `@autoclosure` is implemented?

nickpdemarco: I see your point though - I'm happy to remove the closure overloads until they make more sense for the language.

kyouko-taiga (approved):
LGTM

PR #990, opened 2023-09-11
Support aliases of specialized generics

kyouko-taiga:


dabrahams (commented):
Not: that's not a generic type alias. That's a type, alias for a specialization of a generic.

Issue #989, opened 2023-09-11
Fix overload resolution


camio:
Calling advance with a `PointerToMutable<Never>` should not result in an ambiguity error, but it currently does.

```
public type PointerToMutable<Pointee>: Regular {
  public fun advance( by offset_in_elements: Int ) -> Self {
    //...
  }
}

public extension PointerToMutable where Pointee == Never {
  public fun advance( by offset_in_elements: Int ) -> Self {
    //...
  }
}
```

PR #991, opened 2023-09-11
Fix Builtin.advanced_by_bytes and add PointerToMutable.advance

camio:
- Added missing case for .advancedByBytes in transpilation. This code
  was based on similar code in this file.
- Added PointerToMutable<Never>.advance specialized function. This
  needed to be specialized because of issue #985. The general version
  wasn't added because of issue #989.

This code is required to enable `DynamicBuffer` to provide a pointer
to the beginning of its elements. We need this to ultimately implement
`Array`.

### DIFF
--- a/Library/Hylo/Pointers+DynamicAllocation.hylo
+++ b/Library/Hylo/Pointers+DynamicAllocation.hylo
@@ -7,6 +7,17 @@ public extension PointerToMutable where Pointee == Never {
 
 }
 
+public extension PointerToMutable where Pointee == Never {
+
+  /// Returns a pointer advanced by `offset_in_elements`, which is 1 for `Never`.
+  public fun advance( by offset_in_elements: Int ) -> Self {
+    let offset_in_bytes = 1 * offset_in_elements
+    return PointerToMutable<Pointee>.new(
+      base: Builtin.advanced_by_bytes_word( base, offset_in_bytes.foreign_value() ) )
+  }
+
+}
+
 public extension PointerToMutable {
 
   /// Allocates memory for `count` instances of `Pointee`.
--- a/Library/Hylo/Pointers+DynamicAllocation.hylo
+++ b/Library/Hylo/Pointers+DynamicAllocation.hylo
@@ -13,7 +13,7 @@ public extension PointerToMutable where Pointee == Never {
   public fun advance(by offset_in_elements: Int) -> Self {
     let offset_in_bytes = 1 * offset_in_elements
     return PointerToMutable<Pointee>.new(
-      base: Builtin.advanced_by_bytes_word(base, offset_in_bytes.foreign_value()))
+      base: Builtin.advanced_by_bytes_word(base, offset_in_bytes.value))
   }
 
 }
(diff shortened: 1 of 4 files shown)


Library/Hylo/Pointers+DynamicAllocation.hylo
```diff
@@ -7,6 +7,17 @@ public extension PointerToMutable where Pointee == Never {
 
 }
 
+public extension PointerToMutable where Pointee == Never {
+
+  /// Returns a pointer advanced by `offset_in_elements`, which is 1 for `Never`.
+  public fun advance( by offset_in_elements: Int ) -> Self {
```
kyouko-taiga, line 13:
```suggestion
  public fun advance(by offset_in_elements: Int ) -> Self {
```

Library/Hylo/Pointers+DynamicAllocation.hylo
```diff
@@ -7,6 +7,17 @@ public extension PointerToMutable where Pointee == Never {
...
+    let offset_in_bytes = 1 * offset_in_elements
+    return PointerToMutable<Pointee>.new(
+      base: Builtin.advanced_by_bytes_word( base, offset_in_bytes.foreign_value() ) )
```
kyouko-taiga, line 16:
```suggestion
      base: Builtin.advanced_by_bytes_word(base, offset_in_bytes.foreign_value()))
```

I think it would be better to use the value stored in `offset_in_bytes` directly rather than calling `foreign_value()`. `ForeignConvertible` is a trait meant for FFI conversion and we should probably not use it outside of this use case.

Sources/CodeGen/LLVM/Transpilation.swift
```diff
@@ -899,6 +899,15 @@ extension LLVM.Module {
       case .zeroinitializer(let t):
         register[.register(i)] = ir.llvm(builtinType: t, in: &self).null
 
+      case .advancedByBytes(_):
```
kyouko-taiga, line 902:
```suggestion
      case .advancedByBytes:
```

Tests/LibraryTests/TestCases/PointerToMutableTests.hylo
```diff
@@ -0,0 +1,16 @@
+//- compileAndRun expecting: success
+
+fun test_advance() {
+  var d = PointerToMutable<Never>.allocate_bytes(
+    count: 2,
+    aligned_at: 1 )
```
kyouko-taiga, line 6:
```suggestion
    aligned_at: 1)
```

Tests/LibraryTests/TestCases/PointerToMutableTests.hylo
```diff
@@ -0,0 +1,16 @@
...
+  var e = d.advance(by: 1)
+  var f = e.advance(by:-1)
```
kyouko-taiga, line 8:
```suggestion
  var f = e.advance(by: -1)
```

You can write `+1` if you really want to align the two literals, but I don't think that it's necessary.

kyouko-taiga (commented):
LGTM except for the call to `foreign_value()`. See inline comments and #993.

PR #996, opened 2023-09-12
Add advance(by_bytes:) for PointerToMutable<Never> and advance(by:) for other types

camio:
This works around #989, but allows us to get an advance that works
correctly for non-Never types.

### DIFF
--- a/Library/Hylo/Core/PointerToMutable.hylo
+++ b/Library/Hylo/Core/PointerToMutable.hylo
@@ -33,9 +33,9 @@ public type PointerToMutable<Pointee>: Regular {
     &base = p.base
   }
 
-  /// Returns a pointer advanced by `offset_in_elements`.
-  public fun advance( by offset_in_elements: Int ) -> Self {
-    let offset_in_bytes = MemoryLayout<Pointee>.stride() * offset_in_elements
+  /// Returns `self` offset forward by `n` array elements of `Pointee` type.
+  public fun advance( by n: Int ) -> Self {
+    let offset_in_bytes = MemoryLayout<Pointee>.stride() * n
     return PointerToMutable<Pointee>.new(
       base: Builtin.advanced_by_bytes_word( base, offset_in_bytes.value ) )
   }


Library/Hylo/Core/PointerToMutable.hylo
```diff
@@ -33,13 +33,12 @@ public type PointerToMutable<Pointee>: Regular {
...
-//    let offset_in_bytes = MemoryLayout<Pointee>.stride() * offset_in_elements
-//    return PointerToMutable<Pointee>.new(
-//      base: Builtin.advanced_by_bytes_word( base, offset_in_bytes.foreign_value() ) )
-//  }
+  /// Returns a pointer advanced by `offset_in_elements`.
+  public fun advance( by offset_in_elements: Int ) -> Self {
+    let offset_in_bytes = MemoryLayout<Pointee>.stride() * offset_in_elements
+    return PointerToMutable<Pointee>.new(
+      base: Builtin.advanced_by_bytes_word( base, offset_in_bytes.value ) )
+  }
```
dabrahams, line 41:
```suggestion
  /// Returns `self` offset by `n` array elements of `Pointee` type.
  public fun advance( by n: Int ) -> Self {
    let offset_in_bytes = MemoryLayout<Pointee>.stride() * n
    return PointerToMutable<Pointee>.new(
      base: Builtin.advanced_by_bytes_word( base, offset_in_bytes.value ) )
  }
```

Numerous reasons to make these changes.

1. There's no context for the word "elements"
2. It's not just "a pointer;" it's _this_ pointer.
3. choose parameter names to serve the doc comment.
4. Try not to use the name you're defining to describe its own semantics.

dabrahams: Also I think these should be method bundles so we can say `&p.advance(n)`.

camio: Please add an issue to revisit the interface. I have low confidence that operator overloading isn't going to expose me to a new slew of compiler defects.

PR #999, opened 2023-09-12
Add DynamicBuffer.first_element_address()

camio:
This enables one to access the dynamic buffer's elements.

### DIFF
--- a/Library/Hylo/DynamicBuffer.hylo
+++ b/Library/Hylo/DynamicBuffer.hylo
@@ -31,7 +31,7 @@ public type DynamicBuffer<Header: Deinitializable, Element>: Deinitializable {
     &storage = .null()
   }
 
-  /// Creates an instance for storing `capacity` instances of `Element` and calling `init_header`
+  /// Creates an instance for storing `capacity` instances of `Element` and calls `init_header`
   /// to initialize a `Header`.
   public init<E>(
     capacity: Int,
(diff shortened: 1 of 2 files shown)


Library/Hylo/DynamicBuffer.hylo
```diff
@@ -48,6 +48,14 @@ public type DynamicBuffer<Header: Deinitializable, Element>: Deinitializable {
     })
   }
 
+  /// Returns the address of the initial element.
+  /// - Precondition: `capacity != 0`
```
kyouko-taiga, line 52: Can't we return `.null()` if `capacity == 0`?

kyouko-taiga (approved):
I would consider removing the precondition on `first_element_address` but I understand it is introducing a branch in a function that is likely to be called with the precondition satisfied.

I  _think_ this branch will be written in the caller anyway, so that shouldn't matter too much. Also it seems plausible that the optimizer will be able to remove the branch in most cases, but that is a wild conjecture.

Library/Hylo/DynamicBuffer.hylo
camio: I considered this. Calling this function without meeting that precondition 99% of the time indicates a defect on the caller side. That being the case, the proposed interface (with the check) will move detection of that kind of bug closer to the call site. I don't feel too strongly about it though. I'll merge now. We can always change it later.

Issue #711, opened 2023-07-02
Invalid detection of missing return value


kyouko-taiga:
Compile this:
```swift
public fun f() -> Int {
  if true { return 0 } else { return 1 }
}
```
The compiler complains that `f` doesn't return value on all paths.

esleche:
> There's a third option: analyzing exit paths in the IR. The advantage is that all control-flow subtleties will have been lowered into simple jumps.
> 
> It used to work like that. See [here](https://github.com/val-lang/val/commit/f827f4623151417def8d54473393aedbcb0e360e). I removed the pass because the logic for distinguishing missing returns from implicitly inserted ones was bogus and difficult to debug. Maybe I should have persevered ...
> 
> If we can lay out the assumptions IR analysis can make about what return statements may have been omitted, then I guess we'd have our solution. I'll think about it and report my conclusions.

Any updates on this?

WalterSmuts:
I've given the third option a bash, but just to make sure I understood correctly, here's my approach:
1. Remove the current logic that rejects the test case
2. Add an IR pass that investigates all the reachable blocks of code and...
3. Ensure all reachable blocks return the type you're expecting from the function

Reason I'm investigating only "reachable" blocks is because if you apply this patch:
```diff
--- a/Sources/IR/Emitter.swift
+++ b/Sources/IR/Emitter.swift
@@ -233,9 +233,7 @@ struct Emitter {
   ) -> SourceRange {
     switch emit(braceStmt: b) {
     case .next:
-      if !canonical(returnType).isVoidOrNever {
-        report(.error(missingReturn: returnType, at: .empty(atEndOf: ast[b].site)))
-      } else {
+      if canonical(returnType).isVoidOrNever {
         emitStore(value: .void, to: returnValue!, at: .empty(atEndOf: ast[b].site))
       }
       return ast[b].site
```

You get the following IR from the code:
```
// lib.f
// lib.hylo:4.1-6:2
fun FunctionDecl(2139)() -> Int {
b0(%b0#0 : &Int):
  %i0.0: &Bool = alloc_stack Bool
  %i0.1: &i1 = subfield_view %i0.0, 0
  %i0.2: &i1 = access [set] %i0.1
  store i1(0x1), %i0.2
  %i0.4: &i1 = subfield_view %i0.0, 0
  %i0.5: &i1 = access [sink] %i0.4
  %i0.6: i1 = load %i0.5
  end_access %i0.5
  dealloc_stack %i0.0
[...]

PR #1008, opened 2023-09-15
Exhaustive branch returns

WalterSmuts:
Fixes https://github.com/hylo-lang/hylo/issues/711

### DIFF
--- a/Sources/IR/Analysis/Module+NormalizeObjectStates.swift
+++ b/Sources/IR/Analysis/Module+NormalizeObjectStates.swift
@@ -534,8 +534,8 @@ extension Module {
         if o.value == .full(.initialized) { return }
 
         if k == .set {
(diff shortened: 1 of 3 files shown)


Sources/IR/Analysis/Module+NormalizeObjectStates.swift
```diff
@@ -534,6 +534,16 @@ extension Module {
         if o.value == .full(.initialized) { return }
 
         if k == .set {
+          // If the parameter is a return value (index == 0) we emit specialised diagnostics
+          if case .parameter(_, let index) = p, index == 0 {
```
kyouko-taiga, line 538:
```suggestion
          // If the parameter is a return value (index == 0) we emit specialized diagnostics.
          if case .parameter(_, 0) = p {
```

kyouko-taiga (approved):
Great work. Thank you very much.

PR #1011, opened 2023-09-15
Enable LSP development

koliyo:
- Compile hylo compiler as a library
- Public constructors for `SourcePosition`


Package.swift
```diff
@@ -16,6 +16,7 @@ let package = Package(
   products: [
     .executable(name: "hc", targets: ["hc"]),
     .executable(name: "hylo-demangle", targets: ["hylo-demangle"]),
+    .library(name: "hc-lib", targets: ["Driver"])
```
kyouko-taiga, line 19:
```suggestion
    .library(name: "Hylo", targets: ["Driver"])
```

`Hylo` seems to better fit how Swift libraries are typically named. I am on the fence as to whether it should rather be `HyloCompiler`, but I can easily imagine the library's API will eventually let us interact with the interpreter we'll have to ship for compile-time evaluation.

koliyo: I am open to whatever name you guys prefer. I am very new to Swift in general and don't have a lot of experience with Swift ecosystem and best practices.

kyouko-taiga: Let's go with `Hylo` to start. We can split into smaller specialized libraries in the future if the need emerges.

Issue #997, opened 2023-09-12
Structures without elements produce error in init function


camio:
```
public type Foo {
  public init() { }
}
```
Results in
```
error: set parameter not initialized before function returns
```

kyouko-taiga:
The problem can be fixed in the IR analysis stage by detecting when `self` is the receiver of an initializer for a product type with no fields.

PR #1014, opened 2023-09-15
Empty struct init

WalterSmuts:
Fixes https://github.com/hylo-lang/hylo/issues/997

Suspect there is a better way not to leak abstraction and infer without saving state

### DIFF
--- a/Sources/Core/AST/Decl/InitializerDecl.swift
+++ b/Sources/Core/AST/Decl/InitializerDecl.swift
@@ -70,6 +70,9 @@ public struct InitializerDecl: ExposableDecl, GenericDecl, GenericScope {
   /// Returns whether the declaration is a memberwise initializer.
   public var isMemberwise: Bool { introducer.value == .memberwiseInit }
 
+  /// Returns whether the declaration is a standard initializer. I.e. not a memberwise initializer.
+  public var isInit: Bool { introducer.value == .`init` }
+
   public func validateForm(in ast: AST, reportingDiagnosticsTo log: inout DiagnosticSet) {
     // Parameter declarations must have a type annotation.
     for p in parameters {
--- a/Sources/IR/Analysis/Module+NormalizeObjectStates.swift
+++ b/Sources/IR/Analysis/Module+NormalizeObjectStates.swift
@@ -411,14 +411,33 @@ extension Module {
       return successor(of: i)
     }
 
+    /// Checks if the function is an empty initializer of a type.
+    func isEmptyInitializer(function f: Function.ID) -> Bool {
+        if case .lowered(let d) = f.value, let i = InitializerDecl.ID(d) {
+          if program.ast[i].isInit {
+            let layout = AbstractTypeLayout(
+              of: self[f].inputs[0].type.bareType,
+              definedIn: program
+            )
+            if layout.properties.isEmpty { return true }
+          }
+        }
+        return false
+    }
+
     /// Interprets `i` in `context`, reporting violations into `diagnostics`.
     func interpret(return i: InstructionID, in context: inout Context) -> PC? {
       // Make sure that all non-sink parameters are initialized on exit.
       let entry = entry(of: f)!
-      for (i, p) in self[f].inputs.enumerated() where p.type.access != .sink {
-        ensureInitializedOnExit(
-          .parameter(entry, i), passed: p.type.access, in: &context,
-          reportingDiagnosticsAt: diagnosticSite(for: p, in: f))
+
+      // Unless we're looking at an empty initializer function, we need to make
+      // ensure all non-sink parameters are initialized on exit.
+      if !isEmptyInitializer(function: f) {
+        for (i, p) in self[f].inputs.enumerated() where p.type.access != .sink {
+          ensureInitializedOnExit(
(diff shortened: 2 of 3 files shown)


Sources/Core/AST/Decl/InitializerDecl.swift
```diff
@@ -70,6 +70,9 @@ public struct InitializerDecl: ExposableDecl, GenericDecl, GenericScope {
   /// Returns whether the declaration is a memberwise initializer.
   public var isMemberwise: Bool { introducer.value == .memberwiseInit }
 
+  /// Returns whether the declaration is a standard initializer. I.e. not a memberwise initializer.
+  public var isInit: Bool { introducer.value == .`init` }
```
kyouko-taiga, line 74: Why not use `!isMemberwise`?

WalterSmuts: I see a `enum` and am concerned that the inverse won't always hold (if you add more variants).

Sources/IR/Module.swift
```diff
@@ -312,8 +314,11 @@ public struct Module {
 
     let parameters = program.accumulatedGenericParameters(in: d)
     let inputs = loweredParameters(of: d)
+    let receiverType: AnyType = program[d].receiver.type
+    let layout = AbstractTypeLayout(of: receiverType, definedIn: program)
 
     let entity = Function(
+      isEmptyInit: program.ast[d].isInit && layout.properties.isEmpty,
```
kyouko-taiga, line 321: I don't think there's much to win memoizing this test. We can compute it in `Module+NormalizeObjectStates.swift` and save us the trouble of storing an otherwise needless property.

You can get the declaration from which an IR function was lowered by looking at `FunctionID.value` and you should be able to build the abstract layout of the receiver in the IR pass. If I'm not mistaken, the type of the first argument of a lowered initializer must be `self`.

So we can write something like this (in the IR pass):
```
if case .lowered(let d) = f.value, let i = InitializerDecl.ID(d) {
  if !program.ast[i].isMemberwise {
    let l = AbstractTypeLayout(
      of: self[f].inputs[0].type.bareType,
      definedIn: program)
    if l.properties.isEmpty { ... }
  }
}
```
I would probably implement this code in a function of `IR.Module`, like `isEmptyInitializer(_:)`.

kyouko-taiga (commented):
I think we should avoid storing a property in `IR.Function`.

Another approach you may consider is to have the emitter insert `make_state [initialized] %self` before the `return` instructions of an empty initializer. The instruction will "convince" the borrow checker that `self` is initialized and it will have no effect if `self` is initialized explicitly.

Sources/IR/Module.swift
WalterSmuts: > You can get the declaration from which an IR function was lowered by looking at `FunctionID.value `

These are the things that really slow me down :see_no_evil: Spelunking through a new code-base to do the necessary conversions. I know there must be a path but it takes a while to build up the context to find it easily. This is exactly what I intended :) Just couldn't find it and found a workaround

WalterSmuts:
Weird that this doesn't compile on the CI :thinking: I've successfully recompiled twice now locally and verified I am on the same commit `be6d4684`. I nuked my `.build` to ensure I wasn't using a stale cached artefact. Weird. Suspect it may be the CI system's cache...

kyouko-taiga:
Will try to build on my system.

kyouko-taiga:
I get the same error as CI on my system:

```
/Developer/hc/Sources/IR/Module.swift:297:7: error: missing argument for parameter 'isEmptyInit' in call
      isSubscript: false,
      ^
      isEmptyInit: <#Bool#>, 
```

WalterSmuts:
> Another approach you may consider is to have the emitter insert `make_state [initialized] %self` before the return instructions of an empty initializer. The instruction will "convince" the borrow checker that self is initialized and it will have no effect if self is initialized explicitly.

This approach seems cleaner. Is there any downsides to emitting the `make_state [initialized] %self` directive. Perhaps it has some other side effects? 

Carving out a special case for empty initializers when doing initialization checking looks a bit hacky. I guess the question is whether it's more hacky than the special case for injserting the `make_state [initialized] %self` directive. Will check both (already pushed the first option).

kyouko-taiga:
> This approach seems cleaner. Is there any downsides to emitting the make_state [initialized] %self directive? Perhaps it has some other side effects?

Not really. `mark_state` is a ghost instruction. It has no operational semantics. The only requirement is that we should use it carefully because it's essentially a way to tell the borrow checker "trust me, I know this think is [un]initialized".

> Carving out a special case for empty initializers when doing initialization checking looks a bit hacky.

Unfortunately I think we need to carve out an exception. The rationale is that in other places we do want to track the logical initialization of empty objects.

```
public fun main() {
  var x: Void
  print(x) // error: 'x' is not initialized
}
```

> Will check both (already pushed the first option).

Okay, let me know if you need any help.

Sources/IR/Module.swift
kyouko-taiga: Fully understandable. I'd like to setup live and/or recorded code walks to speed up this process.

WalterSmuts:
Closed in favour of https://github.com/hylo-lang/hylo/pull/1016

PR #1016, opened 2023-09-17
Empty struct init alternate

WalterSmuts:
Fixes: https://github.com/hylo-lang/hylo/issues/997

Alternative approach to https://github.com/hylo-lang/hylo/pull/1014

### DIFF
--- a/Sources/IR/Emitter.swift
+++ b/Sources/IR/Emitter.swift
@@ -316,6 +316,14 @@ struct Emitter {
     insertionPoint = .end(of: entry)
     let bodyFrame = Frame(locals: locals)
     let returnSite = pushing(bodyFrame, { $0.lowerStatements($0.ast[d].body!, expecting: .void) })
+
+    let receiverLayout = AbstractTypeLayout(of: program[d].receiver.type, definedIn: program)
+
+    // If the object is empty, simply mark it initialized.
+    if receiverLayout.properties.isEmpty {
+      let receiver = entry.parameter(0)
+      insert(module.makeMarkState(receiver, initialized: true, at: ast[d].site))
+    }
     insert(module.makeReturn(at: returnSite))
   }
 
--- a/Tests/HyloTests/TestCases/Lowering/EmptyStruct.hylo
+++ b/Tests/HyloTests/TestCases/Lowering/EmptyStruct.hylo
@@ -1,5 +1,5 @@
-//- lowerToFinishedIR expecting: failure
+//- lowerToFinishedIR expecting: success
 
 public type Foo {
-  public init() { }  //! diagnostic set parameter not initialized before function returns
+  public init() { }
 }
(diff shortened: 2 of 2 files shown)


Sources/IR/Emitter.swift
```diff
@@ -316,6 +316,15 @@ struct Emitter {
...
     let bodyFrame = Frame(locals: locals)
     let returnSite = pushing(bodyFrame, { $0.lowerStatements($0.ast[d].body!, expecting: .void) })
+
+    let receiverType: AnyType = program[d].receiver.type
+    let layout = AbstractTypeLayout(of: receiverType, definedIn: program)
+    let receiver = entry.parameter(0)
+
+    // If the object is empty, simply mark it initialized.
+    if layout.properties.isEmpty {
+      insert(module.makeMarkState(receiver, initialized: true, at: ast[d].site))
```
kyouko-taiga, line 326:
```suggestion
    // If the object is empty, simply mark it initialized.
    if layout.properties.isEmpty {
      let receiver = entry.parameter(0)
      insert(module.makeMarkState(receiver, initialized: true, at: ast[d].site))
```

Sources/IR/Emitter.swift
```diff
@@ -316,6 +316,15 @@ struct Emitter {
     insertionPoint = .end(of: entry)
     let bodyFrame = Frame(locals: locals)
     let returnSite = pushing(bodyFrame, { $0.lowerStatements($0.ast[d].body!, expecting: .void) })
+
+    let receiverType: AnyType = program[d].receiver.type
+    let layout = AbstractTypeLayout(of: receiverType, definedIn: program)
```
kyouko-taiga, line 321:
```suggestion
    let layout = AbstractTypeLayout(of: program[d].receiver.type, definedIn: program)
```

kyouko-taiga (approved):
I like this approach much better.

Sources/IR/Emitter.swift
WalterSmuts: Good suggestion. Have included it in my most recent `git push --force`

Sources/IR/Emitter.swift
WalterSmuts: Good suggestion. Have included it in my most recent `git push --force`

WalterSmuts:
> I like this approach much better.

Agreed. There is also a couple of other situations where this approach is already in use:
https://github.com/hylo-lang/hylo/blob/main/Sources/IR/Emitter.swift#L703-L708
https://github.com/hylo-lang/hylo/blob/main/Sources/IR/Emitter.swift#L725-L730
https://github.com/hylo-lang/hylo/blob/main/Sources/IR/Emitter.swift#L2381-L2385
https://github.com/hylo-lang/hylo/blob/main/Sources/IR/Emitter.swift#L2402-L2406

Would be inconsistent not to use the same approach for all these cases.

PR #1040, opened 2023-09-27
Add missing introducer site declarations

### LINKED ISSUE #1025: Missing introducer/introducerSite for nodes
Useful eg for LSP semantic tokens, see #1010

koliyo:
Close #1025 

I created an `Introduced` wrapper for keywords of sub-nodes, eg `else`, and `in`/`where` in for statement.


Sources/Core/AST/Decl/TraitDecl.swift
```diff
@@ -7,6 +7,10 @@ public struct TraitDecl: ExposableDecl, SingleEntityDecl, LexicalScope {
 
   public let site: SourceRange
 
+    /// The site of the `trait` introducer.
+  public let introducerSite: SourceRange
+
+
```
kyouko-taiga, line 13:
```suggestion

```

Sources/Core/AST/Introduced.swift
```diff
@@ -0,0 +1,23 @@
+// Child node with introducer keyword, eg else clause in if-else
+// The NodeIDProtocol conformance is used for BundledNode compatibility
+// eg DoWhileStmt in Emitter.swift: `program[s].condition.site``
```
kyouko-taiga, line 3:
```suggestion
/// A node and the site of its introducer in program sources.
```

Sources/Core/AST/Introduced.swift
```diff
@@ -0,0 +1,23 @@
...
+public struct Introduced<T: NodeIDProtocol>: NodeIDProtocol {
```
kyouko-taiga, line 4: I'm not convinced by the conformance to `NodeIDProtocol`. This type is not an identifier, it's composed of one.

Can you remove the conformance and see how things look like in the emitter?

Sources/Core/AST/Introduced.swift
```diff
@@ -0,0 +1,10 @@
+/// A node and the site of its introducer in program sources.
+public struct Introduced<T: NodeIDProtocol>: Codable {
+  public let introducerSite: SourceRange
+  public let value: T
+
+  public init(introducerSite: SourceRange, value: T) {
+    self.introducerSite = introducerSite
+    self.value = value
+  }
```
kyouko-taiga, line 9:
```suggestion
public struct Introduced<T: NodeIDProtocol>: Codable {

  /// The site of `value`'s introducer.
  public let introducerSite: SourceRange

  /// A node introduced in program sources at `introducerSite`.
  public let value: T

  /// Creates an instance bundling `n` with the site from which its introducer was parsed.
  public init(_ n: T, at introducerSite: SourceRange) {
    self.introducerSite = introducerSite
    self.value = n
  }

```

kyouko-taiga (commented):
One last suggestion to document your new struct. The rest LGTM.

Thanks!

Issue #884, opened 2023-08-14
Implement safe versions of numeric operations


kyouko-taiga:
All operations on `Int` (and its variants) currently do not check for overflow.

PR #1048, opened 2023-09-28
Stand up the trait hierarchy for integer types

kyouko-taiga:
The hierarchy is openly inspired by Swift's standard library.

Some features have not been ported yet due to various blockers. These have been marked with TODOs in the source pointing at a corresponding issue in this repository.

Only the types `Int` and `UInt` have been made conforming to the trait hierarchy. Further, many methods are still thin unsafe wrappers around LLVM operations (see #884).


dabrahams:
Very exciting!

Issue #932, opened 2023-08-25
Copying standard library sources into a resources directory is bad for standard library development.


dabrahams:
You get diagnostics and then you're sent to the copy of the source file, whence you make fixes that don't stick.

- If there's a way to avoid doing that copy, we should do so!
- If we absolutely must do that copy:
    - the files should be made read-only so we don't try to edit them
    - some facility for remapping source locations should be introduced so that diagnostics point to the real source.

kyouko-taiga:
Seems you solved the issue with a PR but I am surprised because it already worked as you expected on my setup.

dabrahams:
@kyouko-taiga I'm confused.  It worked when you build from the command line?  From within Xcode?  Can you double-check your answers by rolling back to an earlier commit and adding a syntax error in the standard library?

kyouko-taiga:
Oh I'm sorry, I got confused because the _tests_ worked as expected (not copied in a bundle), not the standard library. I went back and tested; errors from the stdlib are indeed reported in the bundle. Thanks for your PR.

By the way, I find it simpler to develop new standard library types in a different module before moving them in `Library/Hylo/` once they are complete. That way you don't need to recompile `hc` every time you make a change.

dabrahams:
Yes, that's what I did initially, and we've discussed this before.  It's a trick I've been using since the early days of Swift stdlib development.  But it only works up to a point.

dabrahams:
Somehow this issue came back; we regressed.  I have some ideas that may help.

PR #1097, opened 2023-10-17
Use Bundle.module.resourceURL to point to Hylo standard library

nickpdemarco:
This change removes a hard-coded path to the standard library, which pointed to the pre-build source directory; i.e. directly to `//repo-root/Library/`. Instead, we use `Bundle.module.resourceURL` to point to the _post-build_ location of the standard library resource files.


dabrahams:
I don't know why I approved this.  It totally broke [something](https://github.com/hylo-lang/hylo/issues/932) I had already [fixed](https://github.com/hylo-lang/hylo/pull/947).

PR #1101, opened 2023-10-18
Make sure failing subprocesses cause thrown errors.

dabrahams:
- improve the diagnostic output when there are failures
- factor out synchronous subprocess launching.

### DIFF
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -12,11 +12,13 @@ extension Process {
 
   /// Information about a process that exited with an error.
   struct Failure: Error {
+
     /// The process' exit code.
     let terminationStatus: Int32
 
     /// Any output captured from the process' output pipe.
     let rawOutput: String
+
   }
 
 }
@@ -399,7 +401,8 @@ public struct Driver: ParsableCommand {
     throw EnvironmentError("executable not found: \(executable)")
   }
 
-  /// Executes the program at `path` with the specified arguments in a subprocess, returning
+  /// Runs the executable at `path`, passing `arguments` on the command line, and returns
+  /// its standard output sans any leading or trailing whitespace.
   @discardableResult
   private func runCommandLine(
     _ programPath: String,
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -8,21 +8,6 @@ import IR
 import LLVM
 import Utils
 
-extension Process {
-
-  /// Information about a process that exited with an error.
-  struct Failure: Error {
-
-    /// The process' exit code.
-    let terminationStatus: Int32
-
-    /// Any output captured from the process' output pipe.
-    let rawOutput: String
-
-  }
-
-}
-
 public struct Driver: ParsableCommand {
 
   /// The type of the output files to generate.
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -8,6 +8,19 @@ import IR
 import LLVM
 import Utils
 
+extension Process {
+
+  /// Information about a process that exited with an error.
+  struct Failure: Error {
+    /// The process' exit code.
+    let terminationStatus: Int32
+
+    /// Any output captured from the process' output pipe.
+    let rawOutput: String
+  }
+
+}
+
 public struct Driver: ParsableCommand {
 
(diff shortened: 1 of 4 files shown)


Sources/Driver/Driver.swift
```diff
@@ -8,6 +8,19 @@ import IR
...
+extension Process {
+
+  /// Information about a process that exited with an error.
+  struct Failure: Error {
+    /// The process' exit code.
+    let terminationStatus: Int32
+
+    /// Any output captured from the process' output pipe.
+    let rawOutput: String
+  }
```
kyouko-taiga, line 20:
```suggestion
  struct Failure: Error {

    /// The process' exit code.
    let terminationStatus: Int32

    /// Any output captured from the process' output pipe.
    let rawOutput: String

  }
```

Sources/Driver/Driver.swift
```diff
@@ -386,7 +399,7 @@ public struct Driver: ParsableCommand {
     throw EnvironmentError("executable not found: \(executable)")
   }
 
-  /// Executes the program at `path` with the specified arguments in a subprocess.
+  /// Executes the program at `path` with the specified arguments in a subprocess, returning
```
kyouko-taiga, line 402: The end of the sentence has been trimmed it seems.

Sources/Driver/Driver.swift
```diff
@@ -8,6 +8,19 @@ import IR
 import LLVM
 import Utils
 
+extension Process {
+
+  /// Information about a process that exited with an error.
+  struct Failure: Error {
```
kyouko-taiga, line 14: A reason why this struct is defined here and not in `Process+Extensions.swift`? It is not `fileprivate` and the other file is in the same module. I think it would help if we could all data types related to extensions of `Process` in the same file.

Sources/Driver/Process+Extensions.swift
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+}
+
+extension Process {
+
+  /// The textual output of a process run.
+  public typealias OutputStreams = (standardOutput: Pipe, standardError: Pipe)
+
+  /// The results of a process run that exited with a nonzero code.
```
kyouko-taiga, line 17:
```suggestion
  /// The results of a process run that exited with a non-zero code.
```
nit: Native English speakers will know better whether or not the dash is necessary but I think the form employed should be consistent with the name of the type. `NonZeroExit` suggests that "non" and "zero" are separate.

Sources/Driver/Process+Extensions.swift
```diff
@@ -0,0 +1,95 @@
...
+
+  }
+
+  /// Runs `executable` with the given command line `arguments` and returns its exit status along
+  /// with the text written to its standard output and standard error streams.
+  public static func run(
+    _ executable: URL,
+    arguments: [String] = []
+  ) throws -> OutputStreams {
+
```
kyouko-taiga, line 48:
```suggestion
  public static func run( _ executable: URL, arguments: [String] = []) throws -> OutputStreams {
```

kyouko-taiga (commented):
Left a few comments. Overall LGTM.

Sources/Driver/Driver.swift
dabrahams: Because that file doesn't exist and it seemed a bit much to create a whole file for this extension.

Sources/Driver/Driver.swift
dabrahams:
```suggestion
  /// Returns the standard output generated by the executable at `path` when passed `arguments`
  /// on the command line, sans any leading or trailing whitespace.
```

Sources/Driver/Process+Extensions.swift
dabrahams: Good point!  But I think the name should be changed to `NonzeroExit` to address it; “nonzero” [is a word](https://www.merriam-webster.com/dictionary/nonzero#:~:text=1,nonzero%20affixes).

Sources/Driver/Driver.swift
dabrahams: Oh, oops; this PR adds that file.  I think there was a reason…

dabrahams: The answer is simple: `Failure` is dead code!

PR #1108, opened 2023-10-21
Not operator for conditional compilation

lucteo:
The '&&' and '||' are not implemented, as the logic becomes more involvement (parens, precedence, etc.), and very little gain.


kyouko-taiga (commented):
Just a couple of inline comments.

PR #1109, opened 2023-10-21
Support `feature` for conditional compilation

lucteo:
Allows us to use `#if feature(useLibC) ...` to check the presence of libc.

(at this point, `useLibC` is always on)


Sources/Core/CompilerInfo.swift
```diff
@@ -36,6 +38,7 @@ public struct CompilerInfo {
     compiler = "hc"
     compilerVersion = VersionNumber([0, 1])
     hyloVersion = VersionNumber([0, 1])
+    features = ["useLibC"]
```
kyouko-taiga, line 41:
```suggestion
    features = ["use_libc"]
```

kyouko-taiga:
Is the list of features we'll support defined somewhere? What other option do you envision here? If we're talking about `--no-std` then I think there's no need to mention libc.

Sources/Core/CompilerInfo.swift
lucteo: I would like to fix this in the next patch, where I would make sure that the compiler driver is able to pass flags down to here. The changes are starting to become mode complex, and I would want to separate these changes in another PR.

Here, we agreed that a possible feature would be `freestanding`.

PR #1125, opened 2023-11-06
Standard Library Reorganization and Serialization

dabrahams:
This WIP is part of pre-serializing the standard library ASTs


kyouko-taiga (approved):
LGTM but lately I have been wondering if a name distinction between the standard library and the core library was actually useful. The "core" library is really the standard library and the parts that are dependent on OS-support are just "extensions". I believe this observation goes in the direction of what you wrote on Slack, which I'll copy here for the context:

> I want to rework the way we talk about the standard and core libraries.  In particular, we need a 3rd name so we can talk about
> - The library provided to all user code with no explicit imports (currently "core"), which doesn't depend on OS facilities.
> - The maximal library provided to user code without explicit imports.
> - The difference [between these two artifacts]

Having two separate entities is an abstraction leak because it is almost impossible to cleanly extract all features of the "hosted layer" in separate code. For example, `Array` must be available in the unhosted layer (or today, the core library) but it is unclear how to implement the precondition that makes indexing safe.

dabrahams:
>  The "core" library is really the standard library and the parts that are dependent on OS-support are just "extensions".

That's one valid way of looking at things I suppose, but it's not “the truth:” the fact that something is "just extensions" doesn't make it any less of a library.  That's mostly just about scoping, and I'm sure there *will* one day be entire types, traits, and public functions for which a fully-featured host OS is simply necessary.

You write:
> Having two separate entities is an abstraction leak because

and then go on to give reasons I can't relate to my understanding of the term “abstraction leak”. Care to explain the connection?

The example you give of a problem with the separation doesn't seem very compelling:
> it is unclear how to implement the precondition that makes indexing safe

If by that you mean you don't know how to stop the program, I can think of a bunch of easy answers:
1. We say the language is not supported on platforms without whose LLVM trap instruction doesn't stop the program
2. We require the user to supply a second entry point (other than main) that stops the program and we call that
3. We run an infinite loop (deadlocks are technically safe)
4. no. 2 with a default of no. 3

I prefer no. 1; it's simple.

kyouko-taiga:
> and then go on to give reasons I can't relate to my understanding of the term “abstraction leak”. Care to explain the connection?

It is the fact that the unhosted layer will have to expose parts that I would have expected to be fully encapsulated. For example, the `Array.reserve_capacity` requires privileged access to the internals of `Array`. So we can't really define `Array` in a "core" library and then write an extension that implements `reserve_capacity` without exposing those internals. 

There are other similar examples. The worst case I encountered so far `String` because the hosted layer has to bring essentially extend the union type representing the string's storage. Happy to delve into the details if necessary.

> If by that you mean you don't know how to stop the program, I can think of a bunch of easy answers:

Those answers are great. Last time we chat about it we decided we would do n°2.

The issue is kind of introducing a whole new feature for something that can be solved with conditional compilation, I think. Or we can go with your n°1, which I don't hate.

Issue #1130, opened 2023-11-08
Nondeterministic test failure


dabrahams:
I've seen this happen several times (without parallel testing enabled FWIW):

```
Test Case '-[LibraryTests.LibraryTests test_compileAndRun_OptionalTests]' started.
/Users/dave/src/hylo/Tests/LibraryTests/TestCases/OptionalTests.hylo:1: error: -[LibraryTests.LibraryTests test_compileAndRun_OptionalTests] : processing failed, but success was expected
Test Case '-[LibraryTests.LibraryTests test_compileAndRun_OptionalTests]' failed (7.267 seconds).
```

No further diagnostics are reported.  It's just this one test, as far as I can tell.

dabrahams:
This seems to happen a bit more reliably in CI when parallel testing is enabled, FWIW.

dabrahams:
Here's another example: https://github.com/hylo-lang/hylo/actions/runs/6802288428/job/18495069171#step:4:3081

kyouko-taiga:
Looks like a very nasty bug. I've been completely unable to reproduce it on my machine so far. The thing about parallel testing sounds like a red herring, as I suspect the real culprit is UB.

I'll have to get my hands on Hylo IR and LLVM IR to understand what's going on.

kyouko-taiga:
> I'll have to get my hands on Hylo IR and LLVM IR to understand what's going on.

LLVM IR looks good to me if I compile on my machine.
I am not 100% positive because there's a lot of noise due to unnecessary pointers to `Void` allocas, as mentioned yesterday.

I worry that we may have to insert debug info into the executable to investigate further with lldb.

> I suggest reducing the test case first if possible.

Good idea. I tried on my machine but not on CI.

> What architecture is your machine?

I use an Apple M1.
Am I correct to assume the bug only occurs on Ubuntu? I think CI uses x86 for this OS, but I am not sure.

PR #1137, opened 2023-11-14
Attempt to gather more information about random OptionalTest failure

dabrahams:
z is coming out with the wrong value; let's fix that.

See #1130


Tests/LibraryTests/TestCases/OptionalTests.hylo
```diff
@@ -7,5 +7,6 @@ public fun main() {
 
   &x = None()
   let z = if let i: Int = x { i.copy() } else { 0 }
+  print(z)
   precondition(z == 0)
```
kyouko-taiga, line 11:
```suggestion
  let z = if let i: Int = x { i.copy() } else { 1337 }
  print(z)
  precondition(z == 1337)
```

Hopefully this will make the issue appear consistently, as we're less likely to read 1337 from any random sequence of 8 bytes.

Tests/LibraryTests/TestCases/OptionalTests.hylo
```diff
@@ -6,6 +6,7 @@ public fun main() {
   precondition(y == 42)
```
kyouko-taiga, line 6:
```suggestion
  precondition(y == 42, "#1 failed")
```

dabrahams:
@kyouko-taiga why isn't your error message showing up in the test failure output?

kyouko-taiga:
@dabrahams I got the branch to pass tests and have good reasons to believe that the UB that was causing the errors has been fixed by a0ac6881b24268b5d3f4dd55f730e26b0dd4e92c. Shall we close?

dabrahams:
> @dabrahams I got the branch to pass tests and have good reasons to believe that the UB that was causing the errors has been fixed by [a0ac688](https://github.com/hylo-lang/hylo/commit/a0ac6881b24268b5d3f4dd55f730e26b0dd4e92c). Shall we close?

@kyouko-taiga if you tell me that the fix to switch statements (as that commit describes itself) is also a fix for `if let`, then yes, please go ahead.  Otherwise, there are no switch statements in the failing example, so no.

kyouko-taiga:
It is. The problematic test is unwrapping a union, which involves a switch statement at the LLVM IR level.

PR #1138, opened 2023-11-14
Implement DirectedGraph.isReachable(_:from:)

kyouko-taiga:
This algorithm will be used to detect trait refinement cycles.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -116,7 +116,10 @@ let package = Package(
 
     .target(
       name: "Utils",
-      dependencies: [.product(name: "BigInt", package: "BigInt")],
+      dependencies: [
+        .product(name: "BigInt", package: "BigInt"),
+        .product(name: "Collections", package: "swift-collections"),
+      ],
       swiftSettings: allTargetsSwiftSettings),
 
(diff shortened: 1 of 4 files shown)


dabrahams:
I have an implementation of the Floyd transitive closure algorithm if you want it. , I’m pretty sure That will be faster than asking whether each vertex is reachable from itselfSent from my iPhoneOn Nov 14, 2023, at 5:42 AM, Dimi Racordon ***@***.***> wrote:﻿
Merged #1138 into main.

dabrahams:
[Source](https://github.com/hylo-lang/Lotsawa/blob/refs/heads/main/Sources/Lotsawa/AdjacencyMatrix.swift)
[Reasonably good tests](https://github.com/hylo-lang/Lotsawa/blob/refs/heads/main/Tests/LotsawaTests/AdjacencyMatrixTests.swift)

The use of a bitset makes this pretty fast, but the guaranteed minimum bound is [actually O(|V||E|)](https://www.boost.org/doc/libs/1_83_0/libs/graph/doc/transitive_closure.html)

PR #1140, opened 2023-11-14
Pass compiler flags from driver to AST level

lucteo:
Previously, we used one instance of `CompilerInfo` which would not allow us to pass compiler features from driver to the AST level. This change drops the single instance and passes down a `CompilerInfo` object from driver to AST. It also adds a `freestanding` option to the driver that can now be used with `#if feature(freestanding)`


Sources/Core/AST/AST.swift
```diff
@@ -50,8 +58,15 @@ public struct AST {
     set { storage.coreLibrary = newValue }
   }
 
+  /// Information about the compiler processing `self`.
+  public var compilerInfo: CompilerInfo {
```
kyouko-taiga, line 62:
```suggestion
  public var compiler: CompilerInfo {
```
There's no need to repeat the type of the property verbatim.

Sources/Core/CompilerInfo.swift
```diff
@@ -1,5 +1,5 @@
 /// Describes the compiler; used as source of truth for conditional compilation.
-public struct CompilerInfo {
+public struct CompilerInfo: Codable {
```
kyouko-taiga, line 2: Can we avoid abbreviations?
```suggestion
public struct CompilerConfiguration: Codable {
```

kyouko-taiga (commented):
I left a couple of suggestions inline, but it already looks pretty good.

Sources/Core/AST/AST.swift
lucteo: fixed in all places.

Sources/Core/CompilerInfo.swift
lucteo: fixed in all places.

Issue #1036, opened 2023-09-27
Implement '@autoclosure'


kyouko-taiga:
The feature should work similarly as it does in [Swift](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/closures/#Autoclosures).

Note: `@autoclosure` is necessary to properly implement shortcut behavior in `Bool.infix&&`

kyouko-taiga:
This issue can be split into several steps:

1. Implement support for parameter attributes
2. Make sure that `@autoclosure` attributes have a type `[E]() -> T`
3. Let the type inferrer accept arguments of type `T` to be passed to parameters of type `[E]() -> T`
4. Synthesize the construction of a lambda when an argument is passed to an `@autoclosure` parameter at the IR level.

PR #1151, opened 2023-11-19
Autoclosure: typechecking and code generation (1)

lucteo:
Part of the implementation needed for #1036.

This doesn't properly deal with lambdas that have non-void environments.


Sources/Core/Types/ParameterType.swift
```diff
@@ -9,14 +9,18 @@ public struct ParameterType: TypeProtocol {
   /// The type of the parameter's value.
   public let bareType: AnyType
 
+  /// Indicates if the parameter is marked as autoclosure.
```
kyouko-taiga, line 12:
```suggestion
  /// `true` iff the parameter is an autoclosure.
```

Sources/FrontEnd/TypeChecking/ConstraintSystem.swift
```diff
@@ -516,6 +516,19 @@ struct ConstraintSystem {
       return nil
 
     case let p as ParameterType:
+      if p.isAutoclosure {
```
kyouko-taiga, line 519: I think the whole clause deserves to be in a separate function, so that we can give it a comment.

Tests/EndToEndTests/TestCases/Autoclosure.hylo
```diff
@@ -7,15 +7,38 @@ fun prime() -> Int {
...
+  logic(withDelayed: prime(), counterStartingAt: 0, endingAt: 1)
+
+  // Test with constant.
+  logic(withDelayed: 17, counterStartingAt: 1, endingAt: 1)
+
+  &counter = 0
+  // Test with generics.
+  logic2(withDelayed: prime())
+
+  // Test with generics and non-void environments.
```
kyouko-taiga, line 37:
```suggestion
  // TODO: Test with generics and non-void environments.
```

kyouko-taiga (changes requested):
Nice work, really!

I think the handling of auto closures in the constraint solver deserves to be in its own function.

Sources/Core/AST/Decl/SynthesizedFunctionDecl.swift
```diff
@@ -48,6 +51,7 @@ public struct SynthesizedFunctionDecl: Hashable {
 
   /// Returns the generic parameters of the declaration.
   public var genericParameters: [GenericParameterDecl.ID] {
+    guard !type.captures.isEmpty else { return [] }
```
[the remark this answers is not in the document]
lucteo: For the synthesised autoclosure, we may have no captures. If that's the case, then the code below would fail (there is no receiver). This is similar to how we are testing inside the `description` property.

PR #1176, opened 2023-12-01
Optimize solver

kyouko-taiga:


kyouko-taiga:
> It would be very nice to see some numbers in the PR description and maybe the commit message!

I added numbers in the message. These are just to give an idea and are not comprehensive benchmarks. There are also comments in the code itself to explain the optimization because it's a little tricky.

PR #1184, opened 2023-12-04
Handle implicit conversions on assignments

kyouko-taiga:
This PR fixes IR generation and codegen to properly handle assignments like the one below:
```swift
typealias A = Union<Int, String>

public fun main() {
  var x: A = "Bonjour"
  &x = 42
  if let y: Int = x { print(y) }
}
```
Assigning `42` to `x` requires an implicit conversion of `Int` to `Optional<Int>`. This conversion wasn't performed in the current implementation, which was causing UB.

### DIFF
--- a/Sources/CodeGen/LLVM/Transpilation.swift
+++ b/Sources/CodeGen/LLVM/Transpilation.swift
@@ -1103,14 +1103,14 @@ extension LLVM.Module {
     func insert(switch i: IR.InstructionID) {
       let s = m[i] as! Switch
 
-      // Pick the case 0 as the "default".
-      let cases = s.successors[1...].enumerated().map { (value, destination) in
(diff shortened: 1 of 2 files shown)


dabrahams:
? I thought we were eliminating implicit conversions?

kyouko-taiga:
Still on the fence so I'm trying to keep them for now. I actually wanted to bring this point to the next meeting, as I've been thinking about it a lot lately.

I think implicit conversion can stay if we make move semantics explicit. The main issues I have is that it may be surprising that a `let` parameter gets consumed if a conversion is needed:

```
fun f<T: Regular>(x: Optional<T>) { ... }

public fun main() {
  let a = 42
  f(x: a)
  print(x) // `x` is gone?!
}
```

But if we have to write `a.move()`, then that might be fine.

PR #1189, opened 2023-12-05
Host platform constants cleanups

dabrahams:
Use sites looked overly verbose with `HostPlatform.` qualification so I changed it to `Host.` in the final commit.  We can of course roll that commit back.


Sources/Utils/Host.swift
```diff
@@ -0,0 +1,28 @@
...
+
+  /// The name of the environment variable containing the executable search path.
+  public static let pathEnvironmentVariable = operatingSystem == .windows ? "Path" : "PATH"
+
+  /// The separator between elements of the environment's executable search path.
+  public static let pathEnvironmentSeparator: Character = operatingSystem == .windows ? ";" : ":"
+
+  /// The file extension applied to binary executables.
+  public static let executableSuffix = operatingSystem == .windows ? ".exe" : ""
+}
```
kyouko-taiga, line 28:
```suggestion

}
```

Sources/Utils/Host.swift
```diff
@@ -0,0 +1,28 @@
...
+  #if os(macOS)
+  /// The operating system.
+  public static let operatingSystem: OperatingSystem = .macOS
+  #elseif os(Linux)
+  /// The operating system.
+  public static let operatingSystem: OperatingSystem = .linux
+  #elseif os(Windows)
+  /// The operating system.
+  public static let operatingSystem: OperatingSystem = .windows
+  #endif
```
kyouko-taiga, line 18: Did you run swift-format? IIRC it will indent the contents of `#if ... #endif`.

kyouko-taiga (approved):
LGTM.

Sources/Utils/Host.swift
dabrahams: Obviously I did not!  Thanks for the reminders

PR #1193, opened 2023-12-06
Rename `CompilerConfiguration` to `ConditionalCompilationFactors` & fixes inside the class

lucteo:
- rename `CompilerConfiguration` to `ConditionalCompilationConfiguration`
- fix attribute types
- fix initialization


Sources/Core/ConditionalCompilationConfiguration.swift
```diff
@@ -0,0 +1,35 @@
+import Utils
+
+/// The configuration needed for conditional compilation.
+public struct ConditionalCompilationConfiguration: Codable {
```
dabrahams, line 4:
```suggestion
public struct ConditionalCompilationFactors: Codable {
```

I was racking my brains for this word today and eventually everything became clear.  The name `ConditionalCompilationConfiguration` suggests this thing configures conditional compilation.  I have to stretch the usual sense in which we talk about configuring a language feature quite far in order to make that work.  For a while I thought the word might be `Criteria`, but a criterion would be something like `hyloVersion > 1.0.1` and this does not contain that.  So I was looking for a word meaning “a piece of information based on which a decision is made.”  The definitions I can find for “Factor” fit that idea perfectly, c.f.
- https://dictionary.cambridge.org/us/dictionary/english/factor (1)
- https://www.merriam-webster.com/dictionary/factor (2a)
- https://www.collinsdictionary.com/us/dictionary/english/factor (1)
- https://www.dictionary.com/browse/factor (1)

lucteo: Fixed.

Sources/Core/AST/AST.swift
```diff
@@ -20,12 +20,12 @@ public struct AST {
     /// The module containing Hylo's core library, if any.
     public var coreLibrary: ModuleDecl.ID?
 
-    /// Information about the compiler processing `self`.
-    public let compiler: CompilerConfiguration
+    /// Conditional compilation factors for processing `self`.
+    public let factors: ConditionalCompilationFactors
```
dabrahams, line 24: If you hadn't already chosen this name, would you think it would make sense to refer to this data, in English, as "the factors of the AST?"  I wouldn't.  Another red flag is that the property name contains part of its type name.  That is occasionally unavoidable but rarely.  The name should reflect the role the part plays in the whole.  From what I see here, that might be something like `expansionFilter`.

The ultimate problem is probably that the AST shouldn't contain this property at all, so it may be hard to find a good name, but until you remove it.

Sources/Core/AST/AST.swift
```diff
@@ -20,12 +20,12 @@ public struct AST {
...
 
-    /// Creates an empty AST for given compiler.
-    public init(for compiler: CompilerConfiguration) {
-      self.compiler = compiler
+    /// Creates an empty AST for given conditional compilation factors.
```
dabrahams, line 26: In this case, “for” is just as vacuous in the description as it was as an argument label.  Nobody knows what it means to create an AST for some conditional compilation factors.  You need to tell us what passing a particular set of factors (as opposed to some other set) _means_.

Sources/Core/AST/AST.swift
```diff
@@ -58,14 +58,16 @@ public struct AST {
     set { storage.coreLibrary = newValue }
   }
 
-  /// Information about the compiler processing `self`.
-  public var compiler: CompilerConfiguration {
-    return storage.compiler
+  /// Conditional compilation factors for processing `self`.
```
dabrahams, line 61: Same [problem with "for"](https://github.com/hylo-lang/hylo/pull/1193/files#r1419530199) applies here too.

Sources/Core/AST/AST.swift
dabrahams, line 26:
```suggestion
    /// Creates an empty instance for given conditional compilation factors.
```

The storage is not the AST.

Sources/Core/AST/AST.swift
```diff
@@ -58,14 +58,16 @@ public struct AST {
...
+  public var factors: ConditionalCompilationFactors {
+    return storage.factors
   }
 
   /// Creates an empty AST.
```
dabrahams, line 66: Comment doesn't explain the parameter.

Sources/Core/ConditionalCompilationFactors.swift
```diff
@@ -0,0 +1,35 @@
...
+  /// The target architecture.
+  let architecture: Host.Architecture
+
+  /// The version of the compiler.
+  let compilerVersion: SemanticVersion
+
+  /// The version of the Hylo language version we are targeting.
+  let hyloVersion: SemanticVersion
+
+  /// `true` if we are compiling for freestanding mode.
```
dabrahams, line 18: Try not to write a comment that requires the reader to already understand the meaning of the name you chose.

```suggestion
  /// `true` if the standard library exposes only non-OS-dependent parts.
```

Sources/Core/ConditionalCompilationFactors.swift
```diff
@@ -0,0 +1,35 @@
...
+  /// The target operating system.
+  let operatingSystem: Host.OperatingSystem
+
+  /// The target architecture.
+  let architecture: Host.Architecture
+
+  /// The version of the compiler.
+  let compilerVersion: SemanticVersion
+
+  /// The version of the Hylo language version we are targeting.
```
dabrahams, line 15:
```suggestion
  /// The version of the Hylo language recognized by the language implementation.
```

Sources/Utils/Host.swift
```diff
@@ -2,8 +2,29 @@
 public enum Host {
 
   /// An operating system on which the compiler or interpreter can run.
-  public enum OperatingSystem {
+  public enum OperatingSystem: Codable, CustomStringConvertible {
```
dabrahams, line 5: Nesting most of your additions in `Host` is problematic.  Look at the doc comment for `Host`; it is a namespace representing the platform running the language implementation.   But later you are usingi `Host.OperatingSystem` to describe target platforms.

I would leave `Host` as is.  These perhaps belong nested inside some kind of `Platform` or `Triple` type.

Sources/Utils/Host.swift
```diff
@@ -2,8 +2,29 @@
...
+
     case macOS, linux, windows
+
+    public var description: String {
```
dabrahams, line 9: Is this conformance to `CustomStringConvertible` actually having an effect you care about anywhere?  If not I'd omit it, because the result of`String(describing: someOperatingSystem)` is probably plenty descriptive. If so I'd add a `//` comment explaining why you've done it.

Sources/Utils/Host.swift
```diff
@@ -2,8 +2,29 @@
...
+    }
+
+  }
+
+  /// An architecture on which the compiler or interpreter can run.
+  public enum Architecture: String, Codable, CustomStringConvertible {
+
+    case i386, x86_64, arm, arm64
+
+    public var description: String {
```
dabrahams, line 24: Consider whether this is actually needed, similar to https://github.com/hylo-lang/hylo/pull/1193/files#r1419705945

Sources/Utils/Host.swift
lucteo: Ok. Makes sense. With this in mind, I need to also take `OperatingSystem` out of `Host`, as it's used for two things: to represent the OS for the host, and represent the OS for the target platform.

Sources/Core/AST/AST.swift
```diff
@@ -20,12 +20,12 @@ public struct AST {
     /// The module containing Hylo's core library, if any.
     public var coreLibrary: ModuleDecl.ID?
 
-    /// Information about the compiler processing `self`.
-    public let compiler: CompilerConfiguration
+    /// The expansion filter used while processing conditional compilation statements in `self`.
```
dabrahams, line 23: You need to explain how a set of compilation factors can be interpreted as a filter.

dabrahams (commented):
I think labels could help with comprehensibility here, perhaps?

```swift
    /// Represents "_ >= payload".
    case greaterOrEqual(than: SemanticVersion)

    /// Represents "_ < payload".
    case less(than: SemanticVersion)
```

Sources/Core/AST/AST.swift
lucteo: @dabrahams : can you please help? I'm out of ideas on how to explain this :(

dabrahams: I don't know the answer myself!  Explain it to me, and then maybe I can help with phrasing.

dabrahams: @lucteo I don't want this PR to remain blocked.  How do you want to resolve this? Should we meet?

lucteo: If the Thursday meeting doesn't get too long, let's spend 30 minutes at the end of the meeting to go over this. Does  this fit your schedule?

dabrahams: I think I can do that.

lucteo: use `compilationConditions`; comment: `Conditions for selecting conditional compilation branches`

PR #1194, opened 2023-12-06
Handle closures with environment for autoclosures

lucteo:


Tests/EndToEndTests/TestCases/Autoclosure.hylo
```diff
@@ -37,8 +37,5 @@ public fun main() {
   // TODO: Test with generics and non-void environments.
   // &counter = 0
   // let r = 17
-  // logic2(withDelayed: fun() -> Int {
-  //   &counter += 1
-  //   return r
-  // })
+  // logic2(withDelayed: r.abs())
```
lucteo, line 40: @kyouko-taiga : I would need your help understanding why the code generator inserts a call here to a function that it doesn't exist.

kyouko-taiga: It's a bit hard to tell on the source code only, as I didn't write the logic for lowering autoclosures myself. Can you tell what the emitter believes this function should be?

If you print the IR you should at least see a semi-helpful description of the `FunctionReference` being applied.

PR #1198, opened 2023-12-07
Parse the standard library at most twice for host == target

### LINKED ISSUE #932: Copying standard library sources into a resources directory is bad for standard library development.

dabrahams:
In an attempt to speed up testing, instead of parsing the library each time we need an AST for it as a starting point for testing, store hosted and/or freestanding ASTs in global constants, which are initialized once, lazily and thread-safely in Swift.

Also fixes #932.

## Results

- Debug:
   - Parallel testing (`time swift test --parallel --skip-build`):
      - Before: `440.54s user 23.45s system 834% cpu 55.578 total`
      - After:    `441.36s user 23.85s system 834% cpu 55.723 total`
   - Serial testing  (`time swift test --parallel --skip-build`):
      - Before: `325.20s user 11.28s system 96% cpu 5:49.06 total`
      - After:    `289.28s user 10.43s system 95% cpu 5:13.35 total`
- Release:
   - Parallel testing (`time swift test --parallel --skip-build`):
      - Before: `263.19s user 29.15s system 618% cpu 47.302 total`
      - After:    `264.44s user 28.92s system 624% cpu 46.998 total`
   - Serial testing  (`time swift test --parallel --skip-build`):
      - Before: `211.38s user 15.21s system 144% cpu 2:36.36 total`
      - After:   `186.19s user 14.69s system 155% cpu 2:09.02 total`

The `time` man page is unhelpful but from my experiment (`time sleep 5` yields `0.00s user 0.00s system 0% cpu 5.014 total`), the number we mostly care about is the last one, total.

This change seems to make zero difference if parallel testing is enabled, for reasons I don't understand, and the wins from parallel testing for a developer far outweigh any improvement we get from this change when testing serially.  I think it's still a change worth making.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -136,7 +136,7 @@ let package = Package(
       name: "StandardLibrary",
       dependencies: ["FrontEnd", "Utils"],
       path: "StandardLibrary",
-      resources: [.copy("Sources")],
+      exclude: ["Sources"],
       swiftSettings: allTargetsSwiftSettings),
 
     .plugin(
--- a/Package.swift
+++ b/Package.swift
@@ -88,7 +88,6 @@ let package = Package(
       dependencies: [
         "Utils",
         "Core",
-        "StandardLibrary",
         .product(name: "Collections", package: "swift-collections"),
         .product(name: "Durian", package: "Durian"),
         .product(name: "BigInt", package: "BigInt"),
@@ -135,6 +134,7 @@ let package = Package(
 
     .target(
       name: "StandardLibrary",
+      dependencies: ["FrontEnd", "Utils"],
       path: "StandardLibrary",
       resources: [.copy("Sources")],
       swiftSettings: allTargetsSwiftSettings),
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -157,7 +157,7 @@ public struct Driver: ParsableCommand {
     let productName = makeProductName(inputs)
 
     /// An instance that includes just the standard library.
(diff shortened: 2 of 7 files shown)


kyouko-taiga:
I haven't carefully reviewed the PR yet but I agree that it seems like a change worth making. As I mentioned on Slack, we can probably perform type checking only once (or twice as the title suggest?) for all the tests, which is significantly more expensive than parsing. So this PR is a very nice step toward the hopefully bigger improvement we'll get from reusing the result of type checking.

For information, parsing is about 15% of the total exec time when compiling with `--typecheck` in release, which seems to explain the numbers reported in the PR description.

kyouko-taiga:
Maybe `--parallel` forks the test driver before it got a change to initialize your static lets?

dabrahams:
I never considered that parallel was done by forking rather than by launching threads, but now that you mention it, I'm sure that's the case, since it's very likely the average Swift programmer is testing thread-unsafe constructs (to say nothing of XCTest itself).  Still, to have zero impact it would have to be forking newly for each test.  If that's the case, maybe I can find some kind of testing setup hook that will let me do the initialization earlier.

dabrahams:
Nope, it's [using threads](https://github.com/apple/swift-package-manager/blob/main/Sources/Commands/SwiftTestTool.swift#L929).  And not even using Dispatch, which is a bit shocking!

PR #1199, opened 2023-12-07
Delete unneded Lazy/LazyThrowing

dabrahams:
Globals do this automatically in Swift and if we ever _do_ need something like this for non-globals we should use dispatch_once.


dabrahams:
Oh, whoops; we still need this.

PR #1203, opened 2023-12-09
Throw from AST construction instead of swallowing diagnostics.

dabrahams:
If the standard library fails to parse we want to know why.

### DIFF
--- a/Sources/Utils/ResultOrError.swift
+++ b/Sources/Utils/ResultOrError.swift
@@ -1,0 +1,42 @@
+/// A value of type `T` or the error produced by computing that value.
+public enum ResultOrError<T> {
+
+  /// The successful computation.
+  case result(T)
+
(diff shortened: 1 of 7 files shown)


Sources/Utils/ResultOrError.swift
```diff
@@ -0,0 +1,42 @@
+/// A value of type `T` or the error produced by computing that value.
+public enum ResultOrError<T> {
```
kyouko-taiga, line 2: Doesn't this type exist in the standard library?

dabrahams: There's something called [`Result`](https://developer.apple.com/documentation/swift/result) it seems. Might work if `any Error : Error`, but I thought there were no self-conforming protocols.  OK, it seems to work-ish.  I loathe the design 🤷 but I suppose we should use it…

PR #1208, opened 2023-12-09
Encode/Decode coreTraits as part of AST

dabrahams:
Merging this into #1205 has no effect on the failure.  @kyouko-taiga you might need to fix this problem.  Also I don't know how to write a test for this change.

### DIFF
--- a/Sources/Core/AST/AST.swift
+++ b/Sources/Core/AST/AST.swift
@@ -17,6 +17,9 @@ public struct AST {
     /// - Invariant: All referred modules have a different name.
     public var modules: [ModuleDecl.ID] = []
 
+    /// The traits in Hylo's standard library that are known by the compiler.
+    var coreTraits: CoreTraits?
+
     /// The module containing Hylo's core library, if any.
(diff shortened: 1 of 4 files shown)


Sources/Core/AST/AST.swift
```diff
@@ -17,6 +17,9 @@ public struct AST {
     /// - Invariant: All referred modules have a different name.
     public var modules: [ModuleDecl.ID] = []
 
+    /// The traits in Hylo's standard library that are known by the compiler.
+    var coreTraits: CoreTraits?
```
kyouko-taiga, line 21:
```suggestion
    public var coreTraits: CoreTraits?
```
For consistency with other properties. Or we should make them all internal.

kyouko-taiga (approved):
It is probably fine, though I am not sure we have a clear enough understanding of what we want to write in the serialized representation of a module to tell whether `coreTraits` should be part of the AST. We can always move them back out if necessary so I won't object to merging this PR.

Note that the problem with #1205 is unrelated.

PR #1211, opened 2023-12-10
Initial support for monomorphization of generic value parameters

kyouko-taiga:
This PR stands up support for monomorphizing code involving generic value parameters.

Before depolymorphization, references generic value parameters are represented with a new IR instruction, `generic_argument`, which introduces initialized storage. In monomorphized form, generic value parameters are rewritten as local variables allocated and initialized before the first instruction of the generic function and uses of `generic_argument` are replaced by uses of the corresponding stack allocations.

### DIFF
--- a/Sources/Core/Program.swift
+++ b/Sources/Core/Program.swift
@@ -31,6 +31,11 @@ extension Program {
     return (s.kind == TranslationUnit.self) && isPublic(d) && (n == "main")
   }
 
+  /// `true` iff the `Builtin` module is visible in `s`.
+  public func builtinIsVisible(in s: AnyScopeID) -> Bool {
+    ast[module(containing: s)].canAccessBuiltins
+  }
+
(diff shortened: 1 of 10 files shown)


dabrahams:
s/generic_argument/generic_parameter/. no?

kyouko-taiga:
I hesitated and then chose argument because the instruction only makes sense if we think about the argument that is passed to the generic function. Generic parameters can be abstract, but generic arguments can't.

That said, I won't object if we want to use "parameter".

dabrahams:
I honestly don't know what we want.  I just looked at the language “generic value parameters are represented with a new IR instruction, `generic_argument`” and thought something was amiss.  Is it the argument or the parameter?  These things are different: the argument is an expression and the parameter is the name used inside the parameterized entity.

kyouko-taiga:
I think you were right and the IR instruction should be called `generic_parameter`. I renamed it.

PR #1220, opened 2023-12-13
Implement implicit parameters

kyouko-taiga:
This PR implements implicit parameters. The design of the feature is mostly copied from [Scala3](https://docs.scala-lang.org/tour/implicit-parameters.html). Syntactic choices are motivated by precedent in Scala and Haskell.

In summary:

- Parameters of functions, methods, initializers, and subscripts can be declared "implicit"
- The compiler passes values to implicit parameters automatically
- Bindings elligible as implicit arguments are implicit parameters and annotated local declarations
- Implicit resolution is entirely type-based

The motivation for this feature are twofold:

- Implicit parameters give us a way to alleviate the "argument piping" problem. Many functions in `hc` need "context" to compute properties of objects identified by some position in a collection. The ability to pass such a context implicitly lets us write more concise calls (see https://github.com/orgs/hylo-lang/discussions/731).
- Implicit parameters can be used as "object capabilities" to model a function's ability to perform an unsafe operation or throw an exception. Object capabilities offer an [alternative to effect polymorphism](https://dotty.epfl.ch/docs/reference/experimental/canthrow.html).

Here is an example:

```hylo
type Item: SemiRegular {
  public let cost: Int
  public memberwise init
  public fun infix== (_ other: Item) -> Bool { self.cost == other.cost }
}

/// Returns the cost of item `i`.
fun name(of i: Int, in items? : Array<Item>) -> Int {
  let x = items[i] // See #1219
  return x.cost.copy()
}

public fun main() {
  @given var all_items = Array<Item>()
  &items.append(Item(cost: 28))
  // ...
  let x = 0
  print(name(of: x))
}
```

In the above program, `items` in the function `name(of:in:)` is declared as implicit parameter by appending a question mark to its name. When `name(of:in:)` is called, because the second argument is missing, the compiler looks for a binding suitable to be passed implicitly. It finds `all_items` because it is declared, which is declared with the `@given` attribute. The use of an implicit in this program lets us write more concise calls to `name(of:in:)`.

Unlike in Scala, implicit resolution is not allowed to look further than the scope of the function containing in which the call occurs. Further, the type of the implicit value must match _exactly_ the type of the argument so that no conversion occurs (note: this restriction may be obvious if we remove all implicit conversions, see https://github.com/orgs/hylo-lang/discussions/1165).

Implicit parameters must be `let`. This restriction is meant to avoid surprising effects on the state of local bindings.

### DIFF
--- a/Sources/Core/AST/AST.swift
+++ b/Sources/Core/AST/AST.swift
@@ -281,20 +281,23 @@ public struct AST {
   /// Returns a table mapping each parameter of `d` to its default argument if `d` is a function,
   /// initializer, method or subscript declaration; otherwise, returns `nil`.
   public func defaultArguments(of d: AnyDeclID) -> [AnyExprID?]? {
-    let parameters: [ParameterDecl.ID]
+    runtimeParameters(of: d)?.map({ self[$0].defaultValue })
+  }
+
+  /// Returns the run-time parameters of `d` iff `d` is callable.
+  public func runtimeParameters(of d: AnyDeclID) -> [ParameterDecl.ID]? {
     switch d.kind {
     case FunctionDecl.self:
-      parameters = self[FunctionDecl.ID(d)!].parameters
+      return self[FunctionDecl.ID(d)!].parameters
     case InitializerDecl.self:
-      parameters = self[InitializerDecl.ID(d)!].parameters
+      return self[InitializerDecl.ID(d)!].parameters
     case MethodDecl.self:
-      parameters = self[MethodDecl.ID(d)!].parameters
+      return self[MethodDecl.ID(d)!].parameters
     case SubscriptDecl.self:
-      parameters = self[SubscriptDecl.ID(d)!].parameters
+      return self[SubscriptDecl.ID(d)!].parameters
     default:
       return nil
     }
-    return self[parameters].map(\.defaultValue)
   }
 
   /// Returns the generic parameters introduced by `d`.
--- a/Sources/Core/AST/Decl/BindingDecl.swift
+++ b/Sources/Core/AST/Decl/BindingDecl.swift
@@ -20,6 +20,9 @@ public struct BindingDecl: ExposableDecl {
   /// The initializer of the declaration, if any.
   public let initializer: AnyExprID?
 
+  /// `true` iff the declaration is defining an implicit.
+  public let isGiven: Bool
+
   /// Creates an instance with the given properties.
   public init(
     attributes: [SourceRepresentable<Attribute>] = [],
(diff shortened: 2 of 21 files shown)


dabrahams:
I'm not 100% convinced we need to ban implicit `inout` parameters in the long run.  But can always lift restrictions later.

kyouko-taiga:
Technically we can make any passing convention work, but I'm worried about hiding mutation. But yes, we should definitely keep that option in mind.

dabrahams:
I understood the motivating concern.  But I think if we have a way to explicitly declare that a `@given` thing will be implicitly mutated, e.g.

```
  @given(&) var all_items = Array<Item>()
```

ersump'n, I don't see a lot of danger in it.

kyouko-taiga:
Normally we can see where mutation occurs by scanning through the sources to look for `&`. With implicit mutating givens, call sites will loose their mutation marker:

```
fun foo(x?: inout Int) { &x += 1 }

public fun main() {
  @given(&) var n = 0
  ...
  foo() // seems mutation-free
}
```

That would also cause an issue with the detection of lambda effects, which is currently based on the presence of mutation marker. Reliance on syntax is due to the fact that there's no obvious way to know if a call dispatches to a mutating variant of a bundle during type checking; this information is detected later, during access checking.

```
fun foo(x?: inout Int) { &x += 1 }

public fun main() {
  @given(&) var n = 0
  var f = fun() { foo() }
}
```

It's likely there's a way out of this issue, but it's likely it's not an easy one.

dabrahams:
> call sites will loose their mutation marker

I realize that of course.  I personally don't think it's that bad.  `@given(&) var n` is a signal that any call in scope might mutate `n`.  It's opt-in and it's sufficiently syntactically heavy that it's not likely to be abused.

> detection of lambda effects, which is currently based on the presence of mutation marker

To me that has always seemed upside-down.  We can detect where mutation has to happen or we wouldn't be able to diagnose missing mutation markers.  Since we can diagnose where mutation markers need to be, we can keep a separate record of those spots instead of looking for the markers themselves.

> It's likely there's a way out of this issue, but it's likely it's not an easy one.

Hard to believe it's particularly hard to do, and it's surely not so hard that it needs to be a consideration when discussing what the long-term language design should be, which is what I'm trying to do here.  I'm not saying we need inout implicit arguments today or tomorrow (in fact we didn't need implicits _at all_, especially not right now).  I'm merely suggesting we should be open to the idea of lifting the restriction.

kyouko-taiga:
> I'm merely suggesting we should be open to the idea of lifting the restriction.

And I agreed with my first answer.

> We can detect where mutation has to happen or we wouldn't be able to diagnose missing mutation markers.

The question is _when_ we know that a mutation marker is missing, and the answer is: after type checking. Specifically, it's during object state analysis, which is deep into IR analysis. So before can know whether or not `x.f()` misses a mutation marker, we still have to infer the type of `fun() { x.f() }`. Specifically, we must infer the type of its captures.

We can assume `let` for everything that doesn't have a mutation marker, which leaves us with: "what about implicits?"

PR #1243, opened 2023-12-23
Add `c_void_pointer[to:]` function to easily get `CVoidPointer` objects.

lucteo:


kyouko-taiga:
`CVoidPointer` was a workaround for a limitation that has been lifted (see #1251). One should now use `MemoryAddress` to represent `void*` in FFIs. The helper proposed by this PR is no longer necessary.

You can get a `MemoryAddress` from any pointer like this:
```
let n = ...
let m = MemoryAddress(pointer[to: n])
```

PR #1273, opened 2024-01-04
Fix the specialization of associated types

kyouko-taiga:


dabrahams (commented):
I'd have liked to see a test with this, and not just for “engineering standards” reasons.  At this early stage there are lots of ways to fall off a cliff with the compiler and I'm trying to keep track of what kinds of things work and what don't.  I don't know what didn't work based on the title of the PR, and when there's a test I get to see an example.

Thanks, as always, for fixing stuff and making the compiler work better!

kyouko-taiga:
Sure thing.

Normally I write a test for those kind of fixes but in this particular case the next PR in the pipeline is a change in the standard library that relies on this fix. Would you prefer to have tests even when in this situation? I don't mind if you think it can help.

PR #1274, opened 2024-01-04
Implement 'MutableCollection.rotate'

kyouko-taiga:
This one is for you @sean-parent!

The test is a bit clunky but we don't have a good way to write unit tests yet.

### DIFF
--- a/StandardLibrary/Sources/Core/MutableCollection.hylo
+++ b/StandardLibrary/Sources/Core/MutableCollection.hylo
@@ -12,3 +12,50 @@ trait MutableCollection: Collection {
   fun swap_at(_ i: Position, _ j: Position) inout
 
 }
+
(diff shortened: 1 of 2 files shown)


dabrahams (commented):
The rotate algorithm should return the new midpoint.  Is there a reason you didn't use [this work](https://github.com/apple/swift/blob/6e06e5173753a22702d3cde1caf71110e169793a/test/Prototypes/Algorithms.swift#L53-L154)?

Just looking at the structure of this implementation it seems very unlikely that it can be correct, given what I know about how rotate has to work (it's explained in the referenced code).

kyouko-taiga:
I'll happily merge your own implementation.

dabrahams:
This implementation was much much closer to correct than I thought, for the record.

PR #1277, opened 2024-01-05
Fix the lowering of for loops in generic contexts

kyouko-taiga:
Lowering a for loop in a trait extension caused the emitter to create unspecialized references to the subscript requirement in `Collection`, making the produced IR impossible to compile to LLVM.


dabrahams:
🍾

PR #1279, opened 2024-01-05
Rotate robustness

dabrahams:
This PR adds a much more complete test of Hylo's rotate and makes it handle one important case that used to cause an assertion.

Except for not returning the new pivot, it actually does work!  So I'm sorry I doubted it (@kyouko-taiga). Amazingly, the swift implementation it was derived from failed [this test](https://github.com/hylo-lang/hylo/blob/045f1bb3f3313b532bb67dfc4285cf2fdb6dfce6/Tests/UtilsTests/CollectionExtensionsTests.swift#L68-L89).

As a student of rotate, I was expecting to see nested loops, which I have in my implementation.  But it's been too long since I worked on that algorithm, so I had forgotten the purpose of that nesting; I should have remembered that since it's O(N)  that nesting isn't strictly necessary; it just avoids a bunch of tests/branches.  

----

**update**: this PR now replaces the implementation of rotate with a more efficient one that returns the new pivot.

The comments in the original implementation didn't make sense to me;  I couln't understand how it works and thus I couldn't quite figure out how to get it to return the new pivot.

### DIFF
--- a/StandardLibrary/Sources/Core/MutableCollection.hylo
+++ b/StandardLibrary/Sources/Core/MutableCollection.hylo
@@ -16,8 +16,8 @@ trait MutableCollection: Collection {
 public extension MutableCollection {
 
   /// Swaps the order in which the values `self[..<p]` and `self[p...]` occur, so that the end of
-  /// the latter sequence of element values is the start of the former sequence; returns that
-  /// position, where the sequences coincide.
+  /// the latter sequence of element values is the start of the former sequence, and returns that
+  /// position.
   ///
   /// - Complexity: O(n), where n is the number of elements in `self`.
   public fun rotate(regions_separated_by p: Position) inout -> Position {
--- a/StandardLibrary/Sources/Core/MutableCollection.hylo
+++ b/StandardLibrary/Sources/Core/MutableCollection.hylo
@@ -16,12 +16,8 @@ trait MutableCollection: Collection {
 public extension MutableCollection {
 
   /// Swaps the order in which the values `self[..<p]` and `self[p...]` occur, so that the end of
-  /// the latter sequence of element values is the start of the former sequence, and returns that
-  /// position.
-  ///
-  /// The value originally at `self[p]` (if any) ends up first.  The value originally located at
-  /// `p`'s predecessor (if any) ends up last.  The return value is the new position of the value
-  /// originally at `start_position()` (if any).
+  /// the latter sequence of element values is the start of the former sequence; returns that
+  /// position, where the sequences coincide.
   ///
   /// - Complexity: O(n), where n is the number of elements in `self`.
   public fun rotate(regions_separated_by p: Position) inout -> Position {
--- a/StandardLibrary/Sources/Core/MutableCollection.hylo
+++ b/StandardLibrary/Sources/Core/MutableCollection.hylo
@@ -15,47 +15,80 @@ trait MutableCollection: Collection {
 
 public extension MutableCollection {
 
-  /// Shifts the elements of `self` so that the element currently at `middle` becomes first.
+  /// Swaps the order in which the values `self[..<p]` and `self[p...]` occur, so that the end of
+  /// the latter sequence of element values is the start of the former sequence, and returns that
+  /// position.
   ///
-  /// The contents of `self` is reordered so that the elements in the range [`middle`, `end`) occur
-  /// before the elements in the range [`start`, `middle`) and the orders of the elements in these
-  /// ranges  are preserved.
+  /// The value originally at `self[p]` (if any) ends up first.  The value originally located at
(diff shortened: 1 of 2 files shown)


StandardLibrary/Sources/Core/MutableCollection.hylo
```diff
@@ -15,47 +15,80 @@ trait MutableCollection: Collection {
...
+  /// Swaps the order in which the values `self[..<p]` and `self[p...]` occur, so that the end of
+  /// the latter sequence of element values is the start of the former sequence, and returns that
+  /// position.
   ///
-  /// The contents of `self` is reordered so that the elements in the range [`middle`, `end`) occur
-  /// before the elements in the range [`start`, `middle`) and the orders of the elements in these
-  /// ranges  are preserved.
+  /// The value originally at `self[p]` (if any) ends up first.  The value originally located at
+  /// `p`'s predecessor (if any) ends up last.  The return value is the new position of the value
+  /// originally at `start_position()` (if any).
```
sean-parent, line 24: What if there wasn't any? 
```suggestion
originally at the `start_position()`, if any, otherwise end_position().
```

sean-parent (changes requested):
The code looks good. The contract needs a little work. I suggested an improvement, but I think it could be further improved.

Tests/LibraryTests/TestCases/MutableCollectionTests.hylo
```diff
@@ -1,23 +1,51 @@
 //- compileAndRun expecting: success
 
+/// Appends `0 ..< limit` to `a`
+fun appendCount(up_to limit: Int, to a: inout Array<Int>) {
```
kyouko-taiga, line 4:
```suggestion
fun append_count(up_to limit: Int, to a: inout Array<Int>) {
```

Tests/LibraryTests/TestCases/MutableCollectionTests.hylo
```diff
@@ -1,23 +1,51 @@
...
+fun test_rotate(length N: Int) {
+  var failed = false
+
+  // For all possible pivot points
+  var pivot = 0
+  while pivot <= N {
+
+    // A collection to work on
+    var a = Array<Int>()
+    appendCount(up_to: N, to: &a)
```
kyouko-taiga, line 22:
```suggestion
    append_count(up_to: N, to: &a)
```

kyouko-taiga:
> Amazingly, the swift implementation it was derived from failed this test.

Sadly I can't find back the sources from which I took inspiration to write the my Swift implementation. I didn't understand that code very well, though, so the Hylo one was my attempt at simplifying it. It's interesting that this exercise didn't port the bug from the original!

Anyway, thanks for writing this better-researched version.

StandardLibrary/Sources/Core/MutableCollection.hylo
dabrahams: @sean-parent If I wanted to cover all the cases, I'd have to deal with the other two “(if any)”s.  The full semantics of the function are precisely described by the summary and this paragraph is supplemental; it's designed to help people visualize what happens, so I thought it better to deal with those edge cases by explicitly leaving them out, lest the language get too dense to understand.  Unless you can tell me why this one “(if any)” is special, it seems to me my choices should be:

1. deal with all the (if any)s in the same way you've suggested for the third one.
2. leave the paragraph alone.
3. strike the paragraph.

WDYT?

Tests/LibraryTests/TestCases/MutableCollectionTests.hylo
dabrahams: Oh, good catch!  Old habits die hard (also when switching between codebases).

StandardLibrary/Sources/Core/MutableCollection.hylo
sean-parent: This "if any" is different in that it is describing the result - so I don't think you can leave it out. It is precise in the paragraph above, so you could just strike it.

dabrahams: Struve the (if any) or the whole paragraph?

dabrahams: Another alternative would be to open the supplemental section with, “Except for trivial cases where `p` is at either end of `self`,…”

sean-parent: I would either strike the sentence (don't have supplement text about the result) or strike the paragraph. I'm not sure it adds much. Even the "except for trivial cases..." phrasing needs to state what is returned in the trivial cases. The first "if any" is fine because, in that case, there is no effect, which is what "if any" implies.

PR #1281, opened 2024-01-05
Fix name resolution of generic parameters in trait extensions

kyouko-taiga:


Tests/HyloTests/TestCases/TypeChecking/NameLookupGeneric3.hylo
```diff
@@ -0,0 +1,15 @@
...
+// because it will trigger qualified lookup in the nominal scope of `P` _from_ the scope of `f`.
+
+trait P {}
+
+trait Q {
+  type B
+}
+
+extension P {
+  public fun f<T: Q, v: Int>(x: T.B) {}
```
dabrahams, line 14: Is this the simplest thing that would have failed (e.g. was `v` needed to demonstrate the bug before the fix)?  Mostly just curious.

dabrahams:
🥇

Tests/HyloTests/TestCases/TypeChecking/NameLookupGeneric3.hylo
kyouko-taiga: Not for the implementation that I merged. However, a previous attempt relied on assumptions about what kind of entity we can find after `:`, which failed to account for the fact that one could use arbitrary type expressions to declare value parameters. The declaration of `v` will show why this assumption is too naive.

Issue #1035, opened 2023-09-27
Support conditional compilation


kyouko-taiga:
It should be possible to compile different code depending on the target. For example:

```
public fun main() {
#if os(macOS)
  print("Apple")
#elseif os(Linux)
  print("Pear")
#endif
}
```

See https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Conditional-Compilation-Block for a list of the predicates we may want to support.

Note: this feature is probably necessary to write good `[U]Int` tests.

lucteo:
Implemented with #1105, #1108, #1109, #1140

PR #1282, opened 2024-01-05
Test int64 conditionally on arch

peter-evans:
Wraps Int tests which assume 64-bit arch in a conditional compilation block

This follows up on a `TODO` since https://github.com/hylo-lang/hylo/issues/1035 was completed.

### DIFF
--- a/Tests/LibraryTests/TestCases/IntTests.hylo
+++ b/Tests/LibraryTests/TestCases/IntTests.hylo
@@ -55,6 +55,15 @@ fun test_remainder_reporting_overflow() {
   precondition(a0.1 == true)
 }
 
+fun test_int64() {
+  precondition(Int.bit_width() == 64)
+  precondition(Int.max() == 9223372036854775807)
+  precondition(Int.min() == -9223372036854775808)
+  precondition((-1).nonzero_bit_count() == 64)
+  precondition((10).leading_zeros() == 60)
+  precondition((10).trailing_zeros() == 1)
+}
+
 public fun main() {
   // Int.init()
   precondition(Int() == 0)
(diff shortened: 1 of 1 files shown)


Tests/LibraryTests/TestCases/IntTests.hylo
```diff
@@ -70,13 +79,9 @@ public fun main() {
...
-  // TODO:
-  // These tests assume they are running on a 64-bit architecture.
-  // They should be wrapped in a conditional compilation block (see #1035).
-  precondition(Int.bit_width() == 64)
-  precondition(Int.max() == 9223372036854775807)
-  precondition(Int.min() == -9223372036854775808)
-  precondition((-1).nonzero_bit_count() == 64)
-  precondition((10).leading_zeros() == 60)
-  precondition((10).trailing_zeros() == 1)
+  #if arch(x86_64)
```
dabrahams, line 82: doesn't `||` work in this context?  It would be better to write the call once

dabrahams (commented):
Thanks for the PR!  One question inline

Tests/LibraryTests/TestCases/IntTests.hylo
peter-evans: It doesn't appear to work currently.
```
Tests/LibraryTests/TestCases/IntTests.hylo:82.20-22: error: prefix operator cannot be separated from its operand
  #if arch(x86_64) || arch(arm64)
```
There's a comment in the following PR suggesting that it's not implemented yet.
> The '&&' and '||' are not implemented, as the logic becomes more involvement (parens, precedence, etc.), and very little gain.

https://github.com/hylo-lang/hylo/pull/1108#issue-1955494072

peter-evans: Also no mention of `||` and `&&` here: https://github.com/hylo-lang/specification/pull/15

kyouko-taiga (approved):
LGTM, at least until we have a `||` operator in compiler conditions.

PR #1283, opened 2024-01-06
First concurrency example

lucteo:
Add basic concurrency support to Hylo.


Examples/concurrent_greeting.hylo
```diff
@@ -0,0 +1,25 @@
+
+fun do_greet() -> Int {
+  print("Hello, concurrent world!")
+  return 17
+}
+fun prime_number() -> Int {
+  13
+}
+fun concurrent_greeting() -> Int {
```
kyouko-taiga, line 9:
```suggestion
}

fun prime_number() -> Int {
  13
}

fun concurrent_greeting() -> Int {
```

Examples/concurrent_greeting.hylo
```diff
@@ -0,0 +1,25 @@
+
+fun do_greet() -> Int {
+  print("Hello, concurrent world!")
+  return 17
+}
+fun prime_number() -> Int {
+  13
+}
```
kyouko-taiga, line 8:
```suggestion

fun prime_number() -> Int {
  13
}

```

StandardLibrary/Sources/Concurrency/CancellableFuture.hylo
```diff
@@ -0,0 +1,8 @@
+
+/// A future type that allows cancelling the computation that produces its value.
+public trait CancellableFuture: Future {
```
kyouko-taiga, line 3: There's no use of this trait in this PR. Can we remove it before we can actually support cancellable futures?

StandardLibrary/Sources/Concurrency/Future.hylo
```diff
@@ -0,0 +1,11 @@
+/// A type whose values can be awaited.
+public trait Future {
```
kyouko-taiga, line 2: Why do we need a trait here? The only implementation is `LocalFuture<E>`, at least in this PR.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
```diff
@@ -0,0 +1,86 @@
+/// A future that cannot escape the local scope.
+public type LocalFuture<E>: Future {
```
kyouko-taiga, line 2: The fact that the future can't escape the local scope is encoded in the type system because it's not movable (and therefore not copyable). So I don't think that is the way we should document and name this type.

I would use just `Future` and rename the trait (assuming we can't simply remove it). I believe this kind of future is the one corresponding to the most common case we want to promote, so it would make sense to use an unqualified name. We can always create a "EscapableFuture" later.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
```diff
@@ -0,0 +1,86 @@
...
+
+  // TODO
```
kyouko-taiga, line 4: Documentation missing here.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
```diff
@@ -0,0 +1,86 @@
...
+  public typealias Result = Int
+
+  /// The base frame object needed by the underlying implementation.
```
kyouko-taiga, line 7: I would not mind a couple of sentences in the detailed description telling us a little more about the role of this property. I suspect it is a pointer to the suspended frame but I can't really say for sure with these docs.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
```diff
@@ -0,0 +1,86 @@
...
+  let target_thread_reclaimer: MemoryAddress
+
+  let user_function: MemoryAddress
+
+  init() {}
+
+}
+
+// TODO: #1246
+type FWrapper<E, R>: Movable, Deinitializable {
```
kyouko-taiga, line 69: I understand that is a temporary workaround, but `FWrapper` is a name that can be improved. What about `FunctionWrapper`?

Also

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
```diff
@@ -0,0 +1,86 @@
...
+    &self.f = f
+    let local_f = fun (_ frame: SpawnFrameBase) -> Void {
+      inout self = PointerToMutable<Self>(type_punning: mutable_pointer[to: frame]).unsafe[]
+      &self.do_call()
+    }
+    concore2full_spawn2(self.base_frame, local_f)
+  }
+
+  /// Awaits the result of the computation.
+  /// May return on a different OS thread than the one that called this.
```
kyouko-taiga, line 27:
```suggestion
  /// Awaits the result of the computation.
  ///
  /// - Note: May return on a different OS thread than the one that called this.
```

In an entity's documentation, the first bit of text is the short description, which runs until the first double new lines. It is followed by the detailed description. In principles we should never have more than one sentence in the short description.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
```diff
@@ -0,0 +1,86 @@
...
+  public fun await() sink -> Result {
+    concore2full_await(self.base_frame)
+
+    if let r: Int = self.r {
+      return r.copy()
+    } else {
+      fatal_error("we don't have a result yet")
+      return 0 // TODO: without this we get: "set parameter not initialized before function returns"
```
kyouko-taiga, line 35: Is this path truly unreachable? If so, adding `fatal_error()` _is_ the right way to signal IR gen that there is no need to check the state of the parameters.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  private fun do_call() inout {
+    &r = self.f.f()
+  }
+
+}
+
+/// Describes the frame needed to spawn a computation. Mimics the C implementation.
+type SpawnFrameBase: Deinitializable {
```
kyouko-taiga, line 46: It would be best of we had a short documentation of all these properties.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
```diff
@@ -0,0 +1,86 @@
...
+
+  let task_function: MemoryAddress
+
+  let next_task: MemoryAddress
+
+  let sync_state : Int32
```
kyouko-taiga, line 52:
```suggestion
  let sync_state: Int32
```

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
```diff
@@ -0,0 +1,86 @@
...
+
+}
+
+// @external("concore2full_spawn")
+// fun concore2full_spawn(_ frame: set SpawnFrameBase, _ f: Int); // TODO: f is a pointer to a function
+
+/// Spawns the execution of `f` using control data `frame`.
+@external("concore2full_spawn2")
+fun concore2full_spawn2(_ frame: set SpawnFrameBase, _ f: [](SpawnFrameBase) let -> Void);
+/// Awaits the completion of the computation started in `frame`.
```
kyouko-taiga, line 83:
```suggestion

/// Awaits the completion of the computation started in `frame`.
```

StandardLibrary/Sources/Concurrency/Spawn.hylo
```diff
@@ -0,0 +1,6 @@
+
+/// Spawns a computation and returns a future that can be used to await its completion.
+/// The future cannot escape the current scope.
```
kyouko-taiga, line 3:
```suggestion
/// Spawns a computation and returns a future that can be used to await its completion.
///
/// The future cannot escape the current scope.
```

kyouko-taiga (changes requested):
I left online comments.

Examples/concurrent_greeting.hylo
```diff
@@ -0,0 +1,25 @@
...
+  var future = spawn_(fun() { do_greet() })
+  let x = prime_number()
+  print("Main threads says: hello!")
+  print("# before await")
+  let y = future.await() // switching threads
+  print("# after await")
+  return x + y
+}
+
+public fun main() {
```
kyouko-taiga, line 19: One thing to keep in mind: this example is using a lot of print statements, which has a remote chance of causing UB with the current implementation. The issue is that we currently do not synchronize the initialization of global variables, and the implementation of `print` uses one to get a hold of `stdout`.

lucteo: Thank you. Will keep this in mind.

StandardLibrary/Sources/Concurrency/CancellableFuture.hylo
lucteo: Sure, will do.

StandardLibrary/Sources/Concurrency/Future.hylo
lucteo: I'm soon planning to add more types of futures. But, probably it makes sense to add it later, when we need it.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
lucteo: I can rename this to `Future`, but I believe that, when we have things properly working, the `spawn` keyword can return multiple concrete types, and we would want to teach users about a trait `Future` instead of concrete types.

I'll rename it for now, and see where the future of the Future leads us :)

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
lucteo: I would want to fix the return type not being an `Int` first. I need to figure out the TODO in the `await()` function before.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
lucteo: I would prefer to avoid going into implementation details here. From a Hylo perspective, we just need this object to ensure we can call the C API.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
lucteo: Yes, this is truly unreachable. `do_call()` is guaranteed to be called, and thus we use `Optional` just for delayed initialization. We are paying a small extra cost for the `Optional` but that doesn't matter at this point.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
lucteo: I'd like to avoid going into implementation details here. We have these fields here just to be compatible with the C library

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
lucteo: Works for me.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
kyouko-taiga: `Optional` is quite cheap because the allocation is inline. If need be we can even imagine a branchless unsafe unwrap method in the future.

StandardLibrary/Sources/Concurrency/LocalFuture.hylo
kyouko-taiga: Then I suggest adding some comments in the detailed description to say that. Something like:

> The properties of this type only mimic its corresponding C implementation. They are not meant to be used from Hylo.

PR #1286, opened 2024-01-06
Make `Array` conform to `Movable`

lucteo:


StandardLibrary/Sources/Array.hylo
```diff
@@ -141,6 +141,16 @@ public conformance Array: Collection {
     }
   }
 
+  /// Gives `self` the value of a consumed `source`.
```
kyouko-taiga, line 144: I think the conformance can be trivially synthesized. Did you try not writing any implementation?

lucteo: It appears to work without adding the function here. I did not try it, as I though it will only automatically synthesize it if we have a memberwise init. Removing the code...

kyouko-taiga (approved):

Issue #385, opened 2023-02-13
LambdaType should be renamed ArrowType


kyouko-taiga:
`LambdaType` is used to type all "arrow-like" objects in the language, not only lambdas. The only exception are method bundles, but those already have a dedicated `MethodType`.

dabrahams:
What are some other arrow-like objects?

kyouko-taiga:
```swift
type A {
  init() {}
  fun ham() -> Int { 2 }
}

public fun main() {
  fun foo() -> Int { 0 }
  let bar = fun() -> Int { 1 }
  let a = A()
  ...
}
```

`foo`, `bar`, `A.init`, `A.ham`, and `a.ham` are arrow-like objects, but they're not all the exact same kind of arrows. That has subtle consequences on the semantics.

Note that only `bar` is a lambda.

dabrahams:
Is there any role for the word “callable” in addressing this issue?

kyouko-taiga:
Could be `CallableType` instead of `ArrowType`. The latter is shorter and very a quite common name in type theory though.

dabrahams:
Not just `Callable`?

kyouko-taiga:
I'd like to keep the suffix, at least for now. I still think it's helping distinguishing between important kinds, as we have entities like `TraitType` and `TraitDecl`.

Given the way we tend to name stuff in Swift, I would also reserve `Callable` for a protocol describing a family of arrow types, should we have one.

peter-evans:
I could give this rename a shot if it's helpful. Just want to check this issue is still relevant.

kyouko-taiga:
I think it is.

Note that now we have `CallableType` which is implemented by the arrow types _and_ `SubscriptType`, filling the role of the trait I hypothesized we'd need:

> I would also reserve `Callable` for a protocol describing a family of arrow types, should we have one.

So you'll have to rename that one too.

peter-evans:
Just to clarify, what would you like to rename `CallableType` to?

`LambdaType` -> `ArrowType`
`CallableType` -> ?

kyouko-taiga:
Actually if we rename `LambdaType` to `ArrowType` then there's nothing to change for `CallableType`. I'd accept that change.

@dabrahams had expressed reluctance because (IIUC) "arrow" isn't as established as "callable". But I think this ship has sailed because the distinction is already significant in the compiler in other places:

[...]

PR #1293, opened 2024-01-09
Rename LambdaType to ArrowType

peter-evans:
Closes https://github.com/hylo-lang/hylo/issues/385

I did my best to change all references to "lambda" that aren't referring to *actual* lambda expressions. It wasn't immediately clear in some cases, so I might have made a mistake or missed some references.

I've included grep output below for convenience. Let me know if I've missed references that should be updated.

<details>
<summary>rg Lambda</summary>
<br />

```
rg Lambda
Sources/Core/AST/Expr/LambdaExpr.swift
2:public struct LambdaExpr: Expr {

Sources/Core/AST/Decl/SynthesizedFunctionDecl.swift
24:    /// Lambda generated for an autoclosure argument.

Sources/Core/AST/NodeIDs/NodeKind.swift
115:    LambdaExpr.self,

Sources/Core/AST/AST+Walk.swift
124:    case LambdaExpr.self:
125:      traverse(self[n] as! LambdaExpr, notifying: &o)
503:    _ n: LambdaExpr, notifying o: inout O

Sources/FrontEnd/Parse/Parser.swift
1679:      // Lambda expression.
1680:      return try parseLambdaExpr(in: &state).map(AnyExprID.init)
1962:  private static func parseLambdaExpr(in state: inout ParserState) throws -> LambdaExpr.ID? {
1983:      LambdaExpr(decl: decl, site: state.ast[decl].site))

Sources/FrontEnd/TypeChecking/TypeChecker.swift
2555:    of e: LambdaExpr.ID, withHint hint: ArrowType?
2593:    of e: LambdaExpr.ID, withHint hint: [CallableTypeParameter]?
4700:    case LambdaExpr.self:
4701:      return _inferredType(of: LambdaExpr.ID(e)!, withHint: hint, updating: &obligations)
4947:    of e: LambdaExpr.ID, withHint hint: AnyType? = nil,
5364:    of e: LambdaExpr.ID, withHint hint: AnyType? = nil,

Sources/IR/Emitter.swift
1413:    case LambdaExpr.self:
1414:      emitStore(LambdaExpr.ID(e)!, to: storage)
1597:  private mutating func emitStore(_ e: LambdaExpr.ID, to storage: Operand) {

Docs/MetatypeLowering.md
86:**Lambda types:**

Tests/HyloTests/ParserTests.swift
1275:  func testLambdaExpr() throws {
1278:    let expr = try XCTUnwrap(ast[exprID] as? LambdaExpr)
```
</details>

<details>
<summary>rg lambda</summary>
<br />

```
rg lambda
Sources/Core/Program.swift
459:        return qualification + ".lambda"

Sources/Core/AST/Expr/ParameterTypeExpr.swift
12:  /// `true` if arguments to the parameter are automatically wrapped in lambdas.

Sources/Core/AST/Expr/LambdaExpr.swift
1:/// A lambda.

Sources/FrontEnd/Parse/Parser.swift
1969:    let body = try state.expect("function body", using: lambdaBody)
1986:  private static let lambdaBody = inContext(
2131:      body = try state.expect("function body", using: lambdaBody)

Sources/FrontEnd/TypeChecking/TypeChecker.swift
890:  /// - Requires: `d` is not the underlying declaration of a lambda.
1094:  /// Type checks `d`, which declares a lambda, after its type has been inferred.
1096:    assert(program[d].isInExprContext, "expected lambda")
2548:  /// `hint` is used as contextual information to refine guesses if it is a lambda type with the
2565:    // Slow path: infer elided conventions from the uses in the lambda's body.
2903:  /// underlying declaration of a lambda). Otherwise, `.void` is returned.

StandardLibrary/Sources/Array.hylo
47:    // TODO: Call `self.copy()` directly in the lambda.

Docs/Subscripts.md
70:The caller initiates a projection by calling the ramp function and passing it a lambda representing at least the code using the projection.
71:This continuation is called with another lambda denoting the tail corresponding to the `yield` instruction executed by the ramp.
72:It is expected to call that lambda when it is done with the projection.

Docs/CPlusPlus.org
123:For interop purposes, projections could be translated into functions that accept a C++ lambda

Docs/MetatypeLowering.md
22:- Kind (e.g., product type, lambda type, etc.)
89:struct hylo_lambda_type_mirror {

Tools/gyb.py
320:            in tokenize.generate_tokens(lambda i=iter(source_lines):
367:            in tokenize.generate_tokens(lambda i=iter(source_lines): next(i)):

Tests/HyloTests/TestCases/Lowering/MemberLambda.hylo
4:// call to a member property with a lambda type.

Tests/EndToEndTests/TestCases/Lambda.hylo
8:  // Applies a thin, immutable lambda.
12:  // Applies a thick, immutable lambda.
17:  // Applies a mutable lambda with owned captures.
25:  // Applies a mutable lambda with borrowed captures.

Tests/ManglingTests/ManglingTests.swift
51:        let lambda = fun[let capture = local](x: Int) { local + x }
53:          let z = lambda(x: y)

Tests/LibraryTests/TestCases/PointerToMutableTests.hylo
27:fun test_initialize_pointee_lambda() {
46:  test_initialize_pointee_lambda()
```
</details>

### DIFF
--- a/Sources/Core/Types/ArrowType.swift
+++ b/Sources/Core/Types/ArrowType.swift
@@ -1,18 +1,18 @@
 import Utils
 
-/// The type of a lambda.
-public struct LambdaType: TypeProtocol {
+/// A type denoting arrow-like objects.
+public struct ArrowType: TypeProtocol {
 
-  /// The effect of the lambda's call operator.
+  /// The effect of the arrow's call operator.
   public let receiverEffect: AccessEffect
 
-  /// The environment of the lambda.
+  /// The environment of the arrow.
   public let environment: AnyType
 
-  /// The input labels and types of the lambda.
+  /// The input labels and types of the arrow.
   public let inputs: [CallableTypeParameter]
 
-  /// The output type of the lambda.
+  /// The output type of the arrow.
   public let output: AnyType
 
   public let flags: TypeFlags
@@ -45,7 +45,7 @@ public struct LambdaType: TypeProtocol {
   /// Given an initializer type `t`, creates the corresponding constructor type.
   ///
   /// - Requires: `t` is an initializer type of the form `[](self: set A, B...) -> Void`.
-  public init(constructorFormOf t: LambdaType) {
+  public init(constructorFormOf t: ArrowType) {
     let r = ParameterType(t.inputs.first!.type)!
     precondition(r.access == .set)
     self.init(inputs: Array(t.inputs[1...]), output: r.bareType)
@@ -54,13 +54,13 @@ public struct LambdaType: TypeProtocol {
   /// Given an constructor type `t`, creates the corresponding initializer type.
   ///
   /// - Requires: `t` is a constructor type of the form `[](A...) -> B`.
-  public init(initializerFormOf t: LambdaType) {
+  public init(initializerFormOf t: ArrowType) {
     let r = CallableTypeParameter(label: "self", type: ^ParameterType(.set, t.output))
     self.init(receiverEffect: .let, inputs: [r] + t.inputs, output: .void)
   }
 
   /// Returns a thin type accepting `self`'s environment as parameters.
-  public var lifted: LambdaType {
+  public var lifted: ArrowType {
     let elements = TupleType(environment).map(\.elements) ?? [.init(label: nil, type: environment)]
     let p = elements.map { (e) -> CallableTypeParameter in
       if let t = RemoteType(e.type) {
@@ -73,8 +73,8 @@ public struct LambdaType: TypeProtocol {
   }
 
   /// `self` sans environment.
-  public var strippingEnvironment: LambdaType {
-    LambdaType(receiverEffect: receiverEffect, environment: .void, inputs: inputs, output: output)
+  public var strippingEnvironment: ArrowType {
+    ArrowType(receiverEffect: receiverEffect, environment: .void, inputs: inputs, output: output)
   }
 
   /// `self` transformed as the type of a member of `receiver`, which is existential.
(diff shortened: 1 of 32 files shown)


Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -2543,16 +2543,16 @@ struct TypeChecker {
   }
 
   /// Computes and returns the types of the inputs of `e`'s underlying declaration, using `hint`
-  /// to guess the passing conventions of unnaotated parameters.
+  /// to guess the passing conventions of unanotated parameters.
   ///
-  /// `hint` is used as contextual information to refine guesses iff it is a lambda type with the
+  /// `hint` is used as contextual information to refine guesses if it is a lambda type with the
```
kyouko-taiga, line 2548: `iff` is not a typo. We use it to mean "if and only if", which is a double implication.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -2887,8 +2887,8 @@ struct TypeChecker {
 
     if program[e].isAutoclosure {
       let s = program[program[e].bareType].site
-      guard let u = LambdaType(t), u.inputs.isEmpty else {
-        report(.error(autoclosureExpectsEmptyLambdaAt: s, given: t))
+      guard let u = ArrowType(t), u.inputs.isEmpty else {
+        report(.error(autoclosureExpectsEmptyArrowAt: s, given: t))
```
kyouko-taiga, line 2891: I think the deserve some work but FWIW it does refer to an actual lambda. Maybe `autoclosureExpectsEmptyEnvironment`.

Sources/IR/Callee.swift
```diff
@@ -9,8 +9,8 @@ enum Callee {
   /// A reference to a variant in a method bundle.
   case bundle(BundleReference<MethodDecl>)
 
-  /// A lambda.
-  case lambda(Operand)
+  /// An arrow.
+  case arrow(Operand)
```
kyouko-taiga, line 13: That would be an actual lambda.

kyouko-taiga (commented):
Thanks for this PR! I'm already riding the "arrow" train but seeing the change in context, I really think that is a big improvement.

I left comments for a couple of changes that should not be made.

Sources/FrontEnd/TypeChecking/TypeChecker.swift
peter-evans: https://github.com/hylo-lang/hylo/pull/1293/commits/93713afe5b6b00a276241b45f18971114c94ea1f

Sources/IR/Callee.swift
peter-evans: https://github.com/hylo-lang/hylo/pull/1293/commits/92ac1dee8711669945decffa9a71d4016a8130b1

Sources/Core/Types/ArrowType.swift
```diff
@@ -1,18 +1,18 @@
 import Utils
 
-/// The type of a lambda.
-public struct LambdaType: TypeProtocol {
+/// A type denoting arrow-like objects.
```
kyouko-taiga, line 3:
```suggestion
/// A type expressed with an arrow.
```

kyouko-taiga (approved):
OK for me. "arrow" may not have convinced every one but regardless I think it is fair to say this PR is an improvement over the status quo. We can always rename instances of "ArrowType" by "WhateverType" if we get more inspiration.

Issue #842, opened 2023-08-08
Valc command line option `-l <name>`


jmafc:
The subject command option has a help value of `Link the generated crate(s) to the specified native library.`. Per Dimi in discussion #841, the word `crate` is supposed to be changed.

PR #1297, opened 2024-01-11
Correct the description of the -l option

peter-evans:
Closes https://github.com/hylo-lang/hylo/issues/842

I'm guessing that `module` is what we want here, but please correct me if I'm wrong.


kyouko-taiga:
Yep, `crate` is embarrassing leftover from a very old copy-pasting. I'm not sure that the comment is completely accurate though!

Issue #416, opened 2023-02-17
`using` is a vacuous argument label


dabrahams:
Every function does what it does “using” its arguments.  If we can't find a better label, [this one](https://github.com/val-lang/val/blob/main/Sources/FrontEnd/Parse/ParserState.swift#L256) should just be removed everywhere it appears in our codebase.

peter-evans:
For [these two](https://github.com/hylo-lang/hylo/blob/5b773b79f9ea009c7480625c3b8021cb74817dc9/Sources/FrontEnd/Parse/ParserState.swift#L283-L304) usages, would `applies` make sense? It would then match the use of "Applies" in the function descriptions.

dabrahams:
The goal is to make something that's roughly an english phrase, or just fall back to noun labels if you can't.  If it were up to me I would use "`from:`" in these cases. "`applying:`" is OK, but I would say matching the doc comment is not a primary goal. Once you fix the labels, the doc comment might need an update.

kyouko-taiga:
I think `parsedWith` might be the most meaningful label. It works relatively well at call site IMO:

```
let e = try state.expect("type expression", parsedWith: parseExpr(in:))
```
_Note: IIRC, all uses of `expect` are in the parser._

peter-evans:
Thank you both for the helpful explanations. I've raised a PR for @kyouko-taiga's suggestion of `parsedWith` [here](https://github.com/hylo-lang/hylo/pull/1296).

dabrahams:
FWIW, to me this
```
let e = try state.expect("type expression", parsedWith: parseExpr(in:))
```
says , “I expect that "type expression" (was) parsed successfully with `parseExpr(in:)`”

kyouko-taiga:
You mean that "parsed expression" is the input of `parseExpr(in:)`? Would you understand that the same way if it was written like this?
```
let e = try state.expect(construct: .typeExpression, parsedWith: parseExpr(in:))
```

dabrahams:
> You mean that "parsed expression" is the input of `parseExpr(in:)`? 

I don't think I mean that. I don't see "parsed expression" anywhere in anything either of us wrote before.

I'm reading “_x_ parsed” as “_x_ passed a syntax check,” as in “my program parsed but it didn't type check”

> Would you understand that the same way if it was written like this?
> 
> ```
> let e = try state.expect(construct: .typeExpression, parsedWith: parseExpr(in:))
> ```

[...]

PR #1298, opened 2024-01-12
Refactor expect-using to apply-expecting in ParseState

peter-evans:
Closes https://github.com/hylo-lang/hylo/issues/416

Refactors to @dabrahams' [suggestion](https://github.com/hylo-lang/hylo/issues/416#issuecomment-1888164109).

```
let e = try state.apply(parseExpr(in:), expecting: .typeExpression)
```

### DIFF
--- a/Sources/FrontEnd/Parse/Parser.swift
+++ b/Sources/FrontEnd/Parse/Parser.swift
@@ -282,7 +282,7 @@ public enum Parser {
     wrappedIn context: ParserState.Context
   ) throws -> [AnyDeclID] {
     // Parse the left delimiter.
-    let opener = try state.expect("'{'", using: { $0.take(.lBrace) })
+    let opener = try state.apply({ $0.take(.lBrace) }, expecting: "'{'")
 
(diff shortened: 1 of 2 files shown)


kyouko-taiga (changes requested):
I don't personally feel like the call sites have been improved significantly. Also note that in Swift we typically accept closures as the last parameter, so it should be `state.apply(expecting: "something", parseSomething(in:))`.

To be honest I would leave this issue to rest. The parser requires a complete rewrite if we wanted to make things right. It's based on code that I rushed because we were missing a parser and someone needed it for a related research project (I used to write Hylo ASTs by hand!) The whole thing lacks consistency and has sub-optimal APIs.

Now that we've got some insights into the ways we can write and maintain a MVS-compliant recursive descent parser for Hylo, I think that rewriting a better implementation is a one week effort for someone who knows what they are doing. I suspect that trying incremental changes here and there will cumulatively take way longer.

peter-evans:
Understood. I'll close this then.

PR #1300, opened 2024-01-14
Update the swift-format documentation

adevress:
Update the swift-format documentation to reflect what the script actually does.

- The script uses `swift run` internally with a statically specified version. There seems no need of manual installation
- The version installed with the script is currently different of the one documented


dabrahams:
@adevress Sorry for the slow response time. I think I'm going to move forward with #1318; shall we discuss how the docs should be changed?

dabrahams:
@adevress ping.

PR #1307, opened 2024-01-17
Don't discard any diagnostics from parsing the standard library.

### LINKED ISSUE #1287: Parse errors in the standard library cause silent failures
If you introduce a parse error in the standard library and then try to run the compiler, it will exit with status 1 without emitting any diagnostic.

dabrahams:
fixes #1287


kyouko-taiga:
Nice, thanks!

PR #1316, opened 2024-01-18
Bump swift-codecov-action version to fix deprecation warnings

peter-evans:
These warning are caused by the action using a deprecated command on the GitHub Actions runner. The action author has fixed it in version [`0.7.4`](https://github.com/mattpolzin/swift-codecov-action/releases/tag/0.7.4), but we might as well bump to the latest version.

image


peter-evans:
I think it would be a good idea to enable dependabot for the GitHub Actions eco-system at some point.

Issue #1120, opened 2023-10-27
Support `--version`


nickpdemarco:
For applications like Compiler Explorer, getting a version identifier (even one produced by a git hash) is useful for triaging issues.

peter-evans:
@kyouko-taiga Do you have a good idea for how to manage the version? I'm more familiar with JVM languages where generally this is achieved by passing the version as a string to the compiler at compile time. It looks like only boolean compiler flags are available in swift.

Here's one possible idea.
1. Create a `Version.swift` file containing just `let version = "unknown"`.
2. Add `Version.swift` to `.gitignore` to prevent changes being committed.
3. Create a script under `Tools/set-version.sh` which re-creates the `Version.swift` file with either a version tag argument, or a git hash.
   e.g.
   ```
    echo "let version = \"$VERSION\"" > Version.swift
    ```

CI could call `set-version.sh` before building, and locally it can be called to set the version to a git hash.

kyouko-taiga:
That's a good idea, I think.

I don't know exactly when one should be supposed to run `set-version.sh`. Any idea how things are done in other compilers? I'll ask what people do for Scala and report back here.

peter-evans:
> I don't know exactly when one should be supposed to run `set-version.sh`

I don't know about other compilers specifically, but in general, for CI:

1. Run `build-and-test.yml` on push to tags, as well as branches.
   ```yml
   on:
     push:
       tags:
         - v*
   ```
2. If the run is triggered by a tag, execute `set-version.sh` and pass the tag string as an arg.
3. If the run is not triggered by a tag, execute `set-version.sh` without an arg, and it uses the `HEAD` git hash.

Then you either tag versions on `main` manually, or create a GitHub release, which also adds a tag.

For local development, I don't know of any easy solution. With this idea, developers would just need to exercise discipline to call `set-version.sh` before compiling, otherwise the version is just `unknown`.

kyouko-taiga:
AFAIC we can go ahead with your idea! Thanks a lot, that's a great proposal.

PR #1320, opened 2024-01-19
Support versioning the compiler

peter-evans:
Closes https://github.com/hylo-lang/hylo/issues/1120

- Adds a `-V` / `--version` flag to the compiler.
- Adds a script to set the compiler version. This is mainly for use in CI so that tags can be set as the version before building. The script can also be called during local development to set the version to a git hash. It could help when triaging issues to understand the commit the compiler was built at. This was the motivation for the original [issue](https://github.com/hylo-lang/hylo/issues/1120).
- Updates CI to set the version of the compiler either to a tag (if a `v` prefixed tag is pushed), or the git hash in all other cases.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -27,6 +27,8 @@ on:
       - "README.md"
       - "LICENSE"
       - ".gitignore"
+    tags:
+      - v*
   pull_request:
     branches: [ "**" ]
     paths-ignore:
@@ -37,7 +39,23 @@ on:
       - ".gitignore"
 
 jobs:
+  determine-version:
+    name: "Determine compiler version from git tag"
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get-tag.outputs.version }}
+    steps:
+    - id: get-tag
+      run: |
+        if [[ $GITHUB_REF == refs/tags/* ]]; then
+          version=${GITHUB_REF#refs/tags/}
+        else
+          version=""
+        fi
+        echo "version=$version" >> $GITHUB_OUTPUT
+
   build-devcontainer:
+    needs: determine-version
     name: "Devcontainer: ${{ matrix.host.os }}/${{ matrix.configuration }}"
     strategy:
       fail-fast: false
@@ -63,6 +81,9 @@ jobs:
         submodules: true
         show-progress: false
 
+    - name: Set compiler version
+      run: ./Tools/set-hc-version.sh ${{ needs.determine-version.outputs.version }}
+
     - name: Build and Test
       uses: devcontainers/ci@v0.3
       with:
@@ -77,6 +98,7 @@ jobs:
               (grep -q "[.']EndToEndTests[/. ]test_" testoutput.txt && grep -q "[.']HyloTests[/. ]test_" testoutput.txt) ||
               (echo "error: generated tests failed to run; see
               https://github.com/apple/swift-package-manager/issues/6595" && false) )
+          .build/${{ matrix.configuration }}/hc --version
 
     - name: Check code coverage
       uses: mattpolzin/swift-codecov-action@0.7.3
@@ -86,6 +108,7 @@ jobs:
         CODECOV_JSON: .build/${{ matrix.configuration }}/codecov/*.json
 
   build-native-macos:
+    needs: determine-version
     name: "Native: ${{ matrix.host.os }}/${{ matrix.configuration }}"
     strategy:
       fail-fast: false
@@ -110,6 +133,9 @@ jobs:
       with:
         submodules: true
         show-progress: false
+
+    - name: Set compiler version
+      run: ./Tools/set-hc-version.sh ${{ needs.determine-version.outputs.version }}
 
     - name: Setup swift
       uses: swift-actions/setup-swift@v1
@@ -147,8 +173,12 @@ jobs:
             (grep -q "[.']EndToEndTests[/. ]test_" testoutput.txt && grep -q "[.']HyloTests[/. ]test_" testoutput.txt) ||
             (echo "error: generated tests failed to run; see
             https://github.com/apple/swift-package-manager/issues/6595" && false) )
-          
+
+    - name: Output compiler version
+      run: .build/${{ matrix.configuration }}/hc --version
+
   build-native-windows:
+    needs: determine-version
     name: "Native: windows-latest, debug"
     strategy:
       fail-fast: false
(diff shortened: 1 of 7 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -198,3 +232,6 @@ jobs:
 
     - name: Test
       run: swift test --parallel --skip-build --pkg-config-path . 
+
+    - name: Output compiler version
+      run: .build/debug/hc --version
```
peter-evans, line 237: We can remove these output steps if you don't want them. I was using them to check the version was set correctly, but they might be useful, anyway.

Sources/Driver/Driver.swift
```diff
@@ -107,14 +124,19 @@ public struct Driver: ParsableCommand {
...
+
   @Flag(
     name: [.customShort("O")],
     help: "Compile with optimizations.")
   private var optimize: Bool = false
 
   @Argument(
     transform: URL.init(fileURLWithPath:))
-  private var inputs: [URL]
+  private var inputs: [URL] = []
```
peter-evans, line 139: This change was necessary to allow commands that don't include `inputs`. Without this, `hc --version` throws during parsing the arguments because `inputs` is missing.

Sources/Driver/Driver.swift
```diff
@@ -153,6 +175,15 @@ public struct Driver: ParsableCommand {
...
   private func executeCommand(diagnostics: inout DiagnosticSet) throws {
 
+    if version {
+      standardError.write("\(hcVersion)\n")
+      return
+    }
+
+    guard !inputs.isEmpty else {
+        throw ValidationErrorWithHelp("Missing expected argument '<inputs> ...'")
+    }
```
peter-evans, line 185: I added this custom validation because `inputs` now has a default and is optional during argument parsing.

The `ValidationErrorWithHelp` custom message mimics the output when validation fails during parsing.

Tests/DriverTests/DriverTests.swift
```diff
@@ -29,7 +29,8 @@ final class DriverTests: XCTestCase {
   }
 
   func testNoInput() throws {
-    XCTAssertThrowsError(try Driver.parse([]))
+    let cli = try Driver.parse([])
+    XCTAssertThrowsError(try cli.execute())
```
peter-evans, line 33: Updated this test because now validation fails during execution, not parsing.

Sources/Driver/Driver.swift
```diff
@@ -153,6 +175,15 @@ public struct Driver: ParsableCommand {
   /// Executes the command, accumulating diagnostics in `diagnostics`.
   private func executeCommand(diagnostics: inout DiagnosticSet) throws {
 
+    if version {
+      standardError.write("\(hcVersion)\n")
```
peter-evans, line 179: All output seems to be going to standard error, but perhaps in this case it should be standard out?

Tools/set-hc-version.sh
```diff
@@ -0,0 +1,19 @@
...
+if [ "$#" -ne 1 ]; then
+    version=$(git rev-parse HEAD)
+else
+    version=$1
+fi
+
+echo -e "// DO NOT COMMIT CHANGES TO THIS FILE\nlet hcVersion = \"${version}\"" > $VERSION_FILE
+
+# Ignore local changes to Version.swift
+git update-index --assume-unchanged $VERSION_FILE
```
peter-evans, line 17: This tells git to assume the file is unchanged. It's just to try and prevent it from being committed accidentally.

peter-evans:
You'll notice if you look at the build log that the version is currently `836ec749b5f023a2425d006cff8985b438a8b947`, which doesn't match any of the commits in the PR. The reason for this is that `pull_request` events create a temporary merge commit on the runner. So what is actually being tested in `pull_request` events is a temporary local commit on the runner, which mimics the state of git after this branch is merged to `main`.

There is no problem here. When CI runs on merge to `main` it will correctly pick up the git hash of the `HEAD` of `main`.

Sources/Driver/Driver.swift
```diff
@@ -10,6 +10,23 @@ import Utils
 
 public struct Driver: ParsableCommand {
 
+  /// A validation error that includes the command's full help message.
+  struct ValidationErrorWithHelp: Error, CustomStringConvertible {
+      var message: String
```
kyouko-taiga, line 15:
```suggestion

      var message: String
```

Sources/Driver/Driver.swift
```diff
@@ -10,6 +10,23 @@ import Utils
...
+          self.message = message
+      }
+
+      var description: String {
+          """
+          \(message)
+
+          \(Driver.helpMessage())
+          """
+      }
```
kyouko-taiga, line 27:
```suggestion
      }

```

Sources/Driver/Driver.swift
```diff
@@ -153,6 +175,15 @@ public struct Driver: ParsableCommand {
   /// Executes the command, accumulating diagnostics in `diagnostics`.
   private func executeCommand(diagnostics: inout DiagnosticSet) throws {
 
+    if version {
+      standardError.write("\(hcVersion)\n")
+      return
+    }
+
+    guard !inputs.isEmpty else {
+        throw ValidationErrorWithHelp("Missing expected argument '<inputs> ...'")
```
kyouko-taiga, line 184:
```suggestion
      throw ValidationErrorWithHelp("Missing expected argument '<inputs> ...'")
```

kyouko-taiga (changes requested):
LGTM. Suggested changes are only about style. Perhaps you should also try running swift-format just to be sure.

Sources/Driver/Driver.swift
peter-evans: https://github.com/hylo-lang/hylo/pull/1320/commits/06ef911af8ee735eb294839f1aa5494768c0a841

Sources/Driver/Driver.swift
peter-evans: https://github.com/hylo-lang/hylo/pull/1320/commits/06ef911af8ee735eb294839f1aa5494768c0a841

Sources/Driver/Driver.swift
peter-evans: https://github.com/hylo-lang/hylo/pull/1320/commits/06ef911af8ee735eb294839f1aa5494768c0a841

kyouko-taiga (approved):

kyouko-taiga:
Great work! Thanks a lot.

Issue #76, opened 2022-11-09
Attribute arguments should be expressions


kyouko-taiga:
Attribute arguments should be parsed as expressions, so that we can write e.g.:
```val
@align(MemoryLayout<Int64>.alignment())
public let n: Int16
```

kyouko-taiga:
Interesting idea. I see no reason why attributes couldn't have labels, but do we have use cases in mind to justify supporting this complexity already? I do not expect attributes to be user-definable in the foreseeable future and the attributes we have now do not need labels, I think.

peter-evans:
To start with I decided to try implementing this with labels for attributes.

I have a very rough implementation which I think is working, but I ran into an interesting situation where the following test is now failing.
https://github.com/hylo-lang/hylo/blob/fc02f38c10b37bfb6f398f8c135608cca5c7878d/Tests/HyloTests/ParserTests.swift#L480-L489

I think the parser is getting confused and thinks that what comes after `@autoclosure` is the attribute's arguments, and it tries to treat `[E]() -> Int)` as an expression. Not sure how to fix it yet. 🤔

kyouko-taiga:
Normally attributes should apply on declarations, so you could try to parse `@autoclosure` as a special case in the parameter annotation parser.

kyouko-taiga:
You can parse `@autoclosure` as a special symbol in the beginning of a parameter type annotation without using `parseAttribute`. As I said, this parser is meant to parse declarations attributes and should probably be renamed `parseDeclAttribute`. `@autoclosure` is _not_ a declaration attribute. It applies on type expressions.

I think the solution is to modify `parseParameterTypeExpr` and leave `parseAttributeList` as it is. There's only one "attribute" we expect here (and maybe we'll need `@given` in the future if we change the syntax of implicit parameters).

peter-evans:
> I think the solution is to modify `parseParameterTypeExpr` and leave `parseAttributeList` as it is. There's only one "attribute" we expect here (and maybe we'll need `@given` in the future if we change the syntax of implicit parameters).

Fixed it as you described. https://github.com/peter-evans/hylo/commit/80969a4481e7119ac25a63daee7c10893aeac2da
Just one thing to note, it's now not possible to do this check. If we attempt to parse arguments we'll run into [the issue](https://github.com/hylo-lang/hylo/issues/76#issuecomment-1898820655). Should I remove `.error(attributeTakesNoArgument: a)`, since it's no longer used?
```
[...]

PR #1335, opened 2024-01-26
Parse attribute arguments as expressions

peter-evans:
Implements https://github.com/hylo-lang/hylo/issues/76

- Parses attribute arguments as expressions (`AnyExprID`).
- Parses `@autoclosure` as a special case without arguments in `parseParameterTypeExpr` due to [this issue](https://github.com/hylo-lang/hylo/issues/76#issuecomment-1898820655).
- Removes `error(attributeTakesNoArgument...` and related diagnostic test because they are no longer used.
- Extracts `foreignName` and `externalName` from `FunctionDecl` into functions in the AST


peter-evans:
Let me take a look at this and reopen again later.

PR #1336, opened 2024-01-26
Parse attribute arguments as expressions

peter-evans:
Closes https://github.com/hylo-lang/hylo/issues/76

---

Commit: https://github.com/hylo-lang/hylo/commit/b9cd50c4f3012fc939b9ffe93b521163a263aaab
- Parses attribute arguments as expressions (`AnyExprID`).
- Parses `@autoclosure` as a special case without arguments in `parseParameterTypeExpr` due to [this issue](https://github.com/hylo-lang/hylo/issues/76#issuecomment-1898820655).
- Removes `error(attributeTakesNoArgument...` and related diagnostic test because they are no longer used.
- Extracts `foreignName` and `externalName` from `FunctionDecl` into functions in the AST

Commit: https://github.com/hylo-lang/hylo/commit/cc3047c2d42b9b3dd8c17d5141d40b14000b077b
- Adds a `BundledNode` extension for `foreignName` on `FunctionDecl`

Commit: https://github.com/hylo-lang/hylo/commit/136f946aa77f5b8e942d2b38eca418413cc25818
- Adds a `BundledNode` extension for `externalName` on `FunctionDecl`


Sources/FrontEnd/BundledNode+Extensions.swift
```diff
@@ -44,3 +44,21 @@ extension BundledNode where T == SequenceExpr.ID, P == TypedProgram {
   }
 
 }
+
+extension BundledNode where T == FunctionDecl.ID, P == TypedProgram {
```
kyouko-taiga, line 48:
```suggestion
extension BundledNode where T == FunctionDecl.ID {
```
I think the property lives in the AST, so you should be able to compute it with any kind of program. Same for `externalName`

kyouko-taiga: Ah, now I see that you defined property maps for those attributes.

Sources/FrontEnd/TypedProgram.swift
```diff
@@ -44,6 +44,12 @@ public struct TypedProgram {
   /// A map from sequence expressions to their evaluation order.
   public internal(set) var foldedForm: [SequenceExpr.ID: FoldedSequenceExpr] = [:]
 
+  /// A map from functions to their foreign name.
+  public internal(set) var foreignName: [FunctionDecl.ID: String] = [:]
+
+  /// A map from functions to their external name.
+  public internal(set) var externalName: [FunctionDecl.ID: String] = [:]
```
kyouko-taiga, line 51: It's a good idea to add property maps but I think we should combine these two properties into some kind of abstraction describing all the attributes of the function that have an impact on its API. For instance, we'll soon add information about inlining too.

kyouko-taiga (changes requested):
Great work. I have only one comment about the property maps you added to `TypedProgram`.

Sources/FrontEnd/TypedProgram.swift
peter-evans: Please see this commit: https://github.com/hylo-lang/hylo/pull/1336/commits/b6614874b934384688ef1571cb0431bf17fddc11

I'm not sure if `FunctionAttributes` is the correct name for the abstraction, or even if it's in the right place in the code base.

kyouko-taiga (approved):

PR #1340, opened 2024-01-27
Move non-argument-parsing code from Driver.swift to a new file.

mrael2:
Running [SwiftLint](https://github.com/realm/SwiftLint) against Driver.swift brought up these two issues:

1. hylo/Sources/Driver/Driver.swift:507:1: warning: File Length Violation: File should contain 400 lines or less: currently contains 507 (file_length)
2. hylo/Sources/Driver/Driver.swift:11:8: error: Type Body Length Violation: Type body should span 350 lines or less excluding comments and whitespace: currently spans 351 lines (type_body_length)

To help remedy the above issues, this pull request extracts code from Driver.swift and places it in a new file: Execution.swift.

The code extracted out of Driver.swift is code that is not associated with argument parsing. The code that remains in Driver.swift is associated with argument parsing and conformance to the ParsableCommand protocol. Several unused imports can thus be removed from the top of Driver.swift.

The code placed in Execution.swift is associated with executing the primary compiler code, and the code is wrapped by a new Execution struct. The initializer takes one parameter, which is a Driver struct to get access to the parsed arguments from the Driver struct.

This is essentially a refactor to help prevent Driver.swift from growing too large. I hope it is a useful refactor. Thanks for your consideration.

### DIFF
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -1,34 +1,11 @@
 import ArgumentParser
-import CodeGenLLVM
 import Core
 import Foundation
-import FrontEnd
-import IR
-import LLVM
-import StandardLibrary
-import Utils
 
 public struct Driver: ParsableCommand {
 
-  /// A validation error that includes the command's full help message.
-  struct ValidationErrorWithHelp: Error, CustomStringConvertible {
-    var message: String
-
-    init(_ message: String) {
-      self.message = message
-    }
-
-    var description: String {
-      """
-      \(message)
-
-      \(Driver.helpMessage())
-      """
-    }
-  }
-
   /// The type of the output files to generate.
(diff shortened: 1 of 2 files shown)


dabrahams:
I'm not sure whether hylo's source base should try to satisfy `SwiftLint` or not; there should probably be a discussion and a review of its advice on the entire codebase before the project starts accepting changes to make it happy with one file.  I suggest opening a discussion [here](https://github.com/hylo-lang/hylo/discussions).  Otherwise, accepting a PR like this just creates source code churn to no real benefit.

kyouko-taiga:
A discussion was opened here: https://github.com/orgs/hylo-lang/discussions/1341

As it seems we have not reached consensus on the benefit of using SwiftLint I will close this PR. Anyone feels free to add their thoughts to the GH discussion thread and re-open this PR as necessary.

Issue #965, opened 2023-09-07
Uniformize the creation of function references


kyouko-taiga:
Declaration references to functions are typically turned into function references during IR generation. For example:

https://github.com/hylo-lang/hylo/blob/f73cfecf8024f6a58f37ccb759d6f22743663221/Sources/IR/Emitter.swift#L1267-L1268

Similar code is repeated in several places and should be refactored into a dedicated function.

PR #1344, opened 2024-02-02
Refactor function reference creation

peter-evans:
Closes https://github.com/hylo-lang/hylo/issues/965

I took a shot at the linked issue and tried to refactor creation of references. I didn't see an easy way to refactor it significantly, but perhaps this is a bit of an improvement. Sorry if it's way off base. 😅

### DIFF
--- a/Sources/IR/Analysis/Module+FunctionReferencing.swift
+++ b/Sources/IR/Analysis/Module+FunctionReferencing.swift
@@ -1,0 +1,29 @@
+import Core
+
+extension Module {
+
+  public func reference<T>(to f: T) -> FunctionReference {
+    if let f = f as? Function.ID {
+      return FunctionReference(to: f, in: self)
+    } else if let f = f as? AnyDeclID {
+      var copy = self
+      return FunctionReference(to: f, in: &copy)
+    } else {
+      preconditionFailure("unsupported function reference")
+    }
+  }
+
+  public func reference<T>(
+    to f: T, specializedBy specialization: GenericArguments, in scopeOfUse: AnyScopeID
+  ) -> FunctionReference {
+    if let f = f as? Function.ID {
+      return FunctionReference(to: f, in: self, specializedBy: specialization, in: scopeOfUse)
+    } else if let f = f as? AnyDeclID {
+      var copy = self
+      return FunctionReference(to: f, in: &copy, specializedBy: specialization, in: scopeOfUse)
+    } else {
+      preconditionFailure("unsupported function reference")
+    }
+  }
+
+}
--- a/Sources/IR/Analysis/Module+FunctionReferencing.swift
+++ b/Sources/IR/Analysis/Module+FunctionReferencing.swift
@@ -2,6 +2,7 @@ import Core
 
 extension Module {
 
+  /// Creates a reference to `f`, without specialization.
   public func reference<T>(to f: T) -> FunctionReference {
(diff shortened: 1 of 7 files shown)


Sources/IR/Analysis/Module+FunctionReferencing.swift
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  public func reference<T>(
+    to f: T, specializedBy specialization: GenericArguments, in scopeOfUse: AnyScopeID
+  ) -> FunctionReference {
+    if let f = f as? Function.ID {
+      return FunctionReference(to: f, in: self, specializedBy: specialization, in: scopeOfUse)
+    } else if let f = f as? AnyDeclID {
+      var copy = self
+      return FunctionReference(to: f, in: &copy, specializedBy: specialization, in: scopeOfUse)
```
peter-evans, line 23: The constructor of `FunctionReference` requiring `inout Module` for `AnyDeclID` makes it a little tricky to refactor. To simplify further perhaps it's worth extracting `let f = module.demandDeclaration(lowering: d)!` out and making callers convert to `Function.ID` before reference creation. 🤷‍♂️ 

```swift
public struct FunctionReference: Constant, Hashable {
...

  public init(
    to d: AnyDeclID, in module: inout Module,
    specializedBy specialization: GenericArguments, in scopeOfUse: AnyScopeID
  ) {
    let f = module.demandDeclaration(lowering: d)!
    self.init(to: f, in: module, specializedBy: specialization, in: scopeOfUse)
  }
}
```

kyouko-taiga: Giving a copy of `self` to method is the wrong approach. we do care about the mutation of the module, which is why it's taken inout. The mutation is necessary because it is possible the reference is made to a function whose IR hasn't been declared in the module yet.

Sources/IR/Analysis/Module+FunctionReferencing.swift
```diff
@@ -0,0 +1,31 @@
...
+    } else {
+      preconditionFailure("unsupported function reference")
+    }
+  }
+
+  /// Creates a reference to `f`, specialized by `specialization` in `scopeOfUse`.
+  public func reference<T>(
+    to f: T, specializedBy specialization: GenericArguments, in scopeOfUse: AnyScopeID
+  ) -> FunctionReference {
+    if let f = f as? Function.ID {
```
kyouko-taiga, line 21: We should stay away from dynamic type tests on types. That makes the signature of the method harder to read because `T` is now underspecified. Even with the proper documentation, we'd be losing the benefits of static type checking, as there's nothing to catch me at compile-time if I pass an `Int` to `reference(to:specialization:in:)`.

I would write two distinct methods here, as their implementations share no code. In situations where code is shared, I would consider defining a protocol describing shared requirements. But in any case I would start with distinct implementations.

kyouko-taiga (changes requested):
Most call sites look better, so I think this PR goes in the right direction. However I think `reference(to:specialization:in:)` requires more work.

We have to make sure module mutations are not lost. Further I would write different implementations to avoid dynamic type tests on an unconstrained generic parameter.

peter-evans:
I don't have time to work on this further at the moment so I'll close for now. If anyone wants to pick it up, feel free. I'll try to swing back and take a look again when I have time.

Issue #977, opened 2023-09-11
Interpret `_ = rhs` as a sugar for `let _ = rhs`


kyouko-taiga:
Everything is in the title.

PR #1348, opened 2024-02-07
Interpret wildcard without let as sugar for let wildcard

peter-evans:
Closes https://github.com/hylo-lang/hylo/issues/977

- Returns `.let` as the introducer from `parseBindingIntroducer` when the current token is `_`

### DIFF
--- a/Sources/FrontEnd/Parse/Parser.swift
+++ b/Sources/FrontEnd/Parse/Parser.swift
@@ -2619,6 +2619,9 @@ public enum Parser {
         site: state.range(from: introducer.site.startIndex)))
   }
 
+  /// Parses a binding introducer.
+  ///
+  /// Should not be called before checking if the next token is `_` if another wildcard rule could apply.
   private static func parseBindingIntroducer(
     in state: inout ParserState
   ) throws -> SourceRepresentable<BindingPattern.Introducer>? {
--- a/Sources/FrontEnd/Parse/Parser.swift
+++ b/Sources/FrontEnd/Parse/Parser.swift
@@ -2624,6 +2624,13 @@ public enum Parser {
   ) throws -> SourceRepresentable<BindingPattern.Introducer>? {
     guard let head = state.peek() else { return nil }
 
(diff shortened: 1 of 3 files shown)


Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2624,6 +2624,13 @@ public enum Parser {
   ) throws -> SourceRepresentable<BindingPattern.Introducer>? {
     guard let head = state.peek() else { return nil }
 
+    // Interpret `_ = rhs` as a sugar for `let _ = rhs`
```
kyouko-taiga, line 2627: That's a clever change but perhaps we should add a note somewhere to say that applying `parseBindingIntroducer` should not be called *before* the parser checked whether the next token was `_` if there's another rule that could apply. That's the case in `parsePattern`, for example.

```swift
// `_` is always a wildcard pattern.
private static func parsePattern(in state: inout ParserState) throws -> AnyPatternID? {
   if let u = state.take(.under) {
    return AnyPatternID(state.insert(WildcardPattern(site: u.site)))
  }

  // Attempt to parse a binding pattern.
  if let p = try parseBindingPattern(in: &state) {
    ...
```

Observe that if we move `parseBindingPattern` at the top of the function, the parser will commit to recognize a binding pattern and fail on inputs like `(first: _, second: let s)`.

kyouko-taiga (changes requested):
Very nice, I didn't expect it would be that simple!

I just think you should add a note in the docs of `parseBindingIntroducer` to warn about the fact that this parser now consumes `_` if it's the next token of the stream.

Sources/FrontEnd/Parse/Parser.swift
peter-evans: Added a note https://github.com/hylo-lang/hylo/pull/1348/commits/989caa0e37b00b8ca80a6ea2983a8430d41ea34e

kyouko-taiga (approved):
LGTM

PR #1359, opened 2024-02-15
Fix typos

C-BJ:
Fix spelling problems checked from https://github.com/hylo-lang/hylo/pull/1355
In the meantime, https://github.com/hylo-lang/hylo/pull/1355 should be closed


C-BJ:
OK, but btw what I'm doing in this PR is making the Github Action not clone submodule @dabrahams

dabrahams:
Oh!  In that case maybe it's not obsolete.  But you really shouldn't title the PR just "fix typos" if that's not all it's doing.  Maybe submit a new one?

PR #1379, opened 2024-03-16
update the build instructions for new LLVM-17 requirement

C-BJ:


README.md
```diff
@@ -20,7 +20,7 @@ The repository contains [submodules](https://git-scm.com/book/en/v2/Git-Tools-Su
 You can skip directly to step 3 if you're doing development exclusively in a [devcontainer](#building-a-hylo-devcontainer-with-vscode).
 Otherwise:
 
-1. Install LLVM 15 or later on your system (e.g. `brew install llvm`)
+1. Install LLVM 17 or later on your system (e.g. `brew install llvm`)
```
dabrahams, line 23: It's not "or later" because LLVM versions are not backward-compatible.  Please refer to the updated README for Swifty-LLVM and make similar changes to what you see in the Building with Swift Package Manager section.  Note also that hylo contains its own copy of `make-pkgconfig.sh` so no `swift package resolved` is needed before running it.

dabrahams (changes requested):
In addition to the inline comment, please try to title your PRs more descriptively.  In this case, “update the build instructions for new LLVM-17 requirement” would have been better.

Thanks!

PR #1397, opened 2024-03-23
Concurrency with mutation example

lucteo:
Add example/test for concurrency with mutation.

Make the spawn functionality work with the newest version of `concore2full` library.

### DIFF
--- a/Tests/EndToEndTests/TestCases/Concurrency/mutating_spawn.hylo
+++ b/Tests/EndToEndTests/TestCases/Concurrency/mutating_spawn.hylo
@@ -1,0 +1,23 @@
+//- compileToLLVM expecting: success
+
+
+fun test_mutating_spawn() -> Int {
+  var local_variable = 0
+  let p = mutable_pointer[to: &local_variable]
+
+  var future = spawn_(fun[sink let q=p.copy()] () -> Int {
+    &(q.unsafe[]) = 19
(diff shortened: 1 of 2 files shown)


Tests/EndToEndTests/TestCases/Concurrency/mutating_spawn.hylo
```diff
@@ -0,0 +1,23 @@
+//- compileToLLVM expecting: success
+
+
+fun test_mutating_spawn() -> Int {
+  var local_variable = 0
+  let p = mutable_pointer[to: &local_variable]
+
+  var future = spawn_(fun[sink let q=p.copy()] () -> Int {
+    &(q.unsafe[]) = 19
```
kyouko-taiga, line 9: The compiler should complain about immutability here. We need a copy.
```suggestion
    &(q.copy().unsafe[]) = 19
```

kyouko-taiga:
LGTM except for the small comment that I left. The compiler is too permissive.

Tests/EndToEndTests/TestCases/Concurrency/mutating_spawn.hylo
kyouko-taiga: In fact, I think your code (without the copy) won't compile once we merge #1407.

PR #1408, opened 2024-03-24
Escaping spawn

lucteo:
Add support for `spawn` in which the `await` is made in a different scope.
In this case, the frame object needs to be allocated on the heap.


StandardLibrary/Sources/Concurrency/Future.hylo
```diff
@@ -14,35 +68,32 @@ public type Future<E: Movable & Deinitializable> {
   private var r: Optional<Int>
 
   /// Initializes `self` with `f`, and spawns the computation.
-  public init(_ f: sink [E]() -> Int) {
+  public fun manual_init(_ f: sink [E]() -> Int) set {
```
kyouko-taiga, line 71: Why did you call it `manual_init`. You can call `init` "manually" as a static method if you want:

```
type A: Deinitializable {
  public init(x: Int) {}
}

public fun main() {
  var a: A
  A.`init`(self: &a, x: 1)
}
```

StandardLibrary/Sources/Concurrency/Future.hylo
```diff
@@ -4,6 +4,60 @@ public type Future<E: Movable & Deinitializable> {
...
+
+  // TODO
+  public typealias Result = Int
+
+  /// The underlying frame object that is used to connect the spawned computation with the await point.
+  private var frame: PointerToMutable<SpawnFrame<E>>
+
+  /// Initializes `self` with `f`, and spawns the computation.
+  public init(_ f: sink [E]() -> Int) {
+    &self.frame = alloc_frame(f)
```
kyouko-taiga, line 35:
```suggestion
    &self.frame = allocate_frame(f)
```

StandardLibrary/Sources/Concurrency/Future.hylo
```diff
@@ -4,6 +4,60 @@ public type Future<E: Movable & Deinitializable> {
...
+  /// Awaits the result of the computation.
+  ///
+  /// - Note: May return on a different OS thread than the one that called this.
+  public fun await() sink -> Result {
+    let r = (&frame.unsafe[]).await()
+    frame.deallocate()
+    return r
+  }
+
+  private static fun alloc_frame(_ f: sink [E]() -> Int) -> PointerToMutable<SpawnFrame<E>> {
```
kyouko-taiga, line 47: Let's try to document every function we write in the standard library, even if they are private.

kyouko-taiga (changes requested):
Nice PR, thanks. I'm not convinced that defining an initializer with a custom name is justified, but otherwise it looks good. My other comments are mostly about style.

StandardLibrary/Sources/Concurrency/Future.hylo
lucteo: Fixed. Thank you, I did not know that.

StandardLibrary/Sources/Concurrency/Future.hylo
lucteo: Moved it as internal method of `SpawnFrame`.

StandardLibrary/Sources/Concurrency/Future.hylo
lucteo: Fixed. Thank you.

kyouko-taiga (approved):
LGTM

Issue #1358, opened 2024-02-15
Support `&&` and `||` operators in compiler conditions


peter-evans:
- Support `&&` and `||` operators in compiler conditions

Follow-up to https://github.com/hylo-lang/hylo/pull/1282

Metalymph:
Just open a pull request for this #1410

PR #1410, opened 2024-03-26
Add && and || operators to compile-time compilation conditions (#1358)

Metalymph:
This commit allow the use of something like this:
```
public fun main() {
    #if arch(arm64) && false
        print("False for sure!")
    #elseif os(windows) || os(linux)
        print("Win or linux")
    #elseif arch(x86_64) || arch(arm64) && !false
        let message = "Arm64 or x86_64 (multiple operators condition)"
    #else
        print("unreachable")
    #endif
    print(message)
}
```

### DIFF
--- a/Sources/CodeGen/LLVM/LLVMProgram.swift
+++ b/Sources/CodeGen/LLVM/LLVMProgram.swift
@@ -55,7 +55,8 @@ public struct LLVMProgram {
   /// to `directory`, returning the URL of each written file.
   ///
   /// - Returns: The URL of each written product, one for each module in `self`.
-  public func write(_ type: SwiftyLLVM.CodeGenerationResultType, to directory: URL) throws -> [URL] {
+  public func write(_ type: SwiftyLLVM.CodeGenerationResultType, to directory: URL) throws -> [URL]
+  {
     precondition(directory.hasDirectoryPath)
     var result: [URL] = []
     for m in llvmModules.values {
--- a/Sources/CodeGen/LLVM/LLVMProgram.swift
+++ b/Sources/CodeGen/LLVM/LLVMProgram.swift
@@ -55,8 +55,9 @@ public struct LLVMProgram {
   /// to `directory`, returning the URL of each written file.
   ///
   /// - Returns: The URL of each written product, one for each module in `self`.
-  public func write(_ type: SwiftyLLVM.CodeGenerationResultType, to directory: URL) throws -> [URL]
-  {
+  public func write(
+    _ type: SwiftyLLVM.CodeGenerationResultType, to directory: URL
+  ) throws -> [URL] {
     precondition(directory.hasDirectoryPath)
     var result: [URL] = []
     for m in llvmModules.values {
--- a/Sources/CodeGen/LLVM/TypeLowering.swift
+++ b/Sources/CodeGen/LLVM/TypeLowering.swift
@@ -82,9 +82,9 @@ extension IR.Program {
   /// Returns the LLVM form of `val` in `module`.
   ///
   /// - Requires: `val` is representable in LLVM.
-  func llvm(boundGenericType val: BoundGenericType, in module: inout SwiftyLLVM.Module)
-    -> SwiftyLLVM.IRType
-  {
+  func llvm(
+    boundGenericType val: BoundGenericType, in module: inout SwiftyLLVM.Module
+  ) -> SwiftyLLVM.IRType {
     precondition(val[.isCanonical])
 
     let fields = base.storage(of: val.base).map { (part) in
--- a/Sources/CodeGen/LLVM/TypeLowering.swift
+++ b/Sources/CodeGen/LLVM/TypeLowering.swift
@@ -82,7 +82,9 @@ extension IR.Program {
   /// Returns the LLVM form of `val` in `module`.
   ///
   /// - Requires: `val` is representable in LLVM.
-  func llvm(boundGenericType val: BoundGenericType, in module: inout SwiftyLLVM.Module) -> SwiftyLLVM.IRType {
+  func llvm(boundGenericType val: BoundGenericType, in module: inout SwiftyLLVM.Module)
+    -> SwiftyLLVM.IRType
+  {
     precondition(val[.isCanonical])
 
     let fields = base.storage(of: val.base).map { (part) in
--- a/Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
+++ b/Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
@@ -94,14 +94,14 @@ public struct ConditionalCompilationStmt: Stmt {
 
   }
 
-  /// A sequence of structured conditions to elaborate a 1 to N operand compilation condition
-  public indirect enum SequenceCondition: Codable {
+  /// Abstracts the whole condition structure and its sub-conditions.
+  public indirect enum ConditionTree: Codable {
 
     case operand(Condition)
 
-    case and(Condition, SequenceCondition)
+    case and(ConditionTree, ConditionTree)
 
-    case or(Condition, SequenceCondition)
+    case or(ConditionTree, ConditionTree)
 
     public enum ConditionKind: String {
       case holdOnly
@@ -109,36 +109,21 @@ public struct ConditionalCompilationStmt: Stmt {
       case skipElse
     }
 
-    /// Visit the SequenceCondition tail to calculate the right case for executing the right branch
+    /// Visit the SequenceCondition tail to calculate the right case for executing the right branch.
     private func unroll(
       for factors: ConditionalCompilationFactors, conditionCase: ConditionKind
     ) -> Bool {
       switch self {
-      case .and(let cond, let seq):
-        switch condCase {
-        case .holdOnly:
-          return cond.holds(for: factors) && seq.unroll(for: factors, condCase: condCase)
-        case .skipMain:
-          return cond.mayNotNeedParsing && !cond.holds(for: factors)
-            && seq.unroll(for: factors, condCase: condCase)
-        case .skipElse:
-          return cond.mayNotNeedParsing && cond.holds(for: factors)
-            && seq.unroll(for: factors, condCase: condCase)
-        }
+      case .and(let leftSubtree, let rightSubtree):
+        return leftSubtree.unroll(for: factors, conditionCase: conditionCase)
+          && rightSubtree.unroll(for: factors, conditionCase: conditionCase)
 
-      case .or(let cond, let seq):
-        switch condCase {
-        case .holdOnly:
-          return cond.holds(for: factors) || seq.unroll(for: factors, condCase: condCase)
-        case .skipMain:
-          return cond.mayNotNeedParsing
-            || !cond.holds(for: factors) && seq.unroll(for: factors, condCase: condCase)
-        case .skipElse:
-          return cond.mayNotNeedParsing && cond.holds(for: factors)
-            || seq.unroll(for: factors, condCase: condCase)
-        }
+      case .or(let leftSubtree, let rightSubtree):
+        return leftSubtree.unroll(for: factors, conditionCase: conditionCase)
+          || rightSubtree.unroll(for: factors, conditionCase: conditionCase)
+
       case .operand(let cond):
-        switch condCase {
+        switch conditionCase {
         case .holdOnly: return cond.holds(for: factors)
         case .skipMain: return cond.mayNotNeedParsing && !cond.holds(for: factors)
         case .skipElse: return cond.mayNotNeedParsing && cond.holds(for: factors)
@@ -146,23 +131,26 @@ public struct ConditionalCompilationStmt: Stmt {
       }
     }
 
+    /// Return `true` iff the the full condition is satisfied.
     public func mustSkipMainBranch(for factors: ConditionalCompilationFactors) -> Bool {
-      self.unroll(for: factors, condCase: ConditionKind.skipMain)
+      self.unroll(for: factors, conditionCase: ConditionKind.skipMain)
     }
 
+    /// Return `true` iff the the full condition is unsatisfied.
     public func mustSkipElseBranch(for factors: ConditionalCompilationFactors) -> Bool {
-      self.unroll(for: factors, condCase: ConditionKind.skipElse)
+      self.unroll(for: factors, conditionCase: ConditionKind.skipElse)
     }
 
-    public func allHolds(for factors: ConditionalCompilationFactors) -> Bool {
-      self.unroll(for: factors, condCase: ConditionKind.holdOnly)
+    /// Return `true` iff the the full condition is satisfied.
+    fileprivate func holds(for factors: ConditionalCompilationFactors) -> Bool {
+      self.unroll(for: factors, conditionCase: ConditionKind.holdOnly)
     }
   }
 
   public let site: SourceRange
 
   /// The condition.
-  public let condition: SequenceCondition
+  public let condition: ConditionTree
 
   /// The statements in the block.
   public let stmts: [AnyStmtID]
@@ -172,7 +160,7 @@ public struct ConditionalCompilationStmt: Stmt {
 
   /// Creates an instance with the given properties.
   public init(
-    condition: SequenceCondition, stmts: [AnyStmtID], fallback: [AnyStmtID], site: SourceRange
+    condition: ConditionTree, stmts: [AnyStmtID], fallback: [AnyStmtID], site: SourceRange
   ) {
     self.site = site
     self.condition = condition
@@ -182,7 +170,7 @@ public struct ConditionalCompilationStmt: Stmt {
 
   /// Returns the statements that this expands to.
   public func expansion(for factors: ConditionalCompilationFactors) -> [AnyStmtID] {
-    condition.allHolds(for: factors) ? stmts : fallback
+    condition.holds(for: factors) ? stmts : fallback
   }
 
 }
--- a/Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
+++ b/Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
@@ -96,8 +96,11 @@ public struct ConditionalCompilationStmt: Stmt {
 
   /// A sequence of structured conditions to elaborate a 1 to N operand compilation condition
   public indirect enum SequenceCondition: Codable {
+
     case operand(Condition)
+
     case and(Condition, SequenceCondition)
+
     case or(Condition, SequenceCondition)
 
     public enum ConditionKind: String {
--- a/Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
+++ b/Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
@@ -107,8 +107,9 @@ public struct ConditionalCompilationStmt: Stmt {
     }
 
     /// Visit the SequenceCondition tail to calculate the right case for executing the right branch
-    private func unroll(for factors: ConditionalCompilationFactors, condCase: ConditionKind) -> Bool
-    {
+    private func unroll(
+      for factors: ConditionalCompilationFactors, conditionCase: ConditionKind
+    ) -> Bool {
       switch self {
       case .and(let cond, let seq):
         switch condCase {
--- a/Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
+++ b/Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
@@ -106,9 +106,13 @@ public struct ConditionalCompilationStmt: Stmt {
     case or(ConditionTree, ConditionTree)
 
     public enum ConditionKind: String {
+
       case holdOnly
+
       case skipMain
+
       case skipElse
+
     }
 
     /// Visit the SequenceCondition tail to calculate the right case for executing the right branch.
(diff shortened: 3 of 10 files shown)


Sources/CodeGen/LLVM/LLVMProgram.swift
```diff
@@ -55,7 +55,8 @@ public struct LLVMProgram {
   /// to `directory`, returning the URL of each written file.
   ///
   /// - Returns: The URL of each written product, one for each module in `self`.
-  public func write(_ type: SwiftyLLVM.CodeGenerationResultType, to directory: URL) throws -> [URL] {
+  public func write(_ type: SwiftyLLVM.CodeGenerationResultType, to directory: URL) throws -> [URL]
```
kyouko-taiga, line 58: This change seems unnecessary. Why did it happen?

kyouko-taiga: Ah, I guess it's swift-format doing its job after `LLVM` has become `SwiftyLLVM`.

Sources/CodeGen/LLVM/LLVMProgram.swift
```diff
@@ -55,7 +55,8 @@ public struct LLVMProgram {
...
+  {
```
kyouko-taiga, line 59:
```suggestion
  public func write(
    _ type: SwiftyLLVM.CodeGenerationResultType, to directory: URL
  ) throws -> [URL] {
```

It is unfortunate that swift-format doesn't apply this style (although it accepts it) but that's the one used consistently in the code base for parameter lists that exceed the character limit.

Sources/CodeGen/LLVM/TypeLowering.swift
```diff
@@ -82,7 +82,9 @@ extension IR.Program {
   /// Returns the LLVM form of `val` in `module`.
   ///
   /// - Requires: `val` is representable in LLVM.
-  func llvm(boundGenericType val: BoundGenericType, in module: inout SwiftyLLVM.Module) -> SwiftyLLVM.IRType {
+  func llvm(boundGenericType val: BoundGenericType, in module: inout SwiftyLLVM.Module)
+    -> SwiftyLLVM.IRType
+  {
```
kyouko-taiga, line 87:
```suggestion
  func llvm(
    boundGenericType val: BoundGenericType, in module: inout SwiftyLLVM.Module
  ) -> SwiftyLLVM.IRType {
```

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,70 @@ public struct ConditionalCompilationStmt: Stmt {
 
   }
 
+  /// A sequence of structured conditions to elaborate a 1 to N operand compilation condition
+  public indirect enum SequenceCondition: Codable {
+    case operand(Condition)
+    case and(Condition, SequenceCondition)
+    case or(Condition, SequenceCondition)
```
kyouko-taiga, line 101:
```suggestion
  public indirect enum SequenceCondition: Codable {

    case operand(Condition)

    case and(Condition, SequenceCondition)

    case or(Condition, SequenceCondition)
```

We add spaces after/before the opening/closing braces of types and we add spaces between member declarations. The same rule should be applied to the lines below.

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,70 @@ public struct ConditionalCompilationStmt: Stmt {
...
+
+    public enum ConditionKind: String {
+      case holdOnly
+      case skipMain
+      case skipElse
+    }
+
+    /// Visit the SequenceCondition tail to calculate the right case for executing the right branch
+    private func unroll(for factors: ConditionalCompilationFactors, condCase: ConditionKind) -> Bool
+    {
```
kyouko-taiga, line 111:
```suggestion
    private func unroll(
      for factors: ConditionalCompilationFactors, conditionCase: ConditionKind
    ) -> Bool {
```

We avoid using abbreviations in code and prefer `conditionCase` (which I guess is what you meant here) over `condCase`. The fact that I had to guess is precisely the reason why we stay away from a abbreviations.

kyouko-taiga: Looking at use sites, I suspect using simply `condition` would be a better name.

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,70 @@ public struct ConditionalCompilationStmt: Stmt {
...
+        case .holdOnly:
+          return cond.holds(for: factors) && seq.unroll(for: factors, condCase: condCase)
+        case .skipMain:
+          return cond.mayNotNeedParsing && !cond.holds(for: factors)
+            && seq.unroll(for: factors, condCase: condCase)
+        case .skipElse:
+          return cond.mayNotNeedParsing && cond.holds(for: factors)
+            && seq.unroll(for: factors, condCase: condCase)
+        }
+      case .or(let cond, let seq):
```
kyouko-taiga, line 124:
```suggestion

      case .or(let cond, let seq):
```

I would add a newline here for legibility.

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,70 @@ public struct ConditionalCompilationStmt: Stmt {
...
+    case and(Condition, SequenceCondition)
+    case or(Condition, SequenceCondition)
+
+    public enum ConditionKind: String {
+      case holdOnly
+      case skipMain
+      case skipElse
+    }
+
+    /// Visit the SequenceCondition tail to calculate the right case for executing the right branch
```
kyouko-taiga, line 109: Why not define this abstraction as a tree? We have `SequenceExpr` because the precedence and associativity of operators occurring in expressions in not known during parsing. But I don't think we want to support custom operators in conditional compilation statements. If we only have `||` and `&&`, then we can parse any expression directly into a tree rather than having to post-process a sequence.

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,70 @@ public struct ConditionalCompilationStmt: Stmt {
 
   }
 
+  /// A sequence of structured conditions to elaborate a 1 to N operand compilation condition
```
kyouko-taiga, line 97: Why not define this abstraction as a tree? We have SequenceExpr because the precedence and associativity of operators occurring in expressions in not known during parsing. But I don't think we want to support custom operators in conditional compilation statements. If we only have || and &&, then we can parse any expression directly into a tree rather than having to post-process a sequence.

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,70 @@ public struct ConditionalCompilationStmt: Stmt {
...
+      self.unroll(for: factors, condCase: ConditionKind.skipMain)
+    }
+
+    public func mustSkipElseBranch(for factors: ConditionalCompilationFactors) -> Bool {
+      self.unroll(for: factors, condCase: ConditionKind.skipElse)
+    }
+
+    public func allHolds(for factors: ConditionalCompilationFactors) -> Bool {
+      self.unroll(for: factors, condCase: ConditionKind.holdOnly)
+    }
```
kyouko-taiga, line 154: Please document these functions. Also do they really need to be `public`?

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -115,7 +177,7 @@ public struct ConditionalCompilationStmt: Stmt {
 
   /// Returns the statements that this expands to.
   public func expansion(for factors: ConditionalCompilationFactors) -> [AnyStmtID] {
-    condition.holds(for: factors) ? stmts : fallback
+    condition.allHolds(for: factors) ? stmts : fallback
```
kyouko-taiga, line 180: I don't think the change reads well. `condition.holds` is best, _even_ if `condition` must be described has a sequence. As your documentation specifies:

> A sequence of structured conditions to elaborate a 1 to N operand compilation condition

So, really, `condition` notionally describes a single condition that may or may not be composed of multiple conditions.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2993,16 +2993,38 @@ public enum Parser {
     return head != nil ? try parseCompilerConditionTail(head: head!, in: &state) : nil
   }
 
+  /// Builds the sequence of compiler conditions for the present CompilerConditionTail
+  private static func buildSequenceCondition(
```
kyouko-taiga, line 2997: You can use [precedence climbing](https://en.wikipedia.org/wiki/Operator-precedence_parser#Precedence_climbing_method) to parse a tree directly. Since we only care about `||` and `&&` we can now at parsing how to apply precedence.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2993,16 +2993,38 @@ public enum Parser {
...
+  }
+
   /// Parses a compiler condition structure, after the initial token (#if or #elseif).
   private static func parseCompilerConditionTail(
     head: Token, in state: inout ParserState
   ) throws -> AnyStmtID {
     // Parse the condition.
-    let condition = try parseCompilerCondition(in: &state)
+
+    let conditionSeq = try buildSequenceCondition(in: &state)
```
kyouko-taiga, line 3023: Please avoid using abbreviations and suffixing variables by their type.
```suggestion
    let conditions = try buildSequenceCondition(in: &state)
```

Sources/GenerateHyloFileTests/GenerateHyloFileTests.swift
```diff
@@ -32,13 +32,13 @@ struct GenerateHyloFileTests: ParsableCommand {
...
-            expectSuccess: \(p.expectSuccess))
-        }
+          func test_\(p.methodName)_\(testID)() throws {
+            try \(p.methodName)(
+              \(String(reflecting: source.fileSystemPath)),
+              expectSuccess: \(p.expectSuccess))
+          }
 
-      """
+        """
```
kyouko-taiga, line 41: Were those extra spaces added by swift-format?

kyouko-taiga (changes requested):
Thanks a lot for this PR!

I left a handful of inline comments. The most important one is that I think `SequenceCondition` should be defined as a tree rather than a sequence, and its API should not unnecessary leak details about its internal representation. Specifically, users of the abstraction typically do not care whether it is represented as a sequence, a tree, or any other data structure. For all intents and purposes, it's just a condition.

Other comments are about styling issue.

Sources/GenerateHyloFileTests/GenerateHyloFileTests.swift
Metalymph: Yes

Metalymph:
Thanks a lot for your kind review! 

I've just updated it

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,63 @@ public struct ConditionalCompilationStmt: Stmt {
 
   }
 
+  /// Abstracts the whole condition structure and its sub-conditions.
```
kyouko-taiga, line 97: I think there is a way to improve this docs.
```suggestion
  /// A condition in a conditional compilation statement expressed as a composition of predicates.
```
But at the same time it occurred to me that we could probably modify `Condition` rather than introducing a new abstraction. That's because the most natural way to document this thing is to say that a condition is a propositional statement (as in propositional logic). The propositions are the factors.

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,63 @@ public struct ConditionalCompilationStmt: Stmt {
...
+  public indirect enum ConditionTree: Codable {
+
+    case operand(Condition)
```
kyouko-taiga, line 100:
```suggestion
    /// A proposition or the negation of a proposition.
    case operand(Condition)
```

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,63 @@ public struct ConditionalCompilationStmt: Stmt {
...
+
+    case and(ConditionTree, ConditionTree)
```
kyouko-taiga, line 102:
```suggestion
    /// A conjunction of conditions.
    case and(ConditionTree, ConditionTree)
```

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,63 @@ public struct ConditionalCompilationStmt: Stmt {
...
+
+    case or(ConditionTree, ConditionTree)
```
kyouko-taiga, line 104:
```suggestion
    /// A disjunction of conditions.
    case or(ConditionTree, ConditionTree)
```

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,63 @@ public struct ConditionalCompilationStmt: Stmt {
...
+
+    public enum ConditionKind: String {
+      case holdOnly
+      case skipMain
+      case skipElse
+    }
```
kyouko-taiga, line 110:
```suggestion
    public enum ConditionKind: String {

      case holdOnly

      case skipMain

      case skipElse

    }
```
I'm not sure to understand what this abstraction is representing. Having docs would help.

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,63 @@ public struct ConditionalCompilationStmt: Stmt {
...
+
+    /// Visit the SequenceCondition tail to calculate the right case for executing the right branch.
```
kyouko-taiga, line 112:
```suggestion
    /// Visits the SequenceCondition tail to calculate the right case for executing the right branch.
```
Please document what `conditionCase` mean for this function.

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,63 @@ public struct ConditionalCompilationStmt: Stmt {
...
+      case .operand(let cond):
+        switch conditionCase {
+        case .holdOnly: return cond.holds(for: factors)
+        case .skipMain: return cond.mayNotNeedParsing && !cond.holds(for: factors)
+        case .skipElse: return cond.mayNotNeedParsing && cond.holds(for: factors)
+        }
+      }
+    }
+
+    /// Return `true` iff the the full condition is satisfied.
```
kyouko-taiga, line 134:
```suggestion
    /// Returns `true` iff the the condition is satisfied.
```
I don't think we ever care about a condition being partially satisfied.

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,63 @@ public struct ConditionalCompilationStmt: Stmt {
...
+    public func mustSkipMainBranch(for factors: ConditionalCompilationFactors) -> Bool {
+      self.unroll(for: factors, conditionCase: ConditionKind.skipMain)
+    }
+
+    /// Return `true` iff the the full condition is unsatisfied.
```
kyouko-taiga, line 139:
```suggestion
    /// Returns `true` iff the condition is not unsatisfied.
```

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
```diff
@@ -94,10 +94,63 @@ public struct ConditionalCompilationStmt: Stmt {
...
+    public func mustSkipElseBranch(for factors: ConditionalCompilationFactors) -> Bool {
+      self.unroll(for: factors, conditionCase: ConditionKind.skipElse)
+    }
+
+    /// Return `true` iff the the full condition is satisfied.
```
kyouko-taiga, line 145:
```suggestion
    /// Returns `true` iff the the condition is satisfied.
```
What is the difference between `mustSkipMainBranch(for:)` and `holds(for:)`?

Sources/Core/BuiltinFunction.swift
```diff
@@ -335,7 +335,8 @@ private func mathFlags(_ stream: inout ArraySlice<Substring>) -> NativeInstructi
 }
 
 /// Returns an overflow behavior parsed from `stream` or `.ignore` if none can be parsed.
-private func overflowBehavior(_ stream: inout ArraySlice<Substring>) -> SwiftyLLVM.OverflowBehavior {
+private func overflowBehavior(_ stream: inout ArraySlice<Substring>) -> SwiftyLLVM.OverflowBehavior
+{
```
kyouko-taiga, line 339:
```suggestion
private func overflowBehavior(
  _ stream: inout ArraySlice<Substring>
) -> SwiftyLLVM.OverflowBehavior {
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2993,16 +2993,73 @@ public enum Parser {
     return head != nil ? try parseCompilerConditionTail(head: head!, in: &state) : nil
   }
 
+  /// Parses a ConditionTree for a ConditionalCompilationStmt.
+  private static func parseConditionTree(
+    in state: inout ParserState
+  ) throws -> ConditionalCompilationStmt.ConditionTree {
+
+    /// Calculates operator's precedence.
```
kyouko-taiga, line 3001:
```suggestion
    /// Returns the precedence of `tokenValue`, which is a either `||` or `&&`.
```

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2993,16 +2993,73 @@ public enum Parser {
...
+    func precedence(_ tokenValue: String) -> Int {
+      switch tokenValue {
+      case "||": 0
+      case "&&": 1
+      default: -1
```
kyouko-taiga, line 3006: Do not use `-1` to represent invalid domains. Document your precondition and use `preconditionFailure` or `precondition` to make sure they hold in your function.

But also I suspect there is a type-safer way to write this method. Why do we have to use string input?

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2993,16 +2993,73 @@ public enum Parser {
...
+    /// Parses the tree using Precedence Climbing method.
+    func parseTree(
+      lhs: ConditionalCompilationStmt.ConditionTree,
+      minOpPrecedence: Int = 0,
+      in state: inout ParserState
+    ) throws -> ConditionalCompilationStmt.ConditionTree {
+      var lhs = lhs
+      var backup = state.backup()
+      while let op = state.take() {
+        let tokenValue = state.token(op).value
```
kyouko-taiga, line 3019: Here I think you could translate the token that you parsed into an enum instance representing possible operators (rather than a string) and then use a safer API for `precedence(_:)`.

Sources/FrontEnd/Parse/Parser.swift
```diff
@@ -2993,16 +2993,73 @@ public enum Parser {
...
+        }
+
+        if opPrecedence < minOpPrecedence {
+          break
+        }
+
+        var rhs: ConditionalCompilationStmt.ConditionTree = .operand(
+          try parseCompilerCondition(in: &state))
+
+        backup = state.backup()
```
kyouko-taiga, line 3034: Do we really need to backup all the way? I guess it should be fine to exit the first time you encounter an operator with the wrong precedence.

Sources/Core/AST/Stmt/ConditionalCompilationStmt.swift
Metalymph: Originally I wanted to separated the intents to make more clear the proper action. On this of course I can also remove the 3 specific functions and directly use `unroll` and renaming the latter as `visitWithConditionKind`

kyouko-taiga:
FYI, here's what I mean when I say that we can represent the conjunctions and disjunctions without introducing a new abstraction: https://github.com/hylo-lang/hylo/commit/bc65aac5fb181e16bade3825fbd7c92eda81c150

Metalymph:
> FYI, here's what I mean when I say that we can represent the conjunctions and disjunctions without introducing a new abstraction: [bc65aac](https://github.com/hylo-lang/hylo/commit/bc65aac5fb181e16bade3825fbd7c92eda81c150)

Ok, so you've just embedded it inside the `ConditionalCompilerStmt`. The abstraction could have been avoiding indeed, I was wrong. Thanks

Metalymph:
> @Metalymph this looks like valuable work. Do you think we can move this PR forward and get it merged soon so the value isn't lost?

Hi @dabrahams and thanks a lot to remember me about this. Essentially the correct version of this issue is been rewritten on https://github.com/hylo-lang/hylo/compare/Mtalymph-1410 by @kyouko-taiga. So this one could be also redirected and closed if there's no additional work to do on it.

dabrahams:
@kyouko-taiga I'm sorry I don't understand what @Metalymph is saying.  Perhaps you know how to sort this out?

kyouko-taiga:
I don't remember exactly but I think I opened a branch to illustrate the changes I envisioned and that's what's been linked. If @Metalymph can confirm I think we should close this PR and I can take care of pushing my branch through the finish line.

Metalymph:
> I don't remember exactly but I think I opened a branch to illustrate the changes I envisioned and that's what's been linked. If @Metalymph can confirm I think we should close this PR and I can take care of pushing my branch through the finish line.

Yes, Thanks!

PR #1419, opened 2024-03-30
Concurrent sort example.

lucteo:
Using a mock for `spawn`/`await` so that the example runs without concurrent libraries.

### DIFF
--- a/StandardLibrary/Sources/Concurrency/Future.hylo
+++ b/StandardLibrary/Sources/Concurrency/Future.hylo
@@ -46,23 +46,6 @@ public type EscapingFuture<E: Movable & Deinitializable>: Movable {
 
 }
 
-/// Mimics a future, but runs the computation inplace. Used for testing purposes.
-public type InplaceFuture<E: Movable & Deinitializable> {
-
-  /// The result of the computation.
-  let r: Int
-
-  /// Initializes `self` with `f`, and spawns the computation.
-  public init(_ f: sink [E]() -> Int) {
-    &self.r = f()
-  }
-
-  /// Awaits the result of the computation.
-  public fun await() sink -> Int {
-    return self.r
-  }
-
-}
 /// Describes the frame needed to spawn a computation.
 ///
 /// This object will be shared between the spawned computation (e.g., as it will store the result
--- a/StandardLibrary/Sources/Concurrency/Future.hylo
+++ b/StandardLibrary/Sources/Concurrency/Future.hylo
@@ -46,6 +46,23 @@ public type EscapingFuture<E: Movable & Deinitializable>: Movable {
 
 }
 
+/// Mimics a future, but runs the computation inplace. Used for testing purposes.
+public type InplaceFuture<E: Movable & Deinitializable> {
+
+  /// The result of the computation.
+  let r: Int
+
+  /// Initializes `self` with `f`, and spawns the computation.
+  public init(_ f: sink [E]() -> Int) {
+    &self.r = f()
+  }
+
+  /// Awaits the result of the computation.
+  public fun await() sink -> Int {
+    return self.r
+  }
+
+}
 /// Describes the frame needed to spawn a computation.
 ///
 /// This object will be shared between the spawned computation (e.g., as it will store the result
--- a/Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
+++ b/Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
@@ -1,0 +1,217 @@
+//- compileAndRun expecting: success
+
+let size_threshold = 16
+
+/// Sorts the elements in `a` concurrently.
+fun my_concurrent_sort<Element: Regular & Comparable>(_ a: inout ArraySlice<Element>) -> Int {
+  if a.count() < size_threshold {
+    // Use serial sort under a certain threshold.
+    a.sort()
+  } else {
+    // Partition the data, such as elements [0, mid) < [mid] <= [mid+1, n).
+    let (m1, m2) = sort_partition(&a)
+    inout (lhs, rhs) = &a.split(at: m1)
+    &rhs.drop_first(m2 - m1)
+
+    // Spawn work to sort the right-hand side.
+    let future = spawn_inplace(fun[sink let q=mutable_pointer[to: &rhs].copy()]() -> Int {
+      inout rhs = &(q.copy().unsafe[])
+      return my_concurrent_sort(&rhs)
+    })
+    // Execute the sorting on the left side, on the current thread.
+    let _ = my_concurrent_sort(&lhs)
+    let _ = future.await()
+  }
+  return a.count()
+}
+
+fun sort_partition<Element: Regular & Comparable>(_ a: inout ArraySlice<Element>) -> {Int, Int} {
+  let mid_value = median9(a)
+  return a.partition(on: mid_value)
+}
+/// Returns the median of 3 values.
+fun median3<Element: Regular & Comparable>(_ v1: Element, _ v2: Element, _ v3: Element) -> Element {
+  if v1 < v2 {
+    if v2 < v3 {
+      return v2.copy()
+    }
+    if v1 < v3 {
+      return v3.copy()
+    }
+    return v1.copy()
+  } else {
+    if v1 < v3 {
+      return v1.copy()
+    }
+    if v2 < v3 {
+      return v3.copy()
+    }
+    return v2.copy()
+  }
+}
+/// Returns a median of 9 values taken from the slice `a`.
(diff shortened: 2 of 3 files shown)


Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
```diff
@@ -0,0 +1,217 @@
...
+  precondition(n >= 8)
+  let stride = n / 8
+  let m1 = median3(a[0], a[stride], a[stride * 2])
+  let m2 = median3(a[stride * 3], a[stride * 4], a[stride * 5])
+  let m3 = median3(a[stride * 6], a[stride * 7], a[n - 1])
+  return median3(m1, m2, m3)
+}
+
+
+// TODO: this should be cleaned up and put in the standard library.
```
lucteo, line 64: I would prefer this to be done in a separate PR.

StandardLibrary/Sources/Concurrency/Future.hylo
```diff
@@ -46,6 +46,23 @@ public type EscapingFuture<E: Movable & Deinitializable>: Movable {
 
 }
 
+/// Mimics a future, but runs the computation inplace. Used for testing purposes.
```
kyouko-taiga, line 49: If this type is only used for testing then it shouldn't belong to the standard library.

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
```diff
@@ -0,0 +1,217 @@
...
+    let (m1, m2) = sort_partition(&a)
+    inout (lhs, rhs) = &a.split(at: m1)
+    &rhs.drop_first(m2 - m1)
+
+    // Spawn work to sort the right-hand side.
+    let future = spawn_inplace(fun[sink let q=mutable_pointer[to: &rhs].copy()]() -> Int {
+      inout rhs = &(q.copy().unsafe[])
+      return my_concurrent_sort(&rhs)
+    })
+    // Execute the sorting on the left side, on the current thread.
```
kyouko-taiga, line 21:
```suggestion

    // Execute the sorting on the left side, on the current thread.
```

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
```diff
@@ -0,0 +1,217 @@
...
+    let _ = my_concurrent_sort(&lhs)
+    let _ = future.await()
```
kyouko-taiga, line 23:
```suggestion
    _ = my_concurrent_sort(&lhs)
    _ = future.await()
```
That's the canonical way to discard values.

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
```diff
@@ -0,0 +1,217 @@
...
+  }
+  return a.count()
+}
+
+fun sort_partition<Element: Regular & Comparable>(_ a: inout ArraySlice<Element>) -> {Int, Int} {
+  let mid_value = median9(a)
+  return a.partition(on: mid_value)
+}
+/// Returns the median of 3 values.
```
kyouko-taiga, line 32:
```suggestion

/// Returns the median of 3 values.
```

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
```diff
@@ -0,0 +1,217 @@
...
+      return my_concurrent_sort(&rhs)
+    })
+    // Execute the sorting on the left side, on the current thread.
+    let _ = my_concurrent_sort(&lhs)
+    let _ = future.await()
+  }
+  return a.count()
+}
+
+fun sort_partition<Element: Regular & Comparable>(_ a: inout ArraySlice<Element>) -> {Int, Int} {
```
kyouko-taiga, line 28: Presumably this function is just computing a partition, right? If that's correct, it shouldn't be prefixed by `sort`.

We usually don't need to worry about polluting the global namespace in Hylo because non-public entities won't cross module boundaries. And if really we need namespacing, then we can use namespaces:
```
namespace ConcurrentSortExample {

  fun partition<Element: Regular & Comparable>(_ a: inout ArraySlice<Element>) -> {Int, Int} { ... }

}
```

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
```diff
@@ -0,0 +1,217 @@
...
+  let mid_value = median9(a)
+  return a.partition(on: mid_value)
+}
+/// Returns the median of 3 values.
+fun median3<Element: Regular & Comparable>(_ v1: Element, _ v2: Element, _ v3: Element) -> Element {
```
kyouko-taiga, line 33:
```suggestion
fun median<Element: Regular & Comparable>(_ v1: Element, _ v2: Element, _ v3: Element) -> Element {
```
The fact that there are 3 values is encoded in the signature of the function.

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
```diff
@@ -0,0 +1,217 @@
...
+      return v1.copy()
+    }
+    if v2 < v3 {
+      return v3.copy()
+    }
+    return v2.copy()
+  }
+}
+/// Returns a median of 9 values taken from the slice `a`.
+fun median9<Element: Regular & Comparable>(_ a: ArraySlice<Element>) -> Element {
```
kyouko-taiga, line 53:
```suggestion

/// Returns a median of 9 values taken from the slice `a`.
fun median9<Element: Regular & Comparable>(_ a: ArraySlice<Element>) -> Element {
```
Again, I don't think the number of values should be in the stem of the function. I understand it can't be encoded in the structure of the function here (though I suspect we could eventually take a `Int[9]` that is projected out of the slice at the call site) so I would perhaps consider using an argument label instead.

kyouko-taiga: I'm curious to know if @dabrahams has an opinion.

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
```diff
@@ -0,0 +1,217 @@
...
+  let n = a.count()
+  precondition(n >= 8)
+  let stride = n / 8
+  let m1 = median3(a[0], a[stride], a[stride * 2])
+  let m2 = median3(a[stride * 3], a[stride * 4], a[stride * 5])
+  let m3 = median3(a[stride * 6], a[stride * 7], a[n - 1])
+  return median3(m1, m2, m3)
+}
+
+
```
kyouko-taiga, line 63:
```suggestion
```

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
kyouko-taiga: Agreed. Dave has been working on a safe implementation but I think the support of remote values is not sufficient atm. I've started working on one that uses a `PointerToMutable`, like yours, but is generic over any kind of collection. I hope it will land very soon so you might want to wait a little bit.

If time is an issue you can always work with your implementation on a separate branch before we merge. I don't think it would be shocking even in a conference setting to say "all my examples can be ran on branch XYZ".

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
lucteo: Nice, I didn't know that. Thank you!

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
lucteo: Please note that this is different from the `median3` case. This one also selects the values used for computing the median from the given array slice. There is not indication in the type signature that it takes 9 values from the array slice.

StandardLibrary/Sources/Concurrency/Future.hylo
lucteo: Initially I had this in the example code, and I hesitate it whether to put it here or not.
I wanted this example to be also runnable by Dave, so I can move this back into the example code.

Tests/EndToEndTests/TestCases/Concurrency/concurrent_sort.hylo
lucteo: Fixed.

kyouko-taiga (approved):
LGTM

PR #1420, opened 2024-03-30
Add support for bulk spawn.

lucteo:
We can spawn multiple computation in parallel, at the same time.


kyouko-taiga:
The first question that popped in my mind while starting my review is: why do we need a special type to represent a collection of futures? Is there a way we could use `Array` or `Buffer`?

lucteo:
> The first question that popped in my mind while starting my review is: why do we need a special type to represent a collection of futures? Is there a way we could use `Array` or `Buffer`?

We can, but that would be inefficient. We would essentially be needing escaping futures, and thus we have one memory allocation for each computation. Using `bulk_spawn` we only need one memory allocation for the entire frame.

StandardLibrary/Sources/Concurrency/BulkFuture.hylo
```diff
@@ -0,0 +1,96 @@
+/// A future that is obtained for a bulk spawn.
+public type BulkFuture<E: Movable & Deinitializable> {
```
kyouko-taiga, line 2:
```suggestion
public type FutureSet<E: Movable & Deinitializable> {
```
or
```suggestion
public type FutureArray<E: Movable & Deinitializable> {
```
?

The term "bulk" does not work for me. Even if we can't actually use `Set`, `Array`, or `Buffer`, at least we can probably relate this abstraction to one of these collection using its name.

StandardLibrary/Sources/Concurrency/BulkFuture.hylo
```diff
@@ -0,0 +1,96 @@
...
+
+  /// The underlying frame object that is used to connect the spawned computation with the await point.
+  private var frame: PointerToMutable<BulkSpawnFrame<E>>
+
+  /// Initializes `self` with `f`, and spawns the computation.
```
kyouko-taiga, line 7:
```suggestion
  /// Initializes `self` with `action`, and spawns the computation.
```
I typically try to describe how a callback is called in the docs. We should know what `index` is and what values we can expect receiving in the lambda that we pass.

StandardLibrary/Sources/Concurrency/BulkFuture.hylo
```diff
@@ -0,0 +1,96 @@
...
+  public init(count: Int, action: sink [E](index: Int) -> Void) {
+    &self.frame = BulkSpawnFrame<E>.allocate_and_spawn(count: count, action: action)
+  }
+
+  /// Awaits the result of the computation.
+  ///
+  /// - Note: May return on a different OS thread than the one that called this.
+  public fun await() sink {
```
kyouko-taiga, line 15: I'm a little surprised that this doesn't return anything. Is it because of current compiler limitations?

I would have expected to get a set/array/buffer out of the `await` call.

StandardLibrary/Sources/Concurrency/BulkFuture.hylo
lucteo: This is not an array of futures or a set of futures, but I see why you don't like it. It a future that waits on multiple actions to complete.

I'm thinking at this entire thing as supporting "bulk spawn"; initially, I had the name `bulk_spawn_`, and only later renamed it to `spawn_`. This is why the future still has "bulk" in it.

I would need to think some more about better naming.

lucteo: Renamed to `FutureSet`.

StandardLibrary/Sources/Concurrency/BulkFuture.hylo
lucteo: Fixed.

lucteo:
Aborting. Will reimplement this entirely in Hylo.

PR #1426, opened 2024-04-02
Update .spi.yml

finestructure:
I noticed we hadn't updated our Linux images with the name change from val → hylo and so the builds (including doc generation) were failing.

This should fix everything up and also move doc gen to be run with 5.10.


kyouko-taiga:
Thanks a lot!

PR #1441, opened 2024-04-25
Support arm64 build for latest macos runners

peter-evans:
I opened [this PR](https://github.com/hylo-lang/hylo/pull/1440), to pin to `macos-12`, but then found that adding support for arm64 runners isn't a big change.

~~Note: The build for this PR might fail if this repository hasn't completed the `macos-latest` [migration](https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/) from `macos-12` to `macos-14` yet.~~
**Migration has completed because the macos-build is successful.**

PR #1439, opened 2024-04-23
CMake support

dabrahams:

PR #1443, opened 2024-04-26
Pin to macos-13 to fix build

peter-evans:
Following https://github.com/hylo-lang/hylo/pull/1441, perhaps it makes sense to pin to `macos-13` to fix the build errors on `main`, until https://github.com/hylo-lang/hylo/pull/1439 is ready.


kyouko-taiga:
Seems like a good fix to heal the CI on main for the time being. Let's merge this PR and discuss on #1441 if we want to bring the arm support or if we can simply wait for #1439 to supersede everything.

PR #1448, opened 2024-04-27
Revert "Bump KyleMayes/install-llvm-action from 1 to 2"

kyouko-taiga:
Reverts hylo-lang/hylo#1423


kyouko-taiga:
Revert is not necessary.

PR #1454, opened 2024-05-05
Expose Frontend Library through Package.swfit

tothambrus11:
This PR adds support for importing the Hylo front-end as a library through SPM builds.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -31,6 +31,7 @@ let package = Package(
   products: [
     .executable(name: "hc", targets: ["hc"]),
     .executable(name: "hylo-demangle", targets: ["hylo-demangle"]),
+    .library(name: "FrontEnd", targets: ["FrontEnd"]),
   ],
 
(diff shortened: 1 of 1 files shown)


kyouko-taiga (approved):
LGTM

Package.swift
```diff
@@ -31,6 +31,7 @@ let package = Package(
   products: [
     .executable(name: "hc", targets: ["hc"]),
     .executable(name: "hylo-demangle", targets: ["hylo-demangle"]),
+    .library(name: "FrontEnd", targets: ["FrontEnd"]),
```
kyouko-taiga, line 34: Should we given the library a better name? I guess it would be a little strange to just import "Frontend" in a code that is not hc.

dabrahams: Maybe, but that's a separate issue thst certainly shouldn't stand in the way of this PR.

PR #1460, opened 2024-05-09
Mark the distinction between syntactic and semantic type equality

kyouko-taiga:


Sources/FrontEnd/TypeChecking/TypeChecker.swift
```diff
@@ -129,9 +129,24 @@ struct TypeChecker {
     arguments.mapValues({ canonical($0, in: scopeOfUse) })
   }
 
-  /// Returns `true` iff `t` and `u` are equivalent types in `scopeOfUse`.
+  /// Returns `true` iff `t` and `u` are semantically equivalent in `scopeOfUse`.
```
dabrahams, line 132: Personally I find the change a little confusing (I'd ask myself “are two types with identical definitions but distinct names semantically equivalent?”) and would never ordinarily imagine that "are equivalent types” meant they were somehow “syntactically equivalent,” because the type is a thing independent of the syntax used to express it.

kyouko-taiga: "Syntactic" relates to the representation of the type as a term, not to it's spelling in source code. We still need a notion of syntax during semantic analysis because types are in fact expressions of some value that must be computed in context. In particular, we can think of associated types as function applications. The question is, therefore, to figure out whether `A.X` is equal to `B.Y` (or similarly whether `x(a)` is equal to `y(b)`) in a given context.

It is not enough to test whether two types terms have the same canonical form to establish equality then. That's because we need to keep these unapplied terms for conditional extensions and scoped conformances. Instead, we must ask whether they compute the same type value.

As in the lambda calculus, one subtlety is that the set of values is contained in the set of terms. For example, `Frontend.Builtin.i1` is both a type term and a type value. To make things worse, however, associated types are values _or_ terms depending on context.

dabrahams: All that said, "a and b are equivalent types" has clear meaning that doesn't depend on their spelling either in source code or as terms. The other kind of equivalence is always qualified. It's the difference between having the same value and being called by the same name. When you say "semantically equivalent" I wonder if you mean something at the notional user-model level or something having to do with execution. That's my whole point. It was clearer before the change.

dabrahams: You could say "if a and b name the same type"

PR #1462, opened 2024-05-09
Fix grammar and typos in Docs

Brian-M-J:
This pull request adds various fix-ups to the files in the `Docs` directory:

`CPlusPlus.org`
- Changed "All language rules related to C++ interop captured in that construction." to "All language rules related to C++ interop are captured in that construction."
- Changed "We could create a C++ wrapper that moves-in/moves-out, which would ensure the LoE is upheld at some efficiencly code." to "We could create a C++ wrapper that moves-in/moves-out, which would ensure that the LoE is upheld for some efficient code."
- Changed "However, if someone is wants to initialize a T in place, there may be no optional wrapper available." to "However, if someone wants to initialize a T in place, there may be no optional wrapper available."

`CodeFormatting.md`
- Changed \`swift-format\`\` to \`swift-format\`
- Separated "VisualStudio" into "Visual Studio" and "VSCode" to "VS Code"

`CompilerArchitecture.md`
- Changed "uses LLVM as it's code generation backend" to "uses LLVM as its code generation backend"
- Changed "Transforms hylo source code (Strings) to stream of distinct tokens" to "Transforms Hylo source code (Strings) to a stream of distinct tokens"
- Changed "Depending on the flags passed to the compiler, the compiler can exit early at some of these stages." to "Depending on the flags passed to the compiler, it can exit early at some of these stages."
- Changed "by it's NodeID" to "by its NodeID"
- Changed "creating the IR an inserting it" to "creating the IR and inserting it"
- Changed "theortical" to "theoretical"
- Changed "Some more passes may be necessary dependent on the target." to "Some more passes may be necessary depending on the target."
- Capitalized "Hylo"

`DefiniteInitialization.md`
- Changed "side effects on the memory" to "side effects on memory"
- Changed "DI will not catch such violation" to "DI will not catch such a violation"
- Changed "x is not definitely initialized" to "x may not be initialized" which conveys intent better

`GenericEnvironments.md`
- Changed "Constraints on associated type" to "Constraints on associated types"
- Changed "asks for the traits to which T conforms, and then proceed with quialified" to "asks for the traits to which T conforms, and then proceeds with qualified"
- Changed "As tempting is it may be" to "As tempting as it may be"
- Changed "assign blame when infer" to "assign blame when inferring"
- Changed "the information we already gather" to "the information we already gathered"
- Changed "we now" to "we know"
- Changed "a generic parameters introduced without a bound are assumed type-kinded" to "a generic parameter introduced without a bound is assumed to be type-kinded"
- Changed "generic parameter declaration denote" to "generic parameter declaration denotes"
- Changed "look how" to "look at how"

`LocalBindings.md`
- Changed "The live-range of binding's storage" to "The live-range of a binding's storage"
- Changed "must be operand" to "must be an operand"
- Changed "A var binding always have storage." to "A var binding always has storage."
- Changed "a rvalue" to "an rvalue" and "a lvalue" to "an lvalue"
- Changed "may several possible initializers" to "may have several possible initializers"
- Changed "passed as argument" to "passed as an argument"
- Changed "a optional" to "an optional"
- Changed "binding never have storage" to "binding never has storage"
- Changed "lives in register" to "lives in a register"
- Changed "Objects that live in register" to "Objects that live in registers"
- Changed "and get eventually consumed" to "and eventually get consumed"
- Changed "alloc_stack allocates a memory" to "alloc_stack allocates memory"
- Changed "inserts additional instruction" to "inserts additional instructions"

`MetatypeLowering.md`
- Changed "Name, types, and offsets" to "Names, types, and offsets"
- Changed "This, the question is" to "The question is this" which is less confusing
- Changed "Because every different types" to "Because different types"

`Performance.org`
- Changed "C++ examples were inspired from" to "C++ examples that were inspired by"
- Changed "due the sharing rules" to "due to the sharing rules"

`RemoteParts.md`
- Changed "value that it held" to "value that it held previously"
- Changed "a dotted access is always projection" to "a dotted access is always a projection"
- Capitalize "I" in "Note: it's okay to write"

`Subscripts.md`
- Changed "pre-allocated representing its stack" to "pre-allocated buffer representing its stack"
- Changed "advantage of this approach is that it requires involves" to "disadvantage of this approach is that it requires"
- Changed "identify definitions may escape a continuation" to "identify where definitions may escape a continuation"
- Changed "beyond across" to "beyond"
- Changed "inherant" to "inherent"
- Changed "continuations cannot break control flow and cannot." to "continuations cannot break control flow."


Docs/Subscripts.md
```diff
@@ -74,10 +74,10 @@ It is expected to call that lambda when it is done with the projection.
 *Note: this approach is related to  [yield-once returned-continuation lowering](https://llvm.org/docs/Coroutines.html#returned-continuation-lowering).*
 
 The second approach, which we'll call *subscripts-as-sessions*, also compiles a ramp and its slides, but the ramp returns the yielded value rather than calling a continuation with it.
-To avoid dynamic allocation, the ramp accepts a pointer to a pre-allocated representing its stack, which the caller can push on its stack before the call.
+To avoid dynamic allocation, the ramp accepts a pointer to a pre-allocated buffer representing its stack, which the caller can push on its stack before the call.
 
 The term "session" captures the intuition that the caller initiates a session by calling a ramp and terminates it by calling a slide.
-The advantage of this approach is that it requires involves code transformation.
+The advantage of this approach is that it involves code transformation.
```
Brian-M-J, line 80: Is this an advantage or disadvantage? Because requiring code transformation sounds like a disadvantage to me.

Brian-M-J:
https://github.com/Brian-M-J/hylo-typo-fixes/blob/docs-typo-fixes/Docs/Subscripts.md#using-continuations

> 1. Guarantee that continuations cannot break control flow and cannot.

This sentence seems to be incomplete.

Docs/Subscripts.md
kyouko-taiga: It is a disadvantage

kyouko-taiga:
Thanks a lot for all these changes.

> This sentence seems to be incomplete.

Indeed, we can remove the trailing "and cannot".

Docs/Subscripts.md
Brian-M-J: Updated.

Brian-M-J:
Both suggestions have been applied.

If there's nothing else, I think this pull request can be merged?

kyouko-taiga (approved):
Thanks a lot for all these fixes!

PR #1479, opened 2024-05-22
Update meeting times in README, make contributing section more apparent

tothambrus11:
I moved the contributing section to the beginning of our readme, so that it's not hidden below pages of setup guides. I also updated the meeting times, and added a time converter, so that people all around the world can quickly see when exactly the meetings happen.

I also added the slack and zoom links to CONTRIBUTING.md, which might be helpful on that page too, though I'm not entirely sure whether the redundancy is worth it, so I can remove that if needed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,6 +63,15 @@ Once everything looks good, squash those additional commits before merging.
 
 Do not hesitate to reach out if you are lost in the process.
 
+
+## Join the community
+You can also get in touch with the community by joining our [Slack][join-slack] or one of our [Zoom Meetings][join-zoom]
+(ID: 633 2146 3694, Passcode: 409180) on Tuesdays and Thursdays, 13:00-13:30 Pacific time or 22:00-22:30 Central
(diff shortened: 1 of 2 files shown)


kyouko-taiga:
Thanks!

Do you know what happened with the markdown links?

dabrahams:
Others may disagree, but I don't think moving the contributing section to the top is necessarily desirable.  I don't think it's the first thing people want when they arrive at a project.  They most likely want to try Hylo out before investing any energy in the codebase.  Regardless, IMO the section should simply point at the CONTRIBUTING document, since we have one.

Lastly, @tothambrus11 please either resolve conflicts and move this PR forward so it can be merged, or close it.  Having languishing open issues and PRs is best avoided.

tothambrus11:
Thank you for the feedback and ping.
I moved back the contribution section to its original place. The link link checker was this time actually right, it seems that the website I used for date conversion is not functioning anymore, so I removed the link for now...

dabrahams (approved):

PR #1482, opened 2024-05-31
Implement a variant of Slava Pestov's requirement machine

kyouko-taiga:


Sources/FrontEnd/TypeChecking/Rewriting/RewritingSystem.swift
```diff
@@ -0,0 +1,281 @@
...
+      return (inserted: true, ruleAfterInsertion: q)
+    }
+  }
+
+  /// Turns `self` into a terminating and confluent rewriting system, using `compareOrder` to order
+  /// pairs of terms.
+  ///
+  /// This method uses Knuth-Bendix completion algorithm to transform `self` into a terminating
+  /// confluent system. The completion procedure is semi-decidable: it returns `true` if it
+  /// succeeds or `false` if it suspects that it won't terminate.
```
dabrahams, line 101:
```suggestion
  /// Turns `self` into a terminating and confluent rewriting system using `compareOrder` to order
  /// pairs of terms and returns `true`, or returns `false` indicating that the transformation 
  /// steps reached a fixed limit without succeeding.
  ///
  /// This method uses the semi-decidable Knuth-Bendix completion algorithm, thus the
  /// transformation step limit.  If the transformation fails and returns `false`, `self` is
  /// arbitrarily mutated.
```

The idea of this suggestion is to capture the entire contract in the summary, which is almost always possible.

This is the sort of thing for which I might use a type transformation, and return an optional of the transformed type:

```swift
func completed(orderingTermsWith compareOrder: (Term, Term) -> StrictOrdering) -> CompletedSystem?
```

I don't know if that fits into the rest of your world though.

Sources/FrontEnd/TypeChecking/Rewriting/RewritingSystem.swift
```diff
@@ -0,0 +1,281 @@
...
+  mutating func complete(orderingTermsWith compareOrder: (Term, Term) -> StrictOrdering) -> Bool{
```
dabrahams, line 102:
```suggestion
  mutating func complete(orderingTermsWith compareOrder: (Term, Term) -> StrictOrdering) -> Bool {
```

dabrahams:
Merging main into this is likely to correct failures

PR #1486, opened 2024-06-04
Remove tracker from link in README.md

Brian-M-J:
Watching the "Borrow Checking Hylo" talk shouldn't require Google harvesting your data, so I removed the tracker from the link.


kyouko-taiga:
Yes, indeed!

Issue #1489, opened 2024-06-12
Synthesized SourceFile creation doesn't respect value semantics


tothambrus11:
When creating two SourceFile objects with the same file name, the second one does not get an independent storage from the first one. This was confusing because the next time we initialize a source file, it might have a completely different content at the same file name / url as the old one.

I encountered an issue while writing tests for something that involved creating source files with specific names, and I reduced the problem to the following:

```swift
  func testUsingSourceFileWithSameName() {
    let source1 = SourceFile(synthesizedText: "type A{}", named: "example.hylo")
    XCTAssertEqual(source1.text, "type A{}")

    let source2 = SourceFile(synthesizedText: "type B{}", named: "example.hylo")
    XCTAssertEqual(source2.text, "type B{}") // XCTAssertEqual failed: ("type A{}") is not equal to ("type B{}") - 
  }
```

The problem persists even if we are working within different test cases, which is particularly spooky and breaks value semantics.
```swift
  func test1UsingSourceFileWithSameName() {
    let source = SourceFile(synthesizedText: "type A{}", named: "example.hylo")
    XCTAssertEqual(source.text, "type A{}")
  }
  func test2UsingSourceFileWithSameName() {
    let source = SourceFile(synthesizedText: "type B{}", named: "example.hylo")
    XCTAssertEqual(source.text, "type B{}") // XCTAssertEqual failed: ("type A{}") is not equal to ("type B{}") - 
  }
```

What were the design decisions behind creating a shared storage of source files? Is this a bug or intended behaviour? Can we maybe make this issue less likely to happen, e.g. only making a shared storage when the contents of the two source files equal?

kyouko-taiga:
It is intended behavior. The purpose of the shared storage is to avoid duplicating source contents while deserializing an AST loaded from disk.

Clearly the name of the file should be unique and your code is not upholding this requirement. There's no reason to give explicit names to synthetic files, though. If you look at the constructor you'll see that the default argument is a random UUID.

dabrahams:
@tothambrus11 unless you have some objection we haven't considere, please close the issue.  Thanks.

tothambrus11:
Solved in https://github.com/hylo-lang/hylo/pull/1530 by removing the unnecessary parameter, so now the API is impossible to use wrong.

PR #1490, opened 2024-06-12
Document precondition of SourceFile's constructor

tothambrus11:
I added documentation for the precondition mentioned in: #1489 
When given the same `baseName` for a SourceFile's constructor as before, the provided file content must equal the one before for synthesized source files.

I added a runtime check that only runs in debug builds, so it shouldn't influence the performance of the compiler in release mode.

### DIFF
--- a/Sources/FrontEnd/SourceFile.swift
+++ b/Sources/FrontEnd/SourceFile.swift
@@ -61,9 +61,19 @@ public struct SourceFile {
   }
 
   /// Creates a synthetic source file with the specified contents and base name.
+  ///
+  /// If `baseName` is not provided, a unique name will be generated.
+  ///
+  /// - Precondition: the same baseName must not be used for different text, as
+  ///   synthesized files with the same `baseName` will share storage.
+  ///
   public init(synthesizedText text: String, named baseName: String = UUID().uuidString) {
     let storage = Storage(URL(string: "synthesized://\(baseName)")!) { text[...] }
     self.storage = storage
+
+    // The assertion is not evaluated in release builds.
+    assert(text == self.storage.text, "Precondition violated: synthesized files with the " +
+      "same baseName must have the same text globally.")
   }
 
   /// The name of the source file, sans path qualification or extension.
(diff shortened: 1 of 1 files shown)


Sources/FrontEnd/SourceFile.swift
```diff
@@ -61,9 +61,19 @@ public struct SourceFile {
   }
 
   /// Creates a synthetic source file with the specified contents and base name.
+  ///
+  /// If `baseName` is not provided, a unique name will be generated.
+  ///
+  /// - Precondition: the same baseName must not be used for different text, as
+  ///   synthesized files with the same `baseName` will share storage.
+  ///
```
kyouko-taiga, line 69:
```suggestion
  /// - Precondition: No synthetic source file named `baseName` has been created yet.
```

Sources/FrontEnd/SourceFile.swift
```diff
@@ -61,9 +61,19 @@ public struct SourceFile {
...
   public init(synthesizedText text: String, named baseName: String = UUID().uuidString) {
     let storage = Storage(URL(string: "synthesized://\(baseName)")!) { text[...] }
     self.storage = storage
+
+    // The assertion is not evaluated in release builds.
+    assert(text == self.storage.text, "Precondition violated: synthesized files with the " +
+      "same baseName must have the same text globally.")
```
kyouko-taiga, line 76:
```suggestion
    assert(text == self.storage.text, "duplicate synthesized filename")
```

tothambrus11:
Thanks for the rewording, I applied the changes.

dabrahams:
I'm confused. Didn't we say in Slack that this isn't the right fix? Don't we just want to eliminate the parameter at all allows users to bypass the UUID generation for synthesized files?

tothambrus11:
You are right, I was confused about the conclusion. @kyouko-taiga can you clarify what should be the solution?

kyouko-taiga:
The reason why the default value of the parameter can be overridden is that mangling tests rely on the source file having a specific name. If we remove this parameter we'll also have to change the way mangling is tested.

kyouko-taiga:
The agreed-upon fix has been implemented by #1530.

PR #1491, opened 2024-06-13
Expose lookup functions through TypedProgram

tothambrus11:
Two of the lookup functions from the type checker are exposed in the TypedProgram with non-mutating interface, so that they can be used from the HyloDoc library for name resolution of Hylo code references inside documentation.

The exposed functions are:
- unqualified lookup based on a stem identifier, inside a provided scope
- qualified lookup based on an identifier, a given type as a nominal scope and the scope of use that determines whether some conformances apply or not.


dabrahams:
All newly added code should be covered by tests.

tothambrus11:
Hi, I added tests, can you take a look again?

tothambrus11:
Closing this PR, as the current form of the type checker is going to be replaced in the near future.

PR #1519, opened 2024-07-02
Add support for choosing between standard library sources

tothambrus11:
Add support for choosing between standard library source taken from resources or the original sources.

Developers will need to set the HYLO_USE_BUILTIN_STDLIB environment variable to "true" if they want diagnostics to point to the original sources.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -135,7 +135,7 @@ let package = Package(
       name: "StandardLibrary",
       dependencies: ["FrontEnd", "Utils"],
       path: "StandardLibrary",
-      exclude: ["Sources"],
+      resources: [.copy("Sources")],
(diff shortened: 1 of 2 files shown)


dabrahams (approved):
Approving to get godbolt support unblocked, but needs a follow-up PR with CMake support.

dabrahams:
@tothambrus11 Is there any way we can flip the default on `HYLO_USE_BUILTIN_STDLIB`?  This really messes me up when I try to fix things because my changes get lost.

dabrahams:
If not, can you please find a way to redirect error messages to the original sources when we test in this configuration?

tothambrus11:
Yes, this should be a straightforward fix. I'll make this change tomorrow.

PR #1530, opened 2024-07-20
Remove needless parameter in constructor of 'SourceFile'

kyouko-taiga:
Supersedes #1490


kyouko-taiga:
The code coverage regression reported by codecov isn't due to the change made by this PR.

PR #1535, opened 2024-07-27
Share a typed instance of the standard library across type checking tests

kyouko-taiga:


dabrahams:
A superficial examination says this reduces testing time by 25%.  Nice!

PR #1536, opened 2024-07-27
Add support for parsing arbitrary test runner arguments

kyouko-taiga:
This PR is in preparation for future changes in the test generation framework.

### DIFF
--- a/Sources/GenerateHyloFileTests/GenerateHyloFileTests.swift
+++ b/Sources/GenerateHyloFileTests/GenerateHyloFileTests.swift
@@ -2,8 +2,8 @@ import ArgumentParser
 import Foundation
 import Utils
 
-/// A command-line tool that generates XCTest cases for a list of annotated ".hylo"
-/// files as part of our build process.
+/// A command-line tool that generates XCTest cases for a list of annotated ".hylo" files as part
+/// of our build process.
 @main
 struct GenerateHyloFileTests: ParsableCommand {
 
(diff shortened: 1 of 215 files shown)


Sources/GenerateHyloFileTests/GenerateHyloFileTests.swift
```diff
@@ -23,8 +23,8 @@ struct GenerateHyloFileTests: ParsableCommand {
     transform: URL.init(fileURLWithPath:))
   var hyloSourceFiles: [URL]
 
-  /// Returns the Swift source of the test function for the Hylo file at `source`.
-  func swiftFunctionTesting(valAt source: URL) throws -> String {
+  /// Returns the Swift source of the test function for the Hylo program at `source`.
+  func swiftFunctionTesting(hyloProgramAt source: URL) throws -> String {
```
dabrahams, line 27: I don't love this; it goes from saying something I understand to something I don't understand.  I know what it means for a file to be at a particular URL.  I have no idea what it means for a Hylo program to be at a URL though.

kyouko-taiga: The current comment is wrong. The URL might represent a directory that is the root of a Hylo module.

dabrahams:
It's very un-obvious to me that adding a leading period on the annotated expectations improves anything.  Are you planning to make the expectations valid expressions and compile them into executed Hylo code?

kyouko-taiga:
It is Swift code. The goal is to be able to pass different kinds of arguments and simplify the processing of the values so that we can generate Swift code by just regurgitating the string we have read.

I think it is actually simpler if the syntax we have up there is close to the Swift code that we eventually emit.

PR #1546, opened 2024-08-03
Add Setup Guides for Windows and Linux

tothambrus11:
I wrote detailed setup guides for Windows and Linux, which could be extended further with contributor's tips for setting up IDEs, and the sections for MacOS.

I might have done something wrong with the Windows setup because the SPM builds never succeed, even after creating the llvm.pc and adding it to the environment and reopening the terminal. The CMake builds work, but I couldn't work VSCode and the CMake Tools extension to work properly (it failed with the same message as SPM, not finding LLVM headers).


dabrahams:
SPM isn't well enough supported on Windows for us to make it work there, which is why we don't test Windows+SPM in CI.

Also XCode/SPM doesn't work for release builds on ARM.

I have deep reservations about instructing anyone to use SPM in our guides. I think we should treat it as unsupported. The only reason I haven't pushed to eliminate `Package.swift` is that some IDE tools like LSP are automatically configured to work with it.

dabrahams:
Isn't much of this information redundant with and/or in conflict with the existing build instructions which were very carefully thought out?

tothambrus11:
@dabrahams thanks for the feedback. I wrote this guide because a lot of things could and did go wrong when setting up the environment. Getting the Windows environment up and running took over one full day for me with lots of trial and error, so a linear guide would be definitely valuable there. The first time I used dev containers, it was also giving me quite some trouble, e.g. on Windows the devcontainer only starts up successfully if the repository is cloned into a WSL folder, and WSL integration needs to be manually enabled in Docker, which wasn't obvious for me.

The only possible way to use VSCode Swift toolset with integrated debugging and test case discovery is if it can build through SPM. Outside of XCode, that gave the best DX for me, and it was fairly easy to set up in dev containers. If we can set up a proper VSCode CMake project with the option to debug individual test cases using the integrated debugger, I would happily use CMake but until then, SPM through dev containers gets the job done, even if it is not well-supported on some platforms.

I will try to take another shot at setting up CMake Tools in VSCode. Have you already managed to get the builds working there with the VSCode GUI, or you only tried in XCode and in the terminal?

dabrahams:
I'm not saying you aren't providing valuable information.  I am saying that all this should be integrated somehow with the other information we already have, rather than creating redundancies and conflicts.  

I am not surprised that test discovery doesn't work with CMake, but I am deeply surprised that integrated debugging doesn't work.  We should talk about that.  I know that @compnerd uses CMake and VSCode to develop on Windows.  Also, if better VSCode support is a priority for you IMO you should figure out how to improve the CMake experience there.  I can help; I have gotten some information about how to do test discovery [here](https://github.com/swiftlang/swift-corelibs-xctest/issues/488#issuecomment-2121279923).

tothambrus11:
I agree with this, it's better not to introduce these now. I will close this PR for now and open a new one when I get CMake working, probably editing the README.

dabrahams:
I'm sorry if my comments discouraged you.  I didn't mean to imply the PR should be closed.  I just thought it needed to be updated.

PR #1561, opened 2024-08-18
Implement `MutatingSlice`

kyouko-taiga:
Everything is in the title.

### DIFF
--- a/StandardLibrary/Sources/Core/MutatingSlice.hylo
+++ b/StandardLibrary/Sources/Core/MutatingSlice.hylo
@@ -1,0 +1,74 @@
+/// A view presenting elements between two positions in a collection.
+public type MutatingSlice<Base: MutableCollection>: SemiRegular {
+
+  /// A pointer to the collection of which `self` is a slice.
+  private let base: PointerToMutable<Base>
+
+  /// The first position in `self`.
+  private let lower: Base.Position
+
+  /// The "past-the-end" position in `self`.
+  private let upper: Base.Position
+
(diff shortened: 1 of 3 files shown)


StandardLibrary/Sources/Core/MutatingSlice.hylo
```diff
@@ -0,0 +1,74 @@
...
+  /// A pointer to the collection of which `self` is a slice.
+  private let base: PointerToMutable<Base>
+
+  /// The first position in `self`.
+  private let lower: Base.Position
+
+  /// The "past-the-end" position in `self`.
+  private let upper: Base.Position
+
+  /// Creates an instance with the given source.
```
dabrahams, line 13:
```suggestion
  /// Creates an instance projecting the elements at `lower..<upper` in `base`.
```

FWIW, "the given" is usually a mistake.

StandardLibrary/Sources/Core/MutatingSlice.hylo
```diff
@@ -0,0 +1,74 @@
...
+  memberwise init
+
+  /// Returns `true` iff `self` is equal to `other`.
+  public fun infix== (_ other: Self) -> Bool {
+    (self.base == other.base) && (self.lower == other.lower) && (self.upper == other.upper)
+  }
+
+  /// Projects a slice of `base` presenting its elements from `lower` to `upper`.
```
dabrahams, line 21: A little ambiguous as to whether the range is half-open.

PR #1564, opened 2024-08-22
Add an IR instruction to switch over the contents of a union

### LINKED ISSUE #1504: Generic if-let failurea
on requirements-machine

```
extension Optional where T: Movable {
  public fun postfix!() sink -> T {
    if let r: T = self { return r }
    else { fatal_error("force unwrapping nil") }
  }
}

public fun main() {
  let x = 42 as Optional<Int>
  print(x!) // fatal error
}
```

```
extension Optional where T: Movable {
  public fun postfix!() sink -> T {
    var me = self
    if let r: T = me { return r }
    else { fatal_error("force unwrapping nil") }
  }
}

public fun main() {
  let x = 42 as Optional<Int>
  print(x!) // prints garbage, for me
}
```

kyouko-taiga:
Fixes #1504

### DIFF
--- a/Sources/CodeGen/LLVM/Transpilation.swift
+++ b/Sources/CodeGen/LLVM/Transpilation.swift
@@ -1207,16 +1207,22 @@ extension SwiftyLLVM.Module {
     /// Inserts the transpilation of `i` at `insertionPoint`.
     func insert(unionSwitch i: IR.InstructionID) {
       let s = m[i] as! UnionSwitch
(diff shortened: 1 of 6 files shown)


kyouko-taiga:
Coverage regression is due to the fact that we don't use the `switch` instruction anymore and that we only use `union_discriminator` in synthesized equality operators. We may consider removing both of these instructions but I'd rather do it with separate PRs.

Issue #1099, opened 2023-10-18
Support narrowing projections


kyouko-taiga:
Currently, narrowing conversions are always consuming. That means this program is ill-typed:
```
public fun main() {
  let x: Optional<Int> = 42
  if let y: Int = x { print(y) }
  if let z: Int = x { print(z) } // error: use of consumed object
}
```
`x` is considered consumed because its value may have been moved to initialize `y`. *(note: consumption is always assumed, whether `x` contains a value or not.)*

This behavior is annoying because it makes pattern matching more difficult to use. There are a couple of workarounds, but none of them are satisfactory:
- Copying the subject of the pattern: the copy would be inefficient, especially for large values (e.g., an `Array`). Beside, not all types are copyable.
- Moving the value back: the subject would have to be mutable. Further, this approach is verbose.

In the case of a sum type, it should be possible to project a value out of the union container. For example, given an instance of `Optional<Int>` (aka `Union<Int, None<Int>>`), it should be possible to access the `Int` stored in the container without moving it. A similar kind of narrowing projection should be possible for an existential:
```
public fun main() {
  let x: any Deinitializable = 42
  if let y: Int = x { print(y) } // should be non-consuming
}
```

Note, however, that a narrowing projection is only possible if it doesn't involve any change of representation. Fortunately, there are not so many cases to consider. I think the following conversions can be implemented as projections:
- `Union<A, B, C>` to `Union<A, B>`
- `any T & U` to `any T` (note: widening conversions should also be fine)

I think we're out of luck for narrowing conversions involving lambdas because `[Any](A) -> B` does not have the same layout as `[E](A) -> B`.

*(note: widening projections are also fine for existentials but not for union types. The issue is that there's no guarantee that there are enough spare bits in `Union<A, B>` to store the discriminator of `Union<A, B, C>`.)*
*(note: tuples are not covariant so narrowing/widening conversions do not apply.)*

PR #1568, opened 2024-08-23
Add support for non-consuming projections of union payloads

kyouko-taiga:
The following program is now legal:
```
fun f(_ x: Optional<Int>) {
  if let y: Int = x { print(y) }
}
```
Previously, the compiler would complain that `x` can't be consumed because all narrowing conversions required a `sink` access on the source.


kyouko-taiga:
This PR implements #1099.

PR #1594, opened 2024-09-17
Revert "Update CMakeModules dependency."

dabrahams:
Reverts hylo-lang/hylo#1591

Seemingly this broke the build!


dabrahams:
That was not the problem

PR #1603, opened 2024-10-09
Begin GenericDispatch document

dabrahams:
@kyouko-taiga PTAL. Content below the horizontal line is all unfinished material that may or may not eventually be used.

### DIFF
--- a/Docs/GenericDispatch.md
+++ b/Docs/GenericDispatch.md
@@ -428,7 +428,7 @@ specialization<T: P & Equatable> T: SomeAlgorithm {
 The most significant effect of an overlapping conformance ban is that
 two modules each defining a local conformance `X: P` can't
 coexist. The “orphan rule” described
-[above](#no-generalized-post-hoc-conformances) is probably the most
+[above](#no-generalized-post-hoc-conformance) is probably the most
 effective way to prevent that from happening.  Any other approach we
 know of allows situations where an update to module X can't be used
 because of a conformance conflict with some unrelated module.  These
--- a/Docs/GenericDispatch.md
+++ b/Docs/GenericDispatch.md
@@ -377,7 +377,7 @@ extension<T> T: SomeAlgorithm where T: P, T: Equatable {
 The ability to write a conformance at function scope that uses local
 values in its implementation, along with [implicit function
 parameters](https://docs.scala-lang.org/tour/implicit-parameters.html)—a
-feature Hylo takes from Scala, would solve a convenience problem
+feature Hylo takes from Scala—would solve a convenience problem
 created by strict mutable value semantics: sometimes operations on a
 part of a structure require access to the whole structure.  In a
 reference-semantic languages, one might simply store a reference to
--- a/Docs/GenericDispatch.md
+++ b/Docs/GenericDispatch.md
@@ -240,43 +240,60 @@ For `f1` to use this specialized implementation, there are two possibilities:
 1. `f1` would need to be passed a specialized witness table for
    `Y<Int>: Equatable` that is different from the one for `Y<Bool>` or
    `Y<String>`. That, in turn, would imply that `f`'s
-   separately-compiled implementation needs to look up the witness
-   table for `Y<T>: Equatable` dynamically based on the type of `x`.
+   separately-compiled implementation needs to *dynamically* look up
+   that witness table for `Y<T>: Equatable` based on the type of `x`.
 
 2. The `equals` entry in the generic `Y<T>: Equatable` witness
    table—the one that applies for all `T`s—would need to perform some
    kind of dynamic dispatch based on the type of `x`.
 
+Note that in either case, there's an additional level of dynamism
+implied beyond what is provided by dispatching through a
+statically-selected witness table.
+
 ## Overlapping Conformances
 
-The above is just one simple example of a cluster of issues that can
-arise when conformance declarations overlap.  Two conformance
-declarations overlap when, *taken separately*, each one could imply
-the conformance of a particular type to a given trait. Because of
-visibility boundaries,
-%%
-The hardest
-problems to deal with are ambiguities that are only discovered at
-runtime:
+The above is one simple example of overlapping conformances.  In
+general, two conformance declarations overlap when, *taken
+separately*, each one could imply the conformance of a particular type
+to a given trait. In this case, `Y<T>: Equatable` (for all `T`)
+overlaps with the more-specific conformance `Y<Int>: Equatable`,
+because when `T` is `Int`, either one, taken alone could apply.
+
+Ideally, the language would always select the most specific
+conformance applicable for any given type, but as we saw above, that
+implies additional dynamism.  The algorithm implementing that
+selection is approximately as expensive as doing overload resolution
+at runtime.  That cost can be mitigated by adding a global cache,
+which would require dynamic allocation and synchronization for thread
+safety. In the worst case, the algorithm would need to resolve an
+ambiguity between equally-specific conformances that can only be
+discovered at runtime.  Here's an example:
 
 ```hylo
 trait P {}
 trait Q {}
-trait R {
-  fun f()
+
+// Int conforms to both P and Q
+conformance Int: P {}
+conformance Int: Q {}
+
+trait R { static fun id() -> String }
+type X<T> {}
+
+// X conforms to R differently depending on its conformance to P or Q.
+conformance<T> X<T>: R where T: P { static fun id() { return "P" } }
+conformance<T> X<T>: R where T: Q { static fun id() { return "Q" } }
+
+fun print_id0<SomeR: R>() { print(SomeR.id()) }
+
+fun print_id<U>() {
+  // ambiguous conformance when U == Int
+  print_id0<X<U>>()
 }
-conformance Y: R where T: P { fun f() { print "A"} }
-conformance Y: R where T: Q { fun f() { print "B"} }
 
-fun callf1<T: R>(_ c: T) { c.f() }
-fun callf<U>(_ d: U) {
-  callf1(Y<U>(d)) // conformance of Y<U> to R is ambiguous when U == X
-}
 
-conformance X: P {}
-conformance X: Q {}
-
-callf(X())
+print_id<Int>()
 ```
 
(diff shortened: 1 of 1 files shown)


dabrahams:
@kyouko-taiga I think it makes sense to finish this document but re-frame it in terms of our decision to use committed choice.

kyouko-taiga:
It does. I'm a bit overwhelmed with tasks these days but I'll schedule some time to get to it asap.

dabrahams:
No rush; it's mostly just background so far.

Docs/GenericDispatch.md
```diff
@@ -0,0 +1,616 @@
...
+
+Hylo supports an optimization step that eliminates the costs of
+separate compilation without changing semantics, but that optimization
+can be selectively disabled:
+
+- globally, to improve compile times; typically for a debug build.
+- at select module boundaries, to allow binary-compatible library
+  evolution.
+
+- A concept applies to a single type, not some tuple of types.
```
kyouko-taiga, line 63: It seems like this bullet points do not follow the previous ones.

Docs/GenericDispatch.md
```diff
@@ -0,0 +1,616 @@
...
+- The concept's requirements are specified in `trait`s.
+  - Method requirements are specified in a `trait` by ordinary
+    signatures that must be matched exactly by implementations in a
```
kyouko-taiga, line 66: I wonder about how "exact" we want the implementation to match the corresponding requirement.

First, a couple of precisions:
- Matching is up to renaming of parameter names (generic and non-generic), i.e., `<T, U>(x t: T) -> U` matches `<A, B>(x a: A) -> B`.
- Matching is up to unification of associated types, i.e., `(x: Self.Element) -> Self.Position` matches `(x: T) -> Int` in the implementation of `Array<T>`'s conformance to `Collection`.

Then there's the question of matching a signature with a "subtype". For example, `(A) -> B` can implement `(A) inout -> B`. I also wonder if similar considerations should be made for `(A) -> B` implementing `(inout A) -> B`. For other pairs of passing conventions the picture is even more complicated because we need to consider `A`'s conformance to `Deinitializable`.

Docs/GenericDispatch.md
```diff
@@ -0,0 +1,616 @@
...
+
+  fun equals<Other: Sequence>(_: Other) -> Bool
+    where Self.Element == Other.Element, // type equality bound
+          Self.Element: Equatable
+
+  // requirement with default implementation
+  fun nth(_ n_: Int) -> Optional<Iter.Element> {
+    var i = iterator()
+    for n in 0..<n_ { _ = i.next() }
+    return i.next()
```
kyouko-taiga, line 109:
```suggestion
    return i.next() as _
```

Docs/GenericDispatch.md
```diff
@@ -0,0 +1,616 @@
...
+requirements and the specific types being passed to it.
+
+### Compilation Model
+
+Despite the fact that the *semantics* of generics is thought of as
+“static dispatch,” different types will have different implementations
+of any trait requirement, and therefore a separately-compiled generic
+function—a single implementation in object code that can be used
+regardless of the function's generic parameters—must access these
+implementations via a dynamic *mechanism*.
```
kyouko-taiga, line 164: I think there's something wrong with this sentence. I got lost in the middle.

Docs/GenericDispatch.md
```diff
@@ -0,0 +1,616 @@
...
+```hylo
+fun f<T: Equatable>(_ x: T) -> Bool {
+  return x.equals(x)
+}
+```
+
+desugars to
+
+```hylo
+fun f<T: Equatable>(_ x: T, _ w: EquatableWitness) -> Bool {
```
kyouko-taiga, line 179:
```suggestion
fun f<T>(_ x: T, _ w: EquatableWitness) -> Bool {
```
I think it's relevant to say that the desugaring transforms constraints into parameters. So the constraints disappear.

Docs/GenericDispatch.md
```diff
@@ -0,0 +1,616 @@
...
+the specific types passed is called *monomorphization*.  It involves
+following concrete types passed into a tree of generic component uses
+until they bottom out in concrete code again at the leaves.  Rust
+generics  are separately type checked, so that process happens before
+monomorphization, while in C++, it happens afterwards.
+
+When generics are separately *compiled*, by contrast, the implementation
+details of a generic component are unavailable to the compiler at its
+use-sites.  Only the declaration of that component is available, so
+the compiler cannot follow a type through the tree its leaves;
```
kyouko-taiga, line 207: It seems a preposition is missing here.

Docs/GenericDispatch.md
```diff
@@ -0,0 +1,616 @@
...
+The most significant effect of an overlapping conformance ban is that
+two modules each defining a local conformance `X: P` can't
+coexist. The “orphan rule” described
+[above](#no-generalized-post-hoc-conformance) is probably the most
+effective way to prevent that from happening.  Any other approach we
+know of allows situations where an update to module X can't be used
+because of a conformance conflict with some unrelated module.  These
+conformances might not even be visible in the module's public API, but
+they would have to appear in diagnostics, which is never a good
+look. Avoiding that dynamic is a primary motivator for separate
```
kyouko-taiga, line 437:
```suggestion
look. Avoiding that dynamic check is a primary motivator for separate
```

kyouko-taiga (commented):
Great document!

Docs/GenericDispatch.md
```diff
@@ -0,0 +1,616 @@
...
+    fun g<U>() -> T<Int, U>
+
+    // module C
+    import B
+    fun h<V>() {
+      // Error: result T<Int, V> depends on a conformance `Int: P`
+      // that is not in scope.
+      let x = g<V>()
+    }
+    let x = B.g<Bool>()
```
tothambrus11, line 559: I'm wondering if it would be more suitable to prevent exposing functions like g as public if they can essentially be never used from the outside.

Docs/GenericDispatch.md
dabrahams: Yeah, and like a non-throwing implementation should match one marked `throws`.  I'm dropping “exactly.”
```suggestion
    signatures that must be matched by implementations in a
```

Docs/GenericDispatch.md
dabrahams: Sorry, can you explain why?

Docs/GenericDispatch.md
dabrahams: Thanks, I'll revisit it.  It is certainly long.

Docs/GenericDispatch.md
dabrahams: good point.

Docs/GenericDispatch.md
dabrahams:
```suggestion
the compiler cannot follow a type through the tree to its leaves;
```

Docs/GenericDispatch.md
dabrahams: I really didn't meant that.
```suggestion
look. Avoiding diagnostics that point to implemetnation details
is a primary motivator for separate
```

Docs/GenericDispatch.md
dabrahams: Interesting question (cc @kyouko-taiga); that would be very consistent with what Swift does.

On the other hand, we could also choose to allow them to be used as temporary values that never have names, so, e.g.:

```hylo
    // module C
    import B
    fun ff<T>(_ x: T) {}
    
    fun h<V>() {
      ff(g<V>())  // OK <== HERE
    }
```

I actually would prefer not to, but thought I'd raise it.

Docs/GenericDispatch.md
dabrahams: yes, will fix

Docs/GenericDispatch.md
kyouko-taiga: It's good now.

Docs/GenericDispatch.md
kyouko-taiga: `g` is not public in the example and I guess it is supposed to return `X<Int, U>` (which is defined in `A`) rather than `T<Int, U>`.

So assuming it was declared as `public g<U>() -> T<Int, U>`, the declaration should be reported illegal since `T<Int, U>` is exposing a private value. I think the rationale should be that `T<Int, U>` is really `T<Int, U, w>` where `w` is the value of the private conformance, and so clearly that doesn't satisfy access controls.

Docs/GenericDispatch.md
kyouko-taiga: Certainly. Your function returns an optional and Hylo doesn't allow implicit conversion from `T` to `T?`.

dabrahams: Doesn't i.next also return an optional?

dabrahams: Doesn't i.next also return an optional?

Docs/GenericDispatch.md
kyouko-taiga: If we wanted to support such a pattern I think we'd need a way to explicitly declare an existential type (in the sense of TAPL §24, i.e. not to be confused with existential container). But then we'd be sort of implementing OCaml modules and perhaps that is a lead to explore later.

Docs/GenericDispatch.md
kyouko-taiga: Oh, it does. My bad. Then you need `&i.next()` but the `as _` is indeed useless, sorry.

dabrahams:
```suggestion
    return &i.next()
```

Docs/GenericDispatch.md
dabrahams, line 581:
```suggestion
    import A
    private conformance Int: P {
      public fun boo() -> Int { return 3 }
    }
    // Error: result T<Int, V> depends on a non-public 
    // conformance `Int: P`.
    public fun g<U>() -> X<Int, U>
```

dabrahams: Thanks all; @tothambrus11's suggestion simplifies the example.

PR #1614, opened 2024-11-12
Add a Compiler option to profile the compilation time 

adevress:
Dear all,

I have added an option to the compiler in the spirit of clang `-ftime-trace` to profile and report the time taken by each phase of the compilation of `hc`.

This is an example of such report

```bash
./Sources/hc  ../Examples/factorial.hylo  -o test --profile-compiler  
**Compile time profiling summary**
  Frontend:
    Lexer: 69.934 us
    Parser: 496.462 us
    TypeChecker: 1.537 s
  Backend:
    IR Lowering: 1.042 s
    Depolymorphize: 100.963 ms
    LLVM IR Conversion: 539.563 ms                   
    LLVM Mandatory pass: 90.476 ms                
    LLVM Optimizations: N/A
    LLVM Emit phase: 1.413 s
    Linking phase: 51.975 ms
```

If you find that useful, let me know and I will do the necessary to respect the CONTRIBUTING guidelines and iterate on that.

PS: I am not experienced in Swift. That might currently be my first PR in Swift. Do not hesitate to point any obvious wrong usage of the language.

### DIFF
--- a/Sources/FrontEnd/Parse/Lexer.swift
+++ b/Sources/FrontEnd/Parse/Lexer.swift
@@ -1,3 +1,5 @@
+import Utils
+
 /// A type that tokenize a source file.
 public struct Lexer: IteratorProtocol, Sequence {
 
@@ -7,10 +9,13 @@ public struct Lexer: IteratorProtocol, Sequence {
   /// The current position in the source file.
   private(set) var index: String.Index
 
+  private let profiler : ProfilingMeasurements?
+
   /// Creates a lexer generating tokens from the contents of `source`.
-  public init(tokenizing source: SourceFile) {
+  public init(tokenizing source: SourceFile, profileWith profiler: ProfilingMeasurements? = nil) {
     self.sourceCode = source
     self.index = source.text.startIndex
+    self.profiler = profiler
   }
 
   /// The current location of the lexer in `sourceCode`.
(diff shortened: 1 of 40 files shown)


Sources/FrontEnd/Parse/Lexer.swift
```diff
@@ -348,7 +357,7 @@ public struct Lexer: IteratorProtocol, Sequence {
       index = sourceCode.text.index(after: index)
     }
 
-    return sourceCode.text[start ..< index]
```
kyouko-taiga, line 351: I think you have applied swift-format with the wrong configuration. We're adding spaces around infix operators in this codebase.

kyouko-taiga (changes requested):
Thanks for this PR.

I find the profiling a little too intrusive. It is inconvenient that an instance of the profiler has to be passed down every data type. I would prefer if all the profiling was done in the driver, which would be a little more coarse-grained but I think is good enough.

If it isn't we should perhaps consider another approach. We can discuss it on Slack.

Minor detail: the formatting is wrong.

tothambrus11:
Thanks for this contribution! This has been implemented in the new compiler repo in the [driver functions](https://github.com/hylo-lang/hylo-new/blob/main/Sources/Driver/Driver.swift#L32), which all return a duration. I'll close this PR now as we are slowly transitioning to the new repository.

PR #1622, opened 2024-11-24
Fix semiregular

### LINKED ISSUE #1620: We have `SemiRegular` wrong
I'm not sure I believe `SemiRegular` is a useful concept; it may have fallen out of wanting to support input iterators, which really shouldn't be copyable in the first place, but…

Slide 14 of [this presentation](https://www.stepanovpapers.com/Abstraction.pdf) shows that  `SemiRegular` means, in our current world (I think) `Copyable` + `Deinitializable`.

@sean-parent can you comment on the motivation for `SemiRegular`?

dabrahams:
Fixes #1620.

Another pass to loosen some constraints now that SemiRegular refines
Copyable is warranted.


dabrahams:
@kyouko-taiga feel free to close if you believe this was superseded in the new FE repo

kyouko-taiga:
NFE will define these "umbrella traits" as trait aliases rather than new traits.

PR #1628, opened 2024-12-02
Add intrinsic support for atomic operations.

lucteo:
Add intrinsic support for atomic store and load, atomic Read-Modify-Write, CompareExchange and fence operations.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,12 +1,13 @@
 {
+  "originHash" : "95ff483b2e7dc485aa5e3a5765c4e25dd5ec98b639cdd3ff580e9f41eb484dbb",
   "pins" : [
     {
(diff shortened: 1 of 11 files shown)


kyouko-taiga (changes requested):
I strongly suspect that we can find a simpler way to represent the types of non-returning instructions. What if we say that the return type of a built-in instruction is an enum:

```swift
enum BuiltinReturnType {
  case builtin(BuiltinType)
  case void
}
```

I'll try to propose another implementation ASAP.

kyouko-taiga (approved):
Looks good mostly. Would be nice if we had more tests covering code generation.

lucteo:
Thank you very much, will add more tests in a follow-up PR.

PR #1634, opened 2024-12-20
Add a guide to update exeternal dependencies

lucteo:


Docs/Compiler/Troubleshooting.md
```diff
@@ -0,0 +1,29 @@
...
+
+Let's take the example of Swifty-LLVM, which is a Hylo repository for which we always get the latest version from `main`.
+For this example, making a change to Swifty-LLVM should be picked up automatically, after the change is submitted to `main`.
+But this doesn't always happen.
+
+To properly update the external library, do the following:
+- quit any IDEs (yes, please do that)
+- delete the build folder (e.g., `.build`)
+- delete `Package.resolved`
+- run `swift package resolve`
```
dabrahams, line 18: You don't need that step; it happens automatically.
You _do_ need a potential step that updates `Package.swift` to point at the right version of the external library.  It may be that the symbolic reference is currently what you want and that's fine.

Docs/Compiler/Troubleshooting.md
```diff
@@ -0,0 +1,29 @@
...
+- build with Swift build system (e.g., `swift build -c release`), and ensure everything builds
+- build with the CMake system and ensure that everything builds
```
dabrahams, line 20: You're missing the step that points the CMake builds at the right version of the updated library.  That means updating one of the `FindXXX` modules in https://github.com/hylo-lang/CMakeModules

Currently Swifty-LLVM points at `main` there, which is a little dicey considering that `Package.resolved` points to a specific SHA while `main` can keep changing.  _Probably_ we should repoint the `FindXXX` module for Swifty-LLVM at the `SHA` as long as we don't have stable tags.

Docs/Compiler/Troubleshooting.md
```diff
@@ -0,0 +1,29 @@
...
+- upload the changes to `Package.resolved`; this should contain the commit sha for the latest version of the library.
+
+> **_NOTE:_**
+> VS Code imediatelly re-generates the swift packages, so, depending on the order of operations, this may interfere with the actual resolving of the packages.
+> This is why we recommend closing the IDE first.
+
+> **_NOTE:_**
+> This will most probably update other packages that swift will use.
+> This is expected.
```
dabrahams, line 29: Maybe.  The most “responsible” way to do development would be for our `Package.swift` and CMake Find modules to always point at stable releases, which in turn should be pointing at stable releases of _their_ dependencies.

Docs/Compiler/Troubleshooting.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Some of my CI builds succeed, and some fail.
+
+At this point, Hylo compiler uses both a build system based on CMake, and one based on `swift build`.
+Investigate if just one of them fails, and take it from there.
+
+## Updating an external library
+
+Let's take the example of Swifty-LLVM, which is a Hylo repository for which we always get the latest version from `main`.
+For this example, making a change to Swifty-LLVM should be picked up automatically, after the change is submitted to `main`.
```
dabrahams, line 11: Not really.  `Package.resolved` is designed to lock dependent projects into specific versions of their dependencies, so things don't start unexpectedly failing due to downstream dependency branch updates.

Docs/Compiler/Troubleshooting.md
lucteo: Fixed. Thank you.

Docs/Compiler/Troubleshooting.md
lucteo: I agree that we should not point to *main*, but this is the current situation. I just wanted to describe the issue I had, which took me several hours to figure out what to do.

Docs/Compiler/Troubleshooting.md
lucteo: This is the current state of affairs, so I documented it. The goal here was to improve the documentation, not the process.

Docs/Compiler/Troubleshooting.md
lucteo: Added more clarfication. It was indeed badly written. Thank you!

Docs/Compiler/Troubleshooting.md
```diff
@@ -0,0 +1,15 @@
...
+We use Swifty-LLVM as an example of external library that needs to be updated.
+We assume that there is a change on the `main` branch of Swifty-LLVM that should be used by Hylo compiler.
+
+Steps:
+- In [`CMakeModules` repository](https://github.com/hylo-lang/CMakeModules):
+  - in the corresponding `FindXXX.cmake` file (e.g., `FindSwifty-LLVM.cmake`) edit the `GIT_TAG` value to point to the SHA of the commit that should be included in Hylo;
+  - create a PR with the changes, and ensure it's merged.
+- In `hylo` repository:
+  - in `TopLevelDefaults.cmake`, change the `GIT_TAG` value corresponding to `CMakeModules` repository to point to the SHA containing the change above;
+  - run `swift package update XXX`, where `XXX` is the name of the package that needs to be updated (`Swift-LLVM` in our example); this will update `Package.resolved` file, with the SHA of the change that needs to be brought in;
```
dabrahams, line 14: …which might not work because otherwise you wouldn't need the next step…

Docs/Compiler/Troubleshooting.md
```diff
@@ -0,0 +1,15 @@
...
+  - double check that the SHA in `Package.resolved` matches the one that was written in the `CMakeModules` repository;
```
dabrahams, line 15: Maybe we should just tell people to replace the SHA in Package.resolved… or to write the SHA into Package.swift.  IDK what's best.

Docs/Compiler/Troubleshooting.md
lucteo: I'm not following. Why wouldn't this work?

The next step is just a verification step, as the user is asked to enter two SHAs manually, in two different places. Ultimately we want to ensure that there is consistency between swift build and CMake build.

Docs/Compiler/Troubleshooting.md
lucteo: I'm not sure that would work. For example, Swify-LLVM might require some other libraries and may require other changes in `Package.resolved`.

Docs/Compiler/Troubleshooting.md
dabrahams: It might not work because the symbolic tree-ish used in Package.swift might not correspond to the SHA you've updated CMakeModules to point at.

Docs/Compiler/Troubleshooting.md
dabrahams: OK then, write the SHA into Package.swift and then do swift package resolve?

Docs/Compiler/Troubleshooting.md
lucteo: `swift package update` is independent of CMake build system. CMakeModules is not affecting this command.

dabrahams: Sorry, I think my point was that you can end up with the CMake build using a different version of the dependency than the Swift build does.  You should change them both to use the same SHA and not let `swift package update` choose a SHA for you.

lucteo: that is what the next line covers

Docs/Compiler/Troubleshooting.md
lucteo: @dabrahams : I'm not sure what should I do here. Can we spend a few minutes together during our next call? (or, if you see an easy fix, add it as a suggestion)

PR #1636, opened 2024-12-20
Update compiler architecture documentation.

lucteo:
- make the relative point to the right paths
- Add "Important files" column to the stages of compilation
- Add "Important compiler modules" section with diagram and explanations


kyouko-taiga (approved):
Amazing docs!

Issue #1534, opened 2024-07-25
For loop crashes the compiler


tothambrus11:
Iterating over a range, as well as a hand-written iterator type that conforms to the Iterator trait crashes the compiler during lowering with the attached output.

Examples:
```hylo
public fun main() -> Int {  
  for let i in 0 ..< 3 { 
     print(i)
  }
  
  return 0
}
```

```hylo

type Counter: Iterator, Deinitializable {
  public var n: Int
  public memberwise init
  public typealias Element = Int
  public fun next() inout -> Optional<Int> {
    (n.copy(), &n += 1).0 as _
  }
}

public fun main() -> Int {  
   for let i in Counter(n: 1) { 
     print(i)
     break
   }
   return 0
}
```

```
IR/Emitter.swift:1222: Fatal error: Unexpectedly found nil while unwrapping an Optional value

💣 Program crashed: Illegal instruction at 0x00007f79fa9f719e

Thread 0 "hc" crashed:

 0 0x00007f79fa9f719e closure #1 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) + 318 in libswiftCore.so
 1 0x00007f79fa9f6f4e closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) + 269 in libswiftCore.so
 2 0x00007f79fa9f6dfd closure #1 in _assertionFailure(_:_:file:line:flags:) + 380 in libswiftCore.so
 3 0x00007f79fa9f6986 _assertionFailure(_:_:file:line:flags:) + 229 in libswiftCore.so
 4 Emitter.emit(nonConsumingForStmt:) + 9369 in hc at /workspaces/hylo/Sources/IR/Emitter.swift:1222:72

  1220│     let collection = ast.core.collection
  1221│     let collectionConformance = program.conformance(
  1222│       of: domainType, to: collection.type, exposedTo: program[s].scope)!                                                            
      │                                                                       ▲
  1223│     let collectionWitness = CollectionWitness(collectionConformance, in: &module)
  1224│

 5 Emitter.emit(forStmt:) + 263 in hc at /workspaces/hylo/Sources/IR/Emitter.swift:1212:14

  1210│       UNIMPLEMENTED("consuming for loops")
  1211│     } else {
  1212│       return emit(nonConsumingForStmt: s)                                                                                           
      │              ▲
  1213│     }
  1214│   }

[...]

PR #1651, opened 2025-02-04
Add basic thread support

lucteo:


lucteo:
@kyouko-taiga : updated the argument handling for threads as we discussed (taken externally, not from the closure)

StandardLibrary/Sources/CTypes.hylo
```diff
@@ -0,0 +1,9 @@
+
+/// Type alias compatible with C `size_t` type.
+public typealias c_size_t = UInt64
```
dabrahams, line 3: Hylo types are spelled with `UpperCamelCase`.

lucteo: Fixed. Thank you!

StandardLibrary/Sources/CTypes.hylo
```diff
@@ -0,0 +1,9 @@
+
+/// Type alias compatible with C `size_t` type.
+public typealias CSizeT = UInt64
+/// Type alias compatible with C `c_ptrdiff_t` type.
+public typealias CPtrDiffT = Int64
+/// Type alias compatible with C `int` type.
+public typealias CInt = Int
+/// Type alias compatible with C `long` type.
+public typealias CLong = Int64
```
kyouko-taiga, line 9: What about that instead:

```
namespace CTypes {

  public typealias size_t = UInt64

  // ...

}
```

Also:
- I think `ptrdiff` should be `Int`, at least as long as we keep following the Swift approach of making `Int` the size of a machine word.
- I think `Int32` is a better alias for C's`int`

StandardLibrary/Sources/LibC.hylo
```diff
@@ -51,3 +51,18 @@ public fun memcpy(
 public fun memmove(
   _ destination: MemoryAddress, _ source: MemoryAddress, _ count: Int
 ) -> MemoryAddress
+
+
+/// Get system configuration for `name` at run time.
+///
+/// See https://www.man7.org/linux/man-pages/man3/sysconf.3.html.
+@ffi("sysconf")
+internal fun sysconf(name: CInt) -> CLong
```
kyouko-taiga, line 60: I am not sure I like these "C types". In Swift we just use `Int32` and I think we should do the same in Hylo.

kyouko-taiga: Maybe @dabrahams has a more informed opinion.

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
```diff
@@ -0,0 +1,76 @@
+internal let invalid_thread_id: pthread_t = 0xffffffffffffffff
```
kyouko-taiga, line 1:
```suggestion
private let invalid_thread_id: pthread_t = 0xffffffffffffffff
```
AFAICT it is only used in this file.

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  public fun deinit() {
+    precondition(id < 0, "Thread is not joined before destruction.")
+  }
+
+  public static fun spawn_thread(
+      stack: UInt,
+      executing function_ptr: MemoryAddress,
+      with_argument a: MemoryAddress) -> OSThread {
```
kyouko-taiga, line 21:
```suggestion
  public static fun spawn_thread(
      stack: UInt,
      executing function_ptr: MemoryAddress,
      with_argument a: MemoryAddress
    ) -> OSThread {
```
The rest of the indentation looks weird to me.

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
```diff
@@ -0,0 +1,76 @@
...
+    #if os(Linux) || os(macOS)
+      // TODO: improve error handling
+      var thread_id: pthread_t = invalid_thread_id.copy()
+      var attr: pthread_attr_t = .new()
+      let pattr = mutable_pointer[to: &attr]
+      if pthread_attr_init(pattr) != 0 { return .new() }
+      if stack > 0 {
+        if pthread_attr_setstacksize(pattr, CSizeT(truncating_or_extending: stack)) != 0 { return .new() }
+      }
+      if pthread_create(mutable_pointer[to: &thread_id], pattr, function_ptr, a) != 0 { return .new() }
```
kyouko-taiga, line 31:
```suggestion
      if stack > 0 {
        if pthread_attr_setstacksize(pattr, CSizeT(truncating_or_extending: stack)) != 0 { return .new() }
      } else {
        if pthread_create(mutable_pointer[to: &thread_id], pattr, function_ptr, a) != 0 { return .new() }
      }
```

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
```diff
@@ -0,0 +1,76 @@
...
+
+      return .new(id: thread_id)
+    #else
+      return .new()
+    #endif
+  }
+
+  public fun yield_now() {
+    #if os(Linux) || os(macOS)
+      let _ = sched_yield()
```
kyouko-taiga, line 41:
```suggestion
      _ = sched_yield()
```

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
```diff
@@ -0,0 +1,76 @@
...
+      #if os(Linux)
+        let _SC_NPROCESSORS_ONLN: CInt = 84
+      #elseif os(macOS)
+        let _SC_NPROCESSORS_ONLN: CInt = 58
+      #endif
+      let r = sysconf(name: _SC_NPROCESSORS_ONLN)
+      if r >= 0 {
+        return Int(truncating_or_extending: r)
+      }
+      #endif
```
kyouko-taiga, line 72:
```suggestion
    #endif
```

StandardLibrary/Sources/LowLevel/Threading/Pthread.hylo
```diff
@@ -0,0 +1,51 @@
+internal typealias pthread_t = UInt64
+
+internal type pthread_attr_t : Deinitializable {
```
kyouko-taiga, line 3: Using a buffer rdoesn't work? 
If it doesn't, then using a tuple might cut some code too.

Tests/LibraryTests/TestCases/ThreadTest.hylo
```diff
@@ -0,0 +1,92 @@
...
+}
+
+fun test_can_create_thread_with_mutating_argument() {
+  var arg = 0
+  var future = spawn_thread(executing: fun[](_ x: inout Int) { mutating_function(&x) }, with_inout_argument: mutable_pointer[to: &arg])
+  future.await()
+  precondition(arg == 13, "Expected mutation did not happen.")
+}
+
+//
```
kyouko-taiga, line 34: ?

Tests/LibraryTests/TestCases/ThreadTest.hylo
```diff
@@ -0,0 +1,92 @@
...
+  precondition(tid != 0)
+  precondition(tid != UInt64(truncating_or_extending: -1))
+  precondition(tid != current_thread_id().numeric_value())
+}
+
+fun test_can_yield_thread() {
+  var future = spawn_thread(executing: fun[]() -> Void {
+    let i = 0
+    print("Before yielding loop")
+    while i < 1000 {
```
kyouko-taiga, line 69: I think for loops work, i.e., `for _ in 0 ..< 1000`

kyouko-taiga (changes requested):
Not a fan of the type aliases but the rest is very nice. I left a few inline suggestions.

StandardLibrary/Sources/LibC.hylo
lucteo: Having this type translation helps a lot in interfacing with C. Decisions of which Hylo types maps to a C type should not be taken for each argument; we should take them once, and then just use that decision.

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
lucteo: the `if pthread_create...` needs to happen also in the case of `stack > 0`. We shouldn't have an `else` in here.

StandardLibrary/Sources/LibC.hylo
kyouko-taiga: But we can make this determination once and for all, document it somewhere, and be done with it. Like `int = Int32`, `long= Int64`, etc. We do not need the `CInt` alias.

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
kyouko-taiga: OK, sorry my bad.

StandardLibrary/Sources/LowLevel/Threading/Pthread.hylo
lucteo: created #1680; currently it blocks the progress of using buffers

Tests/LibraryTests/TestCases/ThreadTest.hylo
lucteo: It doesn't work, because of #1534

StandardLibrary/Sources/CTypes.hylo
kyouko-taiga: Also I just realized but if the purpose is to write these aliases in such a way that they are are close as possible to the machine we should use the built-in types, those that are implemented by LLVM.

StandardLibrary/Sources/CTypes.hylo
```diff
@@ -0,0 +1,13 @@
+
+namespace CTypes {
```
kyouko-taiga, line 2:
```suggestion
namespace C {
```

Eventually that will be its own namespace separate from the standard library I suppose. The idea is that we should be able to write `Hylo.Int` and `C.int`.

StandardLibrary/Sources/LowLevel/Threading/Pthread.hylo
kyouko-taiga: Can be resolved right?

StandardLibrary/Sources/LowLevel/Threading/Pthread.hylo
```diff
@@ -0,0 +1,26 @@
...
+@ffi("pthread_attr_destroy")
+internal fun pthread_attr_destroy(_ attr: PointerToMutable<pthread_attr_t>) -> CTypes.int
+@ffi("pthread_create")
+internal fun pthread_create(_ thread: PointerToMutable<pthread_t>, _ attr: PointerToMutable<pthread_attr_t>, _ f: MemoryAddress, _ arg: MemoryAddress) -> CTypes.int
+@ffi("pthread_join")
+internal fun pthread_join(_ thread: pthread_t, _ ret: PointerToMutable<MemoryAddress>) -> CTypes.int
+@ffi("pthread_detach")
+internal fun pthread_detach(_ thread: pthread_t) -> CTypes.int
+@ffi("pthread_self")
+internal fun pthread_self() -> pthread_t
```
kyouko-taiga, line 26:
```suggestion
@ffi("pthread_attr_init")
internal fun pthread_attr_init(_ attr: PointerToMutable<pthread_attr_t>) -> CTypes.int

@ffi("pthread_attr_setstacksize")
internal fun pthread_attr_setstacksize(_ attr: PointerToMutable<pthread_attr_t>, _ size: CTypes.size_t) -> CTypes.int

@ffi("pthread_attr_destroy")
internal fun pthread_attr_destroy(_ attr: PointerToMutable<pthread_attr_t>) -> CTypes.int

@ffi("pthread_create")
internal fun pthread_create(_ thread: PointerToMutable<pthread_t>, _ attr: PointerToMutable<pthread_attr_t>, _ f: MemoryAddress, _ arg: MemoryAddress) -> CTypes.int

@ffi("pthread_join")
internal fun pthread_join(_ thread: pthread_t, _ ret: PointerToMutable<MemoryAddress>) -> CTypes.int

@ffi("pthread_detach")
internal fun pthread_detach(_ thread: pthread_t) -> CTypes.int

@ffi("pthread_self")
internal fun pthread_self() -> pthread_t
```

StandardLibrary/Sources/LowLevel/Threading/Thread.hylo
```diff
@@ -0,0 +1,138 @@
...
+    &id = .new(truncating_or_extending: os_value)
+  }
+
+  /// Returns `self` as a numeric value.
+  public fun numeric_value() -> UInt64 {
+    UInt64(truncating_or_extending: id)
+  }
+
+}
+conformance ThreadId: Regular {}
```
kyouko-taiga, line 22:
```suggestion
public type ThreadId: Regular {

  /// The underlying thread ID, as seen by the operating system.
  var id: pthread_t

  /// Initializes `self` with the given operating system thread ID.
  // TODO: usable from inline; needs to be internal
  public init(os_value: pthread_t) {
    &id = .new(truncating_or_extending: os_value)
  }

  /// Returns `self` as a numeric value.
  public fun numeric_value() -> UInt64 {
    UInt64(truncating_or_extending: id)
  }

}
```

Synthesized conformances declared outside of the primary type will be removed eventually. They are in conflict with the idea that a structure is "opaque".

StandardLibrary/Sources/LowLevel/Threading/Thread.hylo
```diff
@@ -0,0 +1,138 @@
...
+public fun spawn_thread<A>(executing f: sink [](A) -> Void, with_argument a: Pointer<A>) -> ThreadCompletion<A> {
+  .new(stack: 0, executing: f, with_argument: a)
+}
+
+/// Creates a new thread that will execute `f` with inout argument `a`.
+public fun spawn_thread<A>(executing f: sink [](inout A) -> Void, with_inout_argument a: PointerToMutable<A>) -> ThreadCompletion<A> {
+  .new(stack: 0, executing: f, with_inout_argument: a)
+}
+
+
```
kyouko-taiga, line 107:
```suggestion
```

kyouko-taiga (changes requested):
A few minor comments but it looks quite good already!

StandardLibrary/Sources/LibC.hylo
lucteo: resolved the discussion in the meeting. Keeping this as it is for this PR, but we will work out a system of better C types. Resolving conversation.

StandardLibrary/Sources/LibC.hylo
```diff
@@ -51,3 +51,18 @@ public fun memcpy(
 public fun memmove(
   _ destination: MemoryAddress, _ source: MemoryAddress, _ count: Int
 ) -> MemoryAddress
+
+
+/// Get system configuration for `name` at run time.
```
kyouko-taiga, line 56:
```suggestion
/// Returns the system configuration for `name` at runtime.
```

We use the 3rd person in the one-line docs of functions and subscripts.

StandardLibrary/Sources/LowLevel/Threading/Thread.hylo
```diff
@@ -0,0 +1,136 @@
...
+  /// Returns the address of `x` as a `MemoryAddress`.
+  private static fun take_address<T>(_ x: inout T) -> MemoryAddress {
+    return MemoryAddress(type_punning: mutable_pointer[to: &x])
+  }
+
+  /// Dereferences `o` and returns the value as a pointer.
+  private static fun deref(_ p: MemoryAddress) -> MemoryAddress {
+    return PointerToMutable<MemoryAddress>(type_punning: p).unsafe_pointee()
+  }
```
kyouko-taiga, line 87: I guess that's fine for the time being but we should really not commit too strongly about any particular representation of lambdas, especially because all these pointer shenanigans sounds poorly portable.

Could you open an issue explaining the need for a way to extract a function pointer out of a lambda? I think that's rather easy to implement as a built-in feature (famous last words) and that would get rid of this scary code. Not a blocker for this PR, though.

StandardLibrary/Sources/LowLevel/Threading/Thread.hylo
```diff
@@ -0,0 +1,136 @@
...
+  .new(stack: 0, executing: f, with_argument: a)
+}
+
+/// Creates a new thread that will execute `f` with inout argument `a`.
+public fun spawn_thread<A>(executing f: sink [](inout A) -> Void, with_inout_argument a: PointerToMutable<A>) -> ThreadCompletion<A> {
+  .new(stack: 0, executing: f, with_inout_argument: a)
+}
+
+/// Creates a new thread that will execute `f` with no arguments, using a stack size of `stack_size`.
+public fun spawn_thread(stack_size: UInt, executing f: sink []() -> Void) -> ThreadCompletion<Void> {
```
kyouko-taiga, line 107: I am a bit surprised to see `UInt` here. We should commit to a decision but I think we're heading toward a design where `Int` is the general-purpose type we'd use for all numbers, including sizes and indices. Si we'd use `Int` here and apply conversions as necessary to call into a C API expecting unsigned integers.

Or, if this API is supposed to be very low-level, it should use C types and push the conversion requirements to the user.

kyouko-taiga (changes requested):
Looking better and better. I have some concerns about some of the API. I left a few comments but I also think we should discuss some general principles in one of our meetings.

I also think that some of the types and functions require more documentation. That includes the interfaces for lib C. We may also discuss in the next meeting what should be our guidelines for documenting C functions. Maybe that's a topic that may interest @tothambrus11 .

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /// Initializes `self` with `f`.
+  internal init<A>(_ f: sink [](inout A) -> Void) {
+    &self.type_erased = type_erase(&f)
+  }
+
+  /// Type erases `f` transforming it into a `TypeErasedFunctionType`; works for our supported closure types.
+  private static fun type_erase<Closure>(_ f: inout Closure) -> TypeErasedFunctionType {
+    return PointerToMutable<TypeErasedFunctionType>(type_punning: mutable_pointer[to: &f]).unsafe_pointee()
```
kyouko-taiga, line 40:
```suggestion
    PointerToMutable<TypeErasedFunctionType>(type_punning: mutable_pointer[to: &f]).unsafe_pointee()
```

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /// Returns a raw pointer from `self`.
+  internal fun as_raw_pointer() -> MemoryAddress {
+    return Pointer<MemoryAddress>(type_punning: pointer[to: type_erased]).unsafe[].copy()
```
kyouko-taiga, line 45:
```suggestion
    Pointer<MemoryAddress>(type_punning: pointer[to: type_erased]).unsafe[].copy()
```

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+}
+
```
kyouko-taiga, line 49:
```suggestion
```

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
```diff
@@ -0,0 +1,135 @@
...
+    precondition(id < 0, "Thread is not joined before destruction.")
+  }
+
+  /// Spawn a new thread with `stack` bytes of stack space, executing `function_ptr`, and passing the value `a` as an argument.
+  /// - Requires: `stack >= 0`.
+  public static fun spawn_thread(
+      stack: Int,
+      executing function: ThreadFunction,
+      with_argument a: MemoryAddress
+    ) -> OSThread {
```
kyouko-taiga, line 76:
```suggestion
    stack: Int,
    executing function: ThreadFunction,
    with_argument a: MemoryAddress
  ) -> OSThread {
```

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
```diff
@@ -0,0 +1,135 @@
...
+    #if os(Linux) || os(macOS)
+      // TODO: improve error handling
+      var thread_id: pthread_t = invalid_thread_id.copy()
+      var attr: pthread_attr_t = .new()
+      let pattr = mutable_pointer[to: &attr]
+      if pthread_attr_init(pattr) != 0 { return .new() }
+      if stack > 0 {
+        if pthread_attr_setstacksize(pattr, C.size_t(truncating_or_extending: stack)) != 0 { return .new() }
```
kyouko-taiga, line 84: Is the imbrication a workaround for the current lack of short-circuit operator? If so perhaps we should consider this pattern:
```
if a,b { /* `b` evaluated iff `a` is true */ }
```

StandardLibrary/Sources/LowLevel/Threading/Thread.hylo
```diff
@@ -0,0 +1,134 @@
...
+    &id = .new(truncating_or_extending: os_value)
+  }
+
+  /// Returns `self` as a numeric value.
+  public fun numeric_value() -> UInt64 {
+    UInt64(truncating_or_extending: id)
+  }
+
+}
+
```
kyouko-taiga, line 22:
```suggestion
```

kyouko-taiga (approved):
Great job. All the comments I left are nitpicks.

StandardLibrary/Sources/LowLevel/Threading/OSThread.hylo
lucteo: Yes. But for my taste, I would be hesitant to put functions with side-effects in short-circuited boolean expressions, as it somehow masks the side-effect.

PR #1653, opened 2025-02-06
Use SwiftyLab setup-swift everywhere.

dabrahams:


tothambrus11:
Addressed in https://github.com/hylo-lang/hylo/pull/1687

PR #1658, opened 2025-02-13
Simplify IRGen

dabrahams:
This removes the module methods that generate instructions and puts the functionality directly into instruction constructors.  The simplifications probably will be subsumed by whatever EDSL we end up with.  Willing to close if you don't think it's a win, @kyouko-taiga .


kyouko-taiga:
I think we have a design for a better approach already, so perhaps we should not go ahead with this changes. WDYT?

PR #1666, opened 2025-03-10
Fix a bug that would cause deinitializers to run multiple times

kyouko-taiga:


tothambrus11:
I will take a look why the CI is failing on Mac. It seems that the System library is not found for some reason.

PR #1673, opened 2025-04-14
Add `BoundedArray` in the standard library.

lucteo:


StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,280 @@
+/// An ordered, random-access collection that has a fixed maximum size.
```
dabrahams, line 1:
```suggestion
/// A random-access collection with contiguous memory and a fixed maximum size.
```

All collections have an order.

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,280 @@
...
+///
+/// Similar to `Array`, but with a fixed maximum size.
+/// By default, the element type is not required to be `Movable`.
```
dabrahams, line 4:
```suggestion
/// Unlike `Array`, this collection is suitable for non-`Movable` elements.
```

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,280 @@
...
+public type BoundedArray<Element: Deinitializable>: SemiRegular {
+
+  /// The out-of-line storage of the array.
```
dabrahams, line 7:
```suggestion
  /// The dynamically-allocated storage for all parts, including elements.
```

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,280 @@
...
+  ///
+  /// The header of the buffer indicates the number of elements contained in the array.
```
dabrahams, line 9:
```suggestion
  /// The storage for the length and elements.
```

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,280 @@
...
+  var storage: DynamicBuffer<Int, Element>
+
+  /// Creates an empty array.
```
dabrahams, line 12:
```suggestion
  /// An empty instance.
```

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
+/// A random-access collection with contiguous memory, stable element addresses,
+/// and a fixed maximum size.
```
kyouko-taiga, line 2:
```suggestion
/// and a fixed capacity.
```

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+  /// Creates an array that can hold `capacity` elements, initially empty.
+  ///
+  /// - Requires: `capacity >= 0`.
+  public init(capacity: Int) {
+    precondition(capacity >= 0, "negative array capacity")
+    &storage = new_storage(capacity)
+  }
+
+  /// Creates an array whose contents are the results, of calling `initialize` with the values
+  /// `0 ..< count`, in order, passing also the element that needs to be initialized.
```
kyouko-taiga, line 24:
```suggestion
  /// Creates an array whose contents are the results of calling `initialize` with the values
  /// in the range `0 ..< count`, in order, along with the storage of the corresponding element.
```

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+  ///
+  /// - Requires: `count >= 0`.
+  public init<E>(count: Int, initialized_with initialize: inout [E](Int, set Element) inout -> Void) {
+    precondition(count >= 0, "negative array count")
+
+    &storage = new_storage(count)
+
+    var i = 0
+    while i != count {
+      let i_copy = i.copy()
```
kyouko-taiga, line 36: Why is this copy necessary? Is it because of a limitation of the compiler? If so, please add a comment, ideally with a reference to an issue.

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+  public fun with_mutable_contiguous_storage<E, T: Movable>(
+    _ action: inout [E](PointerToMutable<Element>) inout -> T
+  ) inout -> T {
+    if capacity() == 0 { &action(.null()) } else { &action(pointer_to_element(at: 0)) }
+  }
+
+  /// Adds a new element at the end of the array.
+  ///
+  /// Requires: `count() < capacity()`.
+  /// Complexity: O(1).
```
kyouko-taiga, line 92:
```suggestion
  /// - Requires: `count() < capacity()`.
  /// - Complexity: O(1).
```

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+  public fun append<E>(initialized_with initializer: inout [E](set Element) inout -> Void) inout {
+    precondition(count() < capacity(), "array is full")
+    pointer_to_element(at: count()).unsafe_initialize_pointee(&initializer)
+    &storage.header += 1
+  }
+
+  /// Removes al elements in the array.
+  ///
+  /// Effects: `append()` can called `capacity()` times after this call.
+  /// Complexity: O(n), where n is the number of elements in `self`.
```
kyouko-taiga, line 102:
```suggestion
  /// Removes all elements in the array, keeping its capacity.
  ///
  /// - Complexity: O(n), where n is the number of elements in `self`.
```

THB I think the effect is obvious. Even if we wanted to document it explicitly, I would not refer to just `append(_:)` since there are presumably there are other ways to append elements (e.g., `append(all:)`).

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+    let c = count()
+    precondition(c > 0, "array is empty")
+    pointer_to_element(at: c - 1).unsafe_deinitialize_pointee()
+    &storage.header -= 1
+  }
+
+  /// Removes the last `n` elements of the array.
+  ///
+  /// - Requires: `count() >= n`.
+  public fun pop_last_n(_ n: Int) inout {
```
kyouko-taiga, line 128:
```suggestion
  public fun remove_last(count n: Int) inout {
```

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+      &pointer_to_element(at: i).unsafe_deinitialize_pointee()
+      &i += 1
+    }
+    &storage.header = 0
+  }
+
+  /// Removes the last element of the array.
+  ///
+  /// - Requires: `count() > 0`.
+  public fun pop_last() inout {
```
kyouko-taiga, line 118:
```suggestion
  public fun remove_last() inout {
```
`Array.pop_last(:)` returns the last element so I think we should pick a different name for consistency.

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+    // Note: lifetime of `p` is guaranteed by the use of `elements` after the loop.
+    let p = PointerToMutable<Element>(type_punning: mutable_pointer[to: &elements])
+    var i = 0
+    while i < n {
+      &append(value: p.advance(by: i).unsafe_pointee())
+      &i += 1
+    }
+
+    Builtin.mark_uninitialized(elements)
+  }
```
kyouko-taiga, line 170: It's probably much more efficient to call `unsafe_initialize_pointee` in the loop.

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+
+  /// Adds a new element at the end of the array.
+  public fun append(value: sink Element) inout {
+    precondition(count() < capacity(), "array is full")
+    pointer_to_element(at: count()).unsafe_initialize_pointee(value)
+    &storage.header += 1
+  }
+
```
kyouko-taiga, line 197:
```suggestion
```

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+        if self[i] != other[i] { return false }
+        &i += 1
+      }
+      return true
+    }
+  }
+
+}
+
+public conformance BoundedArray: Copyable where Element: Copyable & Movable {
```
kyouko-taiga, line 221:
```suggestion
public conformance BoundedArray: Copyable where Element: Copyable {
```
`Copyable` implies `Movable`

Tests/LibraryTests/TestCases/BoundedArrayTests.hylo
```diff
@@ -0,0 +1,365 @@
...
+  precondition(d.count() == 3)
+  precondition(d.capacity() == 3)
+  precondition(!d.is_empty())
+  precondition(d.is_full())
+  precondition(d[0] == 21)
+  precondition(d[1] == 42)
+  precondition(d[2] == 84)
+}
+
+// TODO: LLVM error
```
kyouko-taiga, line 167: Please open an issue.

kyouko-taiga (changes requested):
Nice work.

StandardLibrary/Sources/BoundedArray.hylo
lucteo: the error is:
```
hylo/StandardLibrary/Sources/BoundedArray.hylo:40.7-9: error: illegal mutable access
      &i += 1
      ~~
```

lucteo: Created #1678

StandardLibrary/Sources/BoundedArray.hylo
lucteo: Fixed, thank you!

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+  public init<E>(count: Int, initialized_with initialize: inout [E](Int, set Element) inout -> Void) {
+    precondition(count >= 0, "negative array count")
+
+    &storage = new_storage(count)
+
+    var i = 0
+    while i != count {
+      // TODO (#1678): without copy, "illegal mutable access" is raised when `i` is incremented.
+      let i_copy = i.copy()
+      &append(initialized_with: fun (_ e: set Element) -> Void {
```
kyouko-taiga, line 38:
```suggestion
      &append(initialized_with: fun[sink let j = i.copy()](_ e: set Element) -> Void {
```
Can you try this? I think that would make the lambda independent and sidestep #1678.

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+  public fun capacity() -> Int {
+    storage.capacity()
+  }
+
+  /// Projects a pointer to the start of the array's contiguous storage.
+  ///
+  /// The projected pointer is valid only for the duration of the projection and can be advanced up
+  /// to `count()`. It may be null if `self` is empty.
+  public property contiguous_storage: Pointer<Element> {
+    yield if capacity() == 0 { .null() } else { .new(pointer_to_element(at: 0)) }
```
kyouko-taiga, line 77: FYI, there are two reasons why it may be better to write `if c { yield a } else { yield b }` instead of `yield if c { a } else { b }`. First, it is more expensive for type inference to process the latter. Second, and more importantly, the latter is more efficient because in general we can yield existing storage rather than creating a temporary.

Even in this example, where both branches are rvalues, we still have to create a temporary and move the result of either branch before we can yield. So there's one more move.

It's possible we'll be able to optimize some of the inefficiency away in the future but currently we do not.

StandardLibrary/Sources/BoundedArray.hylo
```diff
@@ -0,0 +1,278 @@
...
+  public fun remove_last() inout {
+    let c = count()
+    precondition(c > 0, "array is empty")
+    pointer_to_element(at: c - 1).unsafe_deinitialize_pointee()
+    &storage.header -= 1
+  }
+
+  /// Removes the last `n` elements of the array.
+  ///
+  /// - Requires: `count() >= n`.
```
kyouko-taiga, line 127: Why make it a precondition? We could define `remove_last(up_to:)` and then there's one less requirement for the caller.

kyouko-taiga: @dabrahams I'd love to read your opinion.

StandardLibrary/Sources/BoundedArray.hylo
lucteo: This was a 1:1 copy with code from `Array`. Made the same change there.

StandardLibrary/Sources/BoundedArray.hylo
lucteo: Works ok, made the change.
(to me, this is slightly uglier than the previous workaround, but I don't quite care)

Still kept the TODO, as the form without copying is better..

StandardLibrary/Sources/BoundedArray.hylo
lucteo: I think I would prefer exceptions here, but I think for the time being we can use preconditions here. The way I see it, even if we swap this around and make `remove_last(up_tp:)` we would still need one precondition. (`n <= count()`).

(also made a small fix in the preconditions)

kyouko-taiga (approved):
LGTM.

If possible I would still love to have @dabrahams's opinion on the precondition of `remove_last(count:)` before we merge but I have no objection to move forward.

StandardLibrary/Sources/BoundedArray.hylo
dabrahams: Exceptions are a bad fit for this function.  They are for postcondition failures that  can't reasonably be prevented by the caller doing an inexpensive check, and are likely to need to be handled far from the source.

The function that removes the last `min(n, count())` elements is a different one from the one with the precondition, and I don't see a good use case for the former.  It seems like it mostly just masks programming errors.  If there actually is a use case, it needs to return the number of elements actually removed, under the principle of not throwing away information (this is a weak example, but still).  But the latter function has obvious use cases and is more fundamental, so it definitely comes first.

PR #1687, opened 2025-05-05
Upgrade to Swift 6.1 and enable SPM on Windows

tothambrus11:


tothambrus11:
Remaining things to fix:
- CMake builds on MacOS

tothambrus11:
@dabrahams I couldn't get the CMake builds working on MacOS due to a `tgmath.h` not being found.  `tgmath.h` is actually included from a dependency of the swift-algorithms package for handling complex numbers and we don't use any functionality that needs it directly. The issue likely has to do with the Swift SDK/compiler setup, as I've verified the header is indeed found on the machine. Do you have ideas about what could have gone wrong or how to diagnose that?

An approach we could take is to remove our dependency on the `swift-algorithms` package and reimplement the 2-3 algorithms we actually use, we can save some compile time and maintenance effort with a library depending on C interop in exchange of implementing and testing those 2-3 algorithms ourselves.

Maybe we could also fix this as a separate issue after merging this, so that we can build and test other PRs on all operating systems ASAP.

dabrahams:
I suggest trying to update the swift algorithms dependency, and if you've done that already, look at how its own CMake build works. Surely it builds with swift 6

Sources/FrontEnd/AST/Decl/TranslationUnit.swift
```diff
@@ -1,5 +1,5 @@
 /// A collection of declarations at the top-level of a source file.
-public struct TranslationUnit: Node, LexicalScope {
```
dabrahams, line 2: I'd love to get this merged.  But couldn't the size of the diff be reduced dramatically by incorporating `Sendable` into protocols near the root of the refinement hierarchy?  If every `LexicalScope: Sendable` we could drop this change, right?

dabrahams:
How did you get this to build without these changes?

```diff
diff --git a/Sources/Utils/StatefulCoder.swift b/Sources/Utils/StatefulCoder.swift
index d08c97d63..e50a85623 100644
--- a/Sources/Utils/StatefulCoder.swift
+++ b/Sources/Utils/StatefulCoder.swift
@@ -14,15 +14,15 @@ public protocol StatefulCoder {
   associatedtype Encoding = Data
 
   /// The storage vehicle for state.
-  var userInfo: [CodingUserInfoKey: Any] { get set }
+  var userInfo: [CodingUserInfoKey: any Sendable] { get set }
 
 }
 
 extension StatefulCoder {
 
   /// Injects initialState into `self` for use as mutable storage during encoding/decoding.
-  public mutating func setState<T>(_ initialState: T) {
-    userInfo[stateKey] = Box(initialState)
+  public mutating func setState<T: Sendable>(_ initialState: T) {
+    userInfo[stateKey] = SharedMutable(initialState)
   }
 
 }
@@ -43,7 +43,7 @@ extension StatefulEncoder {
 
   /// Returns a serialized representation of `subject`, using `startState` as the initial encoding
   /// state.
-  public mutating func encode<T, State>(_ value: T, startState: State) throws -> Encoding
+  public mutating func encode<T, State: Sendable>(_ value: T, startState: State) throws -> Encoding
   where T: Encodable {
     setState(startState)
     defer { setState(0) }  // Discard state, just in case `self` persists.
@@ -68,7 +68,7 @@ extension StatefulDecoder {
 
   /// Returns the `T` value that was serialized into `archive`, using `startState` as the initial
   /// decoding state.
-  public mutating func decode<T, State>(_ type: T.Type, from encodedT: Encoding, startState: State)
+  public mutating func decode<T, State: Sendable>(_ type: T.Type, from encodedT: Encoding, startState: State)
     throws -> T where T: Decodable
   {
     setState(startState)
```

Sources/FrontEnd/AST/Decl/TranslationUnit.swift
dabrahams: Reminder: testing my work on the interpreter will be very difficult without Swift 6 because I need exit tests from swift-testing.  So, no pressure ;-)

.devcontainer/Dockerfile
```diff
@@ -1,47 +1 @@
...
-# Make and install the llvm.pc file.
-ADD make-pkgconfig.sh /tmp
-RUN <<EOT bash -ex -o pipefail
-
-    export PATH="/opt/llvm-${HYLO_LLVM_BUILD_TYPE}/bin:\$PATH"
-    /tmp/make-pkgconfig.sh /usr/local/lib/pkgconfig/llvm.pc > /dev/null
-    rm /tmp/make-pkgconfig.sh
-
-EOT
+FROM ghcr.io/hylo-lang/hylo-dev-toolchain:v0.1.3
```
dabrahams: Nice.  Why didn't I see this in your Swifty-LLVM patch?  Do we not test that in a devcontainer?

.devcontainer/devcontainer.json
```diff
@@ -36,7 +36,8 @@
             },
             // Add the IDs of extensions you want installed when the container is created.
             "extensions": [
-                "sswg.swift-lang"
+                "swiftlang.swift-vscode",
```
dabrahams, line 39: Why is this the right change?

.devcontainer/devcontainer.json
```diff
@@ -46,9 +47,11 @@
         "HYLO_LLVM_BUILD_TYPE": "${localEnv:HYLO_LLVM_BUILD_TYPE:MinSizeRel}"
     },
     "remoteEnv": {
-        "PATH": "/opt/llvm-${containerEnv:HYLO_LLVM_BUILD_TYPE}/bin:${containerEnv:PATH}"
+        "PATH": "/python-venv/bin:/opt/llvm-${containerEnv:HYLO_LLVM_BUILD_TYPE}/bin:${containerEnv:PATH}",
+        "PKG_CONFIG_PATH": "/opt/llvm-${containerEnv:HYLO_LLVM_BUILD_TYPE}/pkgconfig:${containerEnv:PKG_CONFIG_PATH}",
+        "HYLO_USE_BUILTIN_STDLIB": "true"
```
dabrahams, line 48: What's this?

.gitignore
```diff
@@ -1,12 +1,15 @@
 # Swift Package Manager
 /.build
+/build
+/build_dir
```
dabrahams: Is this change related to the purpose of the patch?

.gitignore
```diff
@@ -1,12 +1,15 @@
...
 /.*-build
 /Packages
 
 # IDEs
 /.vscode
 /.swiftpm
 /.vscode/launch.json
+/.idea
```
dabrahams: Is this change related to the purpose of the patch?

CMakeLists.txt
```diff
@@ -2,6 +2,8 @@
 # Standard project boilerplate
 #
 cmake_minimum_required(VERSION 3.30)
+set(CMAKE_OSX_DEPLOYMENT_TARGET 14.0)
+set(CMAKE_Swift_LANGUAGE_VERSION 6)
```
dabrahams, line 6: It would be awesome to somehow get all the common stuff into HyloCMakeModules. I'm not sure if it's possible with this bit.

CMakeLists.txt
```diff
@@ -12,14 +14,11 @@ project(Hylo
...
 
-set(CMAKE_OSX_DEPLOYMENT_TARGET "13.0")
-
 list(PREPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
-set(CMAKE_Swift_LANGUAGE_VERSION 5)
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ${BUILD_SHARED_LIBS}) # Needed?
 set(CMAKE_Swift_COMPILE_OPTIONS
-  -warnings-as-errors)
+  -warnings-as-errors --verbose)
```
dabrahams: I don't think I want verbose output from the compiler whenever I build with CMake.

README.md
```diff
@@ -24,22 +23,21 @@ for them before checking it out.
...
-comment in `.devcontainer/Dockerfile`.
+If you are using this package for development, we strongly recommend
+the use of an LLVM with assertions enabled; otherwise it's too
+easy to violate LLVM's preconditions without knowing it. 
+You can download such a prebuilt LLVM package from Hylo's 
+[LLVM builds](https://github.com/hylo-lang/llvm-build).
+If you are using development containers, this is already installed preinstalled on the system. Run `llvm-config --prefix` to see where.
+
+*If* you want to build with the Swift Package Manager, you need to have [pkg-config](https://linux.die.net/man/1/pkg-config)
+installed, and `<YOUR LLVM>/pkgconfig` added to the `PKG_CONFIG_PATH` environment variable.
```
dabrahams:
```suggestion
installed, `<YOUR LLVM>/pkgconfig` added to the `PKG_CONFIG_PATH` environment variable,
and you *must* use one of [Hylo's 
LLVM builds](https://github.com/hylo-lang/llvm-build), which provide the necessary pkg-config file.
```

Does pkg-config have a hyphen or not?  We should be consistent.

README.md
```diff
@@ -24,22 +23,21 @@ for them before checking it out.
...
 
-*If* you want to build with the Swift Package Manager and you choose
-to get LLVM some other way, you'll need an installation with an
-`llvm-config` executable, which we will use to create a `pkg-config`
-file for LLVM.
+On **Windows**, make sure to set the environment variable `VSINSTALLDIR` to something like
+`C:/Program Files/Microsoft Visual Studio/2022/Community`, without `/` at the end. This is needed for LLVM to locate the DIA SDK.
```
dabrahams: “this” needs an antecedent, e.g. “this step”

```suggestion
`C:/Program Files/Microsoft Visual Studio/2022/Community`, without `/` at the end, so LLVM can locate the DIA SDK.
```

It's not really LLVM that's locating that SDK though is it?  Do you mean so CMake can, or SPM can, or…?

README.md
```diff
@@ -48,24 +46,17 @@ file for LLVM.
    to the root directory of your LLVM installation,
 
 	```
-	cmake -D CMAKE_BUILD_TYPE=<build-type> \
-	  -D LLVM_DIR=<LLVM>/lib/cmake/llvm   \
-      -G Ninja -S . -B <build-directory>
+	cmake -D CMAKE_BUILD_TYPE=<build-type> -D LLVM_DIR=<LLVM>/lib/cmake/llvm -DBUILD_TESTING=1 -G Ninja -S . -B .build
```
dabrahams: This is not an improvement.  There's a reason I wrapped the lines.  When long lines of code are rendered in markdown on GitHub, they get cut off and to read them you have to scroll.

README.md
```diff
@@ -48,24 +46,17 @@ file for LLVM.
...
     ```
-
-    (on Windows substitute your shell's line continuation character
-    for `\` or just remove the line breaks and backslashes).
-    
-    If you want to run tests, add `-DBUILD_TESTING=1`.
+ 
+    If you don't want to run tests, remove `-DBUILD_TESTING=1`.
```
dabrahams: Rationale for changing the default?  I suppose if you _are_ going to change the default, you could drop this line altogether, and the ability to do so would make a decent rationale.

README.md
```diff
@@ -48,24 +46,17 @@ file for LLVM.
...
     
-    **Note:** on macOS, if you are not using your Xcode's default
-    toolchain, [you may need `-D
+    > **MacOS-specific flags**
```
dabrahams: Why is this part in a quotation?  Why not use a subsection header?  Why the extra spaces before the `>`?

These kinds of changes seem like they exist to satisfy an arbitrary personal aesthetic, and are time consuming to review, which is a bit annoying.

README.md
```diff
@@ -48,24 +46,17 @@ file for LLVM.
...
+   > 
+   > - If you are not using the default Swift toolchain, [you may need `-D
     CMAKE_Swift_COMPILER=swiftc`](https://gitlab.kitware.com/cmake/cmake/-/issues/25750)
     to prevent CMake from using Xcode's default `swift`.
-    
-    If this command fails it could be because you have an LLVM without
```
dabrahams, line 66: There must be a reason I added this.  Why did you delete it?

README.md
```diff
@@ -24,22 +23,21 @@ for them before checking it out.
...
+If you are using development containers, this is already installed preinstalled on the system. Run `llvm-config --prefix` to see where.
+
+*If* you want to build with the Swift Package Manager, you need to have [pkg-config](https://linux.die.net/man/1/pkg-config)
+installed, and `<YOUR LLVM>/pkgconfig` added to the `PKG_CONFIG_PATH` environment variable.
 
-*If* you want to build with the Swift Package Manager and you choose
-to get LLVM some other way, you'll need an installation with an
-`llvm-config` executable, which we will use to create a `pkg-config`
-file for LLVM.
+On **Windows**, make sure to set the environment variable `VSINSTALLDIR` to something like
```
dabrahams:
```suggestion
On **Windows**, set the environment variable `VSINSTALLDIR` to something like
```

Omit needless words.  But "something like” is really insufficient.  I could change one character, and it would be wrong for any installation of VSCode.  Figure out how to tell people the right thing to do.  

Is the community edition required?  I remember something about being locked by LLVM into a particular VSCode edition.

README.md
```diff
@@ -109,38 +100,10 @@ we're considering dropping our use of it, so think about [using CMake
...
-export PKG_CONFIG_PATH=$PWD
-```
-
-or you can put it somewhere that `pkg_config` already searches (needed
-for use with Xcode):
-
-```bash
-sudo mkdir -p /usr/local/lib/pkgconfig && sudo mv llvm.pc /usr/local/lib/pkgconfig/
-```
+Make sure to have followed the instructions regarding LLVM setup above and have your `<LLVM installation>/pkgconfig`
```
dabrahams:
```suggestion
Be sure you've followed the instructions regarding LLVM setup above and have your `<LLVM installation>/pkgconfig` 
```

Do we really want to give redundant directions?  We don't do anything similar for other methods do we?

Sources/Driver/Driver.swift
```diff
@@ -510,7 +514,7 @@ public struct Driver: ParsableCommand {
     for root in path.split(separator: Host.pathEnvironmentSeparator) {
       let candidate = URL(fileURLWithPath: String(root)).appendingPathComponent(executableFileName)
       if FileManager.default.fileExists(atPath: candidate.fileSystemPath) {
-        Driver.executableLocationCache[invocationName] = candidate
+        Driver.executableLocationCache.modify { v in v[invocationName] = candidate }
```
dabrahams, line 517: A bit hard to imagine why this would be necessary.  Can't all that be wrapped into the subscript?  Oh I see, it's now a `SharedMutable<>`.  It's an open question whether most uses of `SharedMutable` should be hidden inside other types.

Sources/Driver/OutputFileHandle.swift
```diff
@@ -1,22 +1,27 @@
 import Foundation
 
-struct OutputFileHandle: TextOutputStream {
+class OutputFileHandle: TextOutputStream, @unchecked Sendable {
```
dabrahams, line 3: I think there should always be an explanation of why something is `@unchecked Sendable` and why it's OK.

Sources/Driver/Process+Extensions.swift
```diff
@@ -16,7 +16,7 @@ extension Process {
   public typealias OutputText = (standardOutput: Lazy<String>, standardError: Lazy<String>)
 
   /// The results of a process run that exited with a nonzero code.
-  public struct NonzeroExit: Error {
+  public struct NonzeroExit: Error, Sendable {
```
dabrahams, line 19: Seeing all these makes me think we should have things like

`protocol Error: Swift.Error, Sendable {}` in `Utils`.

Sources/Driver/SourceLine+Extensions.swift
```diff
@@ -17,8 +17,4 @@ extension SourceLine {
 
 }
 
-#if swift(>=5.11)
-  extension SourceLine: @retroactive ExpressibleByArgument {}
-#else
-  extension SourceLine: ExpressibleByArgument {}
-#endif
+extension SourceLine: ExpressibleByArgument {}
```
dabrahams, line 20: Weird that the need for `@retroactive` went away

Sources/FrontEnd/Terms/ConcreteTerm.swift
```diff
@@ -1,12 +1,21 @@
+import Utils
+
 /// A box wrapping a concrete compile-time value.
 public struct ConcreteTerm: TermProtocol {
 
+  // The value of the term.
+  //
+  // WARNING: Don't use downcasts from here, `as?` will not work properly, only on its wrapped `.anyHashable` object.
```
dabrahams: This is unclear.  If `as?` doesn't work properly, it means there's a bug in Swift.  I think you mean something else.

Sources/TestUtils/AnnotatedHyloFileTest.swift
```diff
@@ -32,16 +32,27 @@ extension Result {
...
-public enum ExpectedTestOutcome {
+public enum ExpectedTestOutcome: Sendable {
 
   case success, failure
 
 }
 
+extension XCTContextHylo {
+
+  // Same as `XCTContext.runActivity(named:block:)`, but without the `@MainActor` requirement.
```
dabrahams, line 43: Needs a proper doc comment that follows Swift API Guidelines.

Sources/TestUtils/XCTest+Shims.swift
```diff
@@ -2,16 +2,15 @@ import XCTest
 
 /// This file declares a lightweight compatibility layer to fill some of the missing parts of
 /// swift-corelibs-xctest API. See https://github.com/apple/swift-corelibs-xctest/issues/348
-///
-/// Do not modify the name of any type or property. Those are meant to match Apple's API.
 
-#if !os(macOS)
+/// Don't modify this part, it's meant to replicate the exact Swift API:
```
dabrahams, line 6: Why did the comment change?  When I said “Apple's API” I could have been clearer. I meant the XCTest API that ships on Apple platforms.

Sources/TestUtils/XCTest+Shims.swift
```diff
@@ -69,27 +68,6 @@ import XCTest
 
   }
 
-  /// A proxy for the current testing context.
```
dabrahams, line 72: The needless reordering of this code is making review hard.

Sources/TestUtils/XCTest+Shims.swift
```diff
@@ -103,3 +81,31 @@ import XCTest
   }
 
 #endif
+
+/// The following code is slightly modified because Swift 6 broke our replicated API of Swift 5, e.g. by
```
dabrahams: Modified from what?  If you mean with respect to the previous version, that note is only useful to a PR reviewer.  You should write something that's useful to a maintainer of the code who isn't looking at diffs.

This should not be a doc comment (`///`).  It is an implementation note.

Sources/TestUtils/XCTest+Shims.swift
```diff
@@ -103,3 +81,31 @@ import XCTest
...
+/// requiring runActivity(named:block:) to be @MainActor and converting the XCTActivity from
+/// a struct to a protocol. See the original commit at
+/// https://github.com/hylo-lang/hylo/commit/5b05ff1b2a359efdcaecb0d9abb86ea4842eb7ad
+//
+// They are deliberately named `XCT*Hylo` to avoid conflicts with the original XCTest API.
```
dabrahams, line 90: The problem was that they were missing from the XCTest API on non-apple platforms.  If you are using them everywhere, that would explain the name change.

Sources/Utils/AnyHashableAndSendable.swift
```diff
@@ -0,0 +1,21 @@
+public struct AnyHashableAndSendable: @unchecked Sendable, Hashable, Equatable {
```
dabrahams, line 1: It's amazing that Swift doesn't already have something like this.

.devcontainer/Dockerfile
tothambrus11: I did that in a later PR, now that's consistent: https://github.com/hylo-lang/Swifty-LLVM/pull/99/files#diff-13bd9d7a30bf46656bc81f1ad5b908a627f9247be3f7d76df862b0578b534fc6

.devcontainer/devcontainer.json
tothambrus11: image
The sswg version is deprecated.

tothambrus11:
> How did you get this to build without these changes?
> 
> ```diff
> diff --git a/Sources/Utils/StatefulCoder.swift b/Sources/Utils/StatefulCoder.swift
> index d08c97d63..e50a85623 100644
> --- a/Sources/Utils/StatefulCoder.swift
> +++ b/Sources/Utils/StatefulCoder.swift
> @@ -14,15 +14,15 @@ public protocol StatefulCoder {
>    associatedtype Encoding = Data
>  
>    /// The storage vehicle for state.
> -  var userInfo: [CodingUserInfoKey: Any] { get set }
> +  var userInfo: [CodingUserInfoKey: any Sendable] { get set }
>  
>  }
>  
>  extension StatefulCoder {
>  
>    /// Injects initialState into `self` for use as mutable storage during encoding/decoding.
> -  public mutating func setState<T>(_ initialState: T) {
> -    userInfo[stateKey] = Box(initialState)
> +  public mutating func setState<T: Sendable>(_ initialState: T) {
> +    userInfo[stateKey] = SharedMutable(initialState)
>    }
>  
>  }
> @@ -43,7 +43,7 @@ extension StatefulEncoder {
>  
>    /// Returns a serialized representation of `subject`, using `startState` as the initial encoding
>    /// state.
> -  public mutating func encode<T, State>(_ value: T, startState: State) throws -> Encoding
> +  public mutating func encode<T, State: Sendable>(_ value: T, startState: State) throws -> Encoding
>    where T: Encodable {
>      setState(startState)
>      defer { setState(0) }  // Discard state, just in case `self` persists.
> @@ -68,7 +68,7 @@ extension StatefulDecoder {
>  
>    /// Returns the `T` value that was serialized into `archive`, using `startState` as the initial
>    /// decoding state.
> -  public mutating func decode<T, State>(_ type: T.Type, from encodedT: Encoding, startState: State)
> +  public mutating func decode<T, State: Sendable>(_ type: T.Type, from encodedT: Encoding, startState: State)
>      throws -> T where T: Decodable
>    {
>      setState(startState)
> ```

Thanks, I think this was the difference between it working on MacOS 14 and not working on MacOS 15. I'll incorporate these and see if we can build with the xcode generator.

README.md
tothambrus11: I agree, I'll change the wording. 

> Is the community edition required?

No, originally the professional version was required but we patched the relevant cmake file in llvm-build to refer to this VSINSTALLDIR variable instead, which can be customized by users.

README.md
tothambrus11: We already suggested to use our llvm builds above, I don't think at this point it's worth supporting arbitrary other LLVM builds.

tothambrus11:
> How did you get this to build without these changes?
> 
> ```diff
> diff --git a/Sources/Utils/StatefulCoder.swift b/Sources/Utils/StatefulCoder.swift
> index d08c97d63..e50a85623 100644
> --- a/Sources/Utils/StatefulCoder.swift
> +++ b/Sources/Utils/StatefulCoder.swift
> @@ -14,15 +14,15 @@ public protocol StatefulCoder {
>    associatedtype Encoding = Data
>  
>    /// The storage vehicle for state.
> -  var userInfo: [CodingUserInfoKey: Any] { get set }
> +  var userInfo: [CodingUserInfoKey: any Sendable] { get set }
>  
>  }
>  
>  extension StatefulCoder {
>  
>    /// Injects initialState into `self` for use as mutable storage during encoding/decoding.
> -  public mutating func setState<T>(_ initialState: T) {
> -    userInfo[stateKey] = Box(initialState)
> +  public mutating func setState<T: Sendable>(_ initialState: T) {
> +    userInfo[stateKey] = SharedMutable(initialState)
>    }
>  
>  }
> @@ -43,7 +43,7 @@ extension StatefulEncoder {
>  
>    /// Returns a serialized representation of `subject`, using `startState` as the initial encoding
>    /// state.
> -  public mutating func encode<T, State>(_ value: T, startState: State) throws -> Encoding
> +  public mutating func encode<T, State: Sendable>(_ value: T, startState: State) throws -> Encoding
>    where T: Encodable {
>      setState(startState)
>      defer { setState(0) }  // Discard state, just in case `self` persists.
> @@ -68,7 +68,7 @@ extension StatefulDecoder {
>  
>    /// Returns the `T` value that was serialized into `archive`, using `startState` as the initial
>    /// decoding state.
> -  public mutating func decode<T, State>(_ type: T.Type, from encodedT: Encoding, startState: State)
> +  public mutating func decode<T, State: Sendable>(_ type: T.Type, from encodedT: Encoding, startState: State)
>      throws -> T where T: Decodable
>    {
>      setState(startState)
> ```

See https://github.com/swiftlang/swift-foundation/issues/1515

dabrahams (changes requested):
One problem with the pkgconfig arrangement is that there's no obvious way to get the file to be found by Xcode.  If such a way exists, it needs to be documented.  If such a way does not exist, we need another scheme.  The only thing I could figure out was to copy it to /usr/local/lib/pkgconfig/ and replace the references to ${pcfiledir} with an absolute path.

tothambrus11:
> One problem with the pkgconfig arrangement is that there's no obvious way to get the file to be found by Xcode. If such a way exists, it needs to be documented. If such a way does not exist, we need another scheme. The only thing I could figure out was to copy it to /usr/local/lib/pkgconfig/ and replace the references to ${pcfiledir} with an absolute path.
> 
> Setting the environment before launching Xcode doesn't work for me.

Isn't setting the PKG_CONFIG_PATH variable not sufficient? That works for SPM at least, since pkg-config executable finds the PC file that way.

Otherwise, the current setup indeed doesn't work. If we can, we should ask xcode to fix their PC file discovery mechanism, otherwise it's inconsistent on their part

dabrahams:
> Isn't setting the PKG_CONFIG_PATH variable not sufficient? That works for SPM at least, since pkg-config executable finds the PC file that way.

SPM runs from your shell.  Xcode doesn't.  But even launching the executable inside the Xcode bundle from the shell didn't work for me.  There's no obvious way to set the environment that will be used for builds within Xcode.

tothambrus11:
> There's no obvious way to set the environment that will be used for builds within Xcode.

Thanks for the context! I'll check if we can add a fix for this in package.swift, by opening a new bash shell, capturing the pkg_confi_path variable and appending a swiftc compiler flag to forward it explicitly (this can be a fallback when the current process's environment variable is not present.

dabrahams:
@tothambrus11 there are many un-replied-to/unresolved comments that I've left on this PR.  I'd really like to see them resolved before it is merged, and I'd really _really_ like to see it merged.  See also #1739

Sources/TestUtils/KnownBugs.swift
```diff
@@ -1,6 +1,6 @@
 /// Whether Swifty-LLVM crashes when running mandatory passes: (https://github.com/hylo-lang/Swifty-LLVM/issues/24).
 #if os(Windows)
-  public let swiftyLLVMMandatoryPassesCrash = true
+  public let swiftyLLVMMandatoryPassesCrash = false
```
dabrahams: This file should be removed now! 🥳

Sources/FrontEnd/AST/Decl/TranslationUnit.swift
tothambrus11: I added sendable to every protocol and removed it from structs when it was possible

Sources/Utils/AnyHashableAndSendable.swift
tothambrus11: Yeah, others have ended up with similar solutions: https://stackoverflow.com/questions/79098989/creating-a-struct-similar-to-anyhashable

.devcontainer/devcontainer.json
tothambrus11: The last line is responsible for setting up the environment variable in the container so that the compiler references the standard library sources instead of the assets within the build directory. See https://github.com/tothambrus11/hylo/blob/swift-601/StandardLibrary/StandardLibrary.swift#L11

I cleaned up the other variables a bit, they didn't seem to be necessary since they are already set up from the docker image.

README.md
tothambrus11: Updated to:
On **Windows**, make sure to set the environment variable `VSINSTALLDIR` to your visual studio installation path, e.g. `C:/Program Files/Microsoft Visual Studio/2022/Community`, without `/` at the end, so LLVM's CMake files can locate the DIA SDK.

README.md
tothambrus11: The compiler developers likely want to build the tests than not, so I changed the default command to include it. I removed the unnecessary sentence afterwards.

README.md
tothambrus11: I merged them into one line due to `\` not being an escape character in Windows shells, so there could be slightly bigger chance of confusion that way. I changed it back for now to a split version since we don't have many windows users anyways.

CMakeLists.txt
tothambrus11: I removed the verbose flag, and left a comment on how to add it back while debugging the build.

Sources/Driver/Driver.swift
tothambrus11: This is a private variable with 2 usages in total, so it's probably not worth wrapping it.

Sources/Driver/Process+Extensions.swift
tothambrus11: There were not enough usages of this to be worth the added complexity, there are around 3-4 error structs in different targets.

Sources/Driver/SourceLine+Extensions.swift
tothambrus11: > _If it is declared in a module that is part of the same package as the conformance, this is not retroactive. Duplicated same-package conformances will be detected at link or load time._

https://github.com/swiftlang/swift-evolution/blob/main/proposals/0364-retroactive-conformance-warning.md#motivation:~:text=If%20it%20is%20declared%20in%20a%20module%20that%20is%20part%20of%20the%20same%20package%20as%20the%20conformance%2C%20this%20is%20not%20retroactive.%20Duplicated%20same%2Dpackage%20conformances%20will%20be%20detected%20at%20link%20or%20load%20time.

tothambrus11:
XCode builds are broken *again*, most likely due to a recent compiler upgrade: https://github.com/swiftlang/swift/issues/84402

CMakeLists.txt
tothambrus11: I wouldn't do that for these flags, these must go before the first `project` declaration, and the relationship is much less obvious if these flags are put somewhere in a remote file (including the CMakeModules would need to have a new, documented precondition)

README.md
tothambrus11: It does, I changed the readme to be consistent. Note, in `pkgconfig` folder we still don't have a hyphen because it would require updating all our dependent CI workflows in multiple projects to accommodate for this tiny change in the path.

README.md
dabrahams: In that case we should upgrade the suggestion to a requirement.

Sources/Utils/AnyHashableAndSendable.swift
dabrahams:
```suggestion
public struct AnyHashableAndSendable: @unchecked Sendable, Hashable {
```

Sources/TestUtils/XCTest+Shims.swift
tothambrus11: Adjusted wording: `Modified compared to Apple's version`

Sources/TestUtils/XCTest+Shims.swift
tothambrus11: Yes, we are using the `XCTContextHylo` everywhere instead of `XCTContext` due to Apple's version being inconsistent, and e.g. `runActivty` requiring MainActor in Swift 6.

Sources/TestUtils/AnnotatedHyloFileTest.swift
tothambrus11: done

Sources/TestUtils/XCTest+Shims.swift
dabrahams: @tothambrus11 Please respond

tothambrus11: Hmm I think I wanted to change referring to a "name of any type or property" to something more generic, e.g. "this part", since not only their name counts towards the API. The ending is now indeed less clear, your phrasing would be better.

What do you think about this? "Do not modify this part. It is meant to match the XCTest API that ships on Apple platforms."

dabrahams: @tothambrus11 I like it!

PR #1695, opened 2025-05-23
Print stacktrace on precondition failure & assertion on pointer dereference

lucteo:
Print stack trace when a precondition is violated.
Also add assertions on pointer derefence, checking for null pointers.

Also remove `concurrent_sort` example as is is flaky.


tothambrus11:
I see that tests are failing because C.int doesn't yet exist. Would it be possible to write a wrapper function in `LibC.c` for backtrace that only uses word-size integers?

Seeing a backtrace would be very valuable for development, so we should merge this if possible.

StandardLibrary/Sources/Core/Assert.hylo
```diff
@@ -45,6 +45,7 @@ fun log_failure_description(_ title: String, _ message: String, in file: String,
       print(title, terminator: ": ")
       print(message)
     }
+    print_backtrace()
```
kyouko-taiga, line 48: I wonder if we should print the backtrace in release mode. Do we have a compiler directive to distinguish `#debug` or `#release`? I don't remember.

StandardLibrary/Sources/Core/Assert.hylo
```diff
@@ -59,9 +60,42 @@ fun trap() -> Never {
...
 
+/// The size of the backtrace buffer.
+private let backtrace_depth = 20
+
+/// Prints the backtrace to the standard error (if OS supports it).
+fun print_backtrace() {
+  #if os(Linux) || os(macOS)
+  #if !feature(freestanding)
+    print("Backtrace:")
+    var buffer: Array<MemoryAddress> = .new(count: backtrace_depth, initialized_with: fun(_ i: Int) -> MemoryAddress {
```
kyouko-taiga:
```suggestion
    var buffer = Array<MemoryAddress>(count: backtrace_depth, initialized_with: fun(_ i: Int) -> MemoryAddress {
```
It's a very subjective suggestion.

StandardLibrary/Sources/Core/Assert.hylo
```diff
@@ -59,9 +60,42 @@ fun trap() -> Never {
...
+      return .null()
+    })
+    {
+      &buffer.with_mutable_contiguous_storage(fun (_ p: PointerToMutable<MemoryAddress>) -> Void {
+        var size = backtrace(p, C.int(truncating_or_extending: backtrace_depth))
+        backtrace_symbols_fd(p, size, 2) // write to stderr
+      })
+    }
```
kyouko-taiga: Is this supposed to delimit a scope? Why is that scope necessary?

StandardLibrary/Sources/Core/Assert.hylo
```diff
@@ -59,9 +60,42 @@ fun trap() -> Never {
...
 /// The definition of this function is provided at linking time for use in free-standing mode.
 /// Programs built for a hosted environment must not call it `halt` and use `trap` instead.
 @external("halt")
 fun halt() -> Never
+
+/// Gets the backtrace of the current thread.
+///
+/// See: https://www.gnu.org/software/libc/manual/html_node/Backtraces.html
+@ffi("backtrace")
+fun backtrace(_ buffer: PointerToMutable<MemoryAddress>, _ size: C.int) -> C.int;
```
kyouko-taiga:
```suggestion
fun backtrace(_ buffer: PointerToMutable<MemoryAddress>, _ size: C.int) -> C.int
```

StandardLibrary/Sources/Core/Assert.hylo
```diff
@@ -59,9 +60,42 @@ fun trap() -> Never {
...
+
+/// Prints the backtrace at `buffer` to file descriptor `fd`.
+///
+/// See: https://www.gnu.org/software/libc/manual/html_node/Backtraces.html
+@ffi("backtrace_symbols_fd")
+fun backtrace_symbols_fd(_ buffer: PointerToMutable<MemoryAddress>, _ size: C.int, _ fd: C.int);
```
kyouko-taiga:
```suggestion
fun backtrace_symbols_fd(_ buffer: PointerToMutable<MemoryAddress>, _ size: C.int, _ fd: C.int)
```

StandardLibrary/Sources/Core/Pointer.hylo
```diff
@@ -11,6 +11,7 @@ public type Pointer<Pointee>: Regular {
   /// - Requires: `self` is the address of an object of type `Pointee` and its storage
   ///   is accessed only through this projection during the projection's lifetime.
   public subscript unsafe(): Pointee {
+    precondition(self != .null(), "dereferencing null pointer")
```
kyouko-taiga: I am not convinced about the value of this precondition. We can't in general test that dereferencing a pointer is safe _and_ (perhaps more importantly) we should guarantee that using the unsafe API is as fast as possible. Therefore adding branches on this path seems wrong.

StandardLibrary/Sources/Core/Assert.hylo
lucteo: We don't seem to have such a feature yet.
(and to be honest, I would keep it in release as well, but add a different flag to disable this)

StandardLibrary/Sources/Core/Assert.hylo
lucteo: Sometimes (probably due to compiler bugs) the scopes are needed. In this case, it seems to work without one. I think this was a leftover from me trying to solve a lifetime issue. Will drop.

StandardLibrary/Sources/Core/Pointer.hylo
lucteo: `null` appears so often in low-level code, that, I think, it deserves a special check. It's just the checks that we do to ensure we are not past the end of a buffer.

In terms of performance, most of these branches should be removed by the compiler once it figures out that the pointer is valid (for most of the cases that I can think of).

(see also: https://security.googleblog.com/2024/11/retrofitting-spatial-safety-to-hundreds.html)

kyouko-taiga: I am still on the fence. At the very least this change should be part of another PR.

kyouko-taiga (changes requested):
I'm fine with the PR as it is but for the precondition in `Pointer`. I don't think we should dismiss the possible performance drawback of adding a branch on pointer dereferencing. We cannot just assume that the compiler will be able to remove this branch in general and this operation is crucial for implementing efficient data structures.

StandardLibrary/Sources/Core/Pointer.hylo
dabrahams: Proposal: create `debug_precondition` which for now just forwards to precondition, but (when we get the time to implement it) is removed from optimized builds, and use that.

lucteo:
> I'm fine with the PR as it is but for the precondition in `Pointer`. I don't think we should dismiss the possible performance drawback of adding a branch on pointer dereferencing. We cannot just assume that the compiler will be able to remove this branch in general and this operation is crucial for implementing efficient data structures.

After the yesterday's discussion I reverted the change to MutablePointer. In the future we might provide extra modes for checking pointer dereferences, but we don't have good support now.

StandardLibrary/Sources/Core/Pointer.hylo
kyouko-taiga: That's an acceptable compromise I think. In the meantime @lucteo and me have discussed and concluded that in any case this change should be part of another PR.
```suggestion
```

PR #1697, opened 2025-05-27
Move to macos-14

lucteo:


dabrahams:
It's possible you need to install gnu tar with brew. Just another WAG

lucteo:
The build failures seem to come from the fact that, changing the triple suffix (new: `apple-darwin24.1.0`), we don't have a corresponding archive in https://github.com/hylo-lang/llvm-build/releases/

dabrahams:
@lucteo I think this PR is obsolete now?  Or at least, the CI portion of it is.

lucteo:
> @lucteo I think this PR is obsolete now? Or at least, the CI portion of it is.

Parts of it are obsolete, but I still need to make sure that the OSes we run here are the same as we use in LLVM-build. I do have patches for Durian and Swifty-LLVM that need to go first.

tothambrus11:
Closing this for now, this has been addressed in https://github.com/hylo-lang/hylo/pull/1687

PR #1712, opened 2025-06-19
C mapping examples

tothambrus11:


MappingExamples/Union2.hylo
```diff
@@ -0,0 +1,231 @@
...
+        &storage = create_storage()
+        &set_from(bytes: bytes[representing: b])
+    }
+
+    public property a: Int32 {
+        let {
+            Pointer<Int32>(type_punning: storage.contiguous_storage).unsafe[]
+        }
+        inout {
+            var a = 0 as Int32
```
dabrahams, line 54: This doesn't look right.  After `for x in 0..<10 { someCUnion.a += 1 }` what is `someCUnion.a`?

Even if this is just a test case you should avoid doing weird things.

MappingExamples/Union2.hylo
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/// A C union type stores its data in a raw byte array, and we expose computed properties
+/// to access the individual members.
+///
+/// @repr(.C)
+type MyCUnion : Deinitializable {
+    internal var storage: Array<Int8> /// This should actually be a statically sized Buffer, but it's not yet implemented in the compiler. 
+    
+    private fun set_from(bytes: PointerToBuffer<Int8>) inout {
```
dabrahams, line 24: All these declarations need doc comments.

MappingExamples/Union2.hylo
```diff
@@ -0,0 +1,231 @@
...
+            precondition(i >= 0 && i < count, "Cannot access end position")
+            yield start.advance(by: i).unsafe[]
+        }
+    }
+}
+
+/// A C union type stores its data in a raw byte array, and we expose computed properties
+/// to access the individual members.
+///
+/// @repr(.C)
```
dabrahams, line 20: This comment is supposed to say what the type _is_ (see Swift API guidelines), e.g.

```suggestion
/// A Hylo representation of some C union type.
// @repr(.C)
```

The second line is not doc comment material.  I'm not actually sure what it is supposed to be, since you would never need to uncomment it… I think.

The rest of this comment belongs in implementation- or doc-comments in the body.

MappingExamples/CFunctionPointer1.hylo
```diff
@@ -0,0 +1,75 @@
+// FFI declaration for a C function
+// that takes two `int32_t`s and returns an `int32_t`
```
dabrahams, line 2: That comment is vacuous; all the information is in the signature.

```suggestion
/// An imported function that we can call directly or pass as a lambda.
```

Maybe?

MappingExamples/ffi-playground/readme.md
```diff
@@ -0,0 +1,3 @@
+This folder demonstrates how to compile and link a Hylo file together with a C library file.
```
dabrahams, line 1: for consistency, this fill should be called `README.md`. (uppercase)

dabrahams (changes requested):
There are several comments inline, but this PR is far too vast for me to evaluate properly, even if I hide whitespace-only changes, and appear to contain several things that could be distinct PRs.  Can it be broken up?

MappingExamples/Union2.hylo
tothambrus11: The Hylo compiler doesn't yet handle @repr(.C), but so far accidentally works as expected. It is indeed not part of the documentation.

tothambrus11:
Thanks for the review! This was mostly meant as a draft PR just for experimentation, so I will make a new one for introducing the C types and conversions after the numeric type code generator is done.

PR #1714, opened 2025-07-15
Analyze projections & check for errors.

lucteo:
Inspect projects and build a "skeleton" out of them, describing the important parts of a projection.

This skeleton is used to check that we have exactly one `yield` on all paths to a `return` instruction, but it is also useful to expand the projection into a ramp + slide function pairs.

(More PRs will follow for expanding projections)


lucteo:
Old PR. Aborting

PR #1715, opened 2025-07-15
Prepare for expanding projections

lucteo:
Add:
- add a new program pass to expand projections; this has two steps:
  - lower projections into ramp and slide pairs
  - lower project calls into calls to ramps and slides
- two new types of synthesized functions: projection ramp and slide
- add support for name mangling & demangling for ramps and slides

### DIFF
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -286,6 +286,9 @@ public struct Driver: ParsableCommand {
 
     var ir = try lower(program: program, reportingDiagnosticsTo: &log)
 
+    logVerbose("begin expandProjections pass.\n")
+    ir.expandProjections()
+
     if outputType == .ir || outputType == .rawIR {
       let m = ir.modules[sourceModule]!
       try m.description.write(to: irFile(productName), atomically: true, encoding: .utf8)
--- a/Sources/IR/Mangling/Mangler.swift
+++ b/Sources/IR/Mangling/Mangler.swift
@@ -363,6 +363,10 @@ struct Mangler {
       append(monomorphized: f, for: a, to: &output)
     case .synthesized(let d):
       append(synthesized: d, to: &output)
+    case .projectionRamp(let b):
+      append(projectionRamp: b, to: &output)
+    case .projectionSlide(let b):
+      append(projectionSlide: b, to: &output)
     case .existentialized:
       UNIMPLEMENTED()
     }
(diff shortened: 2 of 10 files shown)


Sources/Driver/Driver.swift
```diff
@@ -286,6 +286,9 @@ public struct Driver: ParsableCommand {
 
     var ir = try lower(program: program, reportingDiagnosticsTo: &log)
 
+    logVerbose("begin expandProjections pass.\n")
```
kyouko-taiga: I'm surprised to see camelCase here. Is that based on precedent?

Sources/Driver/Driver.swift
```diff
@@ -286,6 +286,9 @@ public struct Driver: ParsableCommand {
...
+    ir.expandProjections()
```
kyouko-taiga: Adding the pass here makes it impossible to output raw IR. The `--emit raw-ir` option is meant to look at the output of emitter. If we want to see IR after just this pass we should add another flag (or generalize to print IR after specific phases).

Sources/IR/Mangling/Mangler.swift
```diff
@@ -377,6 +381,20 @@ struct Mangler {
...
+  /// Writes the mangled representation of `s` to `output`.
+  private mutating func append(projectionRamp s: Function.ID, to output: inout Output
+  ) {
+    append(operator: .projectionRampFunctionDecl, to: &output)
+    append(function: s, to: &output)
+  }
+
+  /// Writes the mangled representation of `s` to `output`.
+  private mutating func append(projectionSlide s: Function.ID, to output: inout Output
+  ) {
```
kyouko-taiga:
```suggestion
  private mutating func append(
    projectionRamp s: Function.ID, to output: inout Output
  ) {
    append(operator: .projectionRampFunctionDecl, to: &output)
    append(function: s, to: &output)
  }

  /// Writes the mangled representation of `s` to `output`.
  private mutating func append(
    projectionSlide s: Function.ID, to output: inout Output
  ) {
```

Sources/Driver/Driver.swift
lucteo: You mean we should use `expand projections` instead of `expendProjections`, right? I can change that.
Previously we have the name of the pass be exactly copied in the strings, but this is the first multi-word pass.

Sources/Driver/Driver.swift
lucteo: I was confused where I should have put this pass (and also, the analysis pass from previous PR, #1714).

I did want to place this before dumping the IR. But, I think I need to separe out the block below to have different IRs written for `raw-ir` and `ir`.

lucteo:
Old PR. Aborting.

PR #1716, opened 2025-08-02
Add `IRWriter` to isolate the part of writing to IR

lucteo:
We want to check constraints while writing to the IR. Currently, most of the functionality of writing IR is distributed between `Module` and its users. We want to reduce the surface for modifying the IR to be able to improve constraint-checking.

Changes:
- add `IRWriter`, moving from `Module` all the functions that modify the IR of functions
- add `IRInspector` protocol to abstract out the inspection of function IR; both `Module` and `IRWriter` implement this protocol, and thus can reuse the functions for inspecting function IR
- other minor changes to support this separation

At this point, we are not adding any new constraint checking. This just prepares the way to consolidate the IR writing logic in one place, so that we can add constraints on top of this.

NOTE: Putting this on pause to investigate if we can use `InstructionID` without function, and move most of the functionality into `Function`.


Sources/IR/Module.swift
```diff
@@ -919,7 +916,7 @@ public struct Module {
   ///
   /// - Requires: The result of `i` have no users.
   mutating func removeInstruction(_ i: InstructionID) {
```
dabrahams: Only a related topic but it seems like this function's guts ought to be in a method on `Function`, and `removeUsesMadeBy` should be moved there as well.

Sources/IR/Analysis/Module+AccessReification.swift
```diff
@@ -143,13 +145,17 @@ extension Module {
     var arguments = s.operands
     for a in arguments.indices where s.parameters[a].access == .yielded {
       let b = makeAccess([k], from: arguments[a], at: s.site)
-      arguments[a] = .register(insert(b, before: i))
+      arguments[a] = .register(modifyIR(of: i.function) { (w) in
+        w.insert(b, at: .before(i))
```
dabrahams, line 149: We should decide whether methods like this one should use the `make`/`insert` pattern or the instruction DSL pattern (`_access(...)`).

Sources/IR/Emitter.swift
```diff
@@ -415,7 +430,10 @@ struct Emitter {
     guard let b = ast[d].body else { return }
 
     // Create the function entry.
-    let entry = module.appendEntry(in: program.scopeContainingBody(of: d)!, to: f)
+    let scope = program.scopeContainingBody(of: d)!
```
dabrahams, line 433: There's almost always a better name than one that simply captures type information.  I might use `s`, especially if it's only used once.  But does this computation need to be hoisted and named, or could it remain inline?

Sources/IR/IRWriter.swift
```diff
@@ -0,0 +1,291 @@
+import FrontEnd
+import Utils
+
+/// Provides functionality to write IR instructions for a function.
+///
+/// Maintains the constraints needed to have a well-formed IR.
+public struct IRWriter {
```
dabrahams, line 7: In order to review this, I need to know what might be new.  If it's a mechanical transplant out of `Emitter`, I don't need to look at it very carefully.  Otherwise it might need more scrutiny.

dabrahams (approved):
Mostly looks great, but I'm concerned about two things.  They might be OK, but would like to discuss:

1. The amount of forwarding from `Emitter` to `IRWriter`.  I realize it keeps lots of `Emitter` code stable, but I wonder if that could shouldn't simply use the `Emitter`'s `IRWriter` to do things directly rather than calling `Emitter` methods.
2. The new protocol.  In general it's better to use-a thing than to be-a thing.  Why isn't IRWriter the thing with inspection powers, which Emitter uses.  OR, why isn't there a single `IRInspector` concrete type that both of them use?

this change.  In general it's better to use-a thing than to be-a thing.  Why is the inspection capability not part

Sources/IR/Analysis/Module+AccessReification.swift
lucteo: my plan was to move more an more logic from Emitter into IRWriter, so that other parts of the code can benefit from it. If we set the "site" and the "insertionPoint" to IRWriter, we can reduce most of  this code to a single function call into IRWriter (i.e., `_access(...)`).

Sources/IR/Emitter.swift
lucteo: one of the downsides of using `IRWriter` is that I can't access any member of `module` (including `program`) inside it, as I would break the law of exclusivity. This implies that I need to read all the properties before modifying IR. This is why I had to extract `scope` from the call to `appendEntry`.

I can change the name to `s`.

Sources/IR/IRWriter.swift
lucteo: This is almost an identical copy from `Module`. There are a few small changes here and there.
Here are some of the changes:
- change how we access the `Function` object needing to change (we have it stored now)
- some minor changes that involved looking at the `Function.ID`
- remove some parameters that don't make sense if we know that we are looking at a specific function.
- `uses` is not part of `Function` instead of `Module`
- no more direct access to `Module` class

Sources/IR/Module.swift
lucteo: I was avoiding this, as most of the mutating functions on the IR of a function are outside of `Function` itself. The `Function` class mainly holds the data for the IR of the function, but we keep the functionality separated.

One of the reasons is that `Function` (mostly) doesn't know about `InstructionID`s. It onl;y knows about block IDs and instruction addresses, but it doesn't know about the ID of itself.

Now that I'm thinking about it, maybe a better way is to remove `Function.ID` from `InstructionID`. We only care about instructions inside of a function, we never have cross-function instructions. If that's the case, then we can move all the functionality from `IRWriter` (and `IRInspector`) into `Function` 🤔

lucteo:
> 1. The amount of forwarding from `Emitter` to `IRWriter`.  I realize it keeps lots of `Emitter` code stable, but I wonder if that could shouldn't simply use the `Emitter`'s `IRWriter` to do things directly rather than calling `Emitter` methods.

The plan is to slowly move a lot of the extra functionality from `Emitter` to `IRWriter`; this way, the passes that don't use `Emitter` can also benefit from them.

> 2. The new protocol.  In general it's better to use-a thing than to be-a thing.  Why isn't IRWriter the thing with inspection powers, which Emitter uses.  OR, why isn't there a single `IRInspector` concrete type that both of them use?
> 
> this change. In general it's better to use-a thing than to be-a thing. Why is the inspection capability not part

The reason why I need to have it split is the law of exclusivity. There are many cases in which you are inspecting the IR while you are changing it. In a temporary patch, I had to duplicate many functions from `Module` into `IRWriter` so I don't have two references to `Function`.

lucteo:
Putting this on pause to investigate if we can use `InstructionID` without function, and move most of the functionality into `Function`.

PR #1719, opened 2025-08-12
Start using local instruction IDs

lucteo:
Currently `InstructionID` contains a `Function.ID`. This means that all the utilities built on top of `InstructionID` need to be places in `Module` (and above); but we can't put anything in `Function`.

The same apply to `Block.ID`.

This PR does the following:
- rename `Block.ID` to `Block.AbsoluteID`
- remove the  `Function.ID` part from  `InstructionID` and `Block.ID`
- most most of the code to use `InstructionID` and `Block.ID`
- move def-use relations in `Function`
- adds a way to get all `InstructionID`s from a function


Sources/CodeGen/LLVM/Transpilation.swift
```diff
@@ -1031,56 +1031,56 @@ extension SwiftyLLVM.Module {
       case .signedAdditionWithOverflow(let t):
         let l = llvm(s.operands[0])
         let r = llvm(s.operands[1])
-        let f = intrinsic(
+        let ff = intrinsic(
```
kyouko-taiga:
```suggestion
        let g = intrinsic(
```

Sources/CodeGen/LLVM/Transpilation.swift
```diff
@@ -683,8 +683,8 @@ extension SwiftyLLVM.Module {
       frame = nil
     }
 
-    for i in context.source[context.source.entry(of: f)!].inputs.indices {
-      let o = Operand.parameter(.init(f, entry), i)
+    for i in context.source[context.source.entry(of: f)!, in: f].inputs.indices {
```
kyouko-taiga, line 686: If the goal is that IR functions are more independent then I think of lot of the APIs that are currently on the context or the program should be moved to `IRFunction`. Even if it has to be mutable we could write most of this code in a function that takes the modified `IRFunction` as an inout parameter.

Sources/CodeGen/LLVM/Transpilation.swift
kyouko-taiga: I'd rather stick to single-letter variables. Longer names typically mean something in the codebase so the general rule of thumb is that it's better to pick an arbitrary letter than start trying to make a name makes without spelling out the full word.

Sources/IR/AbsoluteInstructionID.swift
```diff
@@ -0,0 +1,48 @@
+import Utils
+
+/// The stable identity of an instruction in its module.
+///
+/// - SeeAlso: `InstructionID`
+public struct AbsoluteInstructionID: Hashable {
```
kyouko-taiga: I don't think this type should exist as it creates inconsistencies in the API. There are now many calls in `Transpilation.swift` in which the function containing an instruction must be passed as an extra parameter. I suppose that is fine but then I don't think we should have other APIs adopting a different convention.

There is very little use overall for absolute instruction IDs since referring to an instruction in a different function is illegal anyway. So I think we should work to redesigns APIs around IR functions and simply assume that when an instruction is mentioned it is part of the function present in the argument list.

Sources/IR/Analysis/AbstractContext.swift
```diff
@@ -70,9 +70,9 @@ struct AbstractContext<Domain: AbstractDomain>: Equatable {
 
   /// Adds a new memory cell in `context` and binds its address to `i`, which is in `m`.
   mutating func declareStorage(
-    assignedTo i: InstructionID, in m: Module, initially initialState: Domain
+    assignedTo i: InstructionID, from f: Function.ID, in m: Module, initially initialState: Domain
```
kyouko-taiga, line 73: What's the rationale for taking `InstructionID` and `Function.ID` as separate parameter where instead of using the new `AbsoluteInstructionID` type? I don't think the latter should exist but I find it even more important to keep an API that is as consistent as possible.

Sources/IR/Analysis/Module+AccessReification.swift
```diff
@@ -55,55 +55,55 @@ extension Module {
...
+      let s = self[i, in: f] as! ReifiableAccess
       let available = s.capabilities
       assert(!available.isSingleton, "access already reified")
 
       var lower = AccessEffect.let
       var upper = AccessEffect.let
 
-      forEachClient(of: i) { (u) in
-        let rs = requests(u)
+      forEachClient(of: i, in: f) { (u) in
```
kyouko-taiga, line 65: `forEachClient(of:)` should be called as a method of `f` and I think the same goes for a bunch of other functions in the file.

Sources/IR/AbsoluteInstructionID.swift
lucteo: I think there might still be cases in which we might need `AbsoluteInstructionID`. The one that I have in mind is the interpreter, which might need to jump from function to function.

But, I'm ok removing it until that case comes. However, we still have a case in which we have `Block.AbsoluteID`, in Emitter with `stackOnEntry`.

Sources/IR/Analysis/AbstractContext.swift
lucteo: I'm trying to break my refactoring into smaller PRs (this PR is still huge for my standards). The idea is to allow to move many of the functions declared into `Module` to be declared in `Function`. At that point, we can completely get rid of these extra `Function.ID` parameters.

That is my target full refactoring; although I have a few patches up, there is still work to be done to get to that goal.

Sources/IR/Analysis/Module+AccessReification.swift
lucteo: Indeed, but I did not get there with my refactoring.

Sources/CodeGen/LLVM/Transpilation.swift
lucteo: Yes, that is the goal. But there are a ton of changes to get there, and I need to take an iterative approach to this. This PR is just the first step in that direction (this PR is already huge).

lucteo:
> I understand the direction of this change but I think it's worsening the API in this current form. I wondered if we could not do significantly better by trying to move some (all?) of the algorithms that are currently written on the module into `IRFunction` and then reorganize code so that most modifications are performed in a mutating method of the latter. That would get rid of many, many parameters.

Somehow that is the plan, but we cannot put all the backend into `Function` all at once. I need to take an iterative approach, and this seemed to be a good first step into that direction. By the end of the several PRs that I have, you see that I'm taking steps into that direction, but even after those PRs, there are still things that need to be done.

Hope this makes sense.

Sources/CodeGen/LLVM/Transpilation.swift
lucteo: Fixed.

Sources/IR/Analysis/Module+Depolymorphize.swift
```diff
@@ -35,16 +35,14 @@ extension IR.Program {
...
-          depolymorphize(project: i, definedIn: m)
-        default:
-          continue
-        }
+    for i in modules[m]!.instructions(in: f) {
+      switch modules[m]![i, in: f] {
+      case is Call:
+        depolymorphize(call: i, from:f, definedIn: m)
+      case is Project:
+        depolymorphize(project: i, from:f, definedIn: m)
```
kyouko-taiga:
```suggestion
        depolymorphize(call: i, from: f, definedIn: m)
      case is Project:
        depolymorphize(project: i, from: f, definedIn: m)
```

Sources/IR/Analysis/Module+Depolymorphize.swift
```diff
@@ -35,16 +35,14 @@ extension IR.Program {
...
-        switch modules[m]![i] {
-        case is Call:
-          depolymorphize(call: i, definedIn: m)
-        case is Project:
-          depolymorphize(project: i, definedIn: m)
-        default:
-          continue
-        }
+    for i in modules[m]!.instructions(in: f) {
+      switch modules[m]![i, in: f] {
```
kyouko-taiga, line 39: I'm surprised it's not possible to assign the function identified by `f` to some local variable `x` and then write `x[i]` pretty much everywhere. Perhaps it's what you plan on doing in the next iteration. If so that's fine.

Sources/IR/Analysis/AbstractContext.swift
lucteo: Resolving per our discussion.

Sources/IR/BlockID.swift
```diff
@@ -2,30 +2,30 @@ import Utils
 
 extension Block {
 
-  /// The ID of a basic block.
+  /// The stable ID of a basic block in its function.
```
kyouko-taiga, line 5: This will eventually become a simple type alias, right?

Sources/IR/BlockID.swift
```diff
@@ -41,10 +41,50 @@ extension Block {
 
   }
 
+  /// The absolute ID of a basic block.
```
kyouko-taiga, line 44: Is it correct that the only use of this type is in the emitter to keep track of the state of the stack? If so I'm not sure this type is justified, or at least I think it should be defined privately in the emitter.

Sources/IR/Emitter.swift
```diff
@@ -57,11 +57,14 @@ struct Emitter {
   private var loops = LoopIDs()
 
   /// For each block, the state of `frames` where the block was first entered.
-  private var stackOnEntry: [ Block.ID: Stack ] = [:]
+  private var stackOnEntry: [ Block.AbsoluteID: Stack ] = [:]
```
kyouko-taiga:
```suggestion
  private var stackOnEntry: [Block.AbsoluteID: Stack] = [:]
```

I should have caught this one on Dave's PR.

Sources/IR/Emitter.swift
```diff
@@ -2954,21 +2959,22 @@ struct Emitter {
   /// taking the place of `i`.
   ///
   /// After the call, `insertionPoint` set to `nil`.
-  mutating func replaceMove(_ i: InstructionID, with semantics: AccessEffect) -> InstructionID {
-    let s = module[i] as! Move
-    let predecessor = module.instruction(before: i)
+  mutating func replaceMove(_ i: InstructionID, in f: Function.ID, with semantics: AccessEffect) -> InstructionID {
+    let s = module[i, in: f] as! Move
+    insertionFunction = f
```
kyouko-taiga, line 2965: We should either restore the state of `insertionFunction` or document the fact that `replaceMove(_:with:)` modifies the insertion function. I think the former is better.

Sources/IR/Emitter.swift
```diff
@@ -3581,7 +3590,7 @@ extension Emitter {
   /// This test is used to ensure that all points branching to a block
   /// have consistent stack allocations.
   fileprivate mutating func checkEntryStack(_ b: Block.ID) {
-    modify(&stackOnEntry[b]) { x in
+    modify(&stackOnEntry[Block.AbsoluteID(insertionFunction!, b)]) { x in
```
kyouko-taiga, line 3595: Is that the only use of absolute IDs (see other comments)? If so I think we can just have `stackOnEntry` be defined differently so that we don't need this type. I even suspect that we can come up with a more efficient data structure than the current one because the number of basic blocks is fixed (or at least it's easy to maintain an invariant involving that number in the whole function).

Sources/IR/InstructionID.swift
```diff
@@ -1,13 +1,8 @@
 import Utils
 
-/// The stable identity of an instruction in its module.
-///
-/// - SeeAlso: `InstructionIndex`
+/// The stable identity of an instruction in its function.
```
kyouko-taiga, line 3: Shouldn't that be called an AbsoluteInstructionID for consistency with your block IDs?

Sources/IR/Module.swift
```diff
@@ -171,58 +167,68 @@ public struct Module {
...
+
+  /// Returns the IDs of the instructions in `f`, from all the blocks.
+  public func instructions(in f: Function.ID) -> LazySequence<
+    FlattenSequence<
+      LazyMapSequence<
+        LazySequence<DefaultIndices<Function.Blocks>>.Elements,
+        LazyMapSequence<Block.Instructions.Indices, InstructionID>
+      >
+    >
+  > {
```
kyouko-taiga:
```suggestion
  public func instructions(in f: Function.ID) -> some Collection<InstructionID> {
```
Does that work?

Sources/IR/BlockID.swift
lucteo: Yes. That's what we've agreed, but I haven't yet any PR for this.

Sources/IR/BlockID.swift
lucteo: Yes, that is true.
Is this ok if I do this in a follow-up (very small) PR?

Sources/IR/Emitter.swift
lucteo: Yes. Can we make this change in a dedicated PR?

Sources/IR/InstructionID.swift
lucteo: No.
Both `InstructionID` and `Block.ID` are now missing the `Function.ID` part.
Previously I had `AbsoluteInstructionID` that would correspond to `Block.AbsoluteID` which also contained `Function.ID`. We discussed that we don't need an `AbsoluteInstructionID`. Based on other comments, I understand that we want to get rid of `Block.AbsoluteID` too, so then we just have `InstructionID` and `Block.ID` (which will later become just type aliases)

Sources/IR/Module.swift
lucteo: Yes, it does. I had a trouble with a similar use of `some Collection<T>` but apparently here it works like a charm.

Sources/IR/BlockID.swift
kyouko-taiga: OK

Sources/IR/Emitter.swift
kyouko-taiga: OK

kyouko-taiga:
Did you make the change in the emitter to restore the state of the insertion function? If you did I'm fine with the PR.

kyouko-taiga (approved):
LGTM

PR #1720, opened 2025-08-15
Copy basic utilities for inspecting & writing IR inside of `Function`

lucteo:
Copy basic utilities that work on `IR` inside of `Function`.

Also:
 - Module functions call into the `Function` level functions
 - `Function.entry` now returns a `Block.ID` and is a property

### DIFF
--- a/Sources/IR/Function.swift
+++ b/Sources/IR/Function.swift
@@ -126,7 +126,7 @@ public struct Function {
       parameters.append(.address(output))
     }
 
-    return appendBlock(in: scope, taking: parameters)
+    return append(in: scope, taking: parameters)
   }
 
   /// Removes `block` and updates def-use chains.
@@ -156,6 +156,9 @@ public struct Function {
   }
 
   /// Replaces all uses of `old` with `new` and updates the def-use chains.
+  ///
+  /// The def-use chains are updated so that the uses formerly made by `old` are
+  /// replaced by the uses made by `new` and all former uses of `old` refer to `new`.
   ///
   /// - Requires: `new` as the same type as `old`.
   mutating func replaceUses(of old: Operand, with new: Operand) {
@@ -257,7 +260,7 @@ public struct Function {
   ///   are also in `S`.
   mutating func removeAllInstructions(after i: InstructionID) {
     while let a = self[i.block].instructions.lastAddress, a != i.address {
-      removeInstruction(.init(i.block, a))
+      remove(.init(i.block, a))
     }
   }
 
--- a/Sources/IR/Analysis/Module+Depolymorphize.swift
+++ b/Sources/IR/Analysis/Module+Depolymorphize.swift
@@ -131,7 +131,7 @@ extension IR.Program {
       let inputs = modules[source]![s, in: f].inputs.map { (t) in
         monomorphize(t, for: z, usedIn: scopeOfUse)
       }
-      rewrittenBlock[s] = modules[target]![result].appendBlock(in: modules[source]![s, in: f].scope, taking: inputs)
+      rewrittenBlock[s] = modules[target]![result].append(in: modules[source]![s, in: f].scope, taking: inputs)
     }
 
     let rewrittenGenericValue = modules[target]!.defineGenericValueArguments(z, in: result)
--- a/Sources/IR/Module.swift
+++ b/Sources/IR/Module.swift
@@ -123,7 +123,7 @@ public struct Module {
   /// no particular order.
   public func blocks(
     in f: Function.ID
-  ) -> LazyMapSequence<Function.Blocks.Indices, Block.ID> {
+  ) -> some RandomAccessCollection<Block.ID> {
     functions[f]!.blockIDs
   }
 
@@ -690,7 +690,7 @@ public struct Module {
     taking parameters: [IR.`Type`] = [],
     to f: Function.ID
   ) -> Block.ID {
-    functions[f]!.appendBlock(in: scope, taking: parameters)
+    functions[f]!.append(in: scope, taking: parameters)
   }
 
   /// Removes `block` and updates def-use chains.
@@ -698,7 +698,7 @@ public struct Module {
   /// - Requires: No instruction in `block` is used by an instruction outside of `block`.
   @discardableResult
   mutating func removeBlock(_ block: Block.ID, from f: Function.ID) -> Block {
-    functions[f]!.removeBlock(block)
+    functions[f]!.remove(block)
   }
 
   /// Swaps `old` by `new`.
@@ -753,7 +753,7 @@ public struct Module {
   ///
   /// - Requires: The result of `i` have no users.
   mutating func removeInstruction(_ i: InstructionID, in f: Function.ID) {
-    functions[f]!.removeInstruction(i)
+    functions[f]!.remove(i)
   }
 
   /// Removes all instructions after `i` in its containing block and updates def-use chains.


Sources/IR/Function.swift
```diff
@@ -54,25 +51,50 @@ public struct Function {
...
-    blocks.remove(at: address)
+  /// Returns the IDs of the instructions in `self`, from all the blocks.
+  public var instructions: LazySequence<
+    FlattenSequence<
+      LazyMapSequence<
+        LazySequence<DefaultIndices<Function.Blocks>>.Elements,
+        LazyMapSequence<Block.Instructions.Indices, InstructionID>
+      >
+    >
+  > {
```
kyouko-taiga:
```suggestion
  public var instructions: some Collection<InstructionID>
  > {
```
Does this work?

Sources/IR/Function.swift
```diff
@@ -88,4 +110,369 @@ public struct Function {
     return result
   }
 
+
+  /// Appends to `self` a basic block in `scope` that accepts `parameters`, returning its address.
+  ///
+  /// The new block will become the function's entry if `self` contains no block before
+  /// `appendBlock` is called.
+  ///
+  /// TODO: merge `appendBlock` and `appendEntry`
```
kyouko-taiga, line 111: Isn't that done?

Sources/IR/Function.swift
```diff
@@ -88,4 +110,369 @@ public struct Function {
...
+  @discardableResult
+  mutating func removeBlock(_ block: Block.ID) -> Block {
+    for i in self.instructions(in: block) {
+      precondition(self.allUses(of: i).allSatisfy({ $0.user.block == block.address }))
+      removeUsesMadeBy(i)
+    }
+    return blocks.remove(at: block.address)
+  }
+
+  /// Swaps `old` by `new`.
```
kyouko-taiga:
```suggestion
  /// Replaces `old` for `new`.
```

Sources/IR/Function.swift
```diff
@@ -88,4 +110,369 @@ public struct Function {
...
+  ///
+  /// - Requires: `new` produces results with the same types as `old`.
+  mutating func replace<I: Instruction>(_ old: InstructionID, with new: I) {
+    precondition(self[old].result == new.result)
+    removeUsesMadeBy(old)
+    self[old] = new
+    addUses(for: new, with: old)
+  }
+
+  /// Swaps all uses of `old` in `self` by `new` and updates the def-use chains.
```
kyouko-taiga:
```suggestion
  /// Replaces all uses of `old` in `self` for `new` and updates the def-use chains.
```

Sources/IR/Function.swift
lucteo: I did it in a later PR

Sources/IR/Function.swift
lucteo: Will need to rebase

Sources/IR/Function.swift
```diff
@@ -54,25 +51,43 @@ public struct Function {
     _modify { yield &blocks[i.block].instructions[i.address] }
   }
 
+  /// Accesses the instruction denoted by `o` if it is `.register`; returns `nil` otherwise.
+  public subscript(o: Operand) -> Instruction? {
```
dabrahams:
```suggestion
  public subscript(register o: Operand) -> Instruction? {
```

I don't know for sure, but since every operand corresponds to an instruction, it seems like the semantics of this are surprising without a label.  It's arguable whether this is best as a subscript or a method.

Sources/IR/Function.swift
```diff
@@ -54,25 +51,43 @@ public struct Function {
...
+  /// The entry of the function.
+  public var entry: Block.ID? { blocks.firstAddress.map(Block.ID.init) }
+
   /// `true` iff the function takes generic parameters.
   public var isGeneric: Bool {
     !genericParameters.isEmpty
   }
 
-  /// Appends to `self` a basic block in `scope` that accepts `parameters`, returning its address.
+  /// Returns the operand representing the return value of `self`.
```
dabrahams:
```suggestion
  /// Returns the identity of the return value.
```

Just trying to remove info that's redundant with the signature.

Sources/IR/Function.swift
```diff
@@ -26,14 +26,11 @@ public struct Function {
   public let output: AnyType
 
   /// The blocks in the function.
```
dabrahams:
```suggestion
  /// The basic blocks of the control flow graph.
```

Trying to say more that isn't trivially redundant.

Sources/IR/Function.swift
```diff
@@ -54,25 +51,43 @@ public struct Function {
...
+  /// Returns the operand representing the return value of `self`.
+  public var returnValue: Operand? {
+    if !isSubscript, let e = entry {
+      return Operand.parameter(e, inputs.count)
+    } else {
+      return nil
+    }
+  }
+
+  /// Returns the IDs of the blocks in `self`.
```
dabrahams:
```suggestion
  /// The IDs of all basic blocks in the CFG, beginning with the entry point block.
```

A property doesn't return anything.  Suggest deleting the rest of the comment.

Sources/IR/Function.swift
```diff
@@ -54,25 +51,43 @@ public struct Function {
...
   ///
-  /// The new block will become the function's entry if `self` contains no block before
-  /// `appendBlock` is called.
-  mutating func appendBlock<T: ScopeID>(
-    in scope: T, taking parameters: [IR.`Type`]
-  ) -> Blocks.Address {
-    blocks.append(Block(scope: AnyScopeID(scope), inputs: parameters))
+  /// The first element of the returned collection is the function's entry; other elements are in
+  /// no particular order.
+  public var blockIDs: LazyMapSequence<Function.Blocks.Indices, Block.ID> {
```
dabrahams:
```suggestion
  public var blockIDs: some RandomAccessCollection<Block.ID> {
```

Sources/IR/Function.swift
```diff
@@ -54,25 +51,43 @@ public struct Function {
...
+    blocks.indices.lazy.map({ .init($0.address) })
   }
 
-  /// Removes the block at `address`.
-  @discardableResult
-  mutating func removeBlock(_ address: Blocks.Address) -> Block {
-    blocks.remove(at: address)
+  /// Returns the IDs of the instructions in `self`, from all the blocks.
```
dabrahams:
```suggestion
  /// The IDs of all instructions.
```

Sources/IR/Function.swift
```diff
@@ -88,4 +103,369 @@ public struct Function {
     return result
   }
 
+
+  /// Appends to `self` a basic block in `scope` that accepts `parameters`, returning its address.
+  ///
+  /// The new block will become the function's entry if `self` contains no block before
+  /// `appendBlock` is called.
```
dabrahams:
```suggestion
  /// The first block appended becomes the entry point block.
```

I wonder if that's a guarantee we want to make to callers, since there is a method to explicitly append an entry point?

Sources/IR/Function.swift
```diff
@@ -88,4 +103,369 @@ public struct Function {
...
+  ///
+  /// TODO: merge `appendBlock` and `appendEntry`
+  mutating func appendBlock<T: ScopeID>(
```
dabrahams:
```suggestion
  mutating func append<T: ScopeID>(
```

Keep type information out of non-type names whenever possible.
Is the fact that it's *appending* actually relevant to callers, or should this just be called `add`?

Sources/IR/Function.swift
```diff
@@ -88,4 +103,369 @@ public struct Function {
...
+    }
+
+    return appendBlock(in: scope, taking: parameters)
+  }
+
+  /// Removes `block` and updates def-use chains.
+  ///
+  /// - Requires: No instruction in `block` is used by an instruction outside of `block`.
+  @discardableResult
+  mutating func removeBlock(_ block: Block.ID) -> Block {
```
dabrahams:
```suggestion
  mutating func remove(_ block: Block.ID) -> Block {
```

Sources/IR/Function.swift
```diff
@@ -88,4 +103,369 @@ public struct Function {
     return result
   }
 
+
+  /// Appends to `self` a basic block in `scope` that accepts `parameters`, returning its address.
```
dabrahams:
```suggestion
  /// Appends a basic block in `scope` accepting `parameters`, returning its address.
```
Can avoid using `self`.  Everything is with respect to `self` unless otherwise stated.

Sources/IR/Function.swift
```diff
@@ -88,4 +103,369 @@ public struct Function {
...
+  @discardableResult
+  mutating func removeBlock(_ block: Block.ID) -> Block {
+    for i in self.instructions(in: block) {
+      precondition(self.allUses(of: i).allSatisfy({ $0.user.block == block.address }))
+      removeUsesMadeBy(i)
+    }
+    return blocks.remove(at: block.address)
+  }
+
+  /// Replaces `old` for `new`.
```
dabrahams:
```suggestion
  /// Replaces `old` with `new`.
```

Sources/IR/Function.swift
```diff
@@ -88,4 +103,369 @@ public struct Function {
...
+  ///
+  /// `old` is removed and the def-use chains are updated so that the uses made by `old` are
```
dabrahams:
```suggestion
  /// The def-use chains are updated so that the uses formerly made by `old` are
```

Sources/IR/Function.swift
```diff
@@ -88,4 +103,369 @@ public struct Function {
...
+  /// replaced by the uses made by `new` and all uses of `old` refer to `new`. After the call,
```
dabrahams:
```suggestion
  /// replaced by the uses made by `new` and all former uses of `old` refer to `new`. After the call,
```
 
Ideally this should just be captured in an invariant of `Function` that is documented in its description (or even implied by the meaning of def-use chain!), though that [might require making some variables not publicly settable.](https://val-qs97696.slack.com/archives/C01MEFN3TGE/p1761233302537269)

Sources/IR/Function.swift
```diff
@@ -88,4 +103,369 @@ public struct Function {
...
+  ///
+  /// - Requires: `new` produces results with the same types as `old`.
+  mutating func replace<I: Instruction>(_ old: InstructionID, with new: I) {
+    precondition(self[old].result == new.result)
+    removeUsesMadeBy(old)
+    self[old] = new
+    addUses(for: new, with: old)
+  }
+
+  /// Replaces all uses of `old` in `self` for `new` and updates the def-use chains.
```
dabrahams:
```suggestion
  /// Replaces all uses of `old` with `new` and updates the def-use chains.
```

Sources/IR/Function.swift
dabrahams: Why is it necessary to give such a detailed description of updating def-use chains here but not on line 159?

Sources/IR/Function.swift
```diff
@@ -88,4 +103,369 @@ public struct Function {
...
+      predecessor.block,
+      self[predecessor.block].instructions.insert(newInstruction, after: predecessor.address))
+    addUses(for: newInstruction, with: i)
+    return i
+  }
+
+  /// Removes instruction `i` and updates def-use chains.
+  ///
+  /// - Requires: The result of `i` have no users.
+  mutating func removeInstruction(_ i: InstructionID) {
```
dabrahams:
```suggestion
  mutating func remove(_ i: InstructionID) {
```

PR #1721, opened 2025-08-15
Use IR basic utilities from `Function`

lucteo:
We are not using functions in `Module` for common IR operations (inspect & write). Instead, we directly call `Function` for this.


Sources/IR/BlockID.swift
```diff
@@ -42,7 +42,7 @@ extension Block {
   }
 
   /// The absolute ID of a basic block.
-  public struct AbsoluteID: Hashable {
+  public struct AbsoluteID: Hashable, Sendable {
```
lucteo: This will be removed by #1738

PR #1727, opened 2025-09-06
Improve TupleType API naming

dabrahams:
Scratching an itch

### DIFF
--- a/Sources/FrontEnd/Types/AnyType.swift
+++ b/Sources/FrontEnd/Types/AnyType.swift
@@ -65,7 +65,7 @@ public struct AnyType {
   public static let never = ^UnionType([])
 
   /// Hylo's `Void` type.
-  public static let void = ^TupleType([])
+  public static let void = ^TupleType([] as [AnyType])
 
   /// A shorthand for `^ErrorType()`.
   public static let error = ^ErrorType()
@@ -167,7 +167,7 @@ public struct AnyType {
   public var isBuiltinOrRawTuple: Bool {
     precondition(isCanonical)
     if let b = TupleType(self) {
-      return b.elements.allSatisfy(\.type.isBuiltin)
+      return b.components.allSatisfy(\.type.isBuiltin)
     } else {
       return base is BuiltinType
     }
--- a/Sources/FrontEnd/Types/TupleType.swift
+++ b/Sources/FrontEnd/Types/TupleType.swift
@@ -3,8 +3,8 @@ import Utils
 /// A tuple type.
 public struct TupleType: TypeProtocol {
 
-  /// An element in a tuple type.
-  public struct Element: Hashable {
+  /// A part of a tuple type.
+  public struct Component: Hashable {
 
     /// Creates an instance having the given properties.
     public init(label: String?, type: AnyType) {
@@ -20,36 +20,36 @@ public struct TupleType: TypeProtocol {
 
   }
 
-  /// The elements of the tuple.
-  public let elements: [Element]
+  /// The components of the tuple.
+  public let components: [Component]
 
   public let flags: ValueFlags
 
-  /// Creates a tuple type with a sequence of elements.
-  public init<S: Sequence>(_ elements: S) where S.Element == Element {
-    self.elements = Array(elements)
-    self.flags = ValueFlags(self.elements.map(\.type.flags))
+  /// Creates a tuple type having the given components.
+  public init<S: Sequence>(_ components: S) where S.Element == Component {
+    self.components = Array(components)
+    self.flags = ValueFlags(self.components.map(\.type.flags))
   }
 
-  /// Creates a tuple type with a sequence of label-type pairs.
-  public init<S: Sequence>(labelsAndTypes: S) where S.Element == (String?, AnyType) {
-    self.init(labelsAndTypes.map({ Element(label: $0.0, type: $0.1) }))
+  /// Creates a tuple type having the given components.
+  public init<S: Sequence>(_ components: S) where S.Element == (String?, AnyType) {
+    self.init(components.map({ Component(label: $0.0, type: $0.1) }))
   }
 
-  /// Creates a tuple of unlabeled elements with the types in the given sequence.
-  public init<S: Sequence>(types: S) where S.Element == AnyType {
-    self.init(types.map({ Element(label: nil, type: $0) }))
+  /// Creates a tuple having the given unlabeled components.
+  public init<S: Sequence>(_ components: S) where S.Element == AnyType {
+    self.init(components.map({ Component(label: nil, type: $0) }))
   }
 
   /// The labels of the tuple.
-  public var labels: LazyMapSequence<[Element], String?> {
-    elements.lazy.map(\.label)
+  public var labels: LazyMapSequence<[Component], String?> {
+    components.lazy.map(\.label)
   }
 
   public func transformParts<M>(
     mutating m: inout M, _ transformer: (inout M, AnyType) -> TypeTransformAction
   ) -> Self {
-    let newElements = elements.map { (e) -> Element in
+    let newElements = components.map { (e) -> Component in
       .init(label: e.label, type: e.type.transform(mutating: &m, transformer))
     }
     return TupleType(newElements)
@@ -60,16 +60,16 @@ public struct TupleType: TypeProtocol {
 extension TupleType: CustomStringConvertible {
 
   public var description: String {
-    if (elements.count == 1) && (elements[0].label == nil) {
-      return "{\(elements[0]),}"
+    if (components.count == 1) && (components[0].label == nil) {
+      return "{\(components[0]),}"
     } else {
-      return "{\(list: elements)}"
+      return "{\(list: components)}"
     }
   }
 
 }
 
-extension TupleType.Element: CustomStringConvertible {
+extension TupleType.Component: CustomStringConvertible {
 
   public var description: String {
     if let label = label {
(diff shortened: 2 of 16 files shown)


Sources/FrontEnd/Types/AnyType.swift
```diff
@@ -65,7 +65,7 @@ public struct AnyType {
   public static let never = ^UnionType([])
 
   /// Hylo's `Void` type.
-  public static let void = ^TupleType([])
+  public static let void = ^TupleType([] as [AnyType])
```
kyouko-taiga, line 68: This is precisely an instance of the kind of issue caused by the overloading of the constructor.

Sources/FrontEnd/Types/TupleType.swift
```diff
@@ -20,36 +20,36 @@ public struct TupleType: TypeProtocol {
...
+  /// Creates a tuple type having the given components.
+  public init<S: Sequence>(_ components: S) where S.Element == Component {
+    self.components = Array(components)
+    self.flags = ValueFlags(self.components.map(\.type.flags))
   }
 
-  /// Creates a tuple type with a sequence of label-type pairs.
-  public init<S: Sequence>(labelsAndTypes: S) where S.Element == (String?, AnyType) {
-    self.init(labelsAndTypes.map({ Element(label: $0.0, type: $0.1) }))
+  /// Creates a tuple type having the given components.
```
kyouko-taiga, line 34: I am concerned about this change. I my experience, Swift is not very good at inferring the type of collections in the presence of overloading (to be fair that is kind of a hard problem). It can be extremely helpful to use labels as a way to direct resolution.

I also think the label here is consistent with `Dictionary`'s constructors.

Sources/FrontEnd/Types/TupleType.swift
```diff
@@ -20,36 +20,36 @@ public struct TupleType: TypeProtocol {
...
+  public init<S: Sequence>(_ components: S) where S.Element == (String?, AnyType) {
+    self.init(components.map({ Component(label: $0.0, type: $0.1) }))
   }
 
-  /// Creates a tuple of unlabeled elements with the types in the given sequence.
-  public init<S: Sequence>(types: S) where S.Element == AnyType {
-    self.init(types.map({ Element(label: nil, type: $0) }))
+  /// Creates a tuple having the given unlabeled components.
+  public init<S: Sequence>(_ components: S) where S.Element == AnyType {
```
kyouko-taiga, line 40: The same concern applies on the removal of this label.

kyouko-taiga (changes requested):
Great set of changes overall but I am not convinced that overloading the constructor is the right move. I strongly believe that the argument labels can help and I don't think they add too much noise at call sites.

I could perhaps be favorable to remove the label on one of the constructors, but not both. Specifically, I think it could be fine to remove the label on the constructor for unlabeled tuples, since `labelsAndTypes` can sometimes provide context to read a lambda, e.g.:

```swift
TupleType(labelsAndTypes: xs.map({ ($0, f($0)) }))
```

Sources/FrontEnd/Types/AnyType.swift
dabrahams: Yes, well there really ought to be a no-argument constructor for this case.

Sources/FrontEnd/Types/TupleType.swift
dabrahams: This is following the convention that a value-preserving conversion is unlabeled.  Dictionary's constructor doesn't do that, since the order of the elements in the argument is lost.

Sources/FrontEnd/Types/TupleType.swift
dabrahams: The same rationale applies.  It's jarring at the use site to see an argument label in these cases.  One wonders why one has to write the label and as a reader, what else the constructor might be doing.

dabrahams:
@kyouko-taiga  I understand your concern but it doesn't warrant breaking with the naming convention.

Sources/FrontEnd/Types/AnyType.swift
kyouko-taiga: I really don't like the fact that we have to add the ascription here. I believe this is an issue with this PR.

kyouko-taiga:
I disagree. I think it this change is making it more difficult for humans too to understand what inference is doing in cases where the type of the constructor parameter forms the expected type of the argument's expression. If we can't live with argument labels then we should use named static functions.

I prefer breaking with Swift's convention, though. I believe my concern deserves an exception to the rule in this case.

Maybe a compromise could be to add labels on the constructor that expects the labels. `.init(labelsWithComponents: expr)` is an informative call site, especially because in many cases it will not be obvious what structure `expr` is computing. The reason I labeled the other constructor is precisely because "converting" `[(a, T), (nil, U)]` to a tuple is "value-preserving" whereas converting `[T, U]` is less straightforward.

Sources/FrontEnd/Types/TupleType.swift
kyouko-taiga: I don't think the convention is that strict, especially in the presence of overloads. `String` has a dozen of constructors with labels that are value-preserving conversions.

Sources/FrontEnd/Types/TupleType.swift
kyouko-taiga: A trivial (value-preserving) constructor should expect labels. I do not think it is obvious that one can give a list of types to construct a value that is a list of labels and types.

dabrahams:
> I think it this change is making it more difficult for humans too to understand what inference is doing in cases where the type of the constructor parameter forms the expected type of the argument's expression.

Sorry, I can't quite understand which cases you mean.  Could you show an example?

> Maybe a compromise could be to add labels on the constructor that expects the labels. .init(labelsWithComponents: expr) is an informative call site, especially because in many cases it will not be obvious what structure expr is computing. 

Do we actually have such cases?  I tend to doubt that ends up being the case in practice.  Part of the principle behind Swift's naming conventions is that by the time you get to a function call you tend to know the meanings of the things you pass in to functions.  I say _tend to_ on purpose: it's not always, but it's very much most ways, so we make the tradeoff in favor of the readability of most call sites.  Remember that the author of the call site also has a role in making their code clear. 

> The reason I labeled the other constructor is precisely because "converting" [(a, T), (nil, U)] to a tuple is "value-preserving" whereas converting [T, U] is less straightforward.

They're both value-preserving according to the intended meaning of the API guideline, for what that's worth.  The second one is a widening conversion like `UInt8` -> `UInt`, which is always value-preserving.

Sources/FrontEnd/Types/TupleType.swift
dabrahams: I don't know what it could mean for a  constructor to expect labels.

Sources/FrontEnd/Types/TupleType.swift
dabrahams: I am not sure that's true.  At least, we thought very carefully about `String` conversions in particular when establishing this convention, so maybe you should point out the examples you have in mind.

Now, if we reached a point where we needed multiple different value-preserving conversions from some type `X` to `String`, we'd of course have been forced to break with the convention, but that's not something we just did because we felt like it.

dabrahams:
This is sitting in my languishing PRs bin.  @kyouko-taiga I'm certainly not willing to die on this hill but maybe worth a few minutes discussion, because you haven't addressed the counterarguments here

kyouko-taiga:
I will not die on this hill either but I feel strongly about it. Let me explain my rationale and if you still want to push for this change I won't resist.

IIUC your point is that a value-preserving constructor should not, by convention, have any label.

While I think this convention is great, I also think we should use it more sparingly than you often suggest, especially in situations where (a) the conversion is not necessarily obvious and/or (b) overloading is worsening user experience and/or (c) a label can help writing clearer call sites. The following is expanding on these points:

Regarding (a) I said:

> A trivial (value-preserving) constructor should expect labels.

I meant that the notional value of a tuple is a sequence of optionally labeled types. The fact that a label is or isn't there is significant and therefore I believe that a value-preserving constructor should take (i.e., expect) this information.

I think I understand that you consider the absence of labels to be indeed provided with just the sequence, if one interprets each element as a pair whose first component is simply always `nil`. But if that is the case I think the encoding is too clever to satisfy what I would subjectively consider to be a *trivial* value-preserving constructor.

Regarding (b) I think the cohabitation of two unlabeled constructors is making the API harder to use, which is why I said:

> I think it this change is making it more difficult for humans too to understand what inference is doing in cases where the type of the constructor parameter forms the expected type of the argument's expression.

To give an example, consider this:

```swift
struct S {
  init<T: Sequence<Int>>(_ xs: T) {}
  init<T: Sequence<(String?, Int)>>(_ xs: T) {}
}

let xs = ["a": 1, "b": 2] 
let ys = S(xs)
```

Swift reports that there are "no exact matches in call to initializer". That's because inference is not able to convert `(key: String, value: Int)` to `(String?, Int)`. Unfortunately this issue is hidden in the details of the diagnostic. In my experience, Swift can either overcome the problem entirely if inference can single out a candidate in an overload set before associated type constraints are resolved, or at least it can produce a better error message. As a consequence, I tend to avoid overloading constructors with different constraints on associated types.

Another instance of inference-related issues is in [this very PR](https://github.com/hylo-lang/hylo/pull/1727#discussion_r2335835657). I don't think adding a type ascription is an improvement and the error message we get if we don't is not very good:

```swift
public static let void = ^TupleType([] as [AnyType])
```

Regarding (c), even if inference was better, I would still argue that a constructor label is deserved. I agree when you say this:

> the author of the call site also has a role in making their code clear.

I just think that is simpler to do when there is a label explaining what the argument's expression means. For example, in an expression like this one:

```swift
let t = TupleType(labelsAndTypes: xs)
```

I think the label is acting as a comment explaining *how* we're using `xs`, which IMO is very valuable in situations where `xs` is also used to construct something else, like a parameter list.

dabrahams:
Ok, those are good arguments! I'll see if anything can still be salvaged in this PR and proceed accordingly.

tothambrus11:
@dabrahams Should we close this or are there parts of this that can be salvaged?

PR #1729, opened 2025-09-14
Proof-of-concept for the target of IR refactoring

lucteo:
The current focus is on `IR/Analysis/Module+XXX` files, except NormalizeObjectStates (requires more refactoring) and Deploymorphize (which has more complex changes, that can't be easily assessed by a diff).

There are still some fixes that need to be applied for production code: documentation, access. But the end result passes all the tests.

**These modules show minor usage changes between the old and the target state.**

Notes:
- in a few cases we need, a `TypedProgram`
- in a few cases we need a `Module` to create `FunctionReference`
- in a few places we require an extra block parameter -- this is due to the change that tries to align instructions with Dimi's new frontend.


Sources/IR/Analysis/Module+Depolymorphize.swift
```diff
@@ -119,76 +125,69 @@ extension IR.Program {
...
     let source = module(defining: f)
-    var rewrittenBlock: [Block.ID: Block.ID] = [:]
-    for b in modules[source]![f].blocks.addresses {
-      let s = Block.ID(f, b)
-      let inputs = modules[source]![s].inputs.map { (t) in
-        monomorphize(t, for: z, usedIn: scopeOfUse)
-      }
-      let t = modules[target]![result].appendBlock(in: modules[source]![s].scope, taking: inputs)
-      rewrittenBlock[s] = Block.ID(result, t)
+    var observer = MonomorphizationObserver()
```
kyouko-taiga, line 138: A little subjective but I don't like how that reads. What's this observer observing? I don't have any idea of what this object is supposed to do from just that name. The method calls below are not helping either.

Sources/IR/Analysis/Module+Depolymorphize.swift
```diff
@@ -119,76 +125,69 @@ extension IR.Program {
...
+    var ff = modules[source]![f]
+    ff.observe(ff.instructionIDs, with: &observer)
+
+    // Monomorphize the types.
+    var transformedTypes: [AnyType: AnyType] = [:]
+    for t in observer.types {
+      transformedTypes[t] = monomorphize(t, for: z, usedIn: scopeOfUse)
```
kyouko-taiga, line 145: I understand what's being achieved here and I remember the argument you have made in last week's meeting but I do not believe this approach (i.e., pre-computing the types that have to change to require less context during monomorphization) makes the code simpler. Quite the opposite in fact. I believe it is simpler to understand what is going on when instructions are visited in one pass.

I also don't think that this strategy scales very well. If we later need to refactor monomorphization so that ti does anything slightly cleverer (and that is a very likely proposition) then we'll be stuck with a design that demands we are able to predict in step one all the types that we'll need in step 2.

Finally I think this strategy is inefficient because it requires additional passes over the IR. If we were to generalize this approach to all phases that may compute information requiring access to more than the function being analyzed (especially the ones computing fixed points) I wouldn't be surprised if we started to notice slowdowns.

If we really wanted to make sure the rest of the program cannot change as we're going through the instructions then I would prefer to store the new constructed types in some sort of cache that we merge once we're done. To be clear I'm not advocating for this solution either but I would be less opposed to it.

kyouko-taiga (commented):
It's much, much better than the current live PR but I'm still not seeing the big improvement. Could you direct me to a particular piece of code that you think is being improved?

It seems to me that a lot of the proposed change in the monomorphizer revolve around the two-steps approach that pre-compute monomorphized types to avoid touching the rest of the program mutably during IR generation. But I have reservations about this strategy, as I commented on the code.

lucteo:
> It's much, much better than the current live PR but I'm still not seeing the big improvement. Could you direct me to a particular piece of code that you think is being improved?
> 
> It seems to me that a lot of the proposed change in the monomorphizer revolve around the two-steps approach that pre-compute monomorphized types to avoid touching the rest of the program mutably during IR generation. But I have reservations about this strategy, as I commented on the code.

Please see  IR/Analysis/Module+XXX files, except NormalizeObjectStates (requires more refactoring) and Deploymorphize.

The goal of this version of POC is to show that user experience **does not need to be compromised** if we move this direction. It doesn't show an improvement in user experience, that was not my primary objective. For these examples, the improvements would be hidden -- we don't need to always go from the module to the function in which we are operating.

(I may able to show some improvements, now that I'm getting back to the projection patches).

Sources/IR/Analysis/Module+Depolymorphize.swift
lucteo: The purpose of this type can be summaried as: "Observer that collects the types and function references that need monomorphization"

I don't have a better naming alternative, but I'm happy to change if we can find one.

Sources/IR/Analysis/Module+Depolymorphize.swift
lucteo: The main problem I was trying to solve here is having access to the entire program while reading and writing it. Separating it into two phases (one that just reads, another one that writes) seemed like a reasonable approach.

I agree that your proposal would probably be more efficient. I can change this if we agree that the entire effort is worthwhile.

Sources/IR/Analysis/Module+Depolymorphize.swift
dabrahams: An observer, IIUC, is fundamentally a reference-semantic idea: it's a thing that gets notified when other things change.

PR #1735, opened 2025-10-05
Stop using obsolete compiler option in a test

dabrahams:
I don't know how this ever passed, but with XCode 26.0 we are getting failures.


dabrahams:
Bogus

PR #1736, opened 2025-10-06
Fix CMake testing issues

dabrahams:
1. include(CTest) will set BUILD_TESTING=1 if it isn't already set, so only include CTest conditionally.

2. Update instructions to account for missing CMAKE_OSX_SYSROOT


dabrahams:
Stale

PR #1738, opened 2025-10-07
Remove `Block.AbsoluteID`

lucteo:
Remove `Block.AbsoluteID`

Replace its use in Emitter with a simpler structure.


Sources/IR/Emitter.swift
```diff
@@ -68,6 +68,17 @@ struct Emitter: Sendable {
   /// The source code associated with instructions to be inserted.
   var currentSource: SourceRange
 
+  /// The block's ID across functions.
+  public struct BlockInFunction: Hashable {
```
kyouko-taiga: Is this type supposed to stay or is it there for the transition?
Regardless, I'd like to suggest another name:

```suggestion
  public struct BlockAndFunction: Hashable {
```

FWIW I strongly believe there should be no need to identify basic blocks outside the context of a specific function and so the concept of pairing a basic block with its parent function should likely disappear eventually.

lucteo: Agree, but I didn't want to change this in the scope of this refactoring. I would hope to get it changed with future work around Emitter.

kyouko-taiga: Fine with me.

kyouko-taiga (approved):
LGTM

PR #1741, opened 2025-10-16
Testing conveniences

dabrahams:


Sources/TestUtils/ProgramInstances.swift
```diff
@@ -0,0 +1,77 @@
...
+  /// Returns self parsed.
+  public func parsed(reportingDiagnosticsTo log: inout DiagnosticSet) throws -> AST {
+    var r = AST()
+    _ = try r.loadModule("Main", parsing: [self], reportingDiagnosticsTo: &log)
+    return r
+  }
+
+  /// Returns `self` as a one-file module, lowered to IR.
+  ///
+  /// - Parameter `needsBuiltins`: `true` iff `self` should be allowed access to
```
tothambrus11:
```suggestion
  /// - Parameter `needsBuiltins`: whether `self` should be allowed access to
```

Sources/TestUtils/ProgramInstances.swift
```diff
@@ -0,0 +1,77 @@
...
+        try ast.loadModule("Main", parsing: [self], withBuiltinModuleAccess: needsBuiltins, reportingDiagnosticsTo: &log)
+      }.0
+  }
+
+}
+
+extension AST {
+
+  /// The latest module loaded.
+  public var latestModule: ModuleDecl.ID { modules.last! }
```
tothambrus11: This may result in some hard to diagnose errors if the internals of the module representations change. Would it be worth expressing the relationship between the module that we loaded during parsing and the usage sites of this property? E.g. `.parsed` could return a tuple that has both the AST and the parsed module's id. 

Or we could have `parsing` always parse into "ModuleUnderTest" module, which could be accessible by a computed property `.testedModule` that looks up the module by name.

I don't want to make the testing code look much more verbose, so if this doesn't look like a concern worth addressing, I like this version for brevity.

Sources/TestUtils/ProgramInstances.swift
dabrahams: Yeah… I know the Swift documentation people use “whether.”  It always rubbed me the wrong way.  Both these variables tell you _whether_ the dog is red: `dogIsRed`,  `dogIsNotRed`, but they do it in different ways.  I might have to give up my resistance, because it's common and other people don't seem to have trouble with it.

Sources/TestUtils/ProgramInstances.swift
dabrahams: You make a good point.  The code made me uneasy for exactly that reason, and you have some good ideas, which I might combine.  Let me take another shot at it.

PR #1750, opened 2025-10-26
Interpreter: Make empty program run successfully

RishabhRD:
**Formatting changes are from swift-formatter.**

The PR does the following changes:
1. Make program with empty main method run successfully.
2. Fixes stack.pop implementation.
3. add convenient `executor(for: )` method in InterpreterTests to create interpreter for raw programs.
4. Replace print program with empty program in test, will add back print when function calls will be supported.
5. Used swift-testing for interpreter tests for running program.

~Was experimenting a little with swift-testing. It worked fine for all purposes except, it seems to not produce any binary for debugging purposes, hence stayed with XCTest.~

### DIFF
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -326,8 +326,8 @@ public struct Interpreter {
       a)
   }
 
-  /// Removes topmost stackframe and points `programCounter` to next instruction
-  /// of any previous stackframe, or stops the program if the stack is now empty.
+  /// Removes topmost stack frame and points `programCounter` to next instruction
+  /// of any previous stack frame, or stops the program if the stack is now empty.
   ///
   /// - Precondition: the program is running.
   mutating func popStackFrame() {
--- a/Tests/InterpreterTests/InterpreterTests.swift
+++ b/Tests/InterpreterTests/InterpreterTests.swift
@@ -16,8 +16,8 @@ import Utils
     let module = try input.loweredToIRAsMainWithHostedStandardLibrary();
     let program = IR.Program.init(syntax: module.program, modules: [module.id: module]);
     var executor = Interpreter(program);
-    #expect(throws: Never.self) {
-      while executor.isRunning { try executor.step() }
+    while executor.isRunning {
+      try executor.step()
     }
   }
 


RishabhRD:
UPDATE: Got the solution for swift-testing problem from this [thread](https://forums.swift.org/t/debugging-swift-testing-tests-on-linux/82867/2). Do we want to use swift-testing for new tests? If so I would convert InterpreterTests to swift-testing one.

dabrahams:
> UPDATE: Got the solution for swift-testing problem from this [thread](https://forums.swift.org/t/debugging-swift-testing-tests-on-linux/82867/2). Do we want to use swift-testing for new tests? If so I would convert InterpreterTests to swift-testing one.

I think we do want to ultimately use swift-testing unless the user experience is problematic. Make sure it works well in Xcode (eg you get test diamonds) and if so I think we proceed with that.

RishabhRD:
> Make sure it works well in Xcode (eg you get test diamonds) and if so I think we proceed with that.

Unfortunately, I don't have a mac machine, I develop on Linux. So, its not possible for me to check the same. Will need help of someone here to confirm the same.

kyouko-taiga:
I have started working on migrating the new frontend to Swift Testing. So far Xcode has been compliant and I find the user experience relatively good.

I highly doubt we would be able to painlessly migrate the testing architecture of the current implementation, though. The whole framework is based on a magic macro that I find very difficult to understand 😔 

The good news is that we XCTest and Swift Testing can live along side without any problem, or so it seems from my experience so far.

RishabhRD:
> I have started working on migrating the new frontend to Swift Testing. So far Xcode has been compliant and I find the user experience relatively good.

That's actually nice. Then, I can progress with swift-testing for the PR :smiley:. I have also observed on Linux that XCTest and Swift Testing can live along without any problem.

dabrahams:
It's probably OK for this PR which is small, but if we are going to swift-format the code, we should ensure that everybody has live reformat-on-save functionality in their environment.  We should not include formatting changes in PRs with other purposes; it makes them hard to review.

Sources/Interpreter/Interpreter.swift
```diff
@@ -89,15 +89,18 @@ struct StackAllocation {
...
     size = structure.bytes.size
     storage = .init(repeating: 0, count: max(0, size + structure.bytes.alignment - 1))
-    baseOffset = size == 0 ? 0 : storage.withUnsafeBytes {
-      let b = UInt(bitPattern: $0.baseAddress!)
-      return Int(b.rounded(upToNearestMultipleOf: UInt(structure.bytes.alignment)) - b)
-    }
+    baseOffset =
+      size == 0
+      ? 0
+      : storage.withUnsafeBytes {
```
dabrahams: IMO this formatting change makes the code a bit harder to read, and it was hard to verify that it was just formatting in the webpage (I had to explicitly “hide whitespace”, which at least hid part of the change).

Sources/Interpreter/Interpreter.swift
```diff
@@ -321,6 +329,17 @@ public struct Interpreter {
       a)
   }
 
+  /// Removes topmost stackframe and points `programCounter` to next instruction
+  /// of previous stackframe if present, otherwise marks program as non-running.
```
```suggestion
  /// Removes topmost stackframe and points `programCounter` to next instruction
  /// of any previous stackframe, or stops the program if the stack is now empty.
```

Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -1,30 +1,28 @@
...
-      var executor = Interpreter(ir)
-      while executor.isRunning { try executor.step() }
-      let output = executor.standardOutput
+    return Interpreter(compilation.ir!)
+  }
 
-      // Remember, Windows has a different newline character
-      XCTAssert(output.last?.isNewline ?? false, "Expected a final newline")
-      XCTAssertEqual(output.dropLast(), "Hello, World!")
+  @Test func emptyProgramShouldWork() throws {
```
dabrahams:
```suggestion
  @Test func emptyMain() throws {
```

Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -1,30 +1,28 @@
...
+
+@testable import Interpreter
 
-final class InterpreterTests: XCTestCase {
+@Suite struct InterpreterRunTests {
 
-  func testInterpreter() throws {
-    let source = try FileManager.default.temporaryFile(
-      containing: #"public fun main() { print("Hello, World!") }"#)
+  func executor(for program: String) throws -> Interpreter {
```
dabrahams: Instead, you should be using the facilities of ProgramInstances.swift

Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -1,30 +1,28 @@
...
-      while executor.isRunning { try executor.step() }
-      let output = executor.standardOutput
+    return Interpreter(compilation.ir!)
+  }
 
-      // Remember, Windows has a different newline character
-      XCTAssert(output.last?.isNewline ?? false, "Expected a final newline")
-      XCTAssertEqual(output.dropLast(), "Hello, World!")
+  @Test func emptyProgramShouldWork() throws {
+    var executor = try executor(for: #"public fun main() {  }"#)
```
dabrahams: Use [this function](https://github.com/hylo-lang/hylo/blob/8d9cc1f9872f061ff71a25a925c4c93bf7ead7a7/Sources/FrontEnd/SourceFile.swift#L175-L195) along with ProgramInstances.swift

Sources/Interpreter/Interpreter.swift
```diff
@@ -321,6 +326,17 @@ public struct Interpreter {
       a)
   }
 
+  /// Removes topmost stackframe and points `programCounter` to next instruction
```
dabrahams: Stack frame is two words

Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -1,30 +1,24 @@
...
-      var executor = Interpreter(ir)
+  @Test func emptyMain() throws {
+    let input =
+      """
+        public fun main() { }
+      """.asSourceFile()
+    let module = try input.loweredToIRAsMainWithHostedStandardLibrary();
+    let program = IR.Program.init(syntax: module.program, modules: [module.id: module]);
+    var executor = Interpreter(program);
+    #expect(throws: Never.self) {
```
dabrahams: I don't think it's necessary to expect that something never throws, explicitly. You can just make the test a throwing function, and the framework will fail it if an exception propagates out.

Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -1,30 +1,24 @@
...
       while executor.isRunning { try executor.step() }
```
dabrahams: This is a statement where it might be very useful to be able to set a breakpoint inside the loop, but if you put it all on one line that's impossible

Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -1,30 +1,24 @@
...
 
-    func runAndCheckOutput(_ ir: IR.Program) throws {
-      var executor = Interpreter(ir)
+  @Test func emptyMain() throws {
+    let input =
+      """
+        public fun main() { }
+      """.asSourceFile()
+    let module = try input.loweredToIRAsMainWithHostedStandardLibrary();
+    let program = IR.Program.init(syntax: module.program, modules: [module.id: module]);
```
dabrahams, line 17: Theres a function in ProgramInstances.swift that lowers to IR, applying the mandatory passes. You should use it!

RishabhRD: `loweredToIRAsMainWithHostedStandardLibrary` also calls `lowerToIR(reportingDianosticsTo:)` method that applies mandatory passes.
Are you implying that I should not use withHostStandardLibrary thing?

dabrahams: Nope; I just misread the code; you're good!

dabrahams (commented):
Merge it!

PR #1752, opened 2025-10-28
Interpreter: Make the function call and other basic instructions work

RishabhRD:
**This PR also have changes of my previous PR because those are not merged yet, but I need those for testing.**

## Instructions implemented:
1. Access
2. AllocStack (fix)
3. Branch
4. Call
5. CondBranch
6. DeallocStack
7. Load
8. MemoryCopy
9. Store
10. SubfieldView
11. Load

## Instructions UN-implemented
1. ConstantString

## Design Changes
- Introduced parameters for functions. Parameters can't be part of stack allocations as they always store addresses.
- Introduced instruction result type for type safety.
- Initial support for parsing builtin objects.
- Unify representation of stack and heap memory.

#### Motivation
For supporting pointer operations, we need mechanism to convert a BuiltIn pointer type to Address representation in VM.
However, machine pointer width can be even 32-bits. Conversion of pointer to address that has UUID and things is not possible without huge extra storage. Also discriminator bit would used if there are 2 kind of addresses.

#### Design
During initialization of interpreter, stack is allocated that is nothing but Memory.Allocation. All the local variables live in that allocation and registers, returnAddress, parameters live in stack frames.

## Fixes
- Stack allocation stack frame should return base offset of allocation.
- Removing allocation should also remove allocation id from dictionary.

### DIFF
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -331,8 +331,8 @@ public struct Interpreter {
     return codePointer;
   }
 
-  /// Removes topmost stackframe and points `programCounter` to next instruction
-  /// of any previous stackframe, or stops the program if the stack is now empty.
+  /// Removes topmost stack frame and points `programCounter` to next instruction
+  /// of any previous stack frame, or stops the program if the stack is now empty.
   ///
   /// - Precondition: the program is running.
   mutating func popStackFrame() {


RishabhRD:
UPDATE:
- store instruction works now, that means conditional branch instruction now executes and jumps to correct branches.
- fix type layout calculation for bool in unreal abi

dabrahams:
If you base this PR on the branch of your other PR, the changes of the other PR won't show up here, and when we merge that PR if we delete the branch in this repo, this PR will automatically be relative to main

RishabhRD:
@dabrahams I couldn't base the PR correctly as I raised the PR using the fork I created. I currently don't have access to push branches in this repo, need that access for the same.

RishabhRD:
UPDATE: Implemented unified memory addressing, that was necessary for supporting ConstantString and pointer arithmetic.

Because, its already good change in current design, will wait for some initial feedback before implementing more instructions.

I am planning to implement other instructions in subsequent PRs as this PR already is covering multiple grounds.

RishabhRD:
Closing this PR as it was opened from my fork and would again block me opening incremental PRs.

Raised #1753 from local branch for the same purposes.

PR #1753, opened 2025-10-30
Interpreter: Make function calls and basic instructions working

RishabhRD:
## Design Changes
- Introduced notion of function parameters as it was part of IR.
- Unify the memory representation of stack and heap objects.
- Instructions results in `InstructionResult` type instead of `Any` upon execution.

### Unified Memory Representation
- Pointer instructions require conversion from pointer to address and vice-versa.
- For enabling this mechanism, there should be some way to encode a memory address as pointer.
- With UUID used in stack addressing, this mechanism is not possible without overhead of huge storage as UUID itself is 128 bits and pointer on any machine can be maximum 64 bits. And for supporting any generic target ABI, this is not the way.
- Also different address for stack and heap objects would waste a discriminator bit too.
- Hence, we need to unify the way to address stack and heap objects.

#### Design
- Stack is also a memory allocation now that just stores local variables.
- Stack now has fixed memory. Stack gets created on initialization of interpreter.
- StackFrame is now just responsible for registers, return address and parameters. There are classic techiniques which can also make returnAddress and parameters part of stack. I am not sure if we want that for now.
- Because stack is also a memory allocation now, addresses for heap objects and stack objects are of same type and will be useful for implementing pointer operations.

## Instructions implemented:
1. Access
2. AllocStack
3. Branch
4. Call
5. CondBranch
6. DeallocStack
7. Load
8. MemoryCopy
9. Store [Partial support for only integer types]
10. SubfieldView
11. Load

## Instructions UN-implemented
1. ConstantString

### DIFF
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -215,7 +215,7 @@ public struct Interpreter {
       store(toBuiltinValue(x.object)!, at: toAddress(x.target)!)
     case let x as SubfieldView:
       let parent = toAddress(x.recordAddress)!;
-      currentRegister = address(of: x.subfield, at: parent)
+      currentRegister = address(of: x.subfield, in: parent)
     case let x as Switch:
       _ = x
     case let x as UnionDiscriminator:
@@ -330,7 +330,7 @@ public struct Interpreter {
   }
 
   /// Returns the address of `subField` in the object  at `origin`.
-  mutating func address(of subField: RecordPath, at origin: Address) -> Address
+  mutating func address(of subField: RecordPath, in origin: Address) -> Address
   {
     let memoryAddress = origin.memoryAddress;
     var offset = origin.memoryAddress.offset;


Sources/Interpreter/TargetABI.swift
```diff
@@ -20,9 +20,10 @@ public struct UnrealABI: TargetABI {
     precondition(
       bitWidth > 0 && bitWidth.nonzeroBitCount == 1,
       "bit width \(bitWidth) is not a power of 2.")
+    let sizeInBytes = (bitWidth + 7) / 8
     return .init(
-      alignment: min(bitWidth / 8, maxAlignment),
```
dabrahams, line 24: Good catch!

Those these are equivalent if it width can be no smaller than eight. Should we have a precondition instead?  I think `Builtin.i1` might be a thing, so maybe not.

Sources/Interpreter/Interpreter.swift
```diff
@@ -10,46 +10,53 @@ struct CodePointer {
...
+  /// The storage in little-endian format.
+  public let storage: UInt128
 
-  /// The program counter to which execution should return when
-  /// popping this frame.
-  var returnAddress: CodePointer
+  /// Number of bytes, storage is representing.
+  public let bytes: Int
+
+}
```
dabrahams: Never declare a type without a doc comment that says what it _is_ (see Swift API Guidelines).  I can't evaluate whether this type's definition is correct without knowing what it's supposed to be.  Just guessing here:

```suggestion
/// The [value representation](https://eel.is/c++draft/basic.types#general-4) of a `Builtin` type instance.
struct BuiltinValueRepresentation {

  /// The bytes of the object in little-endian format.
  public let bytes: UInt128
  
  /// The size of the value in bytes.
  public let size: Int
}
```

- Could have used `BuiltinObjectRepresentation` because they are equivalent for `Builtin` types since they have no padding, but `Value` is shorter.
- `storage` is commonly used this way, where some instance _holds_ the value of a type's instances, but doesn't by itself represent the value.  However, it also strongly connotes memory, and at the meta-level, in the interpreter, these are used to represent values separated from memory.
- The name bytes means “these are the bytes;” you'd have to say `byteCount`, but `size` is a standard way of measuring the number of bytes of a type.

I question whether we really want to store the bytes little-endian, which means that the first memory address _of a stored instance_ is in the low byte of `bytes`; that will just make these things hard to use and interpret, because it will make the meaning of an instance dependent on the target architecture.  I'm guessing what we want is more like:

```suggestion
/// The value of a `Builtin` type instance, stripped of type information.
struct UntypedBuiltinValue {

  /// The result of reinterpretation as an unsigned integer value.
  ///
  /// Equivalent to `UInt128(unsafeBitCast<T>(x))`, where `x` is the represented value
  /// and `T` is an unsigned integer of the same size as `x`.  Stores the unsigned
  /// representation of integer types.
  public let asUInt128: UInt128
  
  /// The size of the value in bytes.
  public let size: Int
}
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -10,46 +10,53 @@ struct CodePointer {
...
+
+struct ObjectAddress {
```
dabrahams: Should this be called “Pointer?”  I don't think “Object” is adding anything.

Sources/Interpreter/Interpreter.swift
```diff
@@ -10,46 +10,53 @@ struct CodePointer {
...
+  public let memoryLayout: TypeLayout
+
+}
+
+/// Result of an instruction.
+enum InstructionResult {
 
-  /// This frame's unique identifier throughout time.
-  let id = ID()
+  /// Address of object on stack.
```
dabrahams: Why "on stack?"  can't an instruction return an address that's on the heap?

Sources/Interpreter/Interpreter.swift
RishabhRD: There are instructions like AddressToPointer, thus I didn't pick the name Pointer. But yeah, "object" can be removed from name.

RishabhRD:
> I've only just started with the review, but I must say it's hard to evaluate these diffs, because entities like `StackFrame` do not show up in the same place.

Sorry! I accidentally moved stackframe down, fixed that. I should have observed that while raising PR itself.

RishabhRD:
I tried to resolve all the given comments. Got rid of little-endian thing and made it stack/heap independent.

Sources/Interpreter/Memory.swift
```diff
@@ -251,6 +251,16 @@ public struct Memory {
     throw Error.noDecomposable(t, at: a)
   }
 
+  /// Allocation at given address.
```
```suggestion
  /// The allocation at `i`.
```

"The given" is almost never a win in documentation comments

Do we really want to call an allocation ID an address?

```suggestion
  /// The allocation identified by `i`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -10,46 +10,74 @@ struct CodePointer {
 
 }
 
-/// The local variables, parameters, and return address for a function
-/// call.
+/// The parameters, and return address for a function call.
```
dabrahams: I'm confused by the removal of locals from this comment. The registers are the locals and they appear to still be there.

Sources/Interpreter/Interpreter.swift
```diff
@@ -77,84 +105,33 @@ extension UnsafeRawBufferPointer {
...
-}
-
 /// A virtual machine that executes Hylo's in-memory IR representation.
 public struct Interpreter {
 
   /// The program to be executed.
   private let program: IR.Program
 
-  /// The stack- and dynamically-allocated memory in use by the program.
+  /// The stack and dynamically-allocated memory in use by the program.
```
dabrahams: That was intentional. There's an idiom when you have two hyphenated things with the same suffix and you put the word and between them you leave the suffix off of the first one.

Sources/Interpreter/Interpreter.swift
```diff
@@ -77,84 +105,33 @@ extension UnsafeRawBufferPointer {
...
   private var memory = Memory()
 
-  /// Local variables, parameters, and return addresses.
-  private var stack = Stack()
+  /// function parameters, and return addresses.
+  private var stackFrames: [StackFrame] = []
+
+  /// Local variables.
+  private var stack: Memory.Allocation {
```
dabrahams: I don't understand why the memory of a frame is separated from the stack frame. Calling something “stack” that doesn't include the stack frames doesn't make sense to me.

Sources/Interpreter/Interpreter.swift
```diff
@@ -235,11 +218,17 @@ public struct Interpreter {
     case let x as CloseUnion:
       _ = x
     case let x as CondBranch:
-      _ = x
+      let cond = builtIn(denotedBy: x.condition)!
+      if cond.asUInt128 != 0 {
```
dabrahams: There should be accessors for different basic types on that type so for example, we can get the Bool value. And it probably should have a precondition that the value is zero or one. Almost nothing should use asUInt128 directly

dabrahams:
Why is CI failing?  This can't be ready for review until CI can build it…

With regard to this:

> With UUID used in stack addressing, this mechanism is not possible without overhead of huge storage as UUID itself is 128 bits and pointer on any machine can be maximum 64 bits. And for supporting any generic target ABI, this is not the way.

I had not intended to start out by thinking too much about performance, nor to support different ABIs.  Of course we will need the latter eventually for compile-time computation, so the resulting values have a memory layout suitable for use at runtime.  But assuming we want to solve that problem now, I'm confused about the fact that `Memory.Address` is still 128 bits.  How to solve that problem well is an open question.  We want a unique ID for each allocation throughout time, to detect uses of invalid addresses.

dabrahams (commented):
Just exposing pending comments; more to come.

Sources/Interpreter/Interpreter.swift
dabrahams: I intentionally did not want to use a single `Memory.Allocation` for the whole stack.  That would prevent us from detecting the use of a pointer derived from the address of one local variable to access the memory of another local variable.

Sources/Interpreter/Interpreter.swift
```diff
@@ -168,17 +145,17 @@ public struct Interpreter {
...
-      yield &stack.frames[stack.frames.count - 1]
+      yield &stackFrames[stackFrames.count - 1]
     }
   }
 
   /// An instance executing `p`.
   ///
   /// - Precondition: `p.entry != nil`
-  public init(_ p: IR.Program) {
+  public init(_ p: IR.Program, withStackMemoryOfBytes bytes: Int) {
```
dabrahams: I don't see any good reason that an interpreter should be limited to a fixed amount of stack memory.

Sources/Interpreter/Interpreter.swift
```diff
@@ -275,9 +265,23 @@ public struct Interpreter {
       popStackFrame()
       return
     case let x as Store:
-      _ = x
+      let obj = builtIn(denotedBy: x.object)!
+      let dest = stripAlignment(address(denotedBy: x.target)!)
+      let destAllocIdx = dest.memoryAddress.allocation
+      let val = obj.asUInt128
+      memory[destAllocIdx].withMutableUnsafeStorage(dest.memoryAddress.offset) {
```
dabrahams: We have methods in Memory or Memory.Allocation for doing this don't we?

RishabhRD:
Let me address design changes so that we can be on same page.

> I had not intended to start out by thinking too much about performance, nor to support different ABIs. Of course we will need the latter eventually for compile-time computation, so the resulting values have a memory layout suitable for use at runtime. But assuming we want to solve that problem now, I'm confused about the fact that `Memory.Address` is still 128 bits. How to solve that problem well is an open question. We want a unique ID for each allocation throughout time, to detect uses of invalid addresses.

Currently I am just trying to support UnrealABI. Now regarding `Memory.Address` is still 128 bits, the idea is to use [relocations technique](https://files.solson.me/miri-report.pdf) \[section 4.1\] used by Miri interpreter, that is, store offset inline an every allocation maintain a relocation dictionary that points offset in current allocation to allocation id.
Also `UntypedBuiltinValue` would be an enum with scalar and pointer discriminators. This would allow storing pointers into the memory and reading pointers from memory.

There is still problem with supporting ConstantString instruction as it doesn't create pointer but creates Int64. And with previous strategy, supporting that might not be possible.

Also now I am thinking about it again, I realized UUID was not the problem, I overthought that.

Now coming to stack representation, the confusion came from:
```swift
  /// The stack- and dynamically-allocated memory in use by the program.
  private var memory = Memory()
```
that suggested stack allocation is also part of memory, thus I removed explicit StackAllocation. And to make, local variables contiguous in memory, I used single allocation for local variable. But yes, it would be bad for detecting bad pointer access.

There are 2 options, which one do you think is suitable, will progress with that:
1. Use separate stack allocation as previous i.e., memory would not contain stack allocation.
2. Use memory only for stack allocation and in StackFrame, add allocation id. Stack struct would manage stackframes.

UPDATE: currently CI is failing because cmake builds can't recognize swift-testing framework.

Sources/Interpreter/Interpreter.swift
RishabhRD: If this is about withMutableUnsafeStorage, then that method is coming from Memory.Allocation only. If this is about accessing integer value, it is not exactly the same. It is `unsignedIntValue` method most probably that returns UInt16.
Maybe we want to provide accessors in Memory.Allocation for the same?

Sources/Interpreter/Interpreter.swift
RishabhRD: Okay! removed that for now. However, that means we are not going to have things like stackoverflow then?

Sources/Interpreter/Interpreter.swift
RishabhRD: Totally agree. Made these fields private and using exposed accessors and initializers wherever needed.

Sources/Interpreter/Interpreter.swift
RishabhRD: Yes, detecting pointer use is really useful. So, I am using different Memory.Allocation for different stack variables.

RishabhRD:
@dabrahams I tried to resolve all the comments you raised. Regarding failing CI jobs:
1. swift-testing is not working well with cmake as I highlighted on slack. For the same, do you think I should migrate all interpreter tests to XCTest?
2. I am seeing one SPM job failing: [https://github.com/hylo-lang/hylo/actions/runs/19012119266/job/54294613434?pr=1753](https://github.com/hylo-lang/hylo/actions/runs/19012119266/job/54294613434?pr=1753). I am currently not sure why as I don't see any error in logs, instead I see signal 10 being raised. I am still looking, what maybe the cause.

dabrahams:
1. Yes if swift-testing can't be made to work with CMake, let's switch back to XCTest.  Is it “merely” a matter of some missing CMake code akin to [FindSwiftXCTest.cmake](https://github.com/hylo-lang/CMakeModules/blob/main/FindSwiftXCTest.cmake)?
2. I saw problems like this when working on [gyb-swift](https://github.com/hylo-lang/gyb-swift/), but in my case it was an infijnite recursion. I should try it on my mac to see if I can reproduce.

RishabhRD:
Migrating to xctest fixes all the CI build issues. It seems like for point 2, the CI exits before executing swift-testing tests.

Sources/Interpreter/Interpreter.swift
dabrahams: Just FWIW, I always intended that we could eventually arrange not to do a heap allocation for each of these without changing the high-level structure of the code.

Sources/Interpreter/Interpreter.swift
dabrahams: I think eventually modeling stack overflow could be interesting, but certainly not for many purposes like compile-time computation.  Even most “safe” languages don't actually have guarantees that they will avoid UB in those cases, and exactly how to make it work for real targets is an open question, so I don't think we should encode that constraint today.

Sources/Interpreter/Interpreter.swift
dabrahams: It's about accessing the integer value.  I don't understand what's different about it, other than that you might want to construct a `UInt16` from the result.  I don't know what you're proposing re: accessors, but to the degree possible we should encapsulate unsafe operations inside of `Memory.Allocation`.

dabrahams:
Regarding Miri's Relocations, I'm still working to understand them.  We should always have a discussion before starting down a path like this one.  Looking at the paper,

> In effect, the abstract machine represents pointers as (allocation_id, offset) pairs

We are doing that.  We don't have a relocation table.  Details in the paper of how that is manifested are sparse, but reading between the lines…

- I am guessing it is used to avoid making pointers 2x the integer size.  
- Given that it says

    >  The relocation table for y is {0 => x}

I think this means that every allocation (like the one for y) has a relocation table, which maps offsets of pointers stored in that allocation and represented as offsets into some allocation z (such as the one for x), to z.

This is a reasonable way to handle pointers in memory.  It does mean that pointers in registers need a different representation, but those could be direct (allocation, offset) pairs, AKA an `Address`.  Having a relocation table per allocation makes it easy to free up table space when an allocation is released.

I don't think I understand the issue with `ConstantString` yet.

> that suggested stack allocation is also part of memory, thus I removed explicit StackAllocation.

Yes, that was my intention, since `Memory` can serve both purposes; we don't need separate `Memory` for the stack and the heap; one object will do.  Stack frames contain the identities of allocations that need to be freed when the frame is popped.

The test failure appears to be a stack overflow!  See attached trace
[stacktrace.txt](https://github.com/user-attachments/files/23373478/stacktrace.txt).  I would not be shocked if this went away with the use of `XCTest`.

dabrahams:
Yep, the stack overflow is fixed by using XCTest.  I wonder if XCTest launches processes and swift-testing launches threads?  Threads have more limited stack space.  We need to get off swift-testing ASAP.

Sources/Interpreter/Interpreter.swift
```diff
@@ -13,43 +13,57 @@ struct CodePointer {
...
   /// The program counter to which execution should return when
   /// popping this frame.
   var returnAddress: CodePointer
 
-  /// This frame's unique identifier throughout time.
-  let id = ID()
+}
+
+/// Address to object with type layout.
+struct Address {
```
dabrahams, line 34: This is not just an address, because it's typed.  I think `Pointer` is more appropriate.  It could be `Interpeter.Pointer` if necessary.

Sources/Interpreter/Interpreter.swift
```diff
@@ -13,43 +13,57 @@ struct CodePointer {
...
+  public let memoryLayout: TypeLayout
+
+}
 
-  mutating func allocate(_ t: TypeLayout) -> Stack.Address {
-    let a = StackAllocation(t)
-    allocationIDToIndex[a.id] = allocations.count
-    allocations.append(a)
-    return Stack.Address(memoryLayout: t, frame: id, allocation: a.id, byteOffset: 0)
+/// Result of an instruction.
```
dabrahams: Try to say something that isn't 100% implied by its name.  Not always possible, but try.

Sources/Interpreter/Interpreter.swift
```diff
@@ -13,43 +13,57 @@ struct CodePointer {
...
+enum InstructionResult {
```
dabrahams: Don't instruction results have a type?

Sources/Interpreter/Interpreter.swift
```diff
@@ -150,11 +100,8 @@ public struct Interpreter {
   /// The stack- and dynamically-allocated memory in use by the program.
   private var memory = Memory()
 
-  /// Local variables, parameters, and return addresses.
-  private var stack = Stack()
-
-  /// A mapping from the `id`s of stack frames to their index in the stack.
-  private var frameIDToIndex: [StackFrame.ID: Int] = [:]
+  /// function parameters, and return addresses.
```
```suggestion
  /// function parameters and return addresses.
```

dabrahams: I still think it contains local variables; it just doesn't contain their storage.  Am I mistaken?

Sources/Interpreter/Interpreter.swift
```diff
@@ -216,12 +162,16 @@ public struct Interpreter {
...
 
     case let x as Branch:
-      _ = x
+      jumpTo(block: x.target)
+      return
     case let x as Call:
-      _ = x
+      stackFrames.append(
+        StackFrame(parameters: blockParams(from: x), returnAddress: try nextCodePointer()))
+      jumpToEntry(of: (x.callee.constant as! FunctionReference).function)
```
```suggestion
      jumpTo((x.callee.constant as! FunctionReference).function)
```
It's the same as jumping to the function right?

Sources/Interpreter/Interpreter.swift
```diff
@@ -235,11 +185,17 @@ public struct Interpreter {
     case let x as CloseUnion:
       _ = x
     case let x as CondBranch:
-      _ = x
+      let cond = builtIn(denotedBy: x.condition)!
+      if cond.bool! {
+        jumpTo(block: x.targetIfTrue)
```
```suggestion
        jumpTo(x.targetIfTrue)
```

Sources/Interpreter/UntypedBuiltinValue.swift
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /// Bool value, if present.
+  public var bool: Bool? {
+    if asUInt128 != 0 && asUInt128 != 1 { return nil }
+    return asUInt128 != 0
+  }
+
+  /// UInt8 value, if present.
+  public var uint8: UInt8? { size == 1 ? UInt8(truncatingIfNeeded: asUInt128) : nil }
```
dabrahams:
```suggestion
  public var i8: UInt8? { size == 1 ? UInt8(truncatingIfNeeded: asUInt128) : nil }
```

Because that's what we call it in instructions

Sources/Interpreter/UntypedBuiltinValue.swift
```diff
@@ -0,0 +1,78 @@
+import IR
+
+/// The value of a `Builtin` type instance, stripped of type information.
+struct UntypedBuiltinValue {
```
dabrahams: And now I realize this doesn't represent builtin ptr types I think?  IIRC it was my intention that it would; we'd just encode it into the UInt128…

Sources/Interpreter/Interpreter.swift
```diff
@@ -314,29 +272,191 @@ public struct Interpreter {
...
+  }
+
+  /// Returns builtin object (if any) stored at `a`.
+  mutating func builtIn(at a: Address)
+    -> UntypedBuiltinValue?
+  {
+    if !a.memoryLayout.type.isBuiltin {
+      return nil
+    }
+    let allocIdx = a.memoryAddress.allocation
```
dabrahams: Where did this name come from?  We don't use abbrevs.  Also it's all type information at an even lower level of abstraction than `Allocation.ID` which is already too low.  What *role* is the value playing in this context?  That's what goes in the name.  If you can't do better than to reflect type information, use a single-character name like `a`.

Sources/Interpreter/Interpreter.swift
```diff
@@ -314,29 +272,191 @@ public struct Interpreter {
...
+      case let x as IntegerConstant:
+        return UntypedBuiltinValue(withIntegerConstant: x)
+      default:
+        fatalError("unimplemented constant parsing!!!")
+      }
+    }
+  }
+
+  /// Returns address of subfield denoted by `path` stored in field at `a`.
+  mutating func subField(denotedBy path: RecordPath, of a: Address)
```
dabrahams: `denotedBy` is a lot of big words with little semantic value.  `path` is just type information.


```suggestion
  /// Returns the address of `subField` in the object  at `origin`.
  mutating func address(of subField: RecordPath, in origin: Address)
```

My suggestion is imperfect and should eventually be improved.  The main reason is that I don't think we know how to talk about or properly name the thing called `Address` here.  There may not be an object there yet.  I wonder if it should be called `ValueStorage`, i.e. the storage _for_ a value.

Sources/Interpreter/Interpreter.swift
RishabhRD: The reason I was biased towards name Address, as this is the same term being used in the IR. Currently I am seeing Address being same as IR's address.
Regarding address being typed, there is IR instruction: `pointer_to_address`, whose code representation is:
```swift
public struct PointerToAddress: Instruction {
  //...

  /// The type of the converted address.
  public let target: RemoteType
```
that suggested that somehow Address should have type.

Even I found Pointer term more natural, however, IR treats pointer as "untyped memory address", hence I didn't pick it.

Sources/Interpreter/Interpreter.swift
RishabhRD: Yes, it has an `IR.Type?`. But till now I didn't find it useful except for validating IR correctness.

Sources/Interpreter/Interpreter.swift
RishabhRD: Currently in terms of code it doesn't. `alloc_stack` allocates local variable and `dealloc_stack` deallocates the same. I assumed that IR ensures that `return` is called after all calls to `dealloc_stack` (that actually happens). Thus, I didn't put local variables in stack frame as it was not needed.

But then I always think of standard definition of stackframe includes local variables.

Do you think we should put [Allocation.ID] in stackframe even if it is not being used currently?

Sources/Interpreter/Interpreter.swift
RishabhRD: Yes!

Sources/Interpreter/UntypedBuiltinValue.swift
RishabhRD: UntypedBuiltinValue should represent builtin ptr. But encoding to UInt128 might not be possible. As we could have instruction like this:
```
  %i0.0: &[{}] (sink Int) let -> Int = alloc_stack [{}] (sink Int) let -> Int
  %i0.1: ptr = address_to_pointer @FunctionDecl(524290)
  %i0.2: &ptr = subfield_view %i0.0, 0
  %i0.3: &ptr = access [set] %i0.2
  store %i0.1, %i0.3
  end_access %i0.3
```
Here thing to note is:
- Address should also be able to represent function id.
- We need to support "function pointers" too.

IDK if Function.ID can be represented in 128 bits.

Sources/Interpreter/Interpreter.swift
RishabhRD: Regarding ValueStorage, knowing that Function.ID is also an Address (coming from IR instruction), I am not sure if ValueStorage would be good.

Also, shouldn't we use `at` instead of `in` in function name? Then it would match doc summary too.

Sources/Interpreter/Interpreter.swift
RishabhRD: Ah yeah, this is really a bad habit, mostly coming from work related codebases which normalize the same :disappointed:. Trying to improve on it now!!!

Sources/Interpreter/Interpreter.swift
dabrahams: I see.  Well, we need to [resolve the clash](https://val-qs97696.slack.com/archives/C044SSSHY66/p1762646233534099) with `Memory.Address` somehow.

Sources/Interpreter/Interpreter.swift
dabrahams: We need to dynamically prevent unsafe code from committing type/memory errors.  Don't we need the type for that?

dabrahams: Separate thought: can't results be aggregates or union types?  I think there was a reason I used `Any` (if I remember my original design).  Of course we can extend the enum, but then maybe there's no point in squeezing all the different builtin value types into a `UInt128`.

Sources/Interpreter/UntypedBuiltinValue.swift
dabrahams: I don't think Address should represent function IDs.  However, the need to represent capture-less lambdas is real and leaves me more and more convinced that we shouldn't try and stuff different types in here.

Sources/Interpreter/Memory.swift
```diff
@@ -278,6 +278,33 @@ public struct Memory {
     throw Error.noDecomposable(t, at: a)
   }
 
+  /// Copies `n` bytes from `source` to `destination`.
+  public mutating func copy(
+    bytes n: Int, from source: Memory.Address, to destination: Memory.Address
```
```suggestion
    byteCount n: Int, from source: Memory.Address, to destination: Memory.Address
```

`n` is not the bytes.  However, you could:


```suggestion
    _ n: Int, bytesFrom source: Memory.Address, to destination: Memory.Address
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -250,12 +206,17 @@ public struct Interpreter {
     case let x as GlobalAddr:
       _ = x
     case let x as Load:
-      _ = x
+      let address = address(denotedBy: x.source)!;
```
dabrahams: This `denotedBy` label isn't working for me.

Sources/Interpreter/Interpreter.swift
```diff
@@ -314,29 +276,173 @@ public struct Interpreter {
...
+    return .init(memoryAddress: addr, memoryLayout: t)
+  }
+
+  /// Deallocates object allocated at `a`.
+  mutating func deallocate(_ a: Address) throws {
+    try memory.deallocate(a.memoryAddress)
+  }
+
+  /// Returns address of object denoted by operand, if any.
+  func address(denotedBy operand: Operand) -> Address? {
```
dabrahams: Is it the address _of the object_ denoted by the operand, or the address denoted by the operand?  The name is not consistent with the contract.

Isn't this just the operand, which must denote an address?

Sources/Interpreter/Interpreter.swift
```diff
@@ -37,37 +37,6 @@ struct Address {
 
 }
 
-/// The value produced by executing an instruction.
-enum InstructionResult {
```
dabrahams: Having a named type was a good idea.  Maybe just an alias?

Sources/Interpreter/Interpreter.swift
```diff
@@ -314,29 +252,165 @@ public struct Interpreter {
...
+      return topOfStack.registers[instruction] as? Address
+    case .parameter(_, let i):
+      return topOfStack.parameters[i]
+    case .constant:
+      return nil
+    }
+  }
+
+  /// Returns builtin value referenced by `operand`, if any.
+  func toBuiltinValue(_ operand: Operand) -> BuiltinValue? {
```
dabrahams, line 316: Don't repeat type information in the names of non-types

```suggestion
  /// Returns the value of `x` if it has a builtin type, or `nil` if it does not.
  func builtinValue(_ x: Operand) -> BuiltinValue? {
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -314,29 +252,165 @@ public struct Interpreter {
...
+      .blocks[entryBlock]
+      .instructions
+      .firstAddress!
+    programCounter = CodePointer(
+      module: module,
+      instructionInModule: InstructionID(function, entryBlock, entryInstruction)
+    )
+  }
+
+  /// Store builtin value `v` at address `a`.
```
dabrahams, line 400:
```suggestion
  /// Stores `v` in memory at `a`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -314,29 +252,165 @@ public struct Interpreter {
...
+    let entryBlock = function.entry!;
+    let entryInstruction = function.blocks[entryBlock].instructions.firstAddress!
+    programCounter = CodePointer(
+      module: module,
+      instructionInModule: InstructionID(f, entryBlock, entryInstruction)
+    )
+  }
+
+  /// Moves the program counter to start of `b`.
+  mutating func jumpTo(_ b: Block.ID) {
```
dabrahams, line 386:
```suggestion
  /// Sets the program counter to the start of `b`.
  mutating func jump(to b: Block.ID) {
```

Read the swift API guidelines regarding prepositions.

Sources/Interpreter/Interpreter.swift
```diff
@@ -314,29 +252,165 @@ public struct Interpreter {
...
+    case BuiltinType.i(16): .i16(memory[allocation][offset, type: UInt16.self])
+    case BuiltinType.i(32): .i32(memory[allocation][offset, type: UInt32.self])
+    case BuiltinType.i(64): .i64(memory[allocation][offset, type: UInt64.self])
+    case BuiltinType.i(128): .i128(memory[allocation][offset, type: UInt128.self])
+    default: nil
+    }
+  }
+
+  /// Returns block parameters from operands in `arg`.
+  func blockParams(from arg: Call) -> [Address] {
```
dabrahams, line 369: I don't understand the comment.  It probably should not require me to know what block parameters are, since that is the name of the function, except for the abbrev, which you should nix.

I think it returns the addresses of arguments followed by that of the return value space for `arg`.  

`arg` is a terrible name for a function parameter; of course you're going to pass an argument.  Another abbrev.  Is it not notionally a function call?

dabrahams (commented):
I get the sense that perhaps this PR is trying to do too much all at once.  Partly I think so because I find myself wanting to make too many corrections.
Would it make sense to break it into smaller units, so we can make a really clean PR and learn from that?

RishabhRD:
Closing this PR as it is broken into #1775, #1780, #1781, #1782, #1783, #1784 as suggested.

PR #1755, opened 2025-10-30
Interpreter: Alternate stack allocation technique that don't require stripAlignment

RishabhRD:
## Motivation
In current stack allocation technique, we allocate memory prepended with required allocation and return the address of allocation. However, that requires calls stripAlignment everywhere whenever we need to access value there. This call everywhere can be expensive.

## Idea
We allocate the same way, but instead of returning address of alignment, we return address of actual object.
While deallocating, it would only deallocate the actual object and leave alignment on stack. However, deallocating next object would deallocate this left alignment too.

The tradeoff is when we allocate, deallocate and then allocate, the second allocation might have used the padding left there, but I believe that is not that big of an issue, considering simplicity we get around codebase. Also there are more than 50% cases when this strategy would be optimal strategy too.

### DIFF
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -261,7 +261,7 @@ public struct Interpreter {
       return
     case let x as Store:
       let obj = builtIn(denotedBy: x.object)!
-      let dest = stripAlignment(address(denotedBy: x.target)!)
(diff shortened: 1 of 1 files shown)


RishabhRD:
Allocating local variables in same allocation is not useful for invalid pointer detection. Thus closing this PR.

PR #1773, opened 2025-11-16
Remove the `makeXXX` functions from `Module` that create IR instructions.

lucteo:
Remove the `makeXXX` functions from `Module` that create IR instructions.

Replaced them with direct constructor access.
It is envisioned that most of the IR instruction creation is done through
`Emitter`.

### DIFF
--- a/Sources/IR/Analysis/Module+AccessReification.swift
+++ b/Sources/IR/Analysis/Module+AccessReification.swift
@@ -131,7 +131,12 @@ extension Module {
     // Nothing to do if the request set is already a singleton.
     if s.capabilities == [k] { return }
 
-    let reified = makeAccess([k], from: s.source, correspondingTo: s.binding, in: f, at: s.site)
+    let reified = Access(
+      capabilities: [k],
+      accessedType: self[f].type(of: s.source),
+      source: s.source,
+      binding: s.binding,
+      site: s.site)
     self[f].replace(i, with: reified)
   }
 
@@ -143,13 +148,17 @@ extension Module {
     // Generate the proper instructions to prepare the projection's arguments.
     var arguments = s.operands
     for a in arguments.indices where s.parameters[a].access == .yielded {
-      let b = makeAccess([k], from: arguments[a], in: f, at: s.site)
+      let b = Access(
+        capabilities: [k],
+        accessedType: self[f].type(of: arguments[a]),
+        source: arguments[a],
+        site: s.site)
       arguments[a] = .register(self[f].insert(b, at: .before(i)))
     }
 
     let o = RemoteType(k, s.projection)
-    let reified = makeProject(
-      o, applying: s.variants[k]!, specializedBy: s.bundle.arguments, to: arguments, at: s.site)
+    let reified = Project(
+      projection: o, callee: s.variants[k]!, specialization: s.bundle.arguments, operands: arguments, site: s.site)
     self[f].replace(i, with: reified)
   }
 
--- a/Sources/IR/Analysis/Module+AccessReification.swift
+++ b/Sources/IR/Analysis/Module+AccessReification.swift
@@ -148,9 +148,11 @@ extension Module {
     // Generate the proper instructions to prepare the projection's arguments.
     var arguments = s.operands
     for a in arguments.indices where s.parameters[a].access == .yielded {
+      let t = self[f].type(of: arguments[a])
+      precondition(t.isAddress)
       let b = Access(
         capabilities: [k],
-        accessedType: self[f].type(of: arguments[a]),
+        accessedType: t,
         source: arguments[a],
         site: s.site)
       arguments[a] = .register(self[f].insert(b, at: .before(i)))
--- a/Sources/IR/Emitter.swift
+++ b/Sources/IR/Emitter.swift
@@ -89,6 +89,12 @@ struct Emitter: Sendable {
     program.ast
   }
 
+  /// The IR in which we are currently inserting.
+  private var insertionIR: Function? {
+    guard let f = insertionFunction else { return nil }
+    return module[f]
+  }
+
   /// The basic block in which new instructions are currently inserted.
   private var insertionBlock: Block.ID? {
     insertionPoint?.block
@@ -130,7 +136,7 @@ struct Emitter: Sendable {
   @discardableResult
   private mutating func insert<I: Instruction>(_ newInstruction: I) -> Operand? {
     let i = module[insertionFunction!].insert(newInstruction, at: insertionPoint!)
-    return module[insertionFunction!].result(of: i)
+    return insertionIR!.result(of: i)
   }
 
   // MARK: Top-level entry points
@@ -668,7 +674,7 @@ struct Emitter: Sendable {
   private mutating func assignProjections(of rhs: Operand, to d: BindingPattern.ID) {
     precondition(!program[d].introducer.value.isConsuming)
     let k = AccessEffect(program[d].introducer.value)
-    let request: AccessEffectSet = module[insertionFunction!].isSink(rhs) ? [k, .sink] : [k]
+    let request: AccessEffectSet = insertionIR!.isSink(rhs) ? [k, .sink] : [k]
 
     for (path, name) in ast.names(in: program[d].subpattern) {
       var part = lowering(program[name].decl) {
@@ -744,7 +750,7 @@ struct Emitter: Sendable {
   private mutating func _emitMoveInitRecordParts(
     of receiver: Operand, consuming argument: Operand
   ) {
-    let layout = AbstractTypeLayout(of: module[insertionFunction!].type(of: receiver).ast, definedIn: program)
+    let layout = AbstractTypeLayout(of: insertionIR!.type(of: receiver).ast, definedIn: program)
 
     // If the object is empty, simply mark it initialized.
     if layout.properties.isEmpty {
@@ -768,7 +774,7 @@ struct Emitter: Sendable {
   private mutating func _emitMoveInitUnionPayload(
     of receiver: Operand, consuming argument: Operand
   ) {
-    let t = UnionType(module[insertionFunction!].type(of: receiver).ast)!
+    let t = UnionType(insertionIR!.type(of: receiver).ast)!
 
     // If union is empty, simply mark it initialized.
     if t.elements.isEmpty {
@@ -903,7 +909,7 @@ struct Emitter: Sendable {
   private mutating func _emitCopyRecordParts(
     from source: Operand, to target: Operand
   ) {
-    let layout = AbstractTypeLayout(of: module[insertionFunction!].type(of: source).ast, definedIn: program)
+    let layout = AbstractTypeLayout(of: insertionIR!.type(of: source).ast, definedIn: program)
 
     // If the object is empty, simply mark the target as initialized.
     if layout.properties.isEmpty {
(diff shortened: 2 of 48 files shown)


Sources/IR/Analysis/Module+AccessReification.swift
```diff
@@ -131,7 +131,12 @@ extension Module {
     // Nothing to do if the request set is already a singleton.
     if s.capabilities == [k] { return }
 
-    let reified = makeAccess([k], from: s.source, correspondingTo: s.binding, in: f, at: s.site)
+    let reified = Access(
+      capabilities: [k],
```
dabrahams, line 135: I think the original labels worked; why not carry them forward?
Labeling every parameter with a noun is a last-resort kind of move.  See Swift API guidelines.

lucteo: I haven't changed the labels; I kept them as they were. It's just that I've dropped the `make` functions from `Module` and resorted to the initializers of the types.

dabrahams: I see, sorry.  We should probably change the `init` functions to match the signatures in `Emitter` where possible, and the `makeXXX` functions elsewhere.  I would say that is a completely separate change but the `makeXXX` signatures will be lost after this one and the existing signatures make these changes a lot bigger than they would be.  WDYT?

lucteo: Overall, we said that clients will use the underscore functions from `Emitter`. But, I think that we will still go back and forth a bit on this.

I would keep this PR as much as possible, to just removing the `makeXXX` functions from `Module`. But, I'm fine changing the signatures of the initializers if that makes more sense.

(as typing is not the main bottleneck, and as calling these initializers will be encapsulated anyways (hopefully, eventually), I don't have a strong preference.).

kyouko-taiga: Wouldn't it be better to write a PR that directly moves these functions to emitter?
See my next comment for my rationale.

Sources/IR/Analysis/Module+AccessReification.swift
```diff
@@ -143,13 +148,17 @@ extension Module {
     // Generate the proper instructions to prepare the projection's arguments.
     var arguments = s.operands
     for a in arguments.indices where s.parameters[a].access == .yielded {
-      let b = makeAccess([k], from: arguments[a], in: f, at: s.site)
+      let b = Access(
+        capabilities: [k],
+        accessedType: self[f].type(of: arguments[a]),
```
kyouko-taiga: This line explains why we have `makeXXX` functions.

There are countless invariants in the IR. Almost none of them are checked but at least the `makeXXX` functions ensure that many of them hold by construction. That's the case here because the invariant is that the type of the accessed value is the type of the type of the access origin.

If we merge this PR we'll loose these "correct-by-construction" guarantees because expressions like this one will end up looking like any arbitrary code in the emitter.

I think we need a way to document the invariants that are currently enforced by the `makeXXX` functions somewhere. I would be satisfied if that's only as comments.

Sources/IR/Emitter.swift
```diff
@@ -89,6 +89,12 @@ struct Emitter: Sendable {
     program.ast
   }
 
+  /// The IR in which we are currently inserting.
```
kyouko-taiga: Neither the comment nor the name of the property makes much sense to me. I want to call this property `insertionFunction` but I see that it clashes with another one.

In general I have come to accept that we need a lot of subscripting to interact with the various APIs of the compiler. Even in the new front-end there's a fair amount of `program.types[x]` in various places. I also believe it creates a better debugging experience because I almost never want to enter a getter when I step into an expression like `foo(bar.ham)`. Long story short my point is that I don't think avoiding subscripting is necessarily a clear cut improvement.

Looking at the rest of Emitter it seems this particular change mostly replaces `module[insertionFunction!].type(of: ...)` for `insertionIR!.type(of: ...)`. Then I would suggest to write a function `irType(of:)` and dispense with this property.

Sources/IR/Emitter.swift
```diff
@@ -3406,7 +3412,8 @@ struct Emitter: Sendable {
     if a.mayHoldCaptures {
       _release_capture(a.source)
     }
-    insert(module.makeDeallocStack(for: source, in: insertionFunction!, at: currentSource))
+    precondition(source.instruction.map({ module[$0, in: insertionFunction!] is AllocStack }) ?? false)
```
kyouko-taiga: I dislike the `$n` syntax subjectively but with passion and I find the pattern `optional.map(somePredicate) ?? false` to be difficult to parse.

I think we should add this extension to `Optional`:
```
extension Optional {
  func satisfies(_ predicate: (Wrapped) throws -> Bool) rethrows -> Bool {
    try if let w = self { predicate(w) } else { false }
  }
}
```

Then this statement becomes:
```suggestion
    precondition(source.instruction.satisfies({ (s) in module[s, in: insertionFunction!] is AllocStack }))
```

Sources/IR/Emitter.swift
```diff
@@ -3416,14 +3423,21 @@ struct Emitter: Sendable {
...
-      let s = module.makeSubfieldView(of: r.recordAddress, subfield: p, in: insertionFunction!, at: currentSource)
-      return insert(s)!
+      elementPath = r.subfield + subfield
+      address = r.recordAddress
     } else {
-      let s = module.makeSubfieldView(of: recordAddress, subfield: subfield, in: insertionFunction!, at: currentSource)
-      return insert(s)!
+      elementPath = subfield
+      address = recordAddress
     }
```
kyouko-taiga: This logic deserves its own function.

Sources/IR/Emitter.swift
```diff
@@ -3416,14 +3423,21 @@ struct Emitter: Sendable {
...
+    let ir = insertionIR!
```
kyouko-taiga: This property is very dangerous for performance!

If Swift decides that it needs to copy `ir` because something else is being mutated in the module it may create very, very large temporaries. That's another reason why I am not a fan of removing explicit subscripting.

Sources/IR/Emitter.swift
```diff
@@ -3416,14 +3423,21 @@ struct Emitter: Sendable {
...
+    precondition(ir.type(of: address).isAddress)
```
kyouko-taiga: I really dislike the fact that these preconditions get injected into the emitter's code. I think that's further evidence that it may not be a good idea to just remove the `makeXXX` functions from `Module` without already re-implementing them somewhere else.

Sources/IR/Emitter.swift
```diff
@@ -3715,160 +3734,248 @@ extension Emitter {
...
+      accessedType: insertionIR!.type(of: s),
+      source: s,
+      binding: binding,
+      site: currentSource))!
   }
 
   fileprivate mutating func _end_access(_ x: Operand) {
-    insert(module.makeEndAccess(x, in: insertionFunction!, at: currentSource))
+    let ir = insertionIR!
+    precondition(x.instruction.map({ ir[$0] is Access }) ?? false)
```
kyouko-taiga: Another instance of the `optional.map(predicate) ?? false` that I think should be expressed more concisely.

Sources/IR/Emitter.swift
```diff
@@ -3715,160 +3734,248 @@ extension Emitter {
...
+    precondition(!variants.isEmpty)
+
+    let t = SubscriptType(
+      program.canonicalType(of: b.bundle, specializedBy: b.arguments, in: insertionScope!))!.pure
+
+    return insert(ProjectBundle(
+      bundle: b, variants: variants,
+      parameters: t.inputs.lazy.map({ ParameterType($0.type)! }),
+      projection: RemoteType(t.output)!.bareType,
+      operands: arguments, site: currentSource))!
```
kyouko-taiga: This has to go in a function.

Sources/IR/Operands/Instruction/AdvancedByBytes.swift
```diff
@@ -45,20 +45,3 @@ extension AdvancedByBytes: CustomStringConvertible {
   }
 
 }
-
-extension Module {
-
-  /// Creates an `advanced by bytes` instruction anchored at `site` computing the `source` address
-  /// value advanced by `offset` bytes.
```
kyouko-taiga, line 52: There's valuable documentation in these comments (and in those attached to other `makeXXX` functions) that should be preserved somehow.

kyouko-taiga (changes requested):
I am not convinced this PR is an improvement.

I understand this is an intermediate change but it is loosing valuable information that I thing will be difficult to recover. At the very least it won't be easy to ensure in the next PR that the invariants currently enforced by the `makeXXX` functions as well as their documentation lands in the right place once the dust settles.

My suggestion is to move the `makeXXX` functions directly to the emitter.

Sources/IR/Analysis/Module+AccessReification.swift
lucteo: `makeAccess` had 2 preconditions. The first one is guaranteed to be true in this case. The second one I probably forgot to add it here, so I can add it.

I can double check all the preconditions that we previosuly had, and check if I copied them all over the place.

Sources/IR/Operands/Instruction/AdvancedByBytes.swift
lucteo: Isn't this (advanced by `offset` bytes) documented at in the type documentation?

Sources/IR/Emitter.swift
lucteo: Makes sense. I can revert this change and use `irType(of:)`.

Sources/IR/Emitter.swift
lucteo: Makes sense. I can do this change

BTW. This is one precondition that I ported from `makeDeallocStack`.

Sources/IR/Emitter.swift
lucteo: will do

Sources/IR/Emitter.swift
lucteo: I see.
I've tried to create such properties whenever we had multiple calls to the same subscripts multiple times in the same region of the code.

Maybe I can store just the value of the `irType(of: address)`. WDYT?

Sources/IR/Emitter.swift
lucteo: Wasn't this what we previously agreed?

We said that ideally, most of the instruction creation will be in emitter, and then it makes sense for the classes to just have the initializers. (and this way we can get rid of that extra `Function.ID` parameter from many places.)

Sources/IR/Emitter.swift
lucteo: will go over all these and change to your suggestion

Sources/IR/Emitter.swift
lucteo: I assume you mean everything except the call to `insert`. Is that right?

Sources/IR/Analysis/Module+AccessReification.swift
lucteo: I can. But I'm not sure what the benefits would be. We are just adding another layer to the already-big Emitter.

lucteo: Also, the next place is one place that we are not yet using `Emitter`.
At this point, `Emitter` doesn't seem a good fit for places like this in which we just replace one instruction w/ another.

lucteo:
Abandoning.

PR #1775, opened 2025-11-18
Interpreter Fix: UnrealABI layout calculation and Memory unsafe pointer preconditions

RishabhRD:
UnrealABI should produce correct layout information for builtin integers less than 8-bit.
Example usecase: 1-bit integer that is used in booleans.

Memory.Allocation precondition for providing unsafe pointer access should have equality included in size comparison.

There are additional minimal changes to make CI passing, will need those in my further PRs.

### DIFF
--- a/Tests/InterpreterTests/MemoryTests+internal.swift
+++ b/Tests/InterpreterTests/MemoryTests+internal.swift
@@ -36,12 +36,12 @@ final class InterpreterMemoryInternalTests: XCTestCase {
 
   func testFormingPointerToLastByteOfAllocation() throws {
     var memory = Memory()
-    let allocationAddress = memory.allocate(1, bytesWithAlignment: 1)
+    let a = memory.allocate(1, bytesWithAlignment: 1)
 
-    memory[allocationAddress.allocation].withUnsafeMutablePointer(to: UInt8.self, at: 0) { p in
+    memory[a.allocation].withUnsafeMutablePointer(to: UInt8.self, at: 0) { p in
       p.pointee = 2
     }
-    memory[allocationAddress.allocation].withUnsafePointer(to: UInt8.self, at: 0) { p in
+    memory[a.allocation].withUnsafePointer(to: UInt8.self, at: 0) { p in
       XCTAssertEqual(p.pointee, 2)
     }
 


Tests/InterpreterTests/MemoryTests+internal.swift
```diff
@@ -34,4 +34,11 @@ final class InterpreterMemoryInternalTests: XCTestCase {
     }
   }
 
+  func test_canAccessUnsafePointerToAllocatedBytes() throws {
```
dabrahams: We don't use underscores in Swift names.  I may have used them in the swift-testing code; I seem to remember that it had to do with `swift test` filtering, but that doesn't make much sense to me.

dabrahams:
```suggestion
  func testFormingPointerToLastByteOfAllocation) throws {
```

RishabhRD:
> It looks like your test only checks for one of the two fixes you made

This is because `allocation` variable is `private(set)` inside `Memory` struct. To test mutation, we might need this subscript for `Memory` from my another PR:
```swift
  /// The allocation identified by `i`.
  public subscript(_ i: Allocation.ID) -> Allocation {
    _read {
      yield allocation[i]!
    }
    _modify {
      yield &allocation[i]!
    }
  }
```

dabrahams:
You could use internal(set) but i think private is better. Include the necessary components for performing thorough testing.

RishabhRD:
@dabrahams I added the tests required and component needed for it. You might want to skim through it once. If everything seems good, we can merge this.

Tests/InterpreterTests/MemoryTests+internal.swift
```diff
@@ -34,4 +34,16 @@ final class InterpreterMemoryInternalTests: XCTestCase {
     }
   }
 
+  func testFormingPointerToLastByteOfAllocation() throws {
+    var memory = Memory()
+    let allocationAddress = memory.allocate(1, bytesWithAlignment: 1)
```
dabrahams: This is much too long a name. Try `p` instead and look at the code both ways. Which is easier to understand?

Otherwise LGTM

RishabhRD: Sure! `allocationAddress` is not bringing much here. So, going with single character name only.

RishabhRD:
Merging the PR as all concerns by @dabrahams are now resolved and approved.

PR #1780, opened 2025-11-22
Interpreter: unified allocation for stack and heap memory

RishabhRD:
The comment indicates that `memory` should be used for both stack and heap:
```swift
  /// The stack- and dynamically-allocated memory in use by the program.
  private var memory = Memory()
```
Thus, removing StackAllocation and Stack as StackFrame doesn't own the memory but just contains allocation addresses. All the allocation is owned by memory in interpreter.

### Still using Address instead of Place, Why?
In 18th Nov meeting, we discussed that renaming Address to Place might result in merge conflict with Lucian's current work and thus we might want to hold it a little. I used the term Address for now, just to be consistent with IR.

### Why no unit tests for this PR?
We don't have any program that can just produce alloc_stack and dealloc_stack instructions without producing store, subfield_view and access instructions. So, will cover this test in later PR.

### DIFF
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -21,7 +21,7 @@ struct CodePointer {
 /// The value produced by executing an instruction.
 struct InstructionResult {
   /// The instruction's output as an opaque, type-erased value.
-  var payload: Any
+  public var payload: Any
 }
 
 /// A namespace for notional module-scope declarations that actually
@@ -31,13 +31,13 @@ struct InstructionResult {
 enum ModuleScope {
 
   /// A typed location in memory.
-  struct Address: Regular {
+  public struct Address: Regular {
 
     /// The position in memory.
-    let startLocation: Memory.Address
+    public let startLocation: Memory.Address
 
     /// The type to be accessed at `startLocation`.
-    let type: TypeLayout
+    public let type: TypeLayout
 
   }
 
@@ -50,14 +50,14 @@ typealias Address = ModuleScope.Address
 /// call.
 struct StackFrame {
   /// The results of instructions.
-  var registers: [InstructionID: InstructionResult] = [:]
+  public var registers: [InstructionID: InstructionResult] = [:]
 
   /// The program counter to which execution should return when
   /// popping this frame.
-  var returnAddress: CodePointer
+  public var returnAddress: CodePointer
 
   /// The allocations in this stack frame.
-  var allocations: [Address] = []
+  public var allocations: [Address] = []
 }
 
 extension UnsafeRawPointer {
@@ -102,13 +102,13 @@ struct Stack {
   private var frames: [StackFrame] = []
 
   /// Adds a new frame on top with the given `returnAddress`.
-  mutating func push(returnAddress: CodePointer) {
+  public mutating func push(returnAddress: CodePointer) {
     let f = StackFrame(returnAddress: returnAddress)
     frames.append(f)
   }
 
   /// Removes the top frame and returns its `returnAddress`.
-  mutating func pop() -> CodePointer {
+  public mutating func pop() -> CodePointer {
     let f = frames.last!
     defer {
       frames.removeLast()
@@ -117,7 +117,7 @@ struct Stack {
   }
 
   /// The top stack frame.
-  var top: StackFrame {
+  public var top: StackFrame {
     _read {
       precondition(!isEmpty)
       yield frames[frames.count - 1]
@@ -129,7 +129,7 @@ struct Stack {
   }
 
   /// Boolean indicating whether stack contains atleast 1 stack frame.
-  var isEmpty: Bool {
+  public var isEmpty: Bool {
     frames.isEmpty
   }
 


Sources/Interpreter/Interpreter.swift
```diff
@@ -10,48 +10,32 @@ struct CodePointer {
 
 }
 
+/// The value produced by executing an instruction.
+typealias InstructionResult = Any
+
+/// Address to allocated object with type layout.
```
dabrahams: Objects aren't allocated; memory is.  And I don't think there's necessarily an object there, which would require an initialization to have been performed.

This is the best I could  come up with:

```suggestion
/// A typed location in memory.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -10,48 +10,32 @@ struct CodePointer {
...
+struct Address {
+
+  /// Address of object in memory.
+  public let memoryAddress: Memory.Address
```
dabrahams: Repeating the name of the type in the variable and the comment is an anti-pattern.  And I don't think there's necessarily an object.


```suggestion
  /// The position in memory.
  public let location: Memory.Address
```

```suggestion
  /// The position in memory.
  public let startLocation: Memory.Address
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -10,48 +10,32 @@ struct CodePointer {
...
+
+  /// Type layout of object.
+  public let memoryLayout: TypeLayout
```
dabrahams: I think the role of this is that it's the type part of the typed location.  The whole layout is stored merely as a convenience, but that could instead be looked up at any time.


```suggestion
  /// The type to be accessed at `position`.
  public let type: TypeLayout
```

Sources/Interpreter/Interpreter.swift
dabrahams: I presume we're going to deal with the naming issue in a different PR?

Sources/Interpreter/Interpreter.swift
```diff
@@ -10,48 +10,32 @@ struct CodePointer {
...
-      "The latest allocation that has not been deallocated must be deallocated first.")
-    precondition(a.frame == id, "Can't deallocate address from a different frame.")
-    precondition(a.byteOffset == 0, "Can't deallocate the memory of a subobject.")
-    precondition(
-      a.memoryLayout.type == allocations.last!.structure.type,
-      "Deallocating using address of the wrong type; perhaps this is a subobject?")
-    allocations.removeLast()
-  }
-
+  var allocations: [Address] = []
```
dabrahams: Think about whether these should reallly be `Address` or just `Allocation.ID`.

Sources/Interpreter/Interpreter.swift
```diff
@@ -151,10 +71,7 @@ public struct Interpreter {
   private var memory = Memory()
 
   /// Local variables, parameters, and return addresses.
-  private var stack = Stack()
-
-  /// A mapping from the `id`s of stack frames to their index in the stack.
-  private var frameIDToIndex: [StackFrame.ID: Int] = [:]
+  private var stackFrames: [StackFrame] = []
```
dabrahams: You're just repeating type information.  The role that this thing plays is that of the stack, or the call stack.

```suggestion
  private var stack: [StackFrame] = []
```

or

```suggestion
  private var callStack: [StackFrame] = []
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -216,7 +133,9 @@ public struct Interpreter {
       _ = x
 
     case let x as AllocStack:
-      currentRegister = StackAllocation(typeLayout[x.allocatedType])
+      let allocationAddress = allocate(typeLayout[x.allocatedType])
+      topOfStack.allocations.append(allocationAddress)
+      currentRegister = allocationAddress
```
dabrahams: Long names for variables with short lifetimes tend to obfuscate.  I suggest a standard single character name for the variable that holds the instruction result, _if you need one_.

```suggestion
      let r = allocate(typeLayout[x.allocatedType])
      topOfStack.allocations.append(r)
      currentRegister = r
```

or

```suggestion
      currentRegister = allocate(typeLayout[x.allocatedType])
      topOfStack.allocations.append(currentRegister)
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -239,7 +158,8 @@ public struct Interpreter {
     case let x as ConstantString:
       currentRegister = x.value
     case let x as DeallocStack:
-      _ = x
+      try deallocate(topOfStack.registers[x.location.instruction!]! as! Address)
```
dabrahams: The expression in parentheses is a bit complex.  Is there a higher-level abstraction you can lift out of it to make this more readable?

Sources/Interpreter/Interpreter.swift
```diff
@@ -331,12 +251,23 @@ public struct Interpreter {
...
   mutating func popStackFrame() {
-    programCounter = stack.pop()
-    if stack.frames.isEmpty {
+    programCounter = stackFrames.popLast()!.returnAddress
+    if stackFrames.isEmpty {
       isRunning = false
     }
   }
 
+  /// Allocates object of type layout `t` on `memory` and returns address of start of object.
```
```suggestion
  /// Allocates memory for an object of type `t` and returns the address.
```

`t` is notionally a type; the layout just happens to be handy.  We don't need to say which memory variable the interpreter allocated memory in.  There's only one for the whole interpreter.

Sources/Interpreter/Interpreter.swift
```diff
@@ -331,12 +251,23 @@ public struct Interpreter {
...
+  mutating func allocate(_ t: TypeLayout) -> Address {
+    let addr = memory.allocate(t.size, bytesWithAlignment: t.alignment)
+    return .init(memoryAddress: addr, memoryLayout: t)
+  }
+
+  /// Deallocates object allocated at `a`.
```
```suggestion
  /// Deallocates the memory at `a`.
```

A precondition is missing.

But it seems to me `Address` is still the wrong name and it really denotes a typed _region_ of memory.  If its name reflected that fact, we could say

```suggestion
  /// Deallocates `a`.
```

dabrahams (commented):
See remarks inline

Sources/Interpreter/Interpreter.swift
RishabhRD: > I presume we're going to deal with the naming issue in a different PR?

Yes! we are holding it a little as it would create merge conflict with Lucian's current work on Emitter. Once that's done (which we discussed should not take much time now), I would create a PR that renames Address to Place (or something better if possible).

Sources/Interpreter/Interpreter.swift
RishabhRD: I introduced a function named `addressProduced(by: )` for the same. Please review that too.

Sources/Interpreter/Interpreter.swift
RishabhRD: `Address` is surely the wrong name. Maybe we can rename Address struct for the interpreter now and just hold renaming in IR side until Lucian's work is merged?

Sources/Interpreter/Interpreter.swift
RishabhRD: Yes, `Allocation.ID` seems to be sufficient and thus I changed it accordingly.

I was just wondering if type information in `Address` would be helpful somehow. Maybe we should have an additional `deallocateLocalVariable` function for `dealloc_stack` so that we can enforce in precondition that address being deallocated is address of latest allocated object on stack and has correct type.

dabrahams: The more semantic information and precondition checking you can get into functions the better!

RishabhRD: Sure! added that with name `deallocateStack` to match IR naming. And with this, having `Address` makes sense there as that is bringing the precondition of deallocating with correct type.

Sources/Interpreter/Interpreter.swift
```diff
@@ -14,57 +15,32 @@ struct CodePointer {
 
 }
 
+/// The value produced by executing an instruction.
+typealias InstructionResult = Any
```
dabrahams: At some future time I imagine `Value` could become a clearly more appropriate name.

Sources/Interpreter/Interpreter.swift
```diff
@@ -14,57 +15,32 @@ struct CodePointer {
...
-
-  /// An identifier by which we can look up a frame in a given
-  /// Stack.
-  // We could just use the frame's index on the stack but we'd like
-  // to be able to rigorously detect invalid addresses.
-  typealias ID = UUID
-
   /// The results of instructions.
-  var registers: [InstructionID: Any] = [:]
+  var registers: [InstructionID: InstructionResult] = [:]
```
dabrahams: I wonder if InstructionID is overkill.  I don't remember if it contains the function ID, but if so we might want to eventually pare this key down and store the function ID in the frame.

Sources/Interpreter/Interpreter.swift
dabrahams: We might eventually want a typed allocation ID that doesn't include the offset.

Sources/Interpreter/Interpreter.swift
```diff
@@ -212,9 +95,9 @@ public struct Interpreter {
 
   /// The top stack frame.
   private var topOfStack: StackFrame {
-    get { stack.frames.last! }
+    get { callStack.last! }
```
dabrahams: Since our use of `callStack` never or seldom involves anything but its top, we probably ought to have a `Stack` type for this rather than using a raw array.  Depending on how that's designed, it wouldn't necessarily eliminate the need for this property, which will trap if there is no top.   A `Stack<T>` type might have an optional top.  For efficiency, if you do that, the _modify accessor should temporarily pop the top to avoid needless COWs… wow that's a little messy. 

 I would tell you to make it a separate PR except I see that you've removed the `Stack` I wrote, so this is a step backward IMO

Sources/Interpreter/Interpreter.swift
```diff
@@ -235,13 +118,12 @@ public struct Interpreter {
...
-    stack.push(returnAddress: programCounter)
+    callStack.append(StackFrame(returnAddress: programCounter))
     typeLayout = .init(typesIn: p.base, for: UnrealABI())
   }
 
-  /// The value of the current instruction's result, if it has been computed.
-  private var currentRegister: Any? {
-    get { topOfStack.registers[programCounter.instructionInModule] }
+  private var currentRegister: InstructionResult? {
+    get { topOfStack.registers[programCounter.instructionInModule]! }
```
dabrahams: Why the addition of unwrapping `!` here?

I worry that `Any?` might be too easily implicitly converted from other things.  Maybe `InstructionResult` should be a `struct` (containing an `Any` for now)?

Sources/Interpreter/Interpreter.swift
```diff
@@ -283,7 +167,8 @@ public struct Interpreter {
     case let x as ConstantString:
       currentRegister = x.value
     case let x as DeallocStack:
-      _ = x
+      let a = addressProduced(by: x.location.instruction!)!
```
dabrahams: I'm concerned about both `!`s here.  It's better to put preconditions on things than litter the code with `!`s.  To start with, doesn't every `.location` have an `instruction`?

Sources/Interpreter/Interpreter.swift
```diff
@@ -375,12 +260,42 @@ public struct Interpreter {
   ///
   /// - Precondition: the program is running.
   mutating func popStackFrame() {
-    programCounter = stack.pop()
-    if stack.frames.isEmpty {
+    programCounter = callStack.popLast()!.returnAddress
```
dabrahams: Don't we have to deallocate all the frame's allocations?

Sources/Interpreter/Interpreter.swift
```diff
@@ -375,12 +260,42 @@ public struct Interpreter {
...
+    let a = memory.allocate(t.size, bytesWithAlignment: t.alignment)
+    return .init(startLocation: a, type: t)
+  }
+
+  /// Deallocates `a`.
+  mutating func deallocate(_ a: Address) throws {
+    precondition(a.startLocation.offset == 0, "Can't deallocate the memory of subobject.")
+    precondition(
+      memory.allocation[a.startLocation.allocation]?.size == a.type.size,
+      "Deallocating using address of the wrong type.")
```
dabrahams: These look like preconditions that will be repeated elsewhere.  Can we sink them into `Memory`?

dabrahams (approved):
Overall LGTM. A few changes requested.  Use your own judgement about whether those requests are right and/or we need more discussion before merging.

Sources/Interpreter/Interpreter.swift
RishabhRD: Actually that would not be needed as `dealloc_stack` should deallocate all the local variables. I think its worthy to make this property as precondition (i.e., `allocations.isEmpty`).

Sources/Interpreter/Interpreter.swift
RishabhRD: I don't understand the reasoning behind preferring the name `Value`.  The name `Value` doesn't convey much to me about what to expect from type.

Sources/Interpreter/Interpreter.swift
RishabhRD: I am facing a little problem in accomplishing this. I wanted to move this function as extension of `Memory`. However, then the type name `Address` is conflicting with `Memory.Address`.
The simplest solution would be renaming `Address` (that was our goal too).
Or, if this is possible without renaming, will need little guidance as I am little new to swift.

Sources/Interpreter/Interpreter.swift
RishabhRD: Agreed! Removing `Stack` type was a mistake. I brought it back. I am little confused about "optional top" property for `Stack`. Assuming, you meant: `var top: StackFrame? {...}`, mutation semantics for this property is quite ambiguous. For example, mutating `top` in `nil` case.
For the time being, I have a `top` property whose type is `StackFrame`.

RishabhRD:
Hi @dabrahams, I tried to resolve most of the comments provided. I replied to some comments for which I needed a little discussion.
I am assuming the comments that says "eventually we want something", should be addressed in a different PR, not in this PR.

Sources/Interpreter/Interpreter.swift
dabrahams: You can always explicitly qualify names to disambiguate.  If the name is at module scope you can use `Interpreter.Address`.  Does that help?

Sources/Interpreter/Interpreter.swift
dabrahams: SGTM

Sources/Interpreter/Interpreter.swift
dabrahams: Yeah, you make a good point about the mutation thing.  Preconditions are good :-)

Sources/Interpreter/Interpreter.swift
dabrahams: Doesn't `InstructionResult` need to be able to represent every value that exists at the IR level, and won't we need to use that type in places where it isn't playing the role of the result of an instruction?  If so, `Interpreter.Value` seems right to me.  No?

dabrahams:
Yes, "eventually" comments are speculative and designed to get us thinking in the same direction.

Sources/Interpreter/Interpreter.swift
RishabhRD: Now I got your point. Thanks!

Sources/Interpreter/Interpreter.swift
RishabhRD: I have tried that already. The problem is `Interpreter.Address` confuses compiler to think it as `Interpreter.Interpreter.Address` where the second `Interpreter` type is the `Interpreter` struct we have. (Sorry if it is really a dumb question and there exists some trivial way to handle this.)

dabrahams: OK, I suggest this: create a "namespace" and define `Interpreter.Address` there:

```swift
enum ModuleScope {
  struct Address { ... }
}

typealias Address = ModuleScope.Address
```

Now you can always access `ModuleScope.Address` directly should you need to.

RishabhRD:
Hi @dabrahams, I have resolved all the comments above except the ones labelled with eventually. Mainly:
1. Renaming the `Address` to something better (maybe `Place`).
2. Renaming `InterpreterResult` to `Value`
3. Having a typed allocation ID that doesn't include offset
4. In registers `[InstructionID: InstructionResult]`, storing `FunctionID` for the frame, instead of repeating it all the time.

I am aligned with all of the points mentioned, and am planning to create PRs in those directions in future. Do you think we should cover any point in this PR itself?
If not, I would ask you to skip through the new changes once. Once, it looks good, we can merge the PR.

Sources/Interpreter/Interpreter.swift
```diff
@@ -24,17 +24,24 @@ struct InstructionResult {
   var payload: Any
 }
 
-/// A typed location in memory.
-struct Address: Regular {
+enum ModuleScope {
```
dabrahams, line 31:
```suggestion
/// A namespace for notional module-scope declarations that actually
/// can't be accessed without ambiguity in some cases.
///
/// Aliases for these names are normally placed at module scope for
/// ease-of-use.
enum ModuleScope {
```

…which reminds me, I haven't vetted the use of access control, and I'm not going to other than to suggest you double check that everything has the appropriate access control.  In particular, I don't remember what the rest of the project does w.r.t. explicitly declaring access control that is implied (e.g. `internal` in this case).    Just try to be consistent with that.

Sources/Interpreter/Interpreter.swift
```diff
@@ -72,6 +79,17 @@ extension UnsafeRawBufferPointer {
 
 }
 
+extension Memory {
+  /// Deallocates `a`.
```
```suggestion
  /// Deallocates the allocated memory at `a`, leaving it deallocated and 
  /// rendering `a` unusable for any purpose.
```

1. you aren't deallocating the address, but the memory it refers to 
2. We needed to say something that makes clear there's a precondition on this function; "leaving it deallocated" does that.

Sources/Interpreter/Interpreter.swift
RishabhRD: Yes, I have tried to be consistent with access control conventions already being used in Interpreter code.

RishabhRD:
Resolved the new comments :)

Sources/Interpreter/Interpreter.swift
```diff
@@ -34,10 +34,10 @@ enum ModuleScope {
   struct Address: Regular {
 
     /// The position in memory.
-    public let startLocation: Memory.Address
+    let startLocation: Memory.Address
 
     /// The type to be accessed at `startLocation`.
-    public let type: TypeLayout
+    let type: TypeLayout
```
dabrahams: This is a stylistic nit but I actually pushed for the ability to do this on members of things with non-public access, so we could describe whether they were public with-respect-to their containing scope, and I think it's better when done that way.  Feel free to merge anytime you think that's been resolved.

RishabhRD: Makes sense :-) Followed the same. Thus merging the PR. Thanks!

PR #1781, opened 2025-11-23
Interpreter: Support using local variables in functions

RishabhRD:
The PR does minimal changes to support using local variable in the function. For that we need to support `store`, `subfield_view`, `access` instructions. Considering `alloc_stack` and `dealloc_stack` have been covered in previous PR.

### DIFF
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -18,7 +18,7 @@ struct CodePointer {
 
 }
 
-/// Represents value at IR level.
+/// A value manipulated by the IR.
 struct Value {
   /// The underlying type-erased representation of value.
   public var payload: Any


Sources/Interpreter/BuiltinValue.swift
```diff
@@ -0,0 +1,93 @@
...
+
+/// Methods to create builtin value from IR constants.
+extension BuiltinValue {
+
+  /// Creates instance of builtin value with integer constant `c`.
+  public init(withIntegerConstant c: IntegerConstant) {
+    self =
+      switch c.value.bitWidth {
+      case 1: .i1(Bool(c.value != 0))
+      case 8: .i8(UInt8(truncatingIfNeeded: c.value))
```
dabrahams: Truncating seems very wrong

Sources/Interpreter/BuiltinValue.swift
```diff
@@ -76,7 +76,7 @@ enum BuiltinValue {
 /// Methods to create builtin value from IR constants.
 extension BuiltinValue {
 
-  /// Creates instance of builtin value with integer constant `c`.
+  /// Creates instance of builtin value with `c`.
```
dabrahams: I don't know what it means to create an instance “with” something.  “of builtin value” is redunda tn.

 
```suggestion
  /// An instance with value `c`.
```

Maybe?

Sources/Interpreter/BuiltinValue.swift
```diff
@@ -76,7 +76,7 @@ enum BuiltinValue {
...
   public init(withIntegerConstant c: IntegerConstant) {
```
dabrahams: `withIntegerConstant` is needlessly putting type information in the function name.  Is this a value-preserving conversion?  In that case:


```suggestion
  public init(_ c: IntegerConstant) {
```

I'm not sure the name `IntegerConstant` is good either, but that's a separate issue.

RishabhRD: Assuming value-preserving means preserving mathematical value, then yes its a value preserving conversion.

Sources/Interpreter/Interpreter.swift
```diff
@@ -41,6 +41,9 @@ struct StackFrame {
 
   /// The allocations in this stack frame.
   var allocations: [Address] = []
+
+  /// Function parameters
```
```suggestion
  /// Locations of values passed to the function.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -296,6 +299,54 @@ public struct Interpreter {
     topOfStack.registers[i] as? Address
   }
 
+  /// Returns the value of `x` if it has a address, or `nil` if it does not.
+  func address(_ x: Operand) -> Address? {
```
```suggestion
  func asAddress(_ x: Operand) -> Address? {
```

I guess given the semantics, this should read like a cast.

Sources/Interpreter/Interpreter.swift
```diff
@@ -204,9 +206,10 @@ public struct Interpreter {
       popStackFrame()
       return
     case let x as Store:
-      _ = x
+      store(builtinValue(x.object)!, at: address(x.target)!)
     case let x as SubfieldView:
-      _ = x
+      let p = address(x.recordAddress)!;
```
```suggestion
      let p = address(x.recordAddress)!
```

Instinct, I know

Sources/Interpreter/Interpreter.swift
```diff
@@ -296,6 +299,54 @@ public struct Interpreter {
...
 
+  /// Returns the value of `x` if it has a address, or `nil` if it does not.
+  func address(_ x: Operand) -> Address? {
+    switch x {
+    case .register(let instruction):
+      return topOfStack.registers[instruction] as? Address
+    case .parameter(_, let i):
+      return topOfStack.parameters[i]
+    case .constant:
+      return nil
```
```suggestion
    case .register(let instruction):
      topOfStack.registers[instruction] as? Address
    case .parameter(_, let i):
      topOfStack.parameters[i]
    case .constant:
      nil
```

switch is an expression-statement in Swift these days

Sources/Interpreter/Interpreter.swift
```diff
@@ -296,6 +299,54 @@ public struct Interpreter {
...
+  /// Returns the address of `subField` in the object  at `origin`.
+  mutating func address(of subField: RecordPath, in origin: Address) -> Address {
+    let l = origin.startLocation;
+    var o = l.offset;
+    var t = origin.type;
+    for i in subField {
+      o += t.parts[i].offset
+      t = typeLayout[t.parts[i].type]
+    }
+    return .init(startLocation: .init(allocation: l.allocation, offset: o), type: t)
```
dabrahams: No `;`s. No raw loops..
```suggestion
    let l = origin.startLocation
    let (o, t) = subField.reduce(into: l.offset, origin.type) { (i, t), i in 
      let p = t.parts[i]
      o += p.offset 
      t = typeLayout[p.type]
    }
    
    return .init(
      startLocation: .init(allocation: l.allocation, offset: o), 
      type: typeLayout[t.parts.last?.type ?? t])
```

If `t.parts` is never empty in this case you can simplify the last line further.

dabrahams (commented):
Temporarily ran out of time after my latest comment

Sources/Interpreter/Interpreter.swift
```diff
@@ -296,6 +299,54 @@ public struct Interpreter {
...
+  }
+
+  /// Returns the value of `x` if it has a builtin type, or `nil` if it does not.
+  func builtinValue(_ x: Operand) -> BuiltinValue? {
```
```suggestion
  func asBuiltinValue(_ x: Operand) -> BuiltinValue? {
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -296,6 +299,54 @@ public struct Interpreter {
...
+    case .register(let instruction):
+      return topOfStack.registers[instruction] as? BuiltinValue
+    case .parameter:
+      return nil;
+    case .constant(let c):
+      switch c {
+      case let x as IntegerConstant:
+        return BuiltinValue(x)
+      default:
+        fatalError("unimplemented constant parsing!!!")
```
dabrahams: If we don't have an `unimplemented()` function for this, we should.


```suggestion
        unimplemented("integer constant parsing!!!")
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -296,6 +299,54 @@ public struct Interpreter {
...
+      }
+    }
+  }
+
+  /// Stores `v` at `a`.
+  mutating func store(_ v: BuiltinValue, at a: Address) {
+    let allocation = a.startLocation.allocation
+    let offset = a.startLocation.offset
+    memory[allocation].store(v, at: offset)
```
dabrahams: Seems to me there should be a method on memory that does this part


```suggestion
    memory.store(v, at: offset)
```

Then it can have its own test.

Sources/Interpreter/Interpreter.swift
RishabhRD: I guess you wanted to say `non-integer constant parsing` as currently `IntegerConstant` is the only constant parsing we support.

Sources/Interpreter/Interpreter.swift
RishabhRD: I realized that this whole logic is independent of interpreter and actually is just responsibility of `TypeLayoutCache`. Hence, provided this as extension method and thus can test this logic too :)

Sources/Interpreter/Interpreter.swift
RishabhRD: Yes, this function just requires knowledge of `Memory` and thus we can simply provide it as extension method of `Memory`. However, I am facing a problem to implement this, [I described in my earlier PR](https://github.com/hylo-lang/hylo/pull/1780#discussion_r2616252444). Need your help here to resolve this.

dabrahams: Qualified names work: `Interpreter.Address`, `Memory.Address`, etc.

RishabhRD:
Resolved all the comments provided; Please review the new changes, so we can merge the same.

Sources/Interpreter/BuiltinValue.swift
```diff
@@ -0,0 +1,93 @@
...
+  case i16(UInt16)
+
+  /// 32-bit builtin integer type.
+  case i32(UInt32)
+
+  /// 64-bit builtin integer type.
+  case i64(UInt64)
+
+  /// 128-bit builtin integer type.
+  case i128(UInt128)
```
dabrahams, line 22: Don't allow this to hold up the PR, but consider at some point whether this should just store a UInt128 and a bir width.  You could probably write a generic constructor and value extractor and make the code smaller.

Sources/Interpreter/Interpreter.swift
```diff
@@ -79,6 +82,19 @@ extension UnsafeRawBufferPointer {
 
 }
 
+extension TypeLayoutCache {
+
+  /// Returns the address of `subField` in the object at `origin`.
```
dabrahams, line 104: When Address becomes Place this will be


```suggestion
  /// Returns the address of `subPart` in `whole`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -88,6 +104,13 @@ extension Memory {
       "Deallocating using address of the wrong type.")
     try deallocate(a.startLocation)
   }
+
+  /// Stores `v` at `a`.
```
dabrahams, line 126: And this will be


```suggestion
  /// Stores `v` in `target`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -88,6 +104,13 @@ extension Memory {
...
+  mutating func store(_ v: BuiltinValue, at a: ModuleScope.Address) {
+    let allocation = a.startLocation.allocation
+    let offset = a.startLocation.offset
+    self[allocation].store(v, at: offset)
```
dabrahams: Don't all `store` functions need to throw if the types don't match up?

Sources/Interpreter/Interpreter.swift
```diff
@@ -198,9 +221,8 @@ public struct Interpreter {
   public mutating func step() throws {
     print("\(currentInstruction.site.gnuStandardText): \(currentInstruction)")
     switch currentInstruction {
-    case is Access:
-      // No effect on program state
-      break
+    case let x as Access:
+      currentRegister = .init(payload: asAddress(x.source)!)
```
dabrahams: It strikes me that we will probably never read `currentRegister` which means writing it should probably be a function.  Or you could dispense with the function by simply having the result value of the `switch` statement be what is written.

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -101,3 +101,16 @@ public struct TypeLayoutCache {
     }
   }
 }
+
+public extension TypeLayoutCache {
+
+  /// Returns the byte offset and layout of `subField` within `type`.
+  mutating func layout(of subField: RecordPath, in type: TypeLayout) -> (Int, TypeLayout) {
```
dabrahams:
```suggestion
  mutating func layout(of subField: RecordPath, in type: TypeLayout) -> (offset: Int, structure: TypeLayout) {
```

Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -13,9 +13,25 @@ final class InterpreterRunTests: XCTestCase {
...
       """.asSourceFile()
-    let module = try input.loweredToIRAsMainWithHostedStandardLibrary();
-    let program = IR.Program.init(syntax: module.program, modules: [module.id: module]);
-    var executor = Interpreter(program);
+    let module = try input.loweredToIRAsMainWithHostedStandardLibrary()
+    let program = IR.Program.init(syntax: module.program, modules: [module.id: module])
+    var executor = Interpreter(program)
+    while executor.isRunning {
+      try executor.step()
+    }
```
dabrahams, line 18: DRY: this pattern should be factored out.

Tests/InterpreterTests/MemoryTests+internal.swift
```diff
@@ -46,4 +46,23 @@ final class InterpreterMemoryInternalTests: XCTestCase {
     }
 
   }
+
+  func testStoringBuiltinValueInAllocation() {
+    var m = Memory()
```
dabrahams, line 53: Why isn't this testing all the possible builtin types?

Tests/InterpreterTests/MemoryTests+internal.swift
```diff
@@ -46,4 +46,23 @@ final class InterpreterMemoryInternalTests: XCTestCase {
...
+    m[a.allocation].withUnsafePointer(to: UInt8.self, at: 0) {
+      XCTAssertEqual($0.pointee, 2)
+    }
+  }
+
+  func testStoringBuiltinValueUsingAddress() {
+    var m = Memory()
+    var l = TypeLayoutCache(typesIn: TypedProgram.empty, for: UnrealABI())
+    let a = Address(startLocation: m.allocate(1, bytesWithAlignment: 1), type: l[^BuiltinType.i(8)])
+    m.store(BuiltinValue.i8(2), at: a)
```
dabrahams: Why isn't this testing all the builtin types?

Tests/InterpreterTests/TypeLayoutTests.swift
```diff
@@ -132,4 +132,21 @@ final class TypeLayoutTests: XCTestCase {
...
+    ])
+    let a = m.allocate(c[t].size, bytesWithAlignment: c[t].alignment)
+    let x = c.address(of: [1, 1], in: Address(startLocation: a, type: c[t]))
+    XCTAssertEqual(
+      x,
+      Address(
+        startLocation: .init(allocation: a.allocation, offset: 8),
+        type: c[^BuiltinType.i(8)]
+      )
+    )
```
dabrahams, line 149: Do you need to include a union type in this layout?  If you don't think it adds any real testing value, it's fine not to.

Sources/Interpreter/Interpreter.swift
```diff
@@ -293,7 +294,7 @@ public struct Interpreter {
       popStackFrame()
       return
     case let x as Store:
-      memory.store(asBuiltinValue(x.object)!, at: asAddress(x.target)!)
+      memory.store(asBuiltinValue(x.object)!, at: asAddress(x.target)!, layouts: &typeLayout)
```
dabrahams: At the call site it seems `typeLayouts` would have been a better name

Sources/Interpreter/Memory.swift
```diff
@@ -381,8 +381,48 @@ extension Memory.Allocation {
...
+  public func contains(_ t: AnyType, at o: Int, using layouts: inout TypeLayoutCache) -> Bool {
+    if o + baseOffset >= storage.count {
+      return false
+    }
+    let r = composedRegion(containingOffset: o)
+    return contains(t, in: r.type, at: o - r.offset, memoryOffset: r.offset, using: &layouts)
+  }
+
+  /// Returns true iff `p` contains object of type `t` at offset `o` relative
+  /// to position of `p`, given `p` is at offset `a` in memory.
```
dabrahams: Something is wrong here.  As phrased, `a` should not be relevant, and the words “relative to position of `p`” are unnecessary.

Sources/Interpreter/Memory.swift
```diff
@@ -381,8 +381,48 @@ extension Memory.Allocation {
...
+    if l.isUnionLayout {
+      if o == l.discriminator.offset {
+        return t == l.discriminator.type
+      }
+      let d = Int(
+        unsignedIntValue(
+          at: a + l.discriminator.offset,
+          ofType: l.discriminator.type.base as! BuiltinType
+        ))
+      return contains(
```
dabrahams: It would be good to use a loop rather than recursion here.

Sources/Interpreter/Memory.swift
RishabhRD: `a` was required because I needed to get the current discriminator value to get active type for Union case. And for that I need to know exactly where my current type is in `Allocation` so that I can read it.

Sources/Interpreter/Memory.swift
RishabhRD: Yes! changed it to loop. Recursion was just good starting point for thinking. I am still trying to convert that loop into some possible standard algorithm.

dabrahams:
Feel free to merge when you are ready.

Tests/InterpreterTests/TypeLayoutTests.swift
RishabhRD: I think its valuable, but `Union` is broken in old frontend. So, emitter simply panics for any tuple that contains union (that would be interesting case for this method). Seems like we would need new frontend for fixing the state of unions. Maybe I would put a TODO here.

Sources/Interpreter/BuiltinValue.swift
RishabhRD: This was the path I tried in my old "large" PR. But I think we should hold this decision until we tackle the builtin type like `pointer` and `module`. As their handling would give us a complete picture. Thoughts?

Sources/Interpreter/Interpreter.swift
RishabhRD: Agreed. There are instructions that don't produce any result, and thus we can't go for second option. So, I went with introducing function. I would ask to review the name as I feel it can be improved:
```swift
  /// Sets the result of the current instruction to `r`.
  private mutating func setCurrentRegister(to r: InstructionResult) {
    topOfStack.registers[programCounter.instructionInFunction] = r 
  }
```

Sources/Interpreter/BuiltinValue.swift
dabrahams: Like I said, this can wait.

Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -8,19 +9,23 @@ import Utils
 
 final class InterpreterRunTests: XCTestCase {
 
-  func testEmptyMain() throws {
-    let input =
-      """
-        public fun main() { }
-      """.asSourceFile()
-    let module = try input.loweredToIRAsMainWithHostedStandardLibrary()
+  private func run(_ s: SourceFile) throws {
```
dabrahams, line 12: It would be better if `run()` were a method on `SourceFile`.  I would probably put it [here](https://github.com/hylo-lang/hylo/blob/2b3391d299548f4e49a57e22c579dd40dcb80d14/Sources/TestUtils/ProgramInstances.swift#L40-L48) and name it similarly.

Sources/Interpreter/Interpreter.swift
```diff
@@ -223,18 +223,17 @@ public struct Interpreter {
     typeLayouts = .init(typesIn: p.base, for: UnrealABI())
   }
 
-  /// The value of the current instruction's result, if it has been computed.
-  private var currentRegister: InstructionResult? {
-    get { topOfStack.registers[programCounter.instructionInFunction] }
-    set { topOfStack.registers[programCounter.instructionInFunction] = newValue }
+  /// Sets the result of the current instruction to `r`.
+  private mutating func setCurrentRegister(to r: InstructionResult) {
```
dabrahams: Making this read well is difficult.  `setCurrentRegister` implies we are setting the current register to be this register or that register. Definitely the `to:` argument label isn't adding anything.  And I'm not convinced this needs to be named.

I suggest we remove this helper and instead

```swift
  /// Executes a single instruction.
  public mutating func step() throws {
    topOfStack.registers[programCounter.instructionInFunction] = stepResult()
  }
  
  /// Executes a single instruction without recording its result.
  private mutating func stepResult() -> IntructionResult {
    print("\(currentInstruction.site.gnuStandardText): \(currentInstruction)")
    switch currentInstruction {
    case let x as Access:
      return .init(payload: asAddress(x.source)!)
    case let x as AddressToPointer:
      fatalError("\(x)")
    case let x as AdvancedByBytes
     ...
  }
```

WDYT?

Sources/Interpreter/Interpreter.swift
```diff
@@ -189,23 +219,21 @@ public struct Interpreter {
...
+    topOfStack.registers[programCounter.instructionInFunction] = r 
   }
 
   /// Executes a single instruction.
   public mutating func step() throws {
     print("\(currentInstruction.site.gnuStandardText): \(currentInstruction)")
```
dabrahams: is `ghuStandardText` not the default string representation of a `SourceLocation`?

Sources/Interpreter/Interpreter.swift
```diff
@@ -363,6 +392,35 @@ public struct Interpreter {
...
+    case .register(let instruction):
+      topOfStack.registers[instruction]?.payload as? Address
+    case .parameter(_, let i):
+      topOfStack.parameters[i]
+    case .constant:
+      nil
+    }
+  }
+
+  /// Returns the value of `x` if it has a builtin type, or `nil` if it does not.
```
dabrahams: Would it be better for this function to have a precondition and avoid imposing unwrapping on clients?

Sources/Interpreter/Interpreter.swift
```diff
@@ -363,6 +392,35 @@ public struct Interpreter {
     return a!
   }
 
+  /// Returns the value of `x` if it has a address, or `nil` if it does not.
```
dabrahams: Would it be better for this function to have a precondition and avoid imposing unwrapping on clients?

Sources/Interpreter/Interpreter.swift
```diff
@@ -276,9 +304,10 @@ public struct Interpreter {
       popStackFrame()
       return
     case let x as Store:
-      _ = x
+      try memory.store(asBuiltinValue(x.object)!, at: asAddress(x.target)!, with: &typeLayouts)
```
dabrahams: This would read better, IMO.  Not worth doing unless we find the pattern repeating a lot though:

```suggestion
      try memory.store(
        x.object.asBuiltinValue(in: self), at: x.target.asAddress(in: self), 
        with: &typeLayouts)
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -93,6 +109,20 @@ extension Memory {
       "Deallocating using address of the wrong type.")
     try deallocate(a.startLocation)
   }
+
+  /// Stores `v` at `a`.
+  ///
+  /// - Precondition: The memory at `a` is intended to represent `v` according to
+  ///   the type layouts provided by `layouts`.
+  mutating func store(
+    _ v: BuiltinValue, at a: ModuleScope.Address, with layouts: inout TypeLayoutCache
```
dabrahams: `with` is a vacuous argument label when used this way.  Every function operates “with” its arguments.  
 
 Consider whether the type layout cache should be embedded in `Memory` (perhaps not in this PR)—I wonder if that would allow the interpreter to just traffic in types and leave all the layout mapping to `Memory`?  That would be much cleaner.  In the meantime,

```suggestion
    _ v: BuiltinValue, at a: ModuleScope.Address, per layouts: inout TypeLayoutCache
```

This is not great, but at least it is saying that the way we store the value is according to `layouts`.  Open to better suggestions.

Sources/Interpreter/Interpreter.swift
```diff
@@ -101,9 +131,9 @@ struct Stack {
   /// Local variables, parameters, and return addresses.
   private var frames: [StackFrame] = []
 
-  /// Adds a new frame on top with the given `returnAddress`.
-  public mutating func push(returnAddress: CodePointer) {
-    let f = StackFrame(returnAddress: returnAddress)
+  /// Adds a new frame on top with the given `returnAddress` and `parameters`.
+  mutating func push(returnAddress: CodePointer, parameters: [Address]) {
```
dabrahams: Hard for me to imagine why we want this to be internal rather than public or private.  Intentional?

Sources/Interpreter/Memory.swift
```diff
@@ -17,6 +17,7 @@ public struct Memory {
     case deallocationNotAtStartOfAllocation(Address)
     case noLongerAllocated(Address)
     case noDecomposable(TypeLayout, at: Address)
+    case invalidTypeAccess(AnyType, at: Address)
```
dabrahams: "Type access" suggests we are trying to access a type, not some storage _as_ a type.

```suggestion
    case invalidStore(of: AnyType, at: Address)
```

Though also see my remarks on the function that throws this error.

Sources/Interpreter/Memory.swift
```diff
@@ -81,6 +82,12 @@ public struct Memory {
       self.id = id
     }
 
+    /// Allocation for type having type layout `t` and the given `id`.
```
```suggestion
    /// An allocation for `t` with the given `id`.
```

“and” makes  the two parts of the phrase peers so it sounds like `id` applies to the “type”

Sources/Interpreter/Memory.swift
```diff
@@ -282,6 +289,14 @@ public struct Memory {
     return .init(allocation: a, offset: 0)
   }
 
+  /// Allocates memory for type having type layout `t`.
```
```suggestion
  /// Returns new memory for an instance of `t`.
```

It's usually better if you can avoid describing a thing in terms of words in the thing's name.

Sources/Interpreter/Memory.swift
```diff
@@ -357,3 +372,99 @@ public extension Memory.Address {
   static func -=(l: inout Self, r: Int)  { l = l - r }
 
 }
+
+extension Memory.Allocation {
+
+  /// Returns initaialized region containing offset `o`.
```
```suggestion
  /// Returns the composed region containing offset `o`.
```

composed != initialized.  We decompose a thing before we deinitialize it and we compose it after initializing it.

I know it's subtle, but articles (“the,” “an,” etc.) really matter here.  If it was “an” instead of “the” it would imply we might be creating the composed region.

Should we have a precondition check that the result actually does contain `o`?

Sources/Interpreter/Memory.swift
```diff
@@ -357,3 +372,99 @@ public extension Memory.Address {
...
+  func composedRegion(containingOffset o: Int) -> ComposedRegion {
+    precondition(o + baseOffset < storage.count)
+    let i = composedRegions.partitioningIndex { $0.offset > o } - 1
+    return composedRegions[i]
+  }
+
+  /// Returns the immediate subobject of `t` which contains offset `o` in `self`,
```
```suggestion
  /// Returns the immediate subobject of `t` that contains offset `o` in `self`,
```

“which” is only appropriate after a comma.  Subtle rule of English; it's because of meaning, not punctuation, but the comma is a good guide as to which meaning you want.

Sources/Interpreter/Memory.swift
```diff
@@ -357,3 +372,99 @@ public extension Memory.Address {
...
+      unsignedIntValue(
+        at: i + l.discriminator.offset,
+        ofType: l.discriminator.type.base as! BuiltinType
+      )
+    )
+    let part = l.parts[d]
+    return (part.type, i + part.offset)
+  }
+
+  /// Returns true iff `self` contains object of type `t` starting at offset `o`,
```
```suggestion
  /// Returns true iff `self` contains  an object of type `t` starting at offset `o`,
```

Sources/Interpreter/Memory.swift
```diff
@@ -357,3 +372,99 @@ public extension Memory.Address {
...
+      else { return false }
+      p = n.type
+      i = n.offset
+    }
+  }
+
+  /// Stores `x` at `o`.
+  ///
+  /// - Precondition: The memory at `o` is intended to represent `v` according to
+  ///   the type layouts provided by `layouts`.
```
dabrahams: “Intention” is not something you can measure or restrict with a precondition.  Figure out what the real precondition is.

Sources/Interpreter/Memory.swift
```diff
@@ -357,3 +372,99 @@ public extension Memory.Address {
   static func -=(l: inout Self, r: Int)  { l = l - r }
 
 }
+
+extension Memory.Allocation {
+
+  /// Returns initaialized region containing offset `o`.
+  func composedRegion(containingOffset o: Int) -> ComposedRegion {
```
```suggestion
  /// Returns the initaialized region containing offset `o`.
  func initializedRegion(containingOffset o: Int) -> ComposedRegion {
```

or

```suggestion
  /// Returns the composed region containing offset `o`.
  func composedRegion(containingOffset o: Int) -> ComposedRegion {
```

Composition is not exactly the same as initialization.  Actually I'm not 100% clear on the distinction now, though I invented it only a couple months ago.  Are you?  We might need to discuss with @kyouko-taiga to make sure we know what we're saying.

I know it's subtle, but articles (“the,” “an,” etc.) really matter here.  If it was “an” instead of “the” it would imply we might be creating the composed region.

Should we have a precondition check that the result actually does contain `o`?

dabrahams (commented):
More commentary

Tests/InterpreterTests/TypeLayoutTests.swift
dabrahams: Unions are replaced with enums in the new FE, FWIW.

Sources/Interpreter/Interpreter.swift
RishabhRD: > Consider whether the type layout cache should be embedded in Memory

Yes, I got the same thought while writing this. And then all these preconditions on `layouts` would become invariant of memory.

However,  I guess even after having `TypeLayoutCache` in memory, we would still need to pass `layouts` to some methods of `Memory.Allocation` as cache is kind-of shared between all allocations. (Other than, every allocation can have a pointer to the type layout cache of memory).

Sources/Interpreter/Interpreter.swift
RishabhRD: Merging mistake :smiling_face_with_tear:

Sources/Interpreter/Memory.swift
RishabhRD: IR suggests, they are not the same (still, we need to discuss with Dimi for surety).

> If it was “an” instead of “the” it would imply we might be creating the composed region.

I am not sure what you mean by "creating the composed region" though. IR shows `alloc_stack` would always allocate the "root" type and thus that would be the only composed region according to my understanding. So, in fact currently we always have a single composed region for every allocation (until we do explicit decomposition).
Or I am misunderstanding the composed region thing at all.

Sources/Interpreter/Interpreter.swift
RishabhRD: `stepResult` should return `(InstructionResult?, CodePointer)` as we also need to update program counter in `step` method. I am not loving the line noise it would cause on every case block. However, we can skip some of it, if we handle "jump" instructions(for which noise is necessary) separately than other instructions. Thoughts?

Sources/Interpreter/Interpreter.swift
dabrahams: > I guess even after having `TypeLayoutCache` in memory, we would still need to pass layouts to some methods of `Memory.Allocation` as cache is kind-of shared between all allocations.

If the change would nonetheless create a substantial simplification, we should do it, especially if those methods of `Memory.Allocation` are only called from a few places (and _especially_ especially if they are only called from `Memory`).

On the other hand I have no problem with using `SharedMutable` for the type layouts if that's where we are heading.

Sources/Interpreter/Interpreter.swift
dabrahams: Works for me.  The line noise you are referring to is the construction of this tuple?  I'm not sure we need to do that; we could say instruction results are either new PC values or register values.  We could even say that they are always register values but when they have some particular type/enum case the PC is set to that value in `step()` and otherwise advanced to the next instruction.  I leave it to you which way to sort this out; those are some ideas.  I suggest looking at how each turns out (incuding the status quo) and picking whatever you think is best.

Sources/Interpreter/Memory.swift
dabrahams: I _think_ you're misunderstanding.  I wrote up [this discussion](https://github.com/orgs/hylo-lang/discussions/1821) to explain.  I believe I also recovered the distinction between initialization and composition.  HTH

RishabhRD:
Hi @dabrahams, I have tried to resolve the comments that were there in scope of PR.
Some notes:
- Removed invalid type store validation from `store` instruction handling for now as we need `compose/decompose` instruction for the same. Opened #1822 to track the same. I did put the code in different branch, to avoid repetitive work in future.
- Regarding `TypeLayoutCache` to be part of memory, opening a separate PR for the same.

Sources/Interpreter/Interpreter.swift
```diff
@@ -18,12 +18,25 @@ struct CodePointer {
 
 }
 
-/// The value produced by executing an instruction.
-struct InstructionResult {
-  /// The instruction's output as an opaque, type-erased value.
+/// Represents value at IR level.
+struct Value {
```
RishabhRD, line 22: The name comes from [this](https://github.com/hylo-lang/hylo/pull/1780#discussion_r2617645456) suggestion.

Sources/Interpreter/Interpreter.swift
```diff
@@ -18,12 +18,25 @@ struct CodePointer {
 
 }
 
-/// The value produced by executing an instruction.
-struct InstructionResult {
-  /// The instruction's output as an opaque, type-erased value.
+/// Represents value at IR level.
```
```suggestion
/// A value manipulated by the IR.
```

Guideline: never say “represents X;” it's just “X.”  The API guidelines say to document what a type _is_, not what it represents.

dabrahams (approved):
@RishabhRD you should land this PR ASAP.  There are only a couple of outstanding comments.  I don't need to review it again.

RishabhRD:
Merging the PR. I believe all comments are resolved except the ones which deserve their own PR:
1. `Memory` should own `typeLayouts`. (already opened)
2. `asAddress` should be like a method on `Operand`. (need to evaluate if needed).

PR #1782, opened 2025-11-23
Interpreter: support memory_copy instruction

RishabhRD:
The PR adds capability to copy builtin values using memory_copy instruction.

### DIFF
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -246,9 +246,7 @@ public struct Interpreter {
       // No effect on program state
       return nil
     case let x as MemoryCopy:
-      let s = asPlace(x.source)
-      let t = asPlace(x.target)
-      try memory.copy(s, to: t)
+      try memory.copy(asPlace(x.source), to: asPlace(x.target))
       return nil
     case is Move:
       fatalError("Interpreter: Move instructions have not been removed.")
--- a/Sources/Interpreter/Memory.swift
+++ b/Sources/Interpreter/Memory.swift
@@ -442,19 +442,41 @@ extension Memory {
     try self[target.allocation].store(v, at: target.offset)
   }
 
-  /// Copies bytes of `source` to `destination`.
+  /// Returns the result of calling `body` with raw buffer pointer to bytes in `p`.
+  ///
+  /// - Precondition: `typeLayouts[p.type] == l`.
+  private func withUnsafeBytes<R>(
+    _ p: Place, havingLayout l: TypeLayout, _ body: (UnsafeRawBufferPointer) -> R
+  ) -> R {
+    let o = self[p.allocation].baseOffset
+    return self[p.allocation].storage.withUnsafeBytes {
+      let s = $0.baseAddress!.advanced(by: o + p.offset)
+      return body(UnsafeRawBufferPointer(start: s, count: l.size))
+    }
+  }
+
+  /// Returns the result of calling `body` with mutable raw buffer pointer to bytes in `p`.
+  private mutating func withUnsafeMutableBytes<R>(
+    _ p: Place, _ body: (UnsafeMutableRawBufferPointer) -> R
+  ) -> R {
+    let o = self[p.allocation].baseOffset
+    let n = typeLayouts[p.type].size
+    return self[p.allocation].storage.withUnsafeMutableBytes {
+      let s = $0.baseAddress!.advanced(by: o + p.offset)
+      return body(UnsafeMutableRawBufferPointer(start: s, count: n))
+    }
+  }
+
+  /// Copies the bytes of `source` to `destination`.
   public mutating func copy(_ source: Place, to destination: Place) throws {
     precondition(
-      source.type == destination.type, "source and destination should have same type for copying.")
-    // TODO: throw when source is not in composed regions.
-    let n = typeLayouts[source.type].size
-    let i = self[source.allocation].baseOffset
-    let j = self[destination.allocation].baseOffset
-    self[source.allocation].storage.withUnsafeBytes { a in
-      self[destination.allocation].storage.withUnsafeMutableBytes { b in
-        let s = a.baseAddress! + i + source.offset
-        let d = b.baseAddress! + j + destination.offset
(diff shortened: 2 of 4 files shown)


Sources/Interpreter/Interpreter.swift
```diff
@@ -177,7 +177,13 @@ public struct Interpreter {
       // No effect on program state
       break
     case let x as MemoryCopy:
-      _ = x
+      let sourceAddress = address(x.source)!
+      let destinationAddress = address(x.target)!
```
dabrahams: Please take type information out of names of non types

Sources/Interpreter/Interpreter.swift
```diff
@@ -186,7 +186,13 @@ public struct Interpreter {
       // No effect on program state
       break
     case let x as MemoryCopy:
-      _ = x
+      let s = address(x.source)!
+      let d = address(x.target)!
```
dabrahams: Why `d` and not `t`?

Sources/Interpreter/Memory.swift
```diff
@@ -331,6 +331,23 @@ public struct Memory {
     }
   }
 
+  /// Copies `n` bytes from `source` to `destination`.
+  public mutating func copy(
+    byteCount n: Int, from source: Memory.Address, to destination: Memory.Address
+  ) {
+    precondition(source.offset + n <= self[source.allocation].size)
```
dabrahams: These should be throws, I think.  Anything where unsafe hylo code (and not merely a bug in codegen) could cause UB needs to be checked for and an error thrown.

Also, I think this function needs to take a type layout rather than a size and check that the type is composed at the source address.

Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -37,4 +37,20 @@ final class InterpreterRunTests: XCTestCase {
...
+        }
+      """.asSourceFile()
+    let module = try input.loweredToIRAsMainWithHostedStandardLibrary();
+    let program = IR.Program.init(syntax: module.program, modules: [module.id: module]);
+    var executor = Interpreter(program);
+    while executor.isRunning {
+      try executor.step()
+    }
+  }
+
```
dabrahams, line 41: Naturally, you'll need tests for the throwing cases.

RishabhRD:
The PR now uses `Place` abstraction we introduced a little back. Regarding throwing case for composed regions, I would be adding them in the separate PR itself as we need the same for `store` instruction. I have put them to TODO for now. Due to these changes, requesting for a review.

Sources/Interpreter/Interpreter.swift
```diff
@@ -246,7 +246,10 @@ public struct Interpreter {
       // No effect on program state
       return nil
     case let x as MemoryCopy:
-      _ = x
+      let s = asPlace(x.source)
+      let t = asPlace(x.target)
+      try memory.copy(s, to: t)
```
```suggestion
      try memory.copy(asPlace(x.source), to: asPlace(x.target))
```

I don't think these single-use bindings are helping

Sources/Interpreter/Memory.swift
```diff
@@ -441,5 +441,22 @@ extension Memory {
   mutating func store(_ v: BuiltinValue, in target: Place) throws {
     try self[target.allocation].store(v, at: target.offset)
   }
+
+  /// Copies bytes of `source` to `destination`.
```
```suggestion
  /// Copies the bytes of `source` to `destination`.
```

Sources/Interpreter/Memory.swift
```diff
@@ -441,5 +441,22 @@ extension Memory {
...
+  public mutating func copy(_ source: Place, to destination: Place) throws {
+    precondition(
+      source.type == destination.type, "source and destination should have same type for copying.")
```
```suggestion
      source.type == destination.type, 
      "Copy source type \(source.type) must match destination type \(destination.type).")
```

Sources/Interpreter/Memory.swift
```diff
@@ -441,5 +441,22 @@ extension Memory {
...
+    // TODO: throw when source is not in composed regions.
```
```suggestion
    // TODO: throw when source is not in composed regions and mark destination region composed.
```

Sources/Interpreter/Memory.swift
```diff
@@ -441,5 +441,22 @@ extension Memory {
...
+    let n = typeLayouts[source.type].size
+    let i = self[source.allocation].baseOffset
+    let j = self[destination.allocation].baseOffset
+    self[source.allocation].storage.withUnsafeBytes { a in
+      self[destination.allocation].storage.withUnsafeMutableBytes { b in
+        let s = a.baseAddress! + i + source.offset
+        let d = b.baseAddress! + j + destination.offset
+        d.copyMemory(from: s, byteCount: n)
+      }
+    }
```
```suggestion
    self.withUnsafeBytes(source) { a in
      self.withUnsafeMutableBytes(destination) { b in
        a.copyElements(from: b)
      }
    }
```

Can we not make something like the above work?
The `copyElements` algorithm is embarrassingly missing from the standard library but you could write it.

Tests/InterpreterTests/MemoryTests+internal.swift
```diff
@@ -66,4 +66,35 @@ final class InterpreterMemoryInternalTests: XCTestCase {
     try assertStoring(BuiltinValue.i64(8), asType: BuiltinType.i(64), yields: UInt64(8))
     try assertStoring(BuiltinValue.i128(8), asType: BuiltinType.i(128), yields: UInt128(8))
   }
+
+  func assertCopying<T: Equatable>(_ v: BuiltinValue, asType t: BuiltinType, yields e: T) throws {
```
dabrahams: I try use “check” instead of assert because even though `XCTAssert` exists, `assert` has a precedented meaning in Swift. When you glom `Copying` onto `assert` it gives the impression that it's somehow a “copying assert.”


```suggestion
  func check<T: Equatable>(copying v: BuiltinValue, asType t: BuiltinType, yields e: T) throws {
```

or

```suggestion
  func checkThat<T: Equatable>(copying v: BuiltinValue, asType t: BuiltinType, yields e: T) throws {
```

dabrahams (commented):
Nice; A few notes.  Please make whatever changes you deem appropriate and land the PR w/o re-review.

PR #1783, opened 2025-11-23
Interpreter: support executing if-else statements

RishabhRD:
The PR adds support for minimal instructions so that if/else statements can be executed.

### DIFF
--- a/Sources/Interpreter/BuiltinValue.swift
+++ b/Sources/Interpreter/BuiltinValue.swift
@@ -79,8 +79,8 @@ extension BuiltinValue {
   /// Boolean value, if present.
   var bool: Bool? { i1 }
 
-  /// Creates instance of builtin value with `c`.
-  public init(withIntegerConstant c: IntegerConstant) {
+  /// An instance with value `c`.
+  public init(_ c: IntegerConstant) {
     self =
       switch c.value.bitWidth {
       case 1: .i1(Bool(c.value != 0))
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -2,10 +2,15 @@ import Foundation
 import FrontEnd
 import Collections
 import IR
+import Utils
 
+/// The position of an instruction in the program.
 struct CodePointer {
 
+  /// The module containing `self`.
   var module: Module.ID
+
+  /// The position relative to `module` indicated by `self`.
   var instructionInModule: InstructionID
 
 }
@@ -14,7 +19,7 @@ struct CodePointer {
 typealias InstructionResult = Any
 
 /// A typed location in memory.
-struct Address {
+struct Address: Regular {
 
   /// The position in memory.
   public let startLocation: Memory.Address
@@ -35,7 +40,7 @@ struct StackFrame {
   var returnAddress: CodePointer
 
   /// The allocations in this stack frame.
-  var allocations: [Memory.Allocation.ID] = []
+  var allocations: [Address] = []
 
   /// Function parameters
   var parameters: [Address]
@@ -82,12 +87,16 @@ public struct Interpreter {
   /// True iff the program is still running.
   public private(set) var isRunning: Bool = true
 
+  /// Text written so far to the process' standard output stream.
   public private(set) var standardOutput: String = ""
 
+  /// Text written so far to the process' standard error stream.
   public private(set) var standardError: String = ""
 
+  /// The type layouts that have been computed so far.
   private var typeLayout: TypeLayoutCache
 
+  /// The top stack frame.
   private var topOfStack: StackFrame {
     get { callStack.last! }
     _modify {
@@ -136,7 +145,7 @@ public struct Interpreter {
 
     case let x as AllocStack:
       let a = allocate(typeLayout[x.allocatedType])
-      topOfStack.allocations.append(a.startLocation.allocation)
+      topOfStack.allocations.append(a)
       currentRegister = a
 
     case let x as Branch:
@@ -168,8 +177,7 @@ public struct Interpreter {
       currentRegister = x.value
     case let x as DeallocStack:
       let a = addressProduced(by: x.location.instruction!)!
-      try deallocate(a)
-      topOfStack.allocations.removeLast()
+      try deallocateStack(a)
     case is EndAccess:
       // No effect on program state
       break
@@ -289,6 +297,15 @@ public struct Interpreter {
     try memory.deallocate(a.startLocation)
   }
 
+  /// Deallocates `a` allocated on stack.
+  mutating func deallocateStack(_ a: Address) throws {
+    precondition(
+      a == topOfStack.allocations.last!,
+      "The latest allocation that has not been deallocated must be deallocated first.")
+    try deallocate(a)
+    topOfStack.allocations.removeLast()
+  }
+
   /// Returns the address produced by executing instruction identified by `i` in the current frame,
   /// or `nil` if it didn't produce an address.
   func addressProduced(by i: InstructionID) -> Address? {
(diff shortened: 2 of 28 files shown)


Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -53,4 +53,31 @@ final class InterpreterRunTests: XCTestCase {
...
+          }
+        }
+      """.asSourceFile()
+    let module = try input.loweredToIRAsMainWithHostedStandardLibrary();
+    let program = IR.Program.init(syntax: module.program, modules: [module.id: module]);
+    var executor = Interpreter(program);
+    while executor.isRunning {
+      try executor.step()
+    }
+  }
```
tothambrus11, line 81: Could we add some assertions here? Like checking the value of notOfCondition. This could also be split into 2 test cases or at least 2 independent sets of variables within this test case (just to save time). 

It would be also useful to add a timeout facility for running the interpreter and fail the test if execution takes longer than 1 second.

RishabhRD: Regarding assertions, I am not sure if I can support it with current state of interpreter. Currently I don't want to depend on order of instructions being emitted. That can be changed with changes in components like optimizers.
That's the reason I want to support `print` method ASAP, so that we can inspect the interpreter's standard output for testing purposes after execution of program.

tothambrus11: I see, it wouldn't make sense to look at the values of the variables after completion of the program because by then, their  lifetimes are over. Then indeed, some sort of print or capture functionality would be nice. Or we can implement some test assertions as builtin Hylo functions, so that we can write assertions directly within Hylo code, similarly to `precondition`. Printing would be more robust, because the Hylo based assertions may never get reached if the interpreter is buggy.

Sources/Interpreter/BuiltinValue.swift
```diff
@@ -79,16 +79,16 @@ extension BuiltinValue {
   /// Boolean value, if present.
   var bool: Bool? { i1 }
 
-  /// Creates instance of builtin value with integer constant `c`.
+  /// Creates instance of builtin value with `c`.
   public init(withIntegerConstant c: IntegerConstant) {
```
dabrahams: Why the elaborate argument label?  Isn't this a value-preserving conversion?  If so there should be no label.

Sources/Interpreter/BuiltinValue.swift
```diff
@@ -79,7 +79,7 @@ extension BuiltinValue {
   /// Boolean value, if present.
   var bool: Bool? { i1 }
 
-  /// Creates instance of builtin value with integer constant `c`.
+  /// Creates instance of builtin value with `c`.
```
dabrahams:
```suggestion
  /// An instance with value `c`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -157,8 +157,8 @@ public struct Interpreter {
     case let x as CloseUnion:
       _ = x
     case let x as CondBranch:
-      let condition = builtinValue(x.condition)!
-      if condition.bool! {
+      let c = builtinValue(x.condition)!
+      if c.bool! {
```
dabrahams, line 23: Why create a `BuiltinValue` only to immediately unwrap and discard it?

Sources/Interpreter/Interpreter.swift
```diff
@@ -338,14 +338,14 @@ public struct Interpreter {
...
-    precondition(a.memoryLayout.type.isBuiltin)
-    let allocation = a.memoryAddress.allocation
-    let offset = a.memoryAddress.offset
-    let typeOfValue = a.memoryLayout.type.base as! BuiltinType
-    return memory[allocation].builtinValue(at: offset, ofType: typeOfValue)
+    precondition(a.type.type.isBuiltin)
+    let i = a.startLocation.allocation
+    let o = a.startLocation.offset
+    let t = a.type.type.base as! BuiltinType
+    return memory[i].builtinValue(at: o, ofType: t)
```
dabrahams: I know what you've done here comes from my suggestion, but maybe it would be better to call what is currently `a.type` `a.structure` or simply `a.layout` (sometimes resisting all the type information is counterproductive) and expose `a.type.type` as a computed `a.type` property.

```suggestion
    precondition(a.type.isBuiltin)
    let i = a.startLocation.allocation
    let o = a.startLocation.offset
    let t = a.type.base as! BuiltinType
    return memory[i].builtinValue(at: o, ofType: t)
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -354,16 +354,16 @@ public struct Interpreter {
...
   /// Sets the program counter to the start of `b`.
   mutating func jump(to b: Block.ID) {
-    let function = b.function
-    let module = program.module(defining: function)
-    let entryBlock = b.address;
-    let entryInstruction = program.modules[module]![b.function]
-      .blocks[entryBlock]
+    let f = b.function
+    let m = program.module(defining: f)
+    let a = b.address;
```
dabrahams, line 101: Maybe shouldn't affect this PR, but I wish we could talk about the positions of instructions as something other than `address` (and especially not as anything involving “pointer’).

```suggestion
    let p = b.start
```

Also, please go through your PRs and nix all the semicolons ;-)

Sources/Interpreter/Interpreter.swift
```diff
@@ -354,16 +354,16 @@ public struct Interpreter {
 
   /// Sets the program counter to the start of `b`.
   mutating func jump(to b: Block.ID) {
-    let function = b.function
-    let module = program.module(defining: function)
-    let entryBlock = b.address;
-    let entryInstruction = program.modules[module]![b.function]
-      .blocks[entryBlock]
+    let f = b.function
+    let m = program.module(defining: f)
```
dabrahams: Maybe shouldn't affect this PR, but `defining` seems heavy compared with `of` which would read perfectly well

Sources/Interpreter/Memory.swift
```diff
@@ -329,10 +329,8 @@ public extension Memory.Address {
 
 extension Memory.Allocation {
 
-  /// Returns builtin value stored at offset `o` having type `t`.
-  func builtinValue(at o: Memory.Offset, ofType t: BuiltinType)
-    -> BuiltinValue
-  {
+  /// Returns builtin value stored at `o` having type `t`.
+  func builtinValue(at o: Memory.Offset, ofType t: BuiltinType) -> BuiltinValue {
```
dabrahams: Try not to forget “the.”  I know, it's an english thing.

This is more natural (also an english thing):

```suggestion
  /// Returns the value of type `t` at `o`.
  func value(ofType t: BuiltinType, at o: Memory.Offset) -> BuiltinValue {
```

Don't need to say “builtin value” since that's in the return type and `t`'s parameter type.

But if we have `store` below shouldn't this be `load`?

```suggestion
  /// Returns the value of type `t` at `o`.
  func load(_ t: BuiltinType, at o: Memory.Offset) -> BuiltinValue {
```

dabrahams (approved):
some comments to address

Sources/Interpreter/BuiltinValue.swift
RishabhRD: I see this is a comment on past state of code and was fixed few weeks back itself. I assume, somehow github saved these comments from some past  un-submitted review and thus is appearing now.

dabrahams: My question is not about the coment but about the argument label `withIntegerConstant`

RishabhRD:
There has been a lot of design change in Interpreter and backend from the time of the PR being opened. Porting the PR to new design is same re-writing the whole thing, thus closing this PR and opening a new one for same functionality that respects the new design.

PR #1784, opened 2025-11-23
Interpreter: support calling named functions

RishabhRD:
Add ability to calling named functions that does **NOT** require any sort of function pointer support.

### DIFF
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -2,10 +2,15 @@ import Foundation
 import FrontEnd
 import Collections
 import IR
+import Utils
 
+/// The position of an instruction in the program.
 struct CodePointer {
 
+  /// The module containing `self`.
   var module: Module.ID
+
+  /// The position relative to `module` indicated by `self`.
   var instructionInModule: InstructionID
 
 }
@@ -14,7 +19,7 @@ struct CodePointer {
 typealias InstructionResult = Any
 
 /// A typed location in memory.
-struct Address {
+struct Address: Regular {
 
   /// The position in memory.
   public let startLocation: Memory.Address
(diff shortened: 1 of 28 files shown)


Sources/Interpreter/Interpreter.swift
```diff
@@ -152,7 +152,11 @@ public struct Interpreter {
       jump(to: x.target)
       return
     case let x as Call:
-      _ = x
+      let p = x.arguments.map { address($0)! } + [address(x.output)!]
+      let r = try nextCodePointer()
+      callStack.append(.init(returnAddress: r, parameters: p))
+      jump(to: (x.callee.constant as! FunctionReference).function)
+      return
```
dabrahams, line 159: Experimenting with naming; I'm not committed to any of these changes, except for that I really want to see `callStack.push`.

```suggestion
      let calleeIO = x.arguments.lazy.map { address($0)! } + [address(x.output)!]
      try callStack.push(
        StackFrame(returnAddress: nextCodePointer(), parametersAndResult: calleeIO))
      jump(to: (x.callee.constant as! FunctionReference).function)
      return
```

Why does `nextCodePointer()` throw?  Seems like any UB would have been caught by the time the current program counter was set.

Sources/Interpreter/Interpreter.swift
```diff
@@ -259,16 +263,29 @@ public struct Interpreter {
...
-    programCounter.instructionInModule = InstructionID(
+    return InstructionID(
       programCounter.instructionInModule.function,
       programCounter.instructionInModule.block,
-      a)
+      a
+    )
+  }
+
+  /// Returns code pointer pointing to next instruction.
```
dabrahams, line 284: What does a `CodePointer` add to an `InstructionID`?

dabrahams (commented):
I'm not sure we should be throwing from a lot of the places that can throw.

PR #1785, opened 2025-11-23
Move the `makeXXX` functions from `Module` to `Function`.

lucteo:
This PR is an alternative to #1773 

This allow localizing IR changes to Function, if we don't need anything
else from the Module.

The `makeXXX` functions don't require mutation on `Function`.

The following places required very small refactoring:
- `makeProjectbundle` required inout access to the `module`
- `makeCallBundle` required inout access to the `module`
- `makeSubfieldView` required access to the program
- `makeGlobalAddr` required access to the program


lucteo:
Abandoning.

PR #1786, opened 2025-11-24
Instructions are stored in `Function`

lucteo:
Currently, instructions are stored in blocks. Instead, we want them to be stored in `Function`.
This better aligns with the new frontend.

Other changes:
- external code (outside Function) should not know how instructions are kepy
- merging `appendBlock` and `appendEntry`

### DIFF
--- a/Sources/IR/Emitter.swift
+++ b/Sources/IR/Emitter.swift
@@ -116,7 +116,7 @@ struct Emitter: Sendable {
 
   /// Appends a new basic block at the end of `self.insertionFunction`, defined in s.
   private mutating func appendBlock<T: ScopeID>(in s: T) -> Block.ID {
-    module[insertionFunction!].append(in: s)
+    module[insertionFunction!].appendBlock(in: s)
   }
 
   /// Appends a new basic block at the end of `self.insertionFunction`, defined in the same scope
@@ -232,7 +232,7 @@ struct Emitter: Sendable {
     insertionFunction = f
 
     // Configure the emitter context.
-    let entry = module[f].append(in: program.scopeContainingBody(of: d)!)
+    let entry = module[f].appendBlock(in: program.scopeContainingBody(of: d)!)
     let bodyFrame = outermostFrame(of: d, entering: entry)
     self.insertionPoint = .end(of: entry)
 
@@ -296,7 +296,7 @@ struct Emitter: Sendable {
     insertionFunction = f
 
     // Configure the emitter context.
-    let entry = module[f].append(in: d)
+    let entry = module[f].appendBlock(in: d)
 
     self.insertionPoint = .end(of: entry)
     self.frames.push()
@@ -346,7 +346,7 @@ struct Emitter: Sendable {
     insertionFunction = f
 
     // Create the function entry.
-    let entry = module[f].append(in: ast[d].body!)
+    let entry = module[f].appendBlock(in: ast[d].body!)
 
     // Configure the locals.
     var locals = DeclProperty<Operand>()
@@ -388,7 +388,7 @@ struct Emitter: Sendable {
     insertionFunction = f
 
     // Create the function entry.
-    let entry = module[f].append(in: program.scopeContainingBody(of: d)!)
+    let entry = module[f].appendBlock(in: program.scopeContainingBody(of: d)!)
 
     // Configure the locals.
     var parameters = DeclProperty<Operand>()
@@ -430,7 +430,7 @@ struct Emitter: Sendable {
     insertionFunction = f
 
     // Create the function entry.
-    let entry = module[f].append(in: program.scopeContainingBody(of: d)!)
+    let entry = module[f].appendBlock(in: program.scopeContainingBody(of: d)!)
 
     // Configure the locals.
     var locals = DeclProperty<Operand>()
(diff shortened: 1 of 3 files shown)


Sources/IR/Analysis/Lifetime.swift
```diff
@@ -156,11 +156,11 @@ extension Module {
   /// - Requires: The definition of `l` dominates `u`.
   func extend(lifetime l: Lifetime, toInclude u: Use, in f: Function.ID) -> Lifetime {
     var coverage = l.coverage
-    switch coverage[u.user.block] {
+    switch coverage[functions[f]!.block(of: u.user).address] {
```
kyouko-taiga: Perhaps the expression should be stored in a variable before the switch. My instinct tells me that I should be on the hunt for mutation when code repeats long expressions but in this case it seems no mutation occurs.

Sources/IR/Analysis/Lifetime.swift
```diff
@@ -210,6 +210,8 @@ extension Module {
   }
 
   /// Returns the use that executes last.
+  ///
+  /// If the two uses are not in the same block, this returns `lhs`.
```
kyouko-taiga:
```suggestion
  /// Returns the use that executes last iff both `lhs` and `rhs` are in the same block. Otherwise, returns `lhs`.
```

Sources/IR/Analysis/Module+CloseBorrows.swift
```diff
@@ -8,7 +8,7 @@ extension Module {
   ///
   /// - Requires: `self` is has gone through access reification. `f` is in `self`.
   public mutating func closeBorrows(in f: Function.ID, diagnostics: inout DiagnosticSet) {
-    for i in self[f].instructions {
+    for i in self[f].instructionIDs {
```
kyouko-taiga: We use `ID` in many places but in the new front-end I prefer `Identity`. I'll let you consider if which convention you prefer for this PR.

lucteo: Do you mean we should write `instructionIdentities`? to me that sounds a bit too long. (just using `identities` instead of `instructionIDs` feels confusing to me). But I'm fine either way.

dabrahams: @kyouko-taiga Why change this convention?  `ID` is very much a standard initialism and thus one of the few appropriate abbreviations.

dabrahams: @lucteo I think this is happening because you've added an `instructions` property?  Consider whether that should be named something like `instructionStorage` or something.  If the primary way the instructions are directly addressed is through the `instructionIDs` variable it could be appropriate.  Look at the use-sites and optimize for logical readability.

lucteo: I think that makes a lot of sense. Will do.

Sources/IR/Emitter.swift
```diff
@@ -114,14 +115,14 @@ struct Emitter: Sendable {
   }
 
   /// Appends a new basic block at the end of `self.insertionFunction`, defined in s.
-  private mutating func appendBlock<T: ScopeID>(in s: T) -> Block.ID {
+  private mutating func append<T: ScopeID>(in s: T) -> Block.ID {
```
kyouko-taiga: I find it a bit weird to remove the `Block` suffix. At first I tried to convince myself that it was fine because functions are essentially collections of basic blocks but then I realized this is a method of `Emitter`. It doesn't mean much to me to just "append" some unspecified thing in the middle of `Emitter`.

I think the suffix should stay.

Sources/IR/Function.swift
```diff
@@ -28,6 +31,12 @@ public struct Function: Sendable {
   /// The basic blocks of the control flow graph.
   public var blocks: Blocks
 
+  /// The instructions in the function.
+  public var instructionStorage: Instructions = []
```
kyouko-taiga: This property should be called `instructions`. That is weird otherwise and inconsistent with the rest of the code base.

Sources/IR/Function.swift
```diff
@@ -87,14 +96,14 @@ public struct Function: Sendable {
 
   /// The IDs of all instructions.
   public var instructions: some Collection<InstructionID> {
```
kyouko-taiga: I would call that `instructionIdentities`. The tradeoff is one long property name here or the inconsistent naming of `instructionStorage`. I prefer the long name.

I also think that is somewhat consistent with the way we can get the indices of an array, for example (i.e., `things.indices`).

kyouko-taiga (changes requested):
I am not a fan of the naming convention. See inline comments for my rationale.

Sources/IR/Analysis/Module+CloseBorrows.swift
lucteo: Applied Dimi's suggestion.

Sources/IR/Emitter.swift
```diff
@@ -231,7 +232,7 @@ struct Emitter: Sendable {
     insertionFunction = f
 
     // Configure the emitter context.
-    let entry = module[f].appendEntry(in: program.scopeContainingBody(of: d)!)
+    let entry = module[f].append(in: program.scopeContainingBody(of: d)!)
```
kyouko-taiga: IIUC you changed `appendBlock`. Then this one should be changed back too, for the same reason (i.e., it doesn't make much sense to just append some unspecified thing here).

I'm fine calling it `appendBlock` if you want to avoid two different names. I don't feel very strongly about this specific change.

lucteo: I will name it `appendBlock` for consistency.

History: In a previous patch, Dave suggested to change `appendBlock` to just `append` (https://github.com/hylo-lang/hylo/pull/1720/commits/5a5bbcf27a89008ff2b2623f70348f19ec6a4e40). At that point there was a TODO to unify the old `appendBlock` with `appendEntry`. Fixing this TODO was part of this PR.

kyouko-taiga (approved):

PR #1788, opened 2025-11-25
Interpreter

tothambrus11:


Sources/Driver/Driver.swift
```diff
@@ -188,11 +207,11 @@ public struct Driver: ParsableCommand, Sendable {
 
     // Execute the command.
     var log = DiagnosticSet()
-    try log.capturingErrors { (ds) in
+    let ir = try log.capturingErrors { (ds) in
       try cli.executeCommand(extending: baseProgram, reportingDiagnosticsTo: &ds)
     }
     let status = log.containsError ? ExitCode.failure : ExitCode.success
-    return .init(status: status, output: output, diagnostics: log)
+    return .init(status: status, output: output, diagnostics: log, ir: ir ?? nil)
```
tothambrus11, line 214: `?? nil` looks redundant.

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,342 @@
...
+  }
+
+}
+
+extension UnsafeRawBufferPointer {
+
+  /// Returns the number of bytes from the notional base address to
+  /// the nearest address aligned to `a`.
+  ///
+  /// If `self.baseAddress == 0`, returns `0`.
```
tothambrus11, line 86: Why do we have this default value here instead of propagating the optional?

Sources/Interpreter/TargetABI.swift
```diff
@@ -0,0 +1,57 @@
...
+  func layout(_ t: BuiltinType) ->TypeLayout.Bytes
+
+}
+
+public struct UnrealABI: TargetABI {
+
+  public init() {}
+
+  let bitsInAWord = 64
+  let maxAlignment = 128 / 8
```
tothambrus11: Is this in bytes or bits?

Sources/Driver/Driver.swift
dabrahams: It's not; try compiling without it.

Sources/Interpreter/Interpreter.swift
dabrahams: Because null pointers are real addresses maximally aligned and suitable for zero-length arrays and empty objects.

Sources/Interpreter/TargetABI.swift
dabrahams: Bytes, just like every other `alignment` value in the system.

RishabhRD:
Merging the PR! as we can't get 100% coverage for interpreter at this point. Also my PRs would now be rebased to main.

PR #1793, opened 2025-11-29
Improve `DominatorTree` and `ControlFlowGraph`

lucteo:
`DominatorTree` shall not depend on module anymore.
Use `Block.ID` instead of addresses in `DominatorTree` and `ControlFlowGraph`.

### DIFF
--- a/Sources/IR/Analysis/AbstractInterpreter.swift
+++ b/Sources/IR/Analysis/AbstractInterpreter.swift
@@ -100,7 +100,7 @@ struct AbstractInterpreter<Domain: AbstractDomain>: Sendable {
         changed = true
       }
 
-      if !changed && (sources.count == cfg.predecessors(of: blockToProcess.address).count) {
+      if !changed && (sources.count == cfg.predecessors(of: blockToProcess).count) {
         done.insert(blockToProcess)
       } else {
         work.append(blockToProcess)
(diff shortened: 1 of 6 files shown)


Sources/IR/Analysis/Module+NormalizeObjectStates.swift
```diff
@@ -16,8 +16,7 @@ extension Module {
 
     // Verify that object states are properly initialized/deinitialized in `b` given `context`,
     // updating `self` as necessary and reporting violations in `diagnostics`.
-    machine.fixedPoint { (blockAddress, context) in
-      let b = Block.ID(blockAddress)
+    machine.fixedPoint { (b, context) in
```
tothambrus11, line 19: Previously, `blockAddress` was giving us some context about what `b` is referring to, but by removing it, we now only see `b` on this neighborhood, so I would be wondering what `b` is referring to here. Should we rename it to `block`/add explicit type annotation to `b`?

~~Even if we can guess from the context that it's a block, I'm still wondering if it's a block value, or a block id. Usually I like to not have `id` in the name of variables, and have the type tell me which one it is. In an IDE, we can see the types by hovering, but in a GitHub review, we can only rely on explicit type annotations.~~ Actually, now that I think about it, I shouldn't really be concerned whether it's an id/value because it's not mutable, so there won't be any accidental copies to look out for.

tothambrus11 (changes requested):
The changes look good, they make some nice noise reduction. I left a comment about some abbreviation, which we could improve before merging.

kyouko-taiga (approved):
LGTM, no notes.

Sources/IR/Analysis/Module+NormalizeObjectStates.swift
lucteo: Note that it was previously `b` when I started the refactoring; it become `blockAddress` when I started moving more and more functionalities to use `Block.ID` instead of `Function.Blocks.Address`.

PR #1795, opened 2025-11-30
Lock dependency versions and fix pipeline inconsistencies

tothambrus11:


dabrahams (commented):
I don't think(?) sudo is needed

PR #1800, opened 2025-11-30
Lock dependency versions and use MinSizeRel LLVM build for cmake  builds explicitly

tothambrus11:
Previously, cmake and ninja versions were automatically updated to the latest version, which made builds non-reproducible. I set specific version for now (the current latest versions), and we can upgrade when needed.

I also set the LLVM build type to MinSizeRel explicitly (previous behavior was just defaulting to that). I tried setting it to debug, but the ninja macOS debug builds were running 2+ hours and then failed. I failed to fix that issue, so for now I just set the option explicitly to MinSizeRel. The debug LLVM builds are tested in SwiftyLLVM, there they work without any issue. SwiftyLLVM's build setup is considerably simpler though, as it doesn't have any dependencies.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -45,7 +45,7 @@ env:
   spm-build-options: -Xswiftc -enable-testing --explicit-target-dependency-import-check error
   spm-test-options: --parallel
   swift-version: '6.2.0'
-  HYLO_LLVM_BUILD_RELEASE: 20251018-115119
+  HYLO_LLVM_BUILD_RELEASE: 20251207-115516
   HYLO_LLVM_VERSION: '20.1.6'
 
 jobs:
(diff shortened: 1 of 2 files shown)


tothambrus11:
@dabrahams Could you take a look?

dabrahams (approved):
Works for me.  I think maybe we ought to explicitly say that these are the only LLVM versions supported.

tothambrus11:
> Works for me. I think maybe we ought to explicitly say that these are the only LLVM versions supported.

Yeah, I can update the docs. I'm not sure though if this is the problem of the CI setup. Could you test if the latest Hylo compiler works with Xcode 26, CMake + Xcode generator and the debug version of LLVM?

dabrahams:
> > Works for me. I think maybe we ought to explicitly say that these are the only LLVM versions supported.
> 
> Yeah, I can update the docs. 

And notify all devs that their setup needs to change, please.

> I'm not sure though if this is the problem of the CI setup. Could you test if the latest Hylo compiler works with Xcode 26, CMake + Xcode generator and the debug version of LLVM?

Can you open an issue for that please?  I can't do it today and don't want to lose track of it.  Please be specific about which build of LLVM you want me to test.

RishabhRD (approved):
LGTM

PR #1809, opened 2025-12-08
Analyze projections & check for errors.

lucteo:
Inspect projects and build a "skeleton" out of them, describing the important parts of a projection (yield points, ramp & slide blocks).

The projection is well-formed if there is exactly one yield on each code path though the projection.

### DIFF
--- a/Sources/IR/Analysis/ControlFlowGraph.swift
+++ b/Sources/IR/Analysis/ControlFlowGraph.swift
@@ -110,6 +110,12 @@ struct ControlFlowGraph: Sendable {
     return result
   }
 
+  /// Iterates over all blocks reachable from `start`, in the direction indicated by `forward`.
+  func iterateFrom(_ start: [Block.ID], forward: Bool = true) -> AnySequence<Block.ID> {
+    let f = forward ? { (l: Label) in l != .backward } : { (l: Label) in l != .forward }
+    return AnySequence(relation.bfs(from: start[0], edgeFilter: f))
+  }
+
   /// Explores the graph in a breadth-first manner starting from `start`, calling `action`
   /// for each visited block.
   ///
@@ -118,17 +124,14 @@ struct ControlFlowGraph: Sendable {
   ///
   /// If `forward` is `false`, the graph is explored in the reverse direction, following
   /// predecessor edges instead of successor edges.
+  /// 
+  /// Use this method when a simple exploration with `iterateFrom()` is not sufficient, for example
+  /// to skip parts of the graph or to stop the exploration early.
   func withBFS(
     _ start: [Block.ID], forward: Bool = true, _ action: (Block.ID, [Block.ID]) -> ExplorationContinuation
   ) {
     let f = forward ? { (l: Label) in l != .backward } : { (l: Label) in l != .forward }
     return relation.withBFS(start, edgeFilter: f, action)
-  }
-
-  /// Iterates over all blocks reachable from `start`, in the direction indicated by `Forward`.
-  func iterateFrom(_ start: [Block.ID], forward: Bool = true) -> AnySequence<Block.ID> {
-    let f = forward ? { (l: Label) in l != .backward } : { (l: Label) in l != .forward }
-    return AnySequence(relation.bfs(from: start[0], edgeFilter: f))
   }
 
 }
--- a/Sources/Utils/DirectedGraph.swift
+++ b/Sources/Utils/DirectedGraph.swift
@@ -82,6 +82,9 @@ public struct DirectedGraph<Vertex: Hashable & Sendable, Label: Sendable>: Senda
   ///
   /// When calling `action`, the vertex and its successors are passed as arguments. The
   /// return value of `action` determines how the exploration continues.
+  /// 
+  /// Use this method when a simple exploration with `bfs(from:)` is not sufficient, for example
+  /// to skip parts of the graph or to stop the exploration early.
   public func withBFS<C: Collection<Vertex>>(
     _ start: C,
     edgeFilter: ((Label) -> Bool)? = nil,


Sources/IR/Analysis/ProjectionSkeleton.swift
```diff
@@ -0,0 +1,13 @@
+/// The skeleton of a projection declaration, used to analyze the projection and expand it.
+internal struct ProjectionSkeleton {
+
+  /// The yield points in the projection, in the order they appear in the function.
+  public let yieldPoints: [InstructionID]
+
+  /// The blocks that are part of the ramp, i.e., the blocks that must be executed before an yield instruction.
```
tothambrus11:
```suggestion
  /// The blocks that are part of the ramp, i.e., the blocks that must be executed before a yield instruction.
```

Sources/IR/Analysis/ProjectionSkeleton.swift
```diff
@@ -0,0 +1,13 @@
...
+  public let rampBlocks: [Block.ID]
+
+  /// The blocks that are part of the slide, i.e., the blocks that must be executed after an yield instruction.
```
tothambrus11:
```suggestion
  /// The blocks that are part of the slide, i.e., the blocks that must be executed after a yield instruction.
```

Sources/IR/Analysis/Module+AnalyzeProjection.swift
```diff
@@ -0,0 +1,124 @@
...
+        }
+      }
+    }
+
+    // From now on, we can perform all the checks on blocks.
+    let cfg = cfg()
+
+    // Find the ramp and slide blocks.
+    let rampBlocks = Set(yieldBlocks.lazy.flatMap({ cfg.predecessors(of: $0) }))
+    let slideBlocks = Set(yieldBlocks.lazy.flatMap({ cfg.successors(of: $0) }))
```
lucteo: This doesn't work well. It gets just the first level successors

lucteo: Fixed using Dimi's algorithm.

Sources/IR/Analysis/ControlFlowGraph.swift
```diff
@@ -107,6 +107,57 @@ struct ControlFlowGraph: Sendable {
...
+    forward: Bool = true,
+    handlingBlock: (Block.ID, [Block.ID]) -> ExplorationContinuation
+  ) {
+    var work = start
+    var visited: Set<Block.ID> = []
+
+    while let b = work.popLast() {
+      if visited.contains(b) {
+        continue
+      }
```
kyouko-taiga:
```suggestion
    while let b = work.popLast() where !visited.contains(b) {
```

Sources/IR/Analysis/ControlFlowGraph.swift
```diff
@@ -107,6 +107,57 @@ struct ControlFlowGraph: Sendable {
...
+  ///
+  /// When calling `handlingBlock`, the block's ID and its successors are passed as arguments. The
+  /// return value of `handlingBlock` determines how the exploration continues.
+  ///
+  /// If `forward` is `false`, the graph is explored in the reverse direction, following
+  /// predecessor edges instead of successor edges.
+  func exploreFrom(
+    _ start: [Block.ID],
+    forward: Bool = true,
+    handlingBlock: (Block.ID, [Block.ID]) -> ExplorationContinuation
```
kyouko-taiga:
```suggestion
  /// Explores the graph in a breadth-first manner starting from `start`, calling `handleBlock`
  /// for each visited block.
  ///
  /// When calling `handlingBlock`, the block's ID and its successors are passed as arguments. The
  /// return value of `handlingBlock` determines how the exploration continues.
  ///
  /// If `forward` is `false`, the graph is explored in the reverse direction, following
  /// predecessor edges instead of successor edges.
  func exploreFrom(
    _ start: [Block.ID],
    forward: Bool = true,
    _ handleBlock: (Block.ID, [Block.ID]) -> ExplorationContinuation
```

Most of the time we use trailing-closure syntax in Swift so the label is typically irrelevant. If I had to put a label I would write `handlingBlockWith`.

Inside the function the name of the parameter should sounds more imperative I think.

Sources/IR/Analysis/ControlFlowGraph.swift
```diff
@@ -107,6 +107,57 @@ struct ControlFlowGraph: Sendable {
...
+      }
+      visited.insert(b)
+
+      let successors = forward ? successors(of: b) : predecessors(of: b)
+
+      switch handlingBlock(b, successors) {
+      case .continue:
+        for s in successors where !visited.contains(s) {
+          work.append(s)
+        }
```
kyouko-taiga:
```suggestion
        work.append(contentsOf: successors.filter({ !visited.contains($0) }))
```

I am more and more wanting a `not` predicate constructor for these situations. Then we could write:
```swift
work.append(contentsOf: successors.filter(not(visited.contains(_:))))
```

Sources/IR/Analysis/Module+AnalyzeProjection.swift
```diff
@@ -0,0 +1,149 @@
...
+  /// Check if projection `f` is well-formed, raising diagnostics to `log` as needed, and populates `projectionSkeletons` with
+  /// the relevant information for expanding projections.
+  ///
+  /// A well formed projection must have exactly one yield on each path through the function.
+  mutating func analyzeProjection(
+    _ f: Function.ID, reportingDiagnosticsTo log: inout DiagnosticSet
+  ) {
+    if !self[f].isSubscript {
+      return
+    }
```
kyouko-taiga:
```suggestion
    // Nothing to do if `f` is not a subscript.
    if !self[f].isSubscript { return }
```

Sources/IR/Analysis/Module+AnalyzeProjection.swift
```diff
@@ -0,0 +1,149 @@
...
+    reportingDiagnosticsTo log: inout DiagnosticSet
+  ) throws -> ProjectionSkeleton {
+    // Gather all the yield points in the projection.
+    let yieldPoints: [InstructionID] = instructionIdentities.filter({ self[$0] is Yield })
+    let yieldBlocks = yieldPoints.map({ block(of: $0) })
+
+    // Check that there aren't two yields in the same block.
+    // In that case, we would get two yield blocks one after the other.
+    if yieldBlocks.count >= 2 {
+      for i in 1..<yieldBlocks.count {
```
kyouko-taiga:
```suggestion
      for i in 1 ..< yieldBlocks.count {
```

Sources/IR/Analysis/Module+AnalyzeProjection.swift
```diff
@@ -0,0 +1,149 @@
...
+extension Function {
+
+  /// Check if projection `self` is well-formed, raising diagnostics to `log` as needed,
+  /// and returns the relevant information for expanding projections.
+  fileprivate func analyzeProjection(
+    reportingDiagnosticsTo log: inout DiagnosticSet
+  ) throws -> ProjectionSkeleton {
+    // Gather all the yield points in the projection.
+    let yieldPoints: [InstructionID] = instructionIdentities.filter({ self[$0] is Yield })
+    let yieldBlocks = yieldPoints.map({ block(of: $0) })
```
kyouko-taiga, line 31:
```suggestion
    let yieldBlocks = yieldPoints.map(block(of:_))
```
eta-expansion rocks!

Sources/IR/Analysis/Module+AnalyzeProjection.swift
```diff
@@ -0,0 +1,149 @@
...
+
+      return .continue
+    }
+    try log.throwOnError()
+
+    return ProjectionSkeleton(
+      yieldPoints: yieldPoints,
+      rampBlocks: ramp,
+      slideBlocks: slide,
+    )
```
kyouko-taiga:
```suggestion
      slideBlocks: slide)
```

Sources/IR/Analysis/Module+CheckYieldCoherence.swift
```diff
@@ -0,0 +1,149 @@
...
+  /// - Complexity: O(n) in the number of blocks in the CFG.
+  private func yieldBlockForSlide(_ b: Block.ID, cfg: ControlFlowGraph, yieldBlocks: [Block.ID]) -> Block.ID? {
+    var r: Block.ID? = nil
+    cfg.exploreFrom([b], forward: false) { (b, _) in
+      if yieldBlocks.contains(b) {
+        r = b
+        return .stop
+      }
+      return .continue
+    }
```
kyouko-taiga: This looks very imperative for something that feels like it could be a one-liner. We should probably add another method in the CFG to simply iterates over blocks. Wasn't there one already?

Sources/IR/Analysis/Module+AnalyzeProjection.swift
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+}
+
+extension Array where Element : Equatable {
+  fileprivate mutating func appendUnique(_ element: Element) {
+    if !self.contains(element) {
+      self.append(element)
+    }
+  }
```
kyouko-taiga:
```suggestion

    /// Appends `element` to `self` iff it is not already contained in `self`.
    ///
    /// - Complexity: O(n) where n is the length of `self`.
    fileprivate mutating func appendUnlessContained(_ element: Element) {
    if !self.contains(element) {
      self.append(element)
    }

  }
```

Sources/IR/Analysis/Module+AnalyzeProjection.swift
```diff
@@ -0,0 +1,149 @@
+import FrontEnd
+
+extension Module {
+
+  /// Check if projection `f` is well-formed, raising diagnostics to `log` as needed, and populates `projectionSkeletons` with
+  /// the relevant information for expanding projections.
+  ///
+  /// A well formed projection must have exactly one yield on each path through the function.
+  mutating func analyzeProjection(
```
kyouko-taiga: That is not a very instructive name. What do you think about Dyva's `checkYieldCoherence`?

kyouko-taiga (changes requested):
Nice work.

Sources/IR/Analysis/Module+AnalyzeProjection.swift
lucteo: Initially this pass it was far more heavy on collecting information that helps expanding the projection, but now it's far more on checking. I think Dyva's name fits better.

Sources/IR/Analysis/Module+CheckYieldCoherence.swift
lucteo: I think adding another method to `ControFlowGraph` makes sense. Maybe making `ControlFlowGraph` conform to `Sequence`

Sources/IR/Analysis/ControlFlowGraph.swift
lucteo: `where` clauses don't work with `while` statements; only with `for`s

tothambrus11: Oh, why is that not a compilation error?

lucteo: Imagine the following CFG:
```
A -> B  -> C
 \-> D --^
```
Here, `C` needs to be visited twice, and it's a perfectly valid code. `if A { B } else {D} ; C`

tothambrus11: Ah sorry for the misunderstanding, I was referring to Dimi's snippet. On mobile I didn't realize it's a comment and not checked by the compiler:)

kyouko-taiga: Sorry, my bad. I don't know why I missed that it wouldn't work.

Sources/IR/Analysis/ControlFlowGraph.swift
```diff
@@ -107,6 +107,74 @@ struct ControlFlowGraph: Sendable {
...
+  /// When calling `handleBlock`, the block's ID and its successors are passed as arguments. The
+  /// return value of `handleBlock` determines how the exploration continues.
+  ///
+  /// If `forward` is `false`, the graph is explored in the reverse direction, following
+  /// predecessor edges instead of successor edges.
+  func exploreFrom(
+    _ start: [Block.ID],
+    forward: Bool = true,
+    _ handleBlock: (Block.ID, [Block.ID]) -> ExplorationContinuation
+  ) {
```
kyouko-taiga: A similar method is called `bfs` on `DominatorTree`. Also callbacks are typically named `action` throughout the code base. So I would suggest the following naming for the sake of consistency and discoverability:
```suggestion
  func withBFS(
    _ start: [Block.ID], forward: Bool = true, _ action: (Block.ID, [Block.ID]) -> ExplorationContinuation
  ) {
```

But I also wonder if a call-back API is the best approach. It is more work but I think it's nicer to return a sequence, like `DominatorTree.bfs` does.

Sources/IR/Analysis/ControlFlowGraph.swift
```diff
@@ -107,6 +107,74 @@ struct ControlFlowGraph: Sendable {
...
+      case .skip:
+        continue
+      case .stop:
+        break
+      }
+    }
+  }
+
+  /// Iterates over all blocks reachable from `start`, in the direction indicated by `Forward`.
+  func iterateFrom(_ start: [Block.ID], forward: Bool = true) -> AnySequence<Block.ID> {
```
kyouko-taiga: Ah, here's the sequence! Well then why having the other method as well?

Sources/IR/Analysis/ControlFlowGraph.swift
kyouko-taiga: I see that the sequence-returning method already exists. Then I would remove this one. I don't think performance should be a big concern here. A non-erased sequence in the other method will probably be as efficient as this callback-based approach. Of course I'm saying that without any evidence.

Sources/IR/Module.swift
```diff
@@ -31,6 +31,9 @@ public struct Module: Sendable {
   /// The functions in the module.
   public private(set) var functions: [Function.ID: Function] = [:]
 
+  /// The skeletons of all the projections in the module.
+  internal var projectionSkeletons: [ Function.ID: ProjectionSkeleton ]
```
kyouko-taiga:
```suggestion
  internal var projectionSkeletons: [Function.ID: ProjectionSkeleton]
```

Tests/HyloTests/TestCases/Lowering/ProjectionWithYiedInLoop.hylo
```diff
@@ -0,0 +1,11 @@
+//- lowerToFinishedIR expecting: .failure
+
+public type A: Deinitializable {
+
+  public property p: Int {
+    for sink let i in 0 ..< 10 {
+      yield i
+    }
+  } //! diagnostic not all paths of the projection contain a yield
```
kyouko-taiga, line 9: Is there any test that checking that the analysis works when there is in fact a loop in the function but the yield is before or after?

Also I think we should test a case where a function does not yield on all paths but fatal errors somewhere. Something like:
```hylo
subscript a(x: Int): Int {
  if x > 0 { yield x } else { fatal_error() }
}
```

kyouko-taiga (changes requested):
Great changes.

Is there a reason why you wanted the two ways to iterate over a a control-flow graph? Also we may need more tests.

Sources/IR/Analysis/ControlFlowGraph.swift
lucteo: The other method covers a more general case. As we already know the successors of the current node, we pass it to the action -- this is important in our checking algorithm. Also, the previous method has the ability to stop exploring some of the paths; this doesn't. Both of these features are needed by our projection checking algorithm.

I can implement this method in terms of the previous one, but the implementation  would be (IMHO) harder to read.

Sources/IR/Analysis/ControlFlowGraph.swift
lucteo: I do need this method for the main implementation of the projection checking algorithm (see also next comment).

Tests/HyloTests/TestCases/Lowering/ProjectionWithYiedInLoop.hylo
lucteo: good point

Sources/IR/Analysis/ControlFlowGraph.swift
kyouko-taiga: I'm not sure it would be more complicated.

I am not a fan of `for_each`-like algorithms and I think they are subsumed by either regular for loops or other algorithms on sequences. I also think your `exploreFrom(_:forward:_:)` method is much more complicated to use than this one. We have to go read what `ExplorationContinuation` is doing.

I don't understand why you can't just read the successors or predecessors of a node from the graph in the body of a loop iterating over the result of `iterateFrom(_:forward:)`.

lucteo: The `withBFS` method has 3 options when visiting a node: continue exploring the successors, skip the successors of the current node, and completely stop the exploration. The `iterateFrom` does not have the option to "skip" successors (and their successors) -- all the nodes will eventually be explored. The "skip" option is important for the projection checking algorithm.

Passing the nodes to be explored next as a parameter is a small optimization, as we already know them when we call the closure -- it saves an O(n) operation in some cases. But, yes, we can leave without it.

Please see the latest patch, I've moved the two algorithms in `DirectedGraph` (the `iterateFrom` was already there, but I had to apply 2 fixes to it).

kyouko-taiga: I see. I'm not convinced the optimization is worth the extra complexity but I won't fight the change. However I think you should document why one should use `bfs` over `withBFS`, or vice versa.

lucteo: Improved the documentation.

PR #1819, opened 2026-01-05
Foundation work for lowering projections

lucteo:
Add some foundation work for projection lowering:
- create 2 passes: one for lowering projections and one for lowering projection callers
- add the ability to create 3 synthetic functions: projection.ramp, projection.slide, and caller.plateau
- handle mangling and demangling for the 3 new synthetic functions
- add basic code that generates empty projections ramps and slides as well as caller plateaus


Sources/Driver/Driver.swift
```diff
@@ -412,6 +412,13 @@ public struct Driver: ParsableCommand, Sendable {
     }
 
     var ir = IR.Program(syntax: program, modules: loweredModules)
+
+    if outputType != .rawIR {
```
kyouko-taiga, line 416: Why can't these passes be added to `transforms`?

Sources/IR/Analysis/Program+lowerProjectionCallers.swift
```diff
@@ -0,0 +1,25 @@
+extension IR.Program {
+
+  /// Lowers all functions in `self` that call projections to use regular function calls.
+  public mutating func lowerProjectionCallers() {
```
kyouko-taiga: I think the term "lower" shouldn't be used here because I would like that term to refer to the process of transforming Hylo sources to IR. Similarly I would not use the term "lowering" to denote the transformation carried out by access reification.

If "transform" is not precise enough maybe we can consider "elaborate"? That is also the term I use to describe AST to AST transformations so the idea is that an elaboration is a sort of homomorphism.

Sources/IR/Analysis/ProjectionLoweringUtils.swift
```diff
@@ -0,0 +1,151 @@
...
+        environment: .void,
+        inputs: [
+          CallableTypeParameter(label: nil, type: ^ParameterType(.`let`, addressType))
+        ],
+        output: .void))
+    return AnyType(
+      TupleType([
+        TupleType.Element(label: "resumeFunction", type: t),
+        TupleType.Element(label: "frame", type: addressType),
+      ]))
```
kyouko-taiga: I think you can write this more concisely:

```suggestion
    return ^TupleType([("resumeFunction", t), ("frame", addressType)])
```

Sources/IR/Analysis/ProjectionLoweringUtils.swift
```diff
@@ -0,0 +1,151 @@
...
+    return result
+  }
+
+  /// Returns the IR function representing a plateau region in caller `f`.
+  ///
+  /// Signature:
+  /// > fun Caller.plateauN(let Builtin.ptr, let ProjectionContinuation) -> {}
+  mutating func demandCallerPlateauDeclaration(for f: Function.ID, region index: Int)
+    -> Function.ID
+  {
```
kyouko-taiga:
```suggestion
  mutating func demandCallerPlateauDeclaration(
    for f: Function.ID, region index: Int
  ) -> Function.ID {
```

Sources/Driver/Driver.swift
lucteo: I think we want these to be mandatory passes; that's why I haven't put them in `transforms`.

kyouko-taiga: Gotcha. But then why aren't they in `applyMandatoryPasses`?

lucteo: Because these are methods of the `Program`, not `Module`. My "friend" `InstructionTransformer` forces them to be in the `Program`

kyouko-taiga: I see, I guess that will be improved in NFW then.

PR #1823, opened 2026-01-10
Interpreter: Memory should own TypeLayoutCache

RishabhRD:
## Motivation
Motivation comes from following [PR discussion](https://github.com/hylo-lang/hylo/pull/1781#discussion_r2647289034) where we need to pass `TypeLayoutCache` all the time to functions like `store` so that it can validate correct type store/access by unsafe code (along with preconditions like allocation should be according to passed cache).

## Design Changes
- `Memory` owns `TypeLayoutCache`
- `Interpreter` simply works with `AnyType`, without worrying about `TypeLayout`.
- Introduced `Memory.Place` for typed location and remove `Interpreter.Address`. I know that we need to also rename the address to place in IR, would raise different PR for the same. It was necessary to use `Place` instead of `TypedAddress` kind of name as it would create repetitive work for future.

### DIFF
--- a/Sources/Interpreter/Memory.swift
+++ b/Sources/Interpreter/Memory.swift
@@ -430,23 +430,6 @@ extension Memory.Allocation {
 
 }
 
-public extension Memory.Place {
-  /// Returns `self` offset by `n` bytes having type `t`.
-  func after(_ n: Int, bytesHavingType t: AnyType) -> Self {
-    return Self(allocation: self.allocation, offset: self.offset + n, type: t)
-  }
-
-  /// Returns `self` offset by `-n` bytes having type `t`.
-  func before(_ n: Int, bytesHavingType t: AnyType) -> Self {
-    return Self(allocation: self.allocation, offset: self.offset - n, type: t)
-  }
-
-  /// Returns `self` having type `t`.
-  func withType(_ t: AnyType) -> Self {
-    return Self(allocation: self.allocation, offset: self.offset, type: t)
-  }
-}
-
 extension Memory {
   /// Returns the address of `subPart` in `whole`.
   public mutating func location(of subPart: RecordPath, in whole: Place) -> Place {


Sources/Interpreter/Interpreter.swift
```diff
@@ -324,11 +267,11 @@ public struct Interpreter {
     case is Return:
       return .jump(popStackFrame())
     case let x as Store:
-      try memory.store(asBuiltinValue(x.object), at: asAddress(x.target))
+      try memory.store(asBuiltinValue(x.object), in: asPlace(x.target))
```
dabrahams, line 270: Should `memory` have a method that allows

```suggestion
      try memory.store(asBuiltinValue(x.object), in: x.target)
```

?  Just  a thought.

Sources/Interpreter/Interpreter.swift
```diff
@@ -324,11 +267,11 @@ public struct Interpreter {
...
       return .none
     case let x as SubfieldView:
-      let p = asAddress(x.recordAddress)
-      return .value(.init(payload: typeLayouts.address(of: x.subfield, in: p)))
+      let p = asPlace(x.recordAddress)
+      return .value(.init(payload: memory.place(of: x.subfield, in: p)))
```
dabrahams: I don't know if there's a good solution to this, but this code doesn't read in a way that makes its meaning apparent.  Just a note.

Sources/Interpreter/Interpreter.swift
```diff
@@ -377,46 +320,40 @@ public struct Interpreter {
...
     topOfStack.allocations.removeLast()
   }
 
-  /// Returns the address produced by executing instruction identified by `i` in the current frame,
-  /// or `nil` if it didn't produce an address.
-  func addressProduced(by i: InstructionID) -> Address? {
-    topOfStack.registers[i]?.payload as? Address
+  /// Returns the place produced by executing instruction identified by `i` in the current frame,
+  /// or `nil` if it didn't produce a place.
+  func placeProduced(by i: InstructionID) -> Memory.Place? {
```
```suggestion
  func resultAsPlace(_ i: InstructionID) -> Memory.Place? {
```

Should there be a precondition so we can return non-optional?

Sources/Interpreter/Interpreter.swift
```diff
@@ -377,46 +320,40 @@ public struct Interpreter {
...
+    topOfStack.registers[i]?.payload as? Memory.Place
   }
 
-  /// Returns the address to be deallocated by `i`.
-  func addressToBeDeallocated(by i: DeallocStack) -> Address {
+  /// Returns the place to be deallocated by `i`.
+  func placeToBeDeallocated(by i: DeallocStack) -> Memory.Place {
```
```suggestion
  func placeDeallocated(by i: DeallocStack) -> Memory.Place {
```

It would be nice to get type information out of this name if possible.  The best name would be revealed by the use site.

Sources/Interpreter/Memory.swift
```diff
@@ -4,8 +4,10 @@ import Utils
 /// The memory of an interpreted process.
 public struct Memory {
 
-  /// An empty instance.
-  public init() {}
+  /// An empty instance whose allocation layout semantics are defined by `typeLayouts`.
```
```suggestion
  /// An empty instance whose allocation and access semantics are defined by `typeLayouts`.
```

Sources/Interpreter/Memory.swift
```diff
@@ -384,3 +421,34 @@ extension Memory.Allocation {
   }
 
 }
+
+public extension Memory.Place {
+  /// Returns `self` offset by `n` bytes having type `t`.
```
dabrahams: I can't understand what this means.  Bytes don't have a type, other than maybe `Builtin.RawByte`.

Sources/Interpreter/Memory.swift
```diff
@@ -384,3 +421,34 @@ extension Memory.Allocation {
...
+
+  /// Returns `self` having type `t`.
+  func withType(_ t: AnyType) -> Self {
+    return Self(allocation: self.allocation, offset: self.offset, type: t)
+  }
+}
+
+extension Memory {
+  /// Returns the address of `subPart` in `whole`.
+  public mutating func place(of subPart: RecordPath, in whole: Place) -> Place {
```
```suggestion
  public mutating func location(of subPart: RecordPath, in whole: Place) -> Place {
```

Try to keep type info out of non-type names.

Tests/InterpreterTests/MemoryTests.swift
```diff
@@ -8,23 +8,25 @@ final class InterpreterMemoryTests: XCTestCase {
   let emptyProgram = makeEmptyProgram()
 
   func testAllocation() throws {
-    var m = Memory()
-    var allocations: [Memory.Address] = []
-    for alignmentPower in 0..<8 {
-      let alignment = 1 << alignmentPower
-      let p = m.allocate(3, bytesWithAlignment: alignment)
+    var l = TypeLayoutCache(typesIn: TypedProgram.empty, for: UnrealABI())
+    var m = Memory(l)
```
dabrahams: Unless we have a use for creating `Memory` with an existing non-empty TypeLayoutCache, maybe `Memory`'s `init` should just take a `TypedProgram` and `ABI`.

Tests/InterpreterTests/MemoryTests.swift
```diff
@@ -35,11 +37,11 @@ final class InterpreterMemoryTests: XCTestCase {
...
     assert(i16Pair.alignment > 1, "Need to produce misaligned access for testing")
 
-    var m = Memory()
-    let p = m.allocate(i16Pair.size, bytesWithAlignment: i16Pair.alignment)
+    var m = Memory(layouts)
+    let p = m.allocate(i16Pair.type).address
 
     check(throws: Memory.Error.alignment(p + 1, for: i16)) {
-      try m.compose(i16, at: p + 1)
+      try m.compose(i16.type, at: p + 1)
```
dabrahams: I don't know what caused all these `.type`s to be needed (and perhaps it was the right choice) but this code's readability suffers as a result.

Tests/InterpreterTests/MemoryTests.swift
```diff
@@ -151,4 +153,16 @@ final class InterpreterMemoryTests: XCTestCase {
...
+
+  
+  func testSubPartLayout() throws {
+    var m = Memory(TypeLayoutCache(typesIn: TypedProgram.empty, for: UnrealABI()))
+    let t = ^TupleType(types: [
+      ^BuiltinType.i(32), ^TupleType(types: [^BuiltinType.i(32), ^BuiltinType.i(8)]),
+    ])
+    let a = m.allocate(^t)
+    let x = m.place(of: [1, 1], in: a)
+    XCTAssertEqual(x, .init(allocation: a.allocation, offset: 8, type: ^BuiltinType.i(8)))
```
dabrahams: Why are we only testing one subpart?  Might be my mistake

Sources/Interpreter/Interpreter.swift
RishabhRD: `x.target` is an `IR.Operand`. It seems to me `Interpreter` should be the one interacting with `Operand` directly not `Memory`, its out of scope of `Memory` responsibilities. Do you see any advantage with suggested approach, I might be missing that?

Tests/InterpreterTests/MemoryTests.swift
RishabhRD: Its my mistake

Sources/Interpreter/Memory.swift
RishabhRD: I realized that, these methods are not even needed. Instead provided a method that helps us create a place from address.

Sources/Interpreter/Interpreter.swift
RishabhRD: With strengthening preconditions of `asPlace`, this method just becomes redundant. So, removing it.

Tests/InterpreterTests/MemoryTests.swift
RishabhRD: This is because now `compose` directly takes type and type layout is being handled by `Memory`. But type layout in error is more descriptive.

I have changed tests a little so that I pass layouts explicitly instead of using `.type`. This might increase readability. Please check that once, if you feel this is more readable.

Sources/Interpreter/Interpreter.swift
dabrahams: No I think you're right.

Tests/InterpreterTests/MemoryTests.swift
dabrahams: It helps definitely.  However, I note it's only helping tests.  If you don't anticipate it helping in the regular code consider making an extension just to clean up the tests.  Either way, please apply your judgement and merge the PR without any further review.

RishabhRD: There were no changes done in regular code for improving readability, it was just change in style of writing those tests. So, no action needed here.

PR #1828, opened 2026-01-17
IR: change the notion of "address" with "place"

RishabhRD:
As discussed in past, place is the correct term to denote address with type information of object that should be accessed at address.

Currently, IR instructions use the term place for the same thing. However, in interpreter, recently in a refactoring I needed the term place. So, creating this follow up PR, so that interpreter is consistent with project notations.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -138,7 +138,7 @@ let package = Package(
 
     .target(
       name: "TestUtils",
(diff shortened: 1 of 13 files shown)


RishabhRD:
Okay, so merging the PR then :-)

PR #1834, opened 2026-01-28
Elaborate of projections and projection callers (basic case)

lucteo:
This can generate code that, for simple projects and callers will elaborate the `project`, `end_project` and `yield` instructions to generate simple functions.

Limitations:
- no cross regions allocas/register use (projection & callers)
- callers with only one project
- projections with more than one `yield`

### DIFF
--- a/Sources/IR/Analysis/Function+Split.swift
+++ b/Sources/IR/Analysis/Function+Split.swift
@@ -1,19 +1,19 @@
 /// The positions at which a sequence of instructions is split.
-/// 
+///
 /// An instance defines three contiguous regions of instructions:
 ///   - the "front", at positions in the range `..<splitPoint`
 ///   - the "epilogue", at positions in the range `splitPoint ..< epilogueEnd`
 ///   - the "back", at positions in the range `epilogueEnd...`
-/// 
+///
 /// If we strictly consider the ordering of instructions, a split point divides an array of
 /// instructions into two parts:
 ///   - the instructions at indices `< splitPoint` -- the "front"
-///   - the instructions at indices `> splitPoint` -- the physical "back"
-/// 
+///   - the instructions at indices `> splitPoint` -- the "back"
+///
 /// There are also instructions that are logically part of the front, even though they
 /// appear after the split point. These are the "epilogue" instructions for a split point.
 /// The `epilogueEnd` index indicates where the epilogue ends and the back begins.
-/// 
+///
 /// Invariant: `splitPoint < epilogueEnd`
 internal struct SplitPositions {
 
@@ -37,11 +37,11 @@ extension Function {
   internal func split(
     instructions xs: [InstructionID], where isSplitPoint: (InstructionID) -> Bool
   ) -> [SplitPositions] {
-    var splits: [SplitPositions] = []    
+    var splits: [SplitPositions] = []
     var end = 0
     while let p = xs[end...].firstIndex(where: isSplitPoint) {
       end = xs[(p + 1)...].firstIndex(where: { !self[$0].mayBeEpilogue }) ?? xs.count
-      splits.append(SplitPositions(splitPoint: p + 1, epilogueEnd: end))
+      splits.append(SplitPositions(splitPoint: p, epilogueEnd: end))
     }
     return splits
   }
@@ -54,11 +54,10 @@ extension Function {
   internal func split(
     instructions a: [InstructionID], at splitPoint: InstructionID
   ) -> SplitPositions {
-    let splitPointIndex = a.firstIndex(of: splitPoint)
-    precondition(splitPointIndex != nil, "split point \(splitPoint) not found")
-    let remaining = a.suffix(from: splitPointIndex!).dropFirst()
-    let epilogueEnd = remaining.firstIndex(where: { !Self.mayBeEpilogue(self[$0]) }) ?? a.count
-    return SplitPositions(splitPoint: splitPointIndex!, epilogueEnd: epilogueEnd)
+    let p = a.firstIndex(of: splitPoint)
+    precondition(p != nil, "split point \(splitPoint) not found")
+    let end = a[(p! + 1)...].firstIndex(where: { !self[$0].mayBeEpilogue }) ?? a.count
+    return SplitPositions(splitPoint: p!, epilogueEnd: end)
   }
 
   /// Splits instructions in `b` at `splitPoint`, returning the positions of the resulting partitions.
@@ -70,16 +69,16 @@ extension Function {
     split(instructions: Array(instructions(in: b)), at: splitPoint)
   }
 
+}
+
+extension Instruction {
+
   /// Returns `true` if the instruction `i` can be part of the yield tail,
   /// e.g., it appears after a yield, but still belongs to the ramp of the projection.
   ///
   /// Tail instructions: `MarkState` (uninitialized), `DeallocStack`, `EndAccess`.
-  private static func mayBeEpilogue(_ i: Instruction) -> Bool {
-    if let x = i as? MarkState {
-      return !x.initialized
-    } else {
-      return i is EndAccess || i is DeallocStack
-    }
+  var mayBeEpilogue: Bool {
+    (self as? MarkState)?.initialized == false || (self is EndAccess) || (self is DeallocStack)
   }
 
 }


Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,118 @@
+extension Function {
+
+  /// Split instructions `a` at all the points where `isSplit` returns `true`, returning the parts.
```
dabrahams:
```suggestion
  /// Splits `a` before each element for which `isSplit` , returning the parts.
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,118 @@
...
+  ///
+  /// - Precondition: there is at least one instruction after each `splitPoint` in `a`.
+  /// - Postcondition: if `isSplit` returns `true` for `n` instructions, the returned array has
+  ///   `n + 1` elements.
+  /// - Postcondition: for non-final parts, `before + [splitPoint] + tail` is the slice up to (but
+  ///   not including) the next split point.
+  /// - Postcondition: for the final part, `splitPoint` is `nil` and `tail` is empty.
+  internal func split(
+    instructions a: [InstructionID], when isSplit: (InstructionID) -> Bool
+  ) -> [SplitPart]
```
dabrahams: I don't understand the first precondition.  If `isSplit` is always called on an instruction, isn't that very instruction after the split point?  Maybe I got it wrong and it really splits after the elements where `isSplit` is true?

Can't use `splitPoint` in code voice as it isn't an identifier anywhere. Documenting stuff that's already implied by your summary is not helpful; it just makes the function seem more complicated than it is.

Why isn't this just a use of `split(maxSplits:omittingEmptySubsequences:whereSeparator:)` from the standard library?

Sources/IR/Analysis/Function+Split.swift
lucteo: I would avoid using "before each element for which `isSplit`", as we are also selecting the "tail" for the split instructions. Please see `SplitPart`.

Sources/IR/Analysis/Function+Split.swift
lucteo: 1. Found a more precise way to specify it
2. fixed. thank you!
3. Please see the documentation of the returned type. I need that extra information (especially the tail), that can't be returned by the `split(maxSplits:omittingEmptySubsequences:whereSeparator:)` algorithm.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
+extension Function {
+
+  /// Splits `a` at all the points where `isSplit` returns `true`, returning the parts.
+  ///
+  /// - Precondition: `!a.isEmpty`.
+  /// - Precondition: `!isSplit(a.last!)`
+  /// - Postcondition: if `isSplit` returns `true` for `n` instructions, the returned array has
```
kyouko-taiga:
```suggestion
  /// - Requires:
  ///   * `!a.isEmpty`.
  ///   * `!isSplit(a.last!)`
  ///   * if `isSplit` returns `true` for `n` instructions, the returned array has
```
For consistency with other docs.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+  ///   `n + 1` elements.
+  /// - Postcondition: all elements in `a` appear in exactly one part of the returned array.
+  /// - Postcondition: if `p` is the last returned part, `p.splitPoint` is `nil` and `p.tail` is empty.
+  internal func split(
+    instructions a: [InstructionID], when isSplit: (InstructionID) -> Bool
+  ) -> [SplitPart]
+  {
```
kyouko-taiga:
```suggestion
  ) -> [SplitPart] {
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+    precondition(!a.isEmpty)
+    precondition(!isSplit(a.last!), "final instruction cannot be a split point")
+    var parts: [SplitPart] = []
+    var remaining = a.dropFirst(0)
+    while true {
+      guard let splitPoint = remaining.first(where: isSplit) else {
+        break
+      }
```
kyouko-taiga:
```suggestion
    while let splitPoint = remaining.first(where: isSplit) {
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+      let (splitIndex, tailEndIndex) = splitIndices(in: remaining, splittingAt: splitPoint)
+      parts.append(
+        SplitPart(
+          before: Array(remaining.prefix(upTo: splitIndex)),
+          splitPoint: splitPoint,
+          tail: Array(remaining.suffix(from: splitIndex).dropFirst().prefix(upTo: tailEndIndex))
+        ))
```
kyouko-taiga:
```suggestion
          tail: Array(remaining.suffix(from: splitIndex).dropFirst().prefix(upTo: tailEndIndex))))
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+
+  /// Splits `a` instructions at `s`, returning the split index and tail end index.
+  private func splitIndices(
+    in a: ArraySlice<InstructionID>, splittingAt s: InstructionID
+  ) -> (splitIndex: Int, tailEndIndex: Int) {
+    let splitIndex = a.firstIndex(of: s)
+    precondition(splitIndex != nil, "split point \(s) not found")
+    let remaining = a.suffix(from: splitIndex!).dropFirst()
+    let tailEndIndex = remaining.firstIndex(where: { !Self.canBeTail(self[$0]) })
+    // At least the block terminator must be outside the tail, so `tailEndIndex` is never `nil`.
```
kyouko-taiga:
```suggestion

    // At least the block terminator must be outside the tail, so `tailEndIndex` is never `nil`.
```

I like to add a newline before "blocks" of code, especially when they start with a comment.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+
+  /// Returns `true` if the instruction `i` can be part of the yield tail,
+  /// e.g., it appears after a yield, but still belong to the ramp of the projection.
+  ///
+  /// Tail instructions: `MarkState` (uninitialized), `DeallocStack`, `EndAccess`.
+  private static func canBeTail(_ i: Instruction) -> Bool {
+    if let x = i as? MarkState {
+      return !x.initialized
+    }
+    return i is EndAccess || i is DeallocStack
```
kyouko-taiga:
```suggestion
    } else {
      return i is EndAccess || i is DeallocStack
    }
```
Helps understanding that it's either/or IMO.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+/// the original block. Both `before` and `tail` preserve the original ordering. It also records
+/// the `splitPoint` instruction at which the split occurs; this is `nil` for the last part.
+internal struct SplitPart {
+
+  /// The instructions before the split point.
+  internal let before: [InstructionID]
+  /// The instruction at which the split occurs; nil for the last part.
+  internal let splitPoint: InstructionID?
+  /// The instructions physically after the split point, but logically belonging to before.
+  internal let tail: [InstructionID]
```
kyouko-taiga:
```suggestion

  /// The instruction at which the split occurs; nil for the last part.
  internal let splitPoint: InstructionID?

  /// The instructions physically after the split point, but logically belonging to before.
  internal let tail: [InstructionID]
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,215 @@ extension IR.Program {
...
+    let oldProject = d.regions[index - 1].splitPoint!
+    let p = modules[m]!.addStorage(in: entry, of: plateau, replacing: source[oldProject])
+    let rewrittenRegisters: [Operand: Operand] = [
+      .register(oldProject): .register(p)
+    ]
+
+    // From the tail we need to ignore (not rewrite) the uses of the projected value.
+    // TODO: make this work for the general case
+    let oldProjectAccess = (source[oldProject] as! Project).operands[0].instruction!
+    let toIgnore = source.allUses(of: oldProjectAccess).map({ $0.user })
```
kyouko-taiga, line 68:
```suggestion
    let toIgnore = source.allUses(of: oldProjectAccess).map(\.user )
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,215 @@ extension IR.Program {
...
+
+    var transformer = DictionaryInstructionTransformer(
+      rewrittenBlocks: rewrittenBlocks, rewrittenRegisters: rewrittenRegisters
+    )
```
```suggestion
      rewrittenBlocks: rewrittenBlocks, rewrittenRegisters: rewrittenRegisters)
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,215 @@ extension IR.Program {
...
+  ) {
+    let oldProject = d.regions[index - 1].splitPoint!
+    let oldEndProject = d.regions[index].splitPoint!
+
+    let oldProjectIR = self[d.id][oldProject]
+    let t = oldProjectIR.result!.ast
+    let ramp = Function.ID(projectionRamp: (oldProjectIR as! Project).callee)
+    let rampReference = FunctionReference(to: ramp, in: self)
+
+    let plateauReference = FunctionReference(to: p, in: self)
```
```suggestion
    let plateauReference = FunctionReference(to: p, in: self)
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,215 @@ extension IR.Program {
...
+
+    // Remove old instructions...
+    var toRemove = [oldProject]
+    toRemove.append(contentsOf: d.regions[index].before)
+    toRemove.append(oldEndProject)
+    toRemove.append(contentsOf: d.regions[index].tail)
+    // ... including later `dealloc_stack`, corresponding to `alloc_stack` in our region.
+    let deallocs = toRemove.compactMap({ i in
+      self[d.id].allUses(of: i).map({ $0.user }).filter({ self[d.id][$0] is DeallocStack })
+    }).flatMap({ $0 })
```
kyouko-taiga: I think this would be better expressed with `reduce(into:_:)` and/or with loops.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,215 @@ extension IR.Program {
...
+
+}
+
+extension Emitter {
+
+  /// Allocates and initializes a continuation to call `plateau` with frame pointer `f`.
+  fileprivate mutating func _plateau_continuation(
+    calling plateau: FunctionReference, frame f: Operand
+  ) -> Operand
+  {
```
```suggestion
  ) -> Operand {
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,215 @@ extension IR.Program {
...
+    continuation: Operand
+  ) {
+    let x0 = _access(.let, from: continuation)
+    let x1 = _access(.set, from: _alloc_stack(t))
+    let x2 = _alloc_stack(.void)
+    _emitApply(Operand.constant(r), to: arguments + [x0, x1], writingResultTo: x2)
+    _end_access(x1)
+    _end_access(x0)
+  }
+}
```
kyouko-taiga, line 190:
```suggestion
  }

}
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,215 @@ extension IR.Program {
...
+
+/// The details needed to elaborate a scope using projections.
+private struct ScopeDetails {
+
+  /// The identifier for the original caller function to be elaborated.
+  let id: Function.ID
+
+  /// All the blocks that need to be elaborated.
+  /// Blocks are in dominance order.
```
```suggestion
  /// All the blocks that need to be elaborated, which are in dominance order.
```

I think we should document what a dominance order is somewhere. Maybe together with `DominatorTree`?

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -24,23 +26,264 @@ extension IR.Program {
...
   ) {
-    generateRamp(for: f, skeleton: s, in: m)
-    generateSlide(for: f, skeleton: s, in: m)
-    // TODO: delete old function
+    if modules[m]![f].site.file.baseName != "projection" { return }  // TODO: remove this filter.
+
+    let ramp = modules[m]!.demandProjectionRampDeclaration(for: f)
+    let slide = modules[m]!.demandProjectionSlideDeclaration(for: f)
+
+    let source = modules[m]![f]
```
kyouko-taiga:
```suggestion
    let source = modules[m]![f]
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -24,23 +26,264 @@ extension IR.Program {
...
+    // If all the slide blocks contain just terminators, the whole slide is empty.
+    if modules[m]!.slideIsEmpty(projection: d) {
+      modules[m]!.generateEmptyBody(for: slide, copying: d.id)
+      return
+    }
+
+    // Create the entry block.
+    let slideEntry = modules[m]!.createSlideEntryBlock(slide: slide, for: d)
+
+    var transformer = DictionaryInstructionTransformer()
```
kyouko-taiga:
```suggestion
    var transformer = DictionaryInstructionTransformer()
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -24,23 +26,264 @@ extension IR.Program {
...
-  /// Generate slide function corresponding to `f`, with skeleton `s`, in `m`.
-  private mutating func generateSlide(for f: Function.ID, skeleton s: ProjectionSkeleton, in m: Module.ID) {
-    let s = modules[m]!.demandProjectionSlideDeclaration(for: f)
-    // TODO
-    modules[m]!.generateEmptyBody(for: s, copying: f)
+  /// Creates entry block in `slide`, copying the scope from the projection `d`, returning its identity.
+  fileprivate mutating func createSlideEntryBlock(
+    slide: Function.ID, for d: ProjectionDetails
+  ) -> Block.ID
+  {
```
kyouko-taiga:
```suggestion
  ) -> Block.ID {
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -24,23 +26,264 @@ extension IR.Program {
...
+  }
+
+}
+
+extension Emitter {
+
+  /// Allocates and initializes a continuation to call `slide` with frame pointer `f`.
+  fileprivate mutating func _slide_continuation(calling slide: FunctionReference, frame f: Operand)
+    -> Operand
+  {
```
kyouko-taiga:
```suggestion
  fileprivate mutating func _slide_continuation(
    calling slide: FunctionReference, frame f: Operand
  ) -> Operand {
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -24,23 +26,264 @@ extension IR.Program {
...
+  /// May contain instructions from different blocks.
+  let rampInstructions: [InstructionID]
+
+  /// All the instructions in the slide.
+  /// May contain instructions from different blocks.
+  let slideInstructions: [InstructionID]
+
+  /// An instance describing projection `p`, with IR `source` and skeleton `s`.
+  init(_ p: Function.ID, source: Function, skeleton s: ProjectionSkeleton, of program: TypedProgram)
+  {
```
kyouko-taiga:
```suggestion
  init(
    _ p: Function.ID, source: Function, skeleton s: ProjectionSkeleton, of program: TypedProgram
  ) {
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -24,23 +26,264 @@ extension IR.Program {
...
+      + yieldBlocksSplits.flatMap { (y, split) in split.before + split.tail }
+    self.slideInstructions =
+      s.slideBlocks.flatMap { source.instructions(in: $0) }
+      + yieldBlocksSplits.flatMap { (y, split) in split.after }
+  }
+
+  /// Returns the yield block splits for projection with skeleton `s`, and IR `source`.
+  private static func yieldBlocksSplits(_ s: ProjectionSkeleton, in source: Function)
+    -> [InstructionID: SplitTriplet]
+  {
```
kyouko-taiga:
```suggestion
  private static func yieldBlocksSplits(
    _ s: ProjectionSkeleton, in source: Function
  ) -> [InstructionID: SplitTriplet] {
```

Sources/IR/DictionaryInstructionTransformer.swift
```diff
@@ -0,0 +1,90 @@
...
+  /// Returns a transformed copy of `o`.
+  func transform(_ o: Operand) -> Operand {
+    switch o {
+    case .constant(let c):
+      return .constant(c)
+    case .parameter(let b, let i):
+      if let rewritten = rewrittenParameters[o] {
+        return rewritten
+      }
+      return .parameter(rewrittenBlocks[b]!, i)
```
kyouko-taiga:
```suggestion
      return rewrittenParameters[o] ?? .parameter(rewrittenBlocks[b]!, i)
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: This doesn't work.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: added on my TODO list. I would do this directly on the new frontend

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: Will investigate.
But I'm curious, why you prefer loops rather than algorithms?

kyouko-taiga: It's more nuanced than that. First I find that:
- `map` into `filter` is better expressed with `compactMap`
- compositions of `compactMap` are better expressed with `flatMap` or `reduce`

Then, I think there are many cases where a `for` loop is simpler than the equivalent written in an fp-style. But also I think `for ... in` is actually not an "open loop" as Sean would call it (or I disagree if he thinks it is). I consider that construct to be sugar for `forEach` with better control-flow support.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+  ///
+  /// - Requires:
+  ///   * `!a.isEmpty`.
+  ///   * `!isSplit(a.last!)`
+  ///   * if `isSplit` returns `true` for `n` instructions, the returned array has
+  ///   `n + 1` elements.
+  /// - Postcondition: all elements in `a` appear in exactly one part of the returned array.
+  /// - Postcondition: if `p` is the last returned part, `p.splitPoint` is `nil` and `p.tail` is empty.
+  internal func split(
+    instructions a: [InstructionID], when isSplit: (InstructionID) -> Bool
```
kyouko-taiga:
```suggestion
    instructions a: [InstructionID], where isSplit: (InstructionID) -> Bool
```
For consistency with Swift's usual naming convention.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
+extension Function {
+
+  /// Splits `a` at all the points where `isSplit` returns `true`, returning the parts.
```
kyouko-taiga:
```suggestion
  /// Splits `a` at all the points where `isBoundary` returns `true`, returning the parts.
```

It is not immediately clear to me what is means for an instruction to be (a) "split". Does the term "boundary" works?

Another question that doesn't seem answered in the docs is *where* exactly the split occurs. For example if I have `[a, b, c]` and `isSplit(x)` iff `x == b` then what are the parts?

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+  /// - Postcondition: if `p` is the last returned part, `p.splitPoint` is `nil` and `p.tail` is empty.
+  internal func split(
+    instructions a: [InstructionID], when isSplit: (InstructionID) -> Bool
+  ) -> [SplitPart] {
+    precondition(!a.isEmpty)
+    precondition(!isSplit(a.last!), "final instruction cannot be a split point")
+    var parts: [SplitPart] = []
+    var remaining = a.dropFirst(0)
+    while let splitPoint = remaining.first(where: isSplit) {
+      let (splitIndex, tailEndIndex) = splitIndices(in: remaining, splittingAt: splitPoint)
```
kyouko-taiga:
```suggestion
      let (splitIndex, tailEnd) = splitIndices(in: remaining, splittingAt: splitPoint)
```
The name `tailEnd` is enough to convey the fact that it's a position IMO. I would probably also use just `split` rather than `splitIndex` but again I'm wondering if `boundary` is not a more appropriate name.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+  /// - Requires:
+  ///   * `!a.isEmpty`.
+  ///   * `!isSplit(a.last!)`
+  ///   * if `isSplit` returns `true` for `n` instructions, the returned array has
+  ///   `n + 1` elements.
+  /// - Postcondition: all elements in `a` appear in exactly one part of the returned array.
+  /// - Postcondition: if `p` is the last returned part, `p.splitPoint` is `nil` and `p.tail` is empty.
+  internal func split(
+    instructions a: [InstructionID], when isSplit: (InstructionID) -> Bool
+  ) -> [SplitPart] {
```
kyouko-taiga: Any reason why this API is returning an array of parts rather than an array of positions? I think it would make the meaning of the operation simpler to understand because the meaning of a position in an array is typically well understood.

I suspect this approach would address the issue about not being sure where the split occur.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+    parts.append(SplitPart(before: Array(remaining), splitPoint: nil, tail: []))
+    return parts
+  }
+
+  /// Splits `a` at the instruction `splitPoint`, returning the three parts (before, tail, after).
+  ///
+  /// - Precondition: `splitPoint` is contained in `a`.
+  /// - Precondition: there is at least one instruction after `splitPoint` in `a`.
+  /// - Postcondition: `r.before + [splitPoint] + r.tail + r.after == a`
+  internal func split(instructions a: [InstructionID], at splitPoint: InstructionID) -> SplitTriplet {
```
kyouko-taiga: So I suppose the goal is to create a kind of positionless API for array of instructions? Why is that better than using positions here?

I'm asking because my knee-jerk reaction to these methods is that they seem to introduce a lot of complexity for something that on the surface seems expressible with positions more concisely.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+
+    // At least the block terminator must be outside the tail, so `tailEndIndex` is never `nil`.
+    return (splitIndex!, tailEndIndex!)
+  }
+
+  /// Returns `true` if the instruction `i` can be part of the yield tail,
+  /// e.g., it appears after a yield, but still belong to the ramp of the projection.
+  ///
+  /// Tail instructions: `MarkState` (uninitialized), `DeallocStack`, `EndAccess`.
+  private static func canBeTail(_ i: Instruction) -> Bool {
```
kyouko-taiga:
```suggestion
  private static func mayBeTail(_ i: Instruction) -> Bool {
```
For consistency with the name of other methods in the source.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+
+}
+
+/// One contiguous region produced when a block's instructions are split at one or more split
+/// points.
+///
+/// A `SplitPart` groups together the instructions that logically precede the next split point as
+/// well as the "tail" instructions that must stay with them even though they appear afterward in
+/// the original block. Both `before` and `tail` preserve the original ordering. It also records
+/// the `splitPoint` instruction at which the split occurs; this is `nil` for the last part.
```
kyouko-taiga: Given this description I would have expressed a `SplitPart` as a triple of indices. Two for the lower and upper bounds of the "before" part and one for the end of the "tail" part. I'm assuming that the "split point" is always after the "before. Otherwise I would add additional indices.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+///
+/// The parts are:
+/// - `before`: the instructions before the split point.
+/// - `tail`: the instructions physically after the split point, but logically belonging to before.
+/// - `after`: the instructions after the split point.
+internal struct SplitTriplet {
+
+  /// The instructions before the split point.
+  internal let before: [InstructionID]
+  /// The instructions physically after the split point, but logically belonging to before.
```
kyouko-taiga:
```suggestion

  /// The instructions physically after the split point, but logically belonging to before.
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,119 @@
...
+  internal let tail: [InstructionID]
+  /// The instructions after the split point.
```
kyouko-taiga:
```suggestion

  /// The instructions after the split point.
```

kyouko-taiga (commented):
I have a few questions about the design of the splitting API that I think needs answers before I can continue to carefully review the rest of the PR. Happy to have synchronous discussion if necessary.

Sources/IR/Analysis/Function+Split.swift
lucteo: The main reason for that is that (and actuall,y for this entire file) is the need to have the distinction between "before", "tail", and "after". Without having a "tail", things would have been far simpler.

With the presence of "tail" instructions, there isn't a clear "split position".
I could switch the interface to return positions, but I feel that would complicate the callers.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,87 @@
+/// The position, including epilogue information, at which a sequence of instructions is split.
+/// 
+/// Physically, a split point divides an array of instructions into two parts:
+///   - the instructions at indices `< splitPoint` -- the "front"
```
kyouko-taiga, line 10: What does "physically" mean?
Also, why the back is physical but not the front?

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,87 @@
...
+///   - the instructions at indices `> splitPoint` -- the physical "back"
+/// 
+/// But, there are also instructions that are logically part of the front, even though they
```
kyouko-taiga:
```suggestion
/// There are also instructions that are logically part of the front, even though they
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,87 @@
...
+/// appear after the split point. These are the "epilogue" instructions for a split point.
+/// The `epilogueEnd` index indicates where the epilogue ends and the back begins.
+/// 
+/// Thus, a split point defines three contiguous regions of instructions:
+///   - the "front" instructions at indices `< splitPoint`
+///   - the "epilogue" instructions at indices in range `splitPoint ..< epilogueEnd`
+///   - the "back" instructions at indices `>= epilogueEnd`
```
kyouko-taiga: I would open with that but I think the term "splitPoint" is wrong, no? Below we see that `splitPoint` is a single index. IIUC you meant that "a split position defines three contiguous regions".

Also because there is more than one position, I think this structure should be called `SplitPositions` (plural).

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,87 @@
+/// The position, including epilogue information, at which a sequence of instructions is split.
```
kyouko-taiga:
```suggestion
/// The positions, including epilogue information, at which a sequence of instructions is split.
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,87 @@
...
+  internal let splitPoint: Int
+
+  /// The index of the first instruction that is not part of the epilogue.
+  internal let epilogueEnd: Int
+}
+
+extension Function {
+
+  /// Splits `a`, instructions pointing into `self`, at all the points where `isSplitPoint` returns
+  /// `true`, returning the split positions.
```
kyouko-taiga:
```suggestion
  /// `true`, returning the splits.
```

It doesn't help to make a distinction between a split and a split position in the docs here. We also tend to say "return X" instead of "return the identity/position of X" in almost all other places.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,87 @@
...
+    instructions a: [InstructionID], at splitPoint: InstructionID
+  ) -> SplitPosition {
+    let splitPointIndex = a.firstIndex(of: splitPoint)
+    precondition(splitPointIndex != nil, "split point \(splitPoint) not found")
+    let remaining = a.suffix(from: splitPointIndex!).dropFirst()
+    let epilogueEnd = remaining.firstIndex(where: { !Self.mayBeEpilogue(self[$0]) }) ?? a.count
+    return SplitPosition(splitPoint: splitPointIndex!, epilogueEnd: epilogueEnd)
+  }
+
+  /// Splits instructions in `b` at `splitPoint`, returning the split position.
```
kyouko-taiga:
```suggestion
  /// Splits instructions in `b` at `splitPoint`, returning the positions of the resulting partitions.
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,250 @@ extension IR.Program {
 
   /// Elaborates the projection caller function `f` in module `m`.
   private mutating func elaborateProjectionCaller(_ f: Function.ID, in m: Module.ID) {
-    // TODO: properly implement this
-    let p = modules[m]!.demandCallerPlateauDeclaration(for: f, region: 0)
-    modules[m]!.generateEmptyBody(for: p, copying: f)
+    if modules[m]![f].site.file.baseName != "projection" { return }  // TODO: remove this filter.
```
kyouko-taiga: "projection" is very likely to collide with something else. If you intend to merge this PR to main I think you should pick another filename.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,250 @@ extension IR.Program {
...
+    toRemove.append(contentsOf: d.instructions(region: index))
+    // ... including later `dealloc_stack`, corresponding to `alloc_stack` in our region.
+    let deallocs = toRemove.reduce(into: [InstructionID]()) { r, i in
+      let users = self[d.id].allUses(of: i).map(\.user)
+      r.append(contentsOf: users.filter { self[d.id][$0] is DeallocStack })
+    }
+    toRemove.append(contentsOf: deallocs.filter({ !toRemove.contains($0) }))
+    for i in toRemove.uniqued().reversed() {
+      self[d.id].remove(i)
+    }
```
kyouko-taiga, line 121: This looks very complicated.

```suggestion
    var toRemove = [oldProject]
    var deallocs: [InstructionID] = []
    for i in d.instructions(region: index) {
      toRemove.append(i)
      for u in self[d.id].allUses(of: i) where self[d.id][u.user] is DeallocStack {
        deallocs.append(u.user)
      }
    }
    for i in [toRemove, deallocs].joined().reversed() {
      self[d.id].remove(i)
    }
```

There are ways to make it less imperative but I'm not sure that would help and that would probably be less efficient.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,250 @@ extension IR.Program {
...
+  }
+
+  /// Creates storage in `f` for the value represented by `i` and returns its identity.
```
```suggestion
  /// Creates storage in `f` for the value represented by `i` and returns its address.
```

You're not really returning the identity of the storage. Rather you're returning the identity of the instruction that defines its address.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,250 @@ extension IR.Program {
...
+  ) -> InstructionID {
+    let t = i.result!.ast
+    let r = modifyIR(of: f, at: .end(of: b)) { (e) in
+      e._alloc_stack(t)
+    }
+    return r.instruction!
+  }
+
+  /// Generates the code in block `b` of `f` to call the continuation received as the second
+  /// parameter to `f`, and then return.
```
```suggestion
  /// parameter to `f`, followed by a return instruction.
```

The wording suggests that the return is about the Swift function.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,250 @@ extension IR.Program {
...
+  fileprivate mutating func generateContinuationCall(
+    toBlock b: Block.ID, of f: Function.ID
+  ) {
+    let entryBlock = self[f].entry!
```
kyouko-taiga, line 140: I would probably inline this binding.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,250 @@ extension IR.Program {
...
+    let x2 = _access(.set, from: _subfield_view(x1, at: [0, 0]))
+    _store(x0, x2)
+    _end_access(x2)
+    let x3 = _access(.set, from: _subfield_view(x1, at: [0, 1]))
+    _store(f, x3)
+    _end_access(x3)
+    return x1
+  }
+
+  /// Emit code that jumps to projection continuation `c`.
```
```suggestion
  /// Generates IR for jumping to projection continuation `c`.
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -24,23 +26,256 @@ extension IR.Program {
...
+    let yieldInsertionPoint: InsertionPoint
+    if let x = d.firstTailInstructions[y]! {
+      yieldInsertionPoint = .before(t.rewrittenInstructions[x]!)
+    } else {
+      yieldInsertionPoint = .end(of: b)
+    }
+    modifyIR(of: ramp, at: yieldInsertionPoint) { (e) in
+      // If we have multiple yields, store the index of the current yield point in the state.
+      if d.skeleton.yieldPoints.count > 1 {
+        let _ /*index*/ = d.skeleton.yieldPoints.firstIndex(of: y)!
```
kyouko-taiga:
```suggestion
        _ /*index*/ = d.skeleton.yieldPoints.firstIndex(of: y)!
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: I'm confused. Why isn't `InstructionID` an identity?
The doumentation of the type says:
```
/// The stable identity of an instruction in its function.
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: If I do that, I get an error:
> error: overlapping accesses to 'self', but modification requires exclusive access; consider copying to a local variable

dabrahams:
A for…in not encapsulated in an algorithm is indeed a raw loop. There’s nothing to disagree  with; that’s the definition.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
kyouko-taiga: When you say "its identity" it seams you're returning the identity of the storage, which is confusing. Yes, you are returning the identity of an instruction, but storage is not an instruction.

kyouko-taiga:
> A for…in not encapsulated in an algorithm is indeed a raw loop. There’s nothing to disagree  with; that’s the definition.

Fair enough. The rest of my argument stands, though: I don't believe replacing `for .. in { ... }` by sophisticated compositions of `map` and `filter` helps.

I think I don't like the "no raw loop" argument because it sounds too much like other blanket statements with which I disagree (e.g., "no global variables" or "no goto"). I am skeptical that there exist any construction that has no place in any program. That isn't to say that I think there's no value to Sean's argument–I certainly agree with Sean's points now that I [read them](https://sean-parent.stlab.cc/2018/12/30/cpp-ruminations.html)–only that the slogan rubs be in the wrong way.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
+/// The positions, including epilogue information, at which a sequence of instructions is split.
+/// 
+/// This defines three contiguous regions of instructions:
```
kyouko-taiga:
```suggestion
/// An instance defines three contiguous regions of instructions:
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
+/// The positions, including epilogue information, at which a sequence of instructions is split.
```
kyouko-taiga:
```suggestion
/// The positions at which a sequence of instructions is split.
```
The term only makes sense after it has been defined. I would remove it to avoid confusing the reader who goes through the headlines.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
...
+/// 
+/// This defines three contiguous regions of instructions:
+///   - the "front" instructions at indices `< splitPoint`
+///   - the "epilogue" instructions at indices in range `splitPoint ..< epilogueEnd`
+///   - the "back" instructions at indices `>= epilogueEnd`
```
kyouko-taiga:
```suggestion
///   - the "front", at positions in the range `..<splitPoint`
///   - the "epilogue", at positions in the range `splitPoint ..< epilogueEnd`
///   - the "back", at positions in the range `epilogueEnd...`
```

Sources/IR/Analysis/Function+Split.swift
kyouko-taiga: I still don't understand why the front is not physical.

kyouko-taiga: Is it about those instructions that are part of the epilogue? If so I would not use the term "physical". Also, I think the distinctions between the three parts have been made just above so I'm not sure this precision adds anything.

If the goal is to clarify the definitions, perhaps we should consider adding an example.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
...
+  /// The index of the instruction at which the split occurs.
+  internal let splitPoint: Int
+
+  /// The index of the first instruction that is not part of the epilogue.
+  internal let epilogueEnd: Int
+}
+
+extension Function {
+
+  /// Splits `a`, instructions pointing into `self`, at all the points where `isSplitPoint` returns
```
kyouko-taiga:
```suggestion
  /// Splits `a`, which is a contiguous sequence of instructions in `self`, at all the points where `isSplitPoint` returns
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
...
+  /// `true`, returning the splits.
+  ///
+  /// This is different from `Array.split(separator:)` as it deals with "epilogues". An epilogue is
```
kyouko-taiga:
```suggestion
  /// This method is different from `Array.split(separator:)` as it deals with "epilogues". An epilogue is
```

Sources/IR/Analysis/Function+Split.swift
kyouko-taiga: I am not of fan of the phrase "pointing into `self`".

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
...
+      r.append(SplitPositions(splitPoint: splitPointIndex, epilogueEnd: epilogueEnd))
+
+      remaining = remaining.suffix(from: epilogueEnd)
+    }
+    return r
+  }
+
+  /// Splits `a`, instructions pointing into `self`, at `splitPoint`, returning the split position.
+  ///
+  /// This deals with the epilogue of `splitPoint`.
```
kyouko-taiga:
```suggestion
  /// This method deals with the epilogue of `splitPoint`.
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
...
+    let splitPointIndex = a.firstIndex(of: splitPoint)
+    precondition(splitPointIndex != nil, "split point \(splitPoint) not found")
+    let remaining = a.suffix(from: splitPointIndex!).dropFirst()
+    let epilogueEnd = remaining.firstIndex(where: { !Self.mayBeEpilogue(self[$0]) }) ?? a.count
+    return SplitPositions(splitPoint: splitPointIndex!, epilogueEnd: epilogueEnd)
+  }
+
+  /// Splits instructions in `b` at `splitPoint`, returning the positions of the resulting partitions.
+  ///
+  /// This deals with the epilogue of `splitPoint`.
```
kyouko-taiga:
```suggestion
  /// This method deals with the epilogue of `splitPoint`.
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
...
+  ///
+  /// See also `SplitPositions`.
+  internal func split(block b: Block.ID, at splitPoint: InstructionID) -> SplitPositions {
+    split(instructions: Array(instructions(in: b)), at: splitPoint)
+  }
+
+  /// Returns `true` if the instruction `i` can be part of the yield tail,
+  /// e.g., it appears after a yield, but still belong to the ramp of the projection.
```
kyouko-taiga:
```suggestion
  /// e.g., it appears after a yield, but still belongs to the ramp of the projection.
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -17,9 +20,250 @@ extension IR.Program {
...
+      e._return()
+    }
+  }
+
+}
+
+extension Emitter {
+
+  /// Allocates and initializes a continuation to call `plateau` with frame pointer `f`.
+  fileprivate mutating func _plateau_continuation(
```
```suggestion
  fileprivate mutating func _plateauContinuation(
```

The current naming convention (which may not be ideal) is to use snake case only for instruction constructors.

kyouko-taiga (changes requested):
Great work. We're getting there. I just think we can improve the docs a little more.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
...
+}
+
+extension Function {
+
+  /// Splits `a`, which is a contiguous sequence of instructions in `self`, at all the points where `isSplitPoint` returns
+  /// `true`, returning the splits.
+  ///
+  /// This method is different from `Array.split(separator:)` as it deals with "epilogues". An epilogue is
+  /// a contiguous sequence of instructions that appear immediately after a split point but
+  /// logically belong before the split point.
```
kyouko-taiga:
```suggestion
  /// a contiguous sequence of instructions that appears immediately after a split point but
  /// logically belongs before the split point.
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
...
+    var remaining = a.dropFirst(0)
+    while let splitPointIndex = remaining.firstIndex(where: isSplitPoint) {
+      remaining = a.suffix(from: splitPointIndex).dropFirst()
+      let epilogueEnd = remaining.firstIndex(where: { !Self.mayBeEpilogue(self[$0]) }) ?? a.count
+      r.append(SplitPositions(splitPoint: splitPointIndex, epilogueEnd: epilogueEnd))
+
+      remaining = remaining.suffix(from: epilogueEnd)
+    }
+    return r
+  }
```
kyouko-taiga, line 47:
```suggestion
  internal func split(
    instructions xs: [InstructionID], where isSplitPoint: (InstructionID) -> Bool
  ) -> [SplitPositions] {
    var splits: [SplitPositions] = []    
    var end = 0
    while let p = xs[end...].firstIndex(where: isSplitPoint) {
      end = xs[(p + 1)...].firstIndex(where: { !self[$0].mayBeEpilogue }) ?? xs.count
      splits.append(SplitPositions(splitPoint: p + 1, epilogueEnd: end))
    }
    return splits
  }
```

I think we can do simpler using positions directly rather than slices. Also look at the extension of `Instruction` for the definition of the property that's used in the closure.

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
...
+  /// e.g., it appears after a yield, but still belongs to the ramp of the projection.
+  ///
+  /// Tail instructions: `MarkState` (uninitialized), `DeallocStack`, `EndAccess`.
+  private static func mayBeEpilogue(_ i: Instruction) -> Bool {
+    if let x = i as? MarkState {
+      return !x.initialized
+    } else {
+      return i is EndAccess || i is DeallocStack
+    }
+  }
```
kyouko-taiga: I think function might be better expressed as a property of `Instruction`:

```swift
extension Instruction {

  /// Returns `true` if the instruction `i` can be part of the yield tail,
  /// e.g., it appears after a yield, but still belongs to the ramp of the projection.
  ///
  /// Tail instructions: `MarkState` (uninitialized), `DeallocStack`, `EndAccess`.
  var mayBeEpilogue: Bool {
    (self as? MarkState)?.initialized || (self is EndAccess) || (self is DeallocStack)
  }

}
```

Sources/IR/Analysis/Function+Split.swift
```diff
@@ -0,0 +1,88 @@
...
+    }
+    return r
+  }
+
+  /// Splits `a`, instructions pointing into `self`, at `splitPoint`, returning the split position.
+  ///
+  /// This method deals with the epilogue of `splitPoint`.
+  ///
+  /// See also `SplitPositions`.
+  internal func split(
```
kyouko-taiga, line 54: I suspect this implementation can be improved similarly to the one above.

lucteo:
> There is one question left unanswered about the "physical" back. If you can answer it I think we can converge a the final form for the docs of the splits.

I couldn't find the question (Github seems to be terrible at showing comments nowadays), but I did find a "physical" in one place and removed it. Thank you very much!

PR #1848, opened 2026-02-23
Projected value passed as a parameter to the caller plateau.

lucteo:
Previously, the projected value was assumed to be sent as a `set` parameter to projection's ramp. It requires allocating it in the caller's frame, but the caller's ramp and slide don't need it. A better approach is to pass it as an `inout` parameter to the plateau.

The following code now works:
```hylo
subscript neg(_ x: inout Int): Int {
  {
    let y = -x
    print_projection_ramp()
    yield y
  }
  print_projection_slide()
}

fun plateau_check(_ x: Int) {
  print("check in plateau")
  precondition(x == -42)
}

fun caller() {
  var x = 42
  print_caller_ramp()
  let n = neg[&x]
  plateau_check(n)
  _ = n.copy()
  print_caller_slide()
}

public fun main() {
  caller()
}

fun print_projection_ramp() {
  print("neg.ramp")
}
fun print_projection_slide() {
  print("neg.slide")
}
fun print_caller_ramp() {
  print("caller.ramp")
}
fun print_caller_slide() {
  print("caller.slide")
}
```


Sources/IR/Analysis/ProjectionElaboration.swift
```diff
@@ -86,12 +87,13 @@ extension Module {
...
   }
 
-  /// Returns the IR function representing a plateau region in caller `f`.
+  /// Returns the IR function representing a plateau region in caller `f` corresponding to the
+  /// projected value type `t`.
   ///
   /// Signature:
-  /// > fun Caller.plateauN(let Builtin.ptr, let ProjectionContinuation) -> {}
+  /// > fun Caller.plateauN(inout <yield-type>, let Builtin.ptr, let ProjectionContinuation) -> {}
   mutating func demandCallerPlateauDeclaration(
```
kyouko-taiga:
```suggestion
  mutating func demandPlateau(
```
Is there a plateau that isn't for the caller?

Also I think the name of the function and its return type already convey the fact that the result is a declaration.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -47,19 +47,19 @@ extension IR.Program {
     let source = modules[m]![d.id]
 
     // Generate the declaration.
-    let plateau = modules[m]!.demandCallerPlateauDeclaration(for: d.id, region: plateauIndex)
+    let oldProject = d.splitInstruction(endingRegion: index - 1)
+    let plateau = modules[m]!.demandCallerPlateauDeclaration(
+      for: d.id, region: plateauIndex, projectedType: source[oldProject].result!.ast
+    )
```
```suggestion
    let plateau = modules[m]!.demandCallerPlateauDeclaration(
      for: d.id, region: plateauIndex, projectedType: source[oldProject].result!.ast)
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -102,9 +102,9 @@ extension Module {
 
     modifyIR(of: d.id, at: .before(oldEndProject)) { (e) in
       let nullCall = e._call_builtin(.zeroinitializer(BuiltinType.ptr), [])
-      let c = e._plateauContinuation(calling: plateauReference, frame: nullCall)
+      let c = e._plateauContinuation(calling: plateauReference, frame: nullCall, projectedType: t)
       e._call_projection_ramp(
```
```suggestion
      e._callProjectionRamp(
```
I guess I missed it in the previous PR.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -132,14 +132,14 @@ extension Module {
     return r.instruction!
   }
 
-  /// Generates the code in block `b` of `f` to call the continuation received as the second
+  /// Generates the code in block `b` of `f` to call the continuation received as the third
   /// parameter to `f`, followed by a return instruction.
```
kyouko-taiga, line 135: What are the two first parameters?

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -132,14 +132,14 @@ extension Module {
...
   fileprivate mutating func generateContinuationCall(
     toBlock b: Block.ID, of f: Function.ID
   ) {
     let entryBlock = self[f].entry!
     modifyIR(of: f, at: .end(of: b)) { (e) in
-      e._resume_continuation(Operand.parameter(entryBlock, 1))
+      e._resume_continuation(Operand.parameter(entryBlock, 2))
```
```suggestion
      e._resumeContinuation(Operand.parameter(entryBlock, 2))
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -132,14 +132,14 @@ extension Module {
     return r.instruction!
   }
 
-  /// Generates the code in block `b` of `f` to call the continuation received as the second
+  /// Generates the code in block `b` of `f` to call the continuation received as the third
   /// parameter to `f`, followed by a return instruction.
   fileprivate mutating func generateContinuationCall(
     toBlock b: Block.ID, of f: Function.ID
   ) {
     let entryBlock = self[f].entry!
```
```suggestion
    let entry = self[f].entry!
```
The type says that it's a block, right? Also an "entry" is most likely a block anyway.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -173,17 +175,13 @@ extension Emitter {
     _end_access(x0)
   }
 
-  /// Emit code to call the projection ramp `r` with `arguments`, frame value type `t`,
-  /// and continuation `continuation`.
+  /// Emit code to call the projection ramp `r` with `arguments` and continuation `c`.
```
```suggestion
  /// Emits a call to the projection ramp `r` with `arguments` and continuation `c`.
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -125,10 +133,11 @@ extension Module {
   }
 
   /// Generates a new block in `ramp` that calls the continuation received as parameter, passing to
-  /// it a continuation that calls `slide`; returns the identity of the new block.
+  /// it the value from `p` and a continuation that calls `slide`; returns the identity of the new block.
```
kyouko-taiga:
```suggestion
  /// the value from `p` and a continuation that calls `slide`; returns the identity of the new block.
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -138,28 +147,19 @@ extension Module {
...
-  private func projectedValueParameter(ramp f: Function.ID) -> Operand {
-    let source = self[f]
-    let entry = source.entry!
-    return .parameter(entry, source[entry].inputs.count - 2)
-  }
-
   /// Returns the operand representing the continuation parameter in ramp `f`.
   ///
-  /// This is the previous-to-last parameter to `f`; block has the additional return type.
+  /// This is the last parameter to `f`; block has the additional return type.
```
kyouko-taiga: A little unclear to me.

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -138,28 +147,19 @@ extension Module {
...
   private func continuationParameter(ramp f: Function.ID) -> Operand {
     let source = self[f]
     let entry = source.entry!
-    return .parameter(entry, source[entry].inputs.count - 3)
+    return .parameter(entry, source[entry].inputs.count - 2)
```
kyouko-taiga, line 172: This kind of computation makes me a little worried. Shall we add constants somewhere to document what the numbers are?

kyouko-taiga (changes requested):

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: I'll keep this in mind. Paren on the same line if calling a function, and on a new line if part of the function declaration.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: I was wondering today about this. Can you please explain to me why these functions need to have a different convention (underscore + camel-case) from the ones in emitter (snake case)?

(there are other functions that I need to apply this to)

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: This is specified in the `demandCallerPlateauDeclaration` documentation:
> fun Caller.plateauN(inout <yield-type>, let Builtin.ptr, let ProjectionContinuation) -> {}

Sources/IR/Analysis/Program+ElaborateProjections.swift
lucteo: Note to self: also delete the "to" ending the previous line.

Sources/IR/Analysis/Program+ElaborateProjections.swift
lucteo: Last parameter of the function means: params.count - 1
But the block has and additional parameter, which is the return of the function. Thus this can be found at block.inputs.count - 2.

Maybe I should put the second part of the comment ("block has the additional return type") in the body of the function.

Sources/IR/Analysis/Program+ElaborateProjections.swift
lucteo: I was hoping to capture this in the definition of the function itself. 

BTW, I've seen this in multiple codebases, and I hate it: `let two = 2`

Sources/IR/Analysis/ProjectionElaboration.swift
lucteo: Note to self: also remove "Declaration" from the other two "demand" functions.

Sources/IR/Analysis/Program+ElaborateProjections.swift
kyouko-taiga: Ah, I see. So this function is acting as the abstraction. I guess that is fine.

BTW, I did not suggest to write `let two = 2`.
I was more thinking about some kind of enum that would define the offsets of the different kind of parameters:

```swift
enum ParameterOffset: Int {
  case continuation = 2
}
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
kyouko-taiga: I think it has its place here but it requires some more context; basically what you just answered as a comment.

I like to add examples in the detailed part of the docs when things are a little tricky and require some planet alignment.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
kyouko-taiga: I see a signature but no explanation of what the parameters are. Maybe I am missing something.

> ```
> /// Returns the IR function representing a plateau region in caller `f` corresponding to the
> /// projected value type `t`.
> ///
> /// Signature:
> /// > fun Caller.plateauN(inout <yield-type>, let Builtin.ptr, let ProjectionContinuation) -> {}
> ```

I think I understand what the first parameter is doing thanks to `<yield-type>` but what about the second? `Builtin.ptr` doesn't provide many clues.

Sources/IR/Analysis/ProjectionElaboration.swift
kyouko-taiga:
```
let inputs = [Parameter(decl: nil, type: ParameterType(.`let`, AnyType(BuiltinType.ptr)))]
```
That probably fits the line without the ascription.

kyouko-taiga (changes requested):
I'd just like some clarification about the parameters of the first argument of `demandPlateau` but otherwise I think it looks good!

Sources/IR/Analysis/Program+ElaborateProjections.swift
lucteo: I hope this gets resolved with the extra documentation I've added
(sorry, resolving the comments before submitting the next patch, to make sure I keep track of everything).

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -132,14 +131,17 @@ extension Module {
     return r.instruction!
   }
 
-  /// Generates the code in block `b` of `f` to call the continuation received as the second
+  /// Generates the code in block `b` of `f` to call the continuation received as the third
   /// parameter to `f`, followed by a return instruction.
+  ///
+  /// The plateau function for which this is called has the following signature:
+  /// > fun Caller.plateauN(inout <yield-type>, let Builtin.ptr, let ProjectionContinuation) -> {}
```
```suggestion
  /// The plateau function `f` has the following signature:
  ///
  ///     fun Caller.plateauN(inout <yield-type>, let Builtin.ptr, let ProjectionContinuation) -> {}
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -173,17 +177,13 @@ extension Emitter {
...
   }
 
-  /// Emit code to call the projection ramp `r` with `arguments`, frame value type `t`,
-  /// and continuation `continuation`.
-  fileprivate mutating func _call_projection_ramp(
-    _ r: FunctionReference, with arguments: [Operand], frameValueType t: AnyType,
-    continuation: Operand
+  /// Emits a call to the projection ramp `r` with `arguments` and continuation `c`.
+  fileprivate mutating func _callProjectionRamp(
+    _ r: FunctionReference, with arguments: [Operand], continuation c: Operand
```
```suggestion
  /// Emits a call to the projection ramp `r` with `arguments` and `continuation`.
  fileprivate mutating func _callProjectionRamp(
    _ r: FunctionReference, with arguments: [Operand], continuation: Operand
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -44,17 +44,25 @@ extension IR.Program {
     in m: Module.ID
   ) {
     var transformer = DictionaryInstructionTransformer()
+    let source = modules[m]![d.id]
 
     // Copy the ramp instructions, creating blocks for them as needed.
     rewrite(d.rampInstructions, in: d.id, from: m, transformedBy: &transformer, to: ramp)
 
+    // Add projected value storage at the beginning of the entry block
```
kyouko-taiga:
```suggestion
    // Add the projected value storage at the beginning of the entry block.
```

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -44,17 +44,25 @@ extension IR.Program {
...
+    let entry = modules[m]![ramp].entry!
+    let projectedValueStorage = modules[m]!.modifyIR(of: ramp, at: .start(of: entry)) { (e) in
```
kyouko-taiga:
```suggestion
    let s = modules[m]!.modifyIR(of: ramp, at: .start(of: entry)) { (e) in
```
It seems pretty clear to me what the variable stands for in the remainder of the code and perhaps a shorter name gives you a chance to avoid line breaks.

Sources/IR/Analysis/Program+ElaborateProjections.swift
```diff
@@ -137,29 +146,27 @@ extension Module {
...
-    let source = self[f]
-    let entry = source.entry!
-    return .parameter(entry, source[entry].inputs.count - 2)
-  }
-
   /// Returns the operand representing the continuation parameter in ramp `f`.
   ///
-  /// This is the previous-to-last parameter to `f`; block has the additional return type.
+  /// The ramp function has the following signature:
+  /// > fun Projection.ramp(<parameters>, let PlateauContinuation) -> {}
```
kyouko-taiga:
```suggestion
  ///     fun Projection.ramp(<parameters>, let PlateauContinuation) -> {}
```

That style creates better formatting in Xcode (and I think VSCode too).

Sources/IR/Analysis/ProjectionElaboration.swift
```diff
@@ -35,17 +36,22 @@ extension Module {
   /// Returns the IR function representing the ramp of projection `f`.
   ///
   /// Signature:
-  /// > fun Projection.ramp(<parameters>, let CallerContinuation, set <yield-type>) -> {}
-  mutating func demandProjectionRampDeclaration(for f: Function.ID) -> Function.ID {
+  /// > fun Projection.ramp(<parameters>, let PlateauContinuation) -> {}
+  ///
+  /// Generated function parameters:
+  ///   - `<parameters>`: the original parameter of the projection
+  ///   - `let PlateauContinuation`: the continuation used to resume the flow in the plateau,
```
kyouko-taiga: I would add actual parameter names or use the same syntax for both parameters. It seems odd to me that the first parameter is expressed as a name in angle brackets whereas the second is just written as a type.

Sources/IR/Analysis/ProjectionElaboration.swift
```diff
@@ -86,12 +93,19 @@ extension Module {
...
   /// Signature:
-  /// > fun Caller.plateauN(let Builtin.ptr, let ProjectionContinuation) -> {}
-  mutating func demandCallerPlateauDeclaration(
-    for f: Function.ID, region index: Int
+  /// > fun Caller.plateauN(inout <yield-type>, let Builtin.ptr, let ProjectionContinuation) -> {}
+  ///
+  /// Generated function parameters:
+  ///   - `inout <yield-type>`: the projected value.
+  ///   - `let Builtin.ptr`: the address of the caller frame.
+  ///   - `let ProjectionContinuation`: the continuation used to resume to the projection's slide,
```
kyouko-taiga: Not a fan of referring to parameters only by their type. This convention will fail in many cases.

kyouko-taiga (changes requested):

Sources/IR/Analysis/ProjectionElaboration.swift
lucteo: the `<parameters>` is a placeholder for any parameters that the projection might have. That varies from projection to projection, so I can't properly document them.

I can add a name to the last parameter.

Sources/IR/Analysis/ProjectionElaboration.swift
lucteo: I will add names to these parameters here. (here I know how many parameters I have, but I just don't now one type.

kyouko-taiga (approved):

PR #1849, opened 2026-02-28
Upgrade to LLVM v2 api

tothambrus11:
This PR upgrades the SwiftyLLVM dependency and usages.

The major change is that now `UnsafeReference<T>` is returned instead of `T`. To access `T`'s fields, we must explicitly write `myReference.pointee.someField` or `myReference.with{ $0.someField }`. Initially, [I used the `with` version everywhere](https://github.com/hylo-lang/hylo/compare/main...d114c9aa133f23c411d8c221c997f07961687799#diff-4306f9cc025ba18eb1a6954f99f924f4a1075acf9c3b51ae0bd655fb4c6301fdR491). This added unnecessary syntactic noise, and didn't appear to give any benefits over `pointee`. In the last commit, I changed all usages to `.pointee`.

Since Module is now non-copyable, I had to wrap it in a class so that we can maintain a collection of them. Not the nicest, but it's the best to expose non-copyable types in the library, and remove this wrapper hack when we get non-copyable collections in Swift.

*The CI is failing in CMake because I didn't yet update CMakeModules' dependency references, but I will fix that after review and potential follow-up commits)*

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -44,7 +44,7 @@ on:
 env:
   spm-build-options: -Xswiftc -enable-testing --explicit-target-dependency-import-check error
   spm-test-options: --parallel
-  swift-version: '6.2.0'
+  swift-version: '6.2.3'
   HYLO_LLVM_BUILD_RELEASE: 20251207-115516
   HYLO_LLVM_VERSION: '20.1.6'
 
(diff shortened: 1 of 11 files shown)


tothambrus11:
This was good as a proof of concept, but I'm closing this now so we don't have to review/fix/maintain this with no clear benefit (hylo-new already uses the v2 API)

PR #1853, opened 2026-03-06
Use case-insensitive PATH lookup

inkyo08:
On Windows, environment variable names are case-insensitive at the OS level,
but Swift's Dictionary lookup is case-sensitive. This caused `findExecutable`
to fail in environments like MSYS2, which sets the PATH variable as "PATH"
rather than the native Windows convention "Path".

This change looks up the PATH key case-insensitively, so executable discovery
works correctly regardless of the environment.

### DIFF
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -540,7 +540,11 @@ public struct Driver: ParsableCommand, Sendable {
       ? invocationName : invocationName + Host.executableSuffix
 
     // Search in the PATH.
-    let path = ProcessInfo.processInfo.environment[Host.pathEnvironmentVariable] ?? ""
+    // Use case-insensitive key lookup to support environments like MSYS2 that use "PATH"
+    // instead of the native Windows "Path".
+    let env = ProcessInfo.processInfo.environment
+    let pathKey = env.keys.first(where: { $0.caseInsensitiveCompare(Host.pathEnvironmentVariable) == .orderedSame }) ?? Host.pathEnvironmentVariable
+    let path = env[pathKey] ?? ""
     for root in path.split(separator: Host.pathEnvironmentSeparator) {
       let candidate = URL(fileURLWithPath: String(root)).appendingPathComponent(executableFileName)
       if FileManager.default.fileExists(atPath: candidate.fileSystemPath) {
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -540,10 +540,12 @@ public struct Driver: ParsableCommand, Sendable {
       ? invocationName : invocationName + Host.executableSuffix
 
     // Search in the PATH.
-    // Use case-insensitive key lookup to support environments like MSYS2 that use "PATH"
-    // instead of the native Windows "Path".
+    // On Windows, use case-insensitive key lookup to support environments like MSYS2 that use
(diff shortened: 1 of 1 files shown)


Sources/Driver/Driver.swift
```diff
@@ -540,7 +540,11 @@ public struct Driver: ParsableCommand, Sendable {
       ? invocationName : invocationName + Host.executableSuffix
 
     // Search in the PATH.
-    let path = ProcessInfo.processInfo.environment[Host.pathEnvironmentVariable] ?? ""
+    // Use case-insensitive key lookup to support environments like MSYS2 that use "PATH"
+    // instead of the native Windows "Path".
+    let env = ProcessInfo.processInfo.environment
+    let pathKey = env.keys.first(where: { $0.caseInsensitiveCompare(Host.pathEnvironmentVariable) == .orderedSame }) ?? Host.pathEnvironmentVariable
+    let path = env[pathKey] ?? ""
```
tothambrus11: Thanks for the fix! I'm wondering if it would be more correct if instead of an unconditional case-insensitive search, we just improved the variable name calculation. While Windows is case-insensitive, Linux allows multiple environment variables to coexist that compare equal in a case-insensitive comparison:

Image

So doing an unconditional case-insensitive comparison may lead to errors on Linux and MacOS when someone has set the `pAtH` variable for whatever reason.

What if we modify `Host.pathEnvironmentVariables`'s definition to be something like this?

```swift
  /// The name of the environment variable containing the executable search path.
  public static let pathEnvironmentVariable =
    if Platform.hostOperatingSystem == .windows
      && !ProcessInfo.processInfo.environment.keys.contains("MSYSTEM")
      && !ProcessInfo.processInfo.environment.keys.contains("CYGWIN")
    {
      "Path"
    } else {
      "PATH"
    }
```

I'm not sure if `pathEnvironmentSeparator` also need to be changed to a `:` instead of `;`. If it worked for you while having more than one item on your path, it should be fine.

tothambrus11: It may be even more accurate (and simpler?) to just have the lookup strategy be OS dependent. So on Windows, we can do the iteration-based approach, while on other platforms, do the hashmap lookup.

inkyo08: Good catch! I missed the potential side effects on Linux/macOS. That’s a very valid point regarding compatibility. To avoid any issues, I agree that it's much more reasonable to keep the existing logic for Unix-based platforms and apply the case-insensitive lookup specifically for Windows.

Regarding the path separator (; vs :), I actually did consider it, but I hadn't implemented specific logic for it yet since I didn't encounter any immediate issues during my tests.

I’m planning to update the logic as follows to handle the OS-dependent lookup:

```  swift
    let env = ProcessInfo.processInfo.environment
    // Resolve path key based on the host OS.
    let pathKey = Platform.hostOperatingSystem == .windows
      ? env.keys.first(where: { $0.caseInsensitiveCompare(Host.pathEnvironmentVariable) == .orderedSame }) ?? Host.pathEnvironmentVariable
      : Host.pathEnvironmentVariable
    let path = env[pathKey] ?? ""
```

I'll update the PR with this change shortly. Thanks for the feedback!

tothambrus11: Thanks a lot for the fix! :fire:  (Sorry for the late reply, I missed that you had committed already)

PR #1857, opened 2026-03-25
Remove semantically transparent definition when elaborating projections and projection callers.

lucteo:
This makes elaboration easier in phases that are used just before code generation. There is also no need of complex splitting logic in ElaborateProjections and ElaborateProjectionCallers phases.


Sources/IR/Analysis/Function+RemoveGhostDefinitions.swift
```diff
@@ -0,0 +1,41 @@
...
+    switch self[i] {
+    case let x as Access:
+      replaceUses(of: .register(i), with: x.source)
+    case let x as OpenCapture:
+      replaceUses(of: .register(i), with: x.source)
+    default:
+      let uses = allUses(of: i)
+      precondition(
+        uses.isEmpty,
+        "Instruction \(self[i]) should have no uses, but has uses \(uses.map { self[$0.user] })")
```
tothambrus11, line 27: This could be also useful as a checked postcondition for all cases.

Sources/IR/Analysis/Function+RemoveSemanticallyTransparentDefinitions.swift
```diff
@@ -0,0 +1,41 @@
...
+  /// before code generation.
+  mutating func removeSemanticallyTransparentDefinitions() {
+    // Operate in reverse order to attempt to remove uses of the instructions before removing the instructions themselves.
+    for i in instructionIdentities.reversed() where self[i].isSemanticallyTransparent {
+      fixUses(of: i)
+      remove(i)
+    }
+  }
+
+  /// Ensures that instruction `i` has no uses, replacing them with the appropriate operands if needed.
```
tothambrus11:
```suggestion
  /// Eliminates the semantically transparent `i` from the use-def chain by wiring its uses directly to `i`'s source.
```

Sources/IR/Analysis/Function+RemoveSemanticallyTransparentDefinitions.swift
```diff
@@ -0,0 +1,41 @@
...
+        uses.isEmpty,
+        "Instruction \(self[i]) should have no uses, but has uses \(uses.map { self[$0.user] })")
+    }
+  }
+
+}
+
+extension Instruction {
+
+  /// Returns `true` if this instruction is a semantically transparent definition.
```
tothambrus11: Nit: From the use-site this looks like a property, so I like phrasing these without `Returns`: "`true` iff ..." or "Whether ...".

Maybe you could copy/move the definition of *semantically transparent* here, so that the definition is attached to the property it describes. It's a bit blurry though, since the definition can be phrased in terms of the removal operation.

It was nice that the definition was written on the top of the file, so if you decide to move the definition to this property, it would be useful to move this declaration to the top.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -216,40 +200,35 @@ private struct ScopeDetails {
...
-  /// Returns the "epilogue" instructions of region `r`.
-  internal func epilogueInstructions(region r: Int) -> Array<InstructionID>.SubSequence {
-    precondition(r < regionsCount)
-    let start = regionSplitPositions[r].splitPoint + 1
-    let end = regionSplitPositions[r].epilogueEnd
-    return instructions[start ..< end]
+  /// Returns the split instruction at the start of region `r`.
+  ///
+  /// - Requires: `0 < r`.
+  /// - Requires: `r < regionsCount`.
```
tothambrus11:
```suggestion
  /// - Requires: `0 < r < regionsCount`.
```

Sources/IR/Analysis/Function+RemoveGhostDefinitions.swift
lucteo: I'm not sure if the check is adding more value, as it is directly visible from the calls to `replaceUses`

Sources/IR/Analysis/Function+RemoveSemanticallyTransparentDefinitions.swift
lucteo: Fixed.

I would prefer to keep the definition of semantically transparent at the top, as this is fileprivate, and clients won't probably read this.

Sources/IR/Analysis/Function+RemoveSemanticallyTransparentDefinitions.swift
lucteo: Fixed, thank you. Also added postcondition.

Sources/IR/Analysis/Function+RemoveSemanticallyTransparentDefinitions.swift
```diff
@@ -0,0 +1,44 @@
+extension Function {
+
+  /// Remove from `self` all the definitions that are semantically transparent for code generation.
+  ///
+  /// A semantically transparent definition has no operational semantics, and only serves to
+  /// provide information for IR-level analyses and transformations. They can be safely removed
+  /// before code generation.
```
kyouko-taiga:
```suggestion
  /// Remove from `self` all the definitions that are semantically transparent for code generation.
  ///
  /// A semantically transparent definition has no operational semantics and only informs
  /// IR analyses and transformations. They can be ignored for the purpose of code generation.
```

I'm not sure we can say "safely removed". These instructions also serve guaranteeing safety.

Sources/IR/Analysis/Function+RemoveGhostDefinitions.swift
```diff
@@ -0,0 +1,44 @@
+extension Function {
+
+  /// Remove from `self` all the definitions that are semantically transparent for code generation.
+  ///
+  /// A semantically transparent definition has no operational semantics, and only serves to
+  /// provide information for IR-level analyses and transformations. They can be safely removed
+  /// before code generation.
+  mutating func removeSemanticallyTransparentDefinitions() {
+    // Operate in reverse order to attempt to remove uses of the instructions before removing the instructions themselves.
```
kyouko-taiga: That will not work in general.

There are two ways to remove uses before their definition soundly:
- Visit the dominator tree from its roots in a breadth-first manner.
- Recursively remove uses of an instruction before the instruction itself, using the def-use chain.

Sources/IR/Analysis/Function+RemoveGhostDefinitions.swift
```diff
@@ -0,0 +1,44 @@
...
+    switch self[i] {
+    case let x as Access:
+      replaceUses(of: .register(i), with: x.source)
+    case let x as OpenCapture:
+      replaceUses(of: .register(i), with: x.source)
+    default:
+      let uses = allUses(of: i)
+      precondition(
+        uses.isEmpty,
+        "Instruction \(self[i]) should have no uses, but has uses \(uses.map { self[$0.user] })")
```
kyouko-taiga, line 27: This precondition is checked in the remove function of the module/instruction (don't remember exactly the API of OFE).

Sources/IR/Analysis/Function+RemoveSemanticallyTransparentDefinitions.swift
```diff
@@ -0,0 +1,44 @@
...
+        "Instruction \(self[i]) should have no uses, but has uses \(uses.map { self[$0.user] })")
+    }
+  }
+
+}
+
+extension Instruction {
+
+  /// `true` iff this instruction is a semantically transparent definition.
+  fileprivate var isSemanticallyTransparent: Bool {
```
kyouko-taiga: I would probably define this property in the `Instruction` protocol.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -22,6 +22,9 @@ extension IR.Program {
   private mutating func elaborateProjectionCaller(_ f: Function.ID, in m: Module.ID) {
     if modules[m]![f].site.file.baseName != "feature-test-projection" { return }  // TODO: remove this filter.
 
+    // Before doing anything, remove all the instructions not needed for code generation.
```
```suggestion
    // Remove instructions that can be ignored for the purpose of code generation.
```

kyouko-taiga (changes requested):
We had a discussion with @dabrahams and we were no longer sure why we need to apply this pass. Now I think it's because we want to simplify the splitting logic, right? Still, I think we can do that directly during the splitting pass. You can filter out semantically transparent uses when you are copying instructions into the new functions.

There are two reasons why I think this approach would be better.
1. It makes me uncomfortable to generate IR functions that will break many of the invariants of refined IR just before code generation. An important benefit of having these invariants is that it reduces the number of defensive checks we need make when we're generating LLVM IR.
2. I think it is more efficient.

lucteo:
> We had a discussion with @dabrahams and we were no longer sure why we need to apply this pass. Now I think it's because we want to simplify the splitting logic, right? Still, I think we can do that directly during the splitting pass. You can filter out semantically transparent uses when you are copying instructions into the new functions.
> 
> There are two reasons why I think this approach would be better.
> 
> 1. It makes me uncomfortable to generate IR functions that will break many of the invariants of refined IR just before code generation. An important benefit of having these invariants is that it reduces the number of defensive checks we need make when we're generating LLVM IR.
> 2. I think it is more efficient.

The reason is that it makes the logic of extracting frame variables much easier. Previously, I had cases in which an access was in one region, and the actual instruction that uses the access in another regions. Also, I had the case in which end-access was in a different region than the instruction itself (which, theoretically can be in another region of its `access`). There were so many edge cases that we had to deal with. Simplifying the splitting logic was just a byproduct.

I also thought about applying this as a filter. This means:
- each time we access an instruction
- each time we check the instructions of a block
- each time we look at the def-use chains for an instruction
- each time we remove an instruction
- each time we want to see the previous/next instruction of an instruction.

That would have made the logic far more complex. And, in terms of performance, because we have to have a conditional all the time (not just for projections and projection callers), my estimates that it would be slower than just removing the instructions when we don't need them.

Sources/IR/Analysis/Function+RemoveGhostDefinitions.swift
lucteo: If we look at the algorithm, we correctly remove instructions and their uses. However, traversing in reverse order tends to reduce the number of operations we are performing. The comment included "attempt" as a way to signal that we might not always succeed.

Sources/IR/Analysis/Function+RemoveGhostDefinitions.swift
lucteo: If I remove it from here, there is no way to check the postcondition, as the removal happens afterwards.

Sources/IR/Analysis/Function+RemoveSemanticallyTransparentDefinitions.swift
lucteo: I'm hesitant of doing so, because it won't be clear at that point what semantically transparent instruction means. (`end_access` and `dealloc_stack` carry have some semantic infromation, just not something we need in code generation.) This is why I prefer to encapsulate this logic here, behind the definition of "semantically transparent" from this file.

Sources/IR/Analysis/Function+RemoveGhostDefinitions.swift
```diff
@@ -0,0 +1,42 @@
+extension Function {
+
+  /// Remove from `self` all the ghost definitions.
```
kyouko-taiga:
```suggestion
  /// Removes ghost definitions from `self`.
```

Sources/IR/Analysis/Function+RemoveGhostDefinitions.swift
kyouko-taiga: I don't think that is more efficient. The backing data structure is a doubly linked list so if you visit the instructions in order and then remove the uses before you remove it, in general I think that will have the same efficiency.

kyouko-taiga (approved):

Sources/IR/Analysis/Function+RemoveGhostDefinitions.swift
lucteo: Fine. I don't think the subject is important at this point, so I will do the in-order pass.

PR #1861, opened 2026-04-01
Fix Windows CI

tothambrus11:
Swift now requires a different WinSDK version: 10.0.26100.0

See reported issue at: https://github.com/swiftlang/swift/issues/88237

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -44,7 +44,7 @@ on:
 env:
   spm-build-options: -Xswiftc -enable-testing --explicit-target-dependency-import-check error
   spm-test-options: --parallel
-  swift-version: '6.2.0'
+  swift-version: '6.2.4'
   HYLO_LLVM_BUILD_RELEASE: 20251207-115516
   HYLO_LLVM_VERSION: '20.1.6'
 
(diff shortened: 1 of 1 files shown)


lucteo (approved):
LGTM

kyouko-taiga:
Just out of curiosity, I had to remove the explicit SDK version from the CI in NFE to make it work. I was under the impression that the bug that caused the action to fail had been resolved by the Swift team. 

Did I operate under a faulty assumption?
Should we pin the SDK version in NFE too?

tothambrus11:
Ah thanks for mentioning. If you don't write the version number, the action defaults to what I just pinned here. I'm not sure what the Swift team did, but I would have guessed if they make a "fix" like that, they would release it in a minor version bump to not break existing workflows.

In general I like pinning version numbers to avoid exactly the cases like this - when we don't change anything and something gets broken. It seems there is still something I was not aware of.

kyouko-taiga:
Pinning doesn't avoid the issue here. My CI failure was caused by the pinning.

tothambrus11:
The fact that it doesn't avoid the issue is a problem. But pinning the new version works just as well, and it avoids implicitly relying on a default.

kyouko-taiga:
I'm not so sure. Presumably the maintainers of gha-setup-vsdevenv will face the same issue as we do every time there's an update from Windows that breaks their system and they will adapt accordingly, thus breaking the pin we're putting in our own project.

The way I see it we're essentially opting to a model where we have to regularly manually update the SDK version to avoid breakage rather than automatically benefitting from the updates of the people maintaining our dependencies.

tothambrus11:
One Windows SDK works with several Windows versions. An upgrade to Windows shouldn't break the link between the Windows SDK and the Swift compiler. The Swift compiler has a fairly strict dependency on the Windows SDK version because they rely on an API that doesn't guarantee stability. Therefore not pinning one of Swift or the Windows SDK would just increase the gap between the two over time.

kyouko-taiga:
Fair enough. I trust your judgement on that issue =)

PR #1865, opened 2026-04-06
Interpreter: Runtime safety checks -> implement data structures

RishabhRD:
The PR adds data structures required for implementing runtime safety checks in interpreter to detect UB.

### The idea
We can divide the problem into 3 orthogonal parts and thus 3 data structures handle those 3 problems:
1. `ReservedTypeRegions`: Adds capability to dynamically reserve/unreserve a region of an allocation to be interpreted as specific type and thus handles the type safety aspect of safety.
2. `ComposedRegions`: Adds capability to build composed type that represent initialized regions in allocation. Thus, it detects the uninitialized reads UB. Also, as type information is preserved, would also detect invalid type reads.
3. `AccessTracker`: Allows observing accesses on objects.

These 3 data structures can be composed together in `MemorySafetyValidator` to provide runtime safety checks in interpreter.

### Some changes to already existing code
- `ComposedRegions` was part of `Allocation` previously. However, putting all these 3 structures into `Allocation` would make the invariants of `Allocation` really blur. So, now `Allocation` only handles raw memory, and  these are independent components having their own invariants and thus own tests too. This enable testing on different level of problem.


Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+import Utils
+
+/// The kind of access to a `Place`.
+public enum AccessKind: UInt8, Regular {
```
RishabhRD: better word for access kind that doesn't include the word "kind". Maybe mode?

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
+import Utils
+
+/// Tracks active accesses to parts of an object. Enforces that no two
```
RishabhRD: Summary should tell what thing is, not what it does.

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,39 @@
...
+  /// A unique `Access` identifier.
+  public typealias ID = UUID
+
+  /// A unique `Access` identifier.
+  public let id: ID
+
+  /// Kind of access.
+  public let kind: AccessKind
+
+  /// An access of kind `k`.
```
RishabhRD: Create an instance, for consistency

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+/// overlapping active exclusive accesses exist simultaneously.
+///
+/// Invariants:
+///  - Accesses are active when they are created.
+///  - If `x` is an active exclusive access (`sink`, `set`, or `inout`) of an
+///    object `y`, no other access that includes `y` is active.
+public struct AccessTracker<Component: Regular> {
+
+  // An object A[B[C, D], F] is represented as tree like:
```
RishabhRD: replace representation with tuple

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+  /// Nodes form a tree where each node corresponds to a sequence of components
+  /// (a prefix of a `Path`). Each node stores the accesses at that part
+  /// and links to nodes representing its subparts.
+  private struct Node {
+
+    /// The component that identifies this node relative to its parent.
+    let component: Component
+
+    /// The accesses on this node.
+    var accesses: [Access]
```
RishabhRD: order of accesses should be documented as it is a stack

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+
+  /// Creates a tracker for `object` with an initial access of kind `a`.
+  public init(_ object: Component, with a: AccessKind) {
+    self.storage.append(Node(object))
+    storage[0].accesses.append(Access(kind: a))
+  }
+
+  /// An invalid access operation.
+  public enum Error: Swift.Error, Regular {
+    case overlappingExclusiveAccessExists(for: Component)
```
RishabhRD: The name should have InterpreterError like construct

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+  public mutating func begin(_ a: AccessKind, at p: Path) throws -> Access {
+    let n = createNodesIfNeeded(for: p)
+    return try begin(a, at: n)
+  }
+
+  /// Ends `a` at `p`.
+  ///
+  /// - Precondition: `a` is present at the part identified by `p`.
+  public mutating func end(_ a: Access, at p: Path) throws {
+    try requireIsActive(a, at: p)
```
RishabhRD: Is is not necessary

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
+import Utils
+
+/// Tracks active accesses to parts of an object. Enforces that no two
+/// overlapping active exclusive accesses exist simultaneously.
+///
+/// Invariants:
+///  - Accesses are active when they are created.
+///  - If `x` is an active exclusive access (`sink`, `set`, or `inout`) of an
+///    object `y`, no other access that includes `y` is active.
+public struct AccessTracker<Component: Regular> {
```
RishabhRD: this should be internal

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+        storage[i].children.first { storage[$0].component == c }
+        ?? addChild(c, to: i)
+    }
+  }
+
+  /// Returns the index of the node corresponding to `p`.
+  ///
+  /// - Precondition: Nodes for all prefixes of `p` exist in the tree.
+  private func index(of p: Path) -> Index {
+    p.reduce(into: 0) { i, c in
```
RishabhRD: (i, c) in

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+  ///
+  /// - Precondition: `p` corresponds to a valid path.
+  public func accesses(along p: Path) -> [[Access]] {
+    nodePath(of: p).lazy.map { storage[$0].accesses }
+  }
+
+  /// Returns the node corresponding to path `p`, creating any missing
+  /// intermediate nodes along the path if needed.
+  private mutating func createNodesIfNeeded(for p: Path) -> Index {
+    p.reduce(into: 0) { i, c in
```
RishabhRD: reduce(into: ) -> reduce

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+  /// The returned array has length `p.count + 1`, where the first element
+  /// is the root node.
+  private func nodePath(of p: Path) -> [Index] {
+    p.reduce(into: [0]) { r, c in
+      let i = storage[r.last!].children.first { storage[$0].component == c }!
+      r.append(i)
+    }
+  }
+
+  /// Adds `c` as child to node at `i`.
```
RishabhRD: Document return

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+      storage.append(Node(c))
+    }
+
+    storage[i].children.append(r)
+    return r
+  }
+
+  /// Starts a new access of kind `a` for node at `i`.
+  private mutating func begin(_ a: AccessKind, at i: Index) throws -> Access {
+    let hasNoOverlap =
```
RishabhRD: hasNoOverlap should be a function

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+      if a == .let {
+        storage[i].children.allSatisfy {
+          subtree(at: $0) { storage[$0].accesses.first?.kind == .let }
+        }
+      } else {
+        storage[i].children.isEmpty
+      }
+    if !hasNoOverlap {
+      throw Error.overlappingExclusiveAccessExists(for: storage[i].component)
+    }
```
RishabhRD: else should be explicit, it improves readability

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+  private mutating func begin(_ a: AccessKind, at i: Index) throws -> Access {
+    let hasNoOverlap =
+      if a == .let {
+        storage[i].children.allSatisfy {
+          subtree(at: $0) { storage[$0].accesses.first?.kind == .let }
+        }
+      } else {
+        storage[i].children.isEmpty
+      }
+    if !hasNoOverlap {
```
RishabhRD: throw should be on single line

Sources/Interpreter/AccessTracker.swift
tothambrus11: We should describe what `Component` is in docs, I couldn't guess without looking at usages.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+  /// A `Path` is a sequence of `Component` values.
+  ///
+  /// - An empty sequence identifies the whole object being tracked.
+  /// - Each subsequent element identifies a component of the region selected
+  ///   by the previous elements.
+  ///
+  /// A path therefore describes a sequence of nested components, starting
+  /// from a specific object and refining down to a subcomponent.
+  ///
+  /// For example, given a structure `A { B { C, D }, E }`, the path
```
tothambrus11: If this is the same kind of structure as in line 12, we could use the same notation (e.g. Swift tuple)

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+  /// `[B, D]` identifies `D`.
+  public typealias Path = ArraySlice<Component>
+
+  /// Starts a new access of kind `a` at `p`.
+  public mutating func begin(_ a: AccessKind, at p: Path) throws -> Access {
+    let n = createNodesIfNeeded(for: p)
+    return try begin(a, at: n)
+  }
+
+  /// Ends `a` at `p`.
```
tothambrus11: It could help the reader if we prepend the type name before a one-letter symbol, otherwise we cannot interpret the summary's meaning at a glance, we have to look up the API. What about "Ends access `a` for the part identified by `p`".?

Sources/Interpreter/AccessTracker.swift
tothambrus11:
```suggestion
  /// Adds component `c` as a child to the node at index `i`.
  ///
  /// - Returns: The index of the newly created child node.
```

Sources/Interpreter/AccessTracker.swift
tothambrus11:
```suggestion
public struct AccessTracker<PathComponent: Regular> {
```

This would make it a bit more obvious how `PathComponent` and `Path` are related.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+  /// Returns the accesses encountered along the path `p`, including the root.
+  ///
+  /// - Precondition: `p` corresponds to a valid path.
+  public func accesses(along p: Path) -> [[Access]] {
+    nodePath(of: p).lazy.map { storage[$0].accesses }
+  }
+
+  /// Returns the node corresponding to path `p`, creating any missing
+  /// intermediate nodes along the path if needed.
+  private mutating func createNodesIfNeeded(for p: Path) -> Index {
```
tothambrus11:
```suggestion
  private mutating func demandNode(for p: Path) -> Index {
```
The new frontend uses this naming convention for things like this.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+      r = storage.endIndex
+      storage.append(Node(c))
+    }
+
+    storage[i].children.append(r)
+    return r
+  }
+
+  /// Starts a new access of kind `a` for node at `i`.
+  private mutating func begin(_ a: AccessKind, at i: Index) throws -> Access {
```
tothambrus11: You could document when exactly this throws.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+    self.storage[i].accesses.append(r)
+    return r
+  }
+
+  /// Returns true iff all nodes in subtree at `i` satisfy `predicate`.
+  private func subtree(
+    at i: Index,
+    satisfies predicate: (Index) throws -> Bool
+  ) rethrows -> Bool {
+    if try !predicate(i) { return false }
```
tothambrus11: The API suggests that a subtree is one entity, and the predicate is run on it once. Somehow it would be nice to put the verb/action as the main name of the function declaration. E.g. `descendantsSatisfy(at subtreeRoot: Index, _ predicate)`

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+
+    return try storage[i].children.allSatisfy {
+      try subtree(at: $0, satisfies: predicate)
+    }
```
tothambrus11:
```suggestion
    try predicate(i) && try storage[i].children.allSatisfy {
      try subtree(at: $0, satisfies: predicate)
    }
```

Reducing control flow and explicit returns can help readability (&& uses autoclosure, so the behavior is equivalently short-circuiting)

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  /// Removes the maximal suffix of `p` consisting of nodes that have neither
+  /// active accesses nor children.
```
tothambrus11: If `storage[i].accessesses`contain an access, does that mean that it's necessarily active?

Sources/Interpreter/ComposedRegions.swift
```diff
@@ -0,0 +1,269 @@
...
+  private let allocationID: Memory.Allocation.ID
+
+  /// The allocation for which composed regions belong to.
+  private var allocation: Memory.Allocation {
+    _read {
+      yield memory.pointee.allocation[allocationID]!
+    }
+  }
+
+  /// The composed regions, in ascending order.
```
tothambrus11: Ascending based on what? Offset? It could be also interesting to describe if this is a list of all typed regions in the hierarchy put in a flat list, or the top-level non-overlapping set of regions.

Sources/Interpreter/ComposedRegions.swift
```diff
@@ -0,0 +1,269 @@
...
+
+  /// ID of allocation for which composed regions belong to.
+  private let allocationID: Memory.Allocation.ID
+
+  /// The allocation for which composed regions belong to.
+  private var allocation: Memory.Allocation {
+    _read {
+      yield memory.pointee.allocation[allocationID]!
+    }
+  }
```
tothambrus11: Is there a benefit of making this a `_read` access rather than a simple computed property?

Sources/Interpreter/ComposedRegions.swift
```diff
@@ -0,0 +1,269 @@
...
+
+/// The composed regions of an `Allocation`.
+public struct ComposedRegions {
+
+  /// A position in some allocation.
+  public typealias Offset = ComposedRegion.Offset
+
+  /// Memory where `allocation` lives.
+  private let memory: UnsafeMutablePointer<Memory>
+
```
tothambrus11: We should justify the usage of any unsafe code with a comment. Why do we need unsafe pointer to Memory? Can't we use classes or indices instead?

Sources/Interpreter/Memory.swift
```diff
@@ -12,14 +12,10 @@ public struct Memory {
```
tothambrus11, line 12: I think it's nice to start a struct declaration by listing all stored properties, and not mixing them with type definitions/methods/initializers.

Sources/Interpreter/Memory.swift
```diff
@@ -338,13 +238,6 @@ public struct Memory {
     }
   }
```
tothambrus11, line 245: Describe the requirements for successful deallocation or the conditions for when the function throws.

Sources/Interpreter/Memory.swift
tothambrus11, line 233: Please add documentation.

Tests/InterpreterTests/AccessTrackerTests.swift
```diff
@@ -0,0 +1,149 @@
...
+import Utils
+import XCTest
+
+extension String: Regular {}
+
+final class AccessTrackerTests: XCTestCase {
+
+  typealias Error = AccessTracker<String>.Error
+
+  func testBeginAccessWhenSubPartProjectionNotExists() throws {
```
tothambrus11:
```suggestion
  func testBeginAccessWhenSubPartProjectionDoesNotExist() throws {
```

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+    p.reduce(into: 0) { i, c in
+      i = storage[i].children.first { storage[$0].component == c }!
+    }
+  }
+
+  /// Returns the indices of nodes along the path `p`, starting from the root.
+  ///
+  /// The returned array has length `p.count + 1`, where the first element
+  /// is the root node.
+  private func nodePath(of p: Path) -> [Index] {
```
tothambrus11: The difference between `nodePath` and `Path` is not so obvious. Can we find some more unique distinction between the two types of paths? E.g. one might be `ComponentPath`, the other `indexPath`.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+    if !storage[i].children.isEmpty || storage[i].accesses.last != a {
+      throw Error.overlappingExclusiveAccessExists(for: storage[i].component)
+    }
+  }
+
+  /// Returns the accesses encountered along the path `p`, including the root.
+  ///
+  /// - Precondition: `p` corresponds to a valid path.
+  public func accesses(along p: Path) -> [[Access]] {
+    nodePath(of: p).lazy.map { storage[$0].accesses }
```
tothambrus11: Does `lazy` have a benefit here? It seems that the result is immediately materialized by creating an array from the result.

Tests/InterpreterTests/AccessTrackerTests.swift
```diff
@@ -0,0 +1,149 @@
...
+    for p in [.let, .inout, .sink, .set] as [AccessKind] {
+      var t = AccessTracker("a", with: .sink)
+      _ = try t.begin(p, at: ["b"])
+      for p in [.let, .inout, .sink, .set] as [AccessKind] {
+        _ = try t.begin(p, at: ["b", "x"])
+        _ = try t.begin(p, at: ["b", "x"])
+      }
+    }
+  }
```
tothambrus11: This is quite counterintuitive why it should work like this, so you could document the expectations more verbosely here.

tothambrus11: I would recommend avoiding name shadowing (`p` in loops), it can lead to surprises, or just make it harder to talk about code.

Sources/Interpreter/AccessTracker.swift
tothambrus11: It's nice if two overloads of the `begin` function are next to each other.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,217 @@
...
+  // where each node has list of accesses it currently has.
+
+  // Class invariants:
+  //   1. Every node has at least one access or one child.
+  //
+  //   2. If a node contains an exclusive access `a` at index i in `accesses`,
+  //      then `a` is active iff:
+  //        - there is no access at any index `j` > `i` in `accesses`, and
+  //        - the node has no descendants.
+  //      Otherwise, `a` is inactive.
```
tothambrus11:
```suggestion
  //   2. A `let` access is always active.
  //   3. An `inout`/`set`/`sink` access is active iff:
  //      - it's the last element of `accesses` of the corresponding node, and
  //      - the node has no descendants.
```

Tests/InterpreterTests/AccessTrackerTests.swift
```diff
@@ -0,0 +1,149 @@
...
+
+  func testBeginAccessWhenSubPartProjectionExists() throws {
```
tothambrus11: You could break this into 3 separate test cases.

Tests/InterpreterTests/ComposedRegionsTests.swift
```diff
@@ -0,0 +1,220 @@
+import FrontEnd
+import TestUtils
+import XCTest
+
+@testable import Interpreter
+
+final class ComposedRegionsTests: XCTestCase {
+
+  func test_requireInitialized() {
```
tothambrus11:
```suggestion
  func testRequireInitialized() {
```

Tests/InterpreterTests/ComposedRegionsTests.swift
```diff
@@ -0,0 +1,220 @@
...
+    var layouts = TypeLayoutCache(typesIn: TypedProgram.empty, for: UnrealABI())
+    let void_ = AnyType.void
```
tothambrus11:
```suggestion
    let void = AnyType.void
```

Tests/InterpreterTests/ComposedRegionsTests.swift
```diff
@@ -0,0 +1,220 @@
...
+    var c = ComposedRegions(
+      memory: withUnsafeMutablePointer(to: &m) { $0 },
+      allocation: p.allocation
+    )
+
+    XCTAssertFalse(c.canCompose(voidPair, at: 0))
+
+    // It should be possible to initialize both parts at the same address
+    XCTAssertTrue(c.canCompose(void_, at: 0))
+    c.compose(void_, at: p.offset + layouts[voidPair].parts[0].offset)
```
tothambrus11: Test failures shouldn't cause precondition failures. `XCTAssertTrue` doesn't short-circuit, so it could crash the testing process in some subsequent steps that rely on that precondition (e.g. `compose`).

A solution could be to make a throwing testing utility:

```swift

/// Fails the test case with `message` if `condition` is false, and throws an error to terminate the test case early.
func require(_ condition: Bool, @autoclosure _ message: () -> String = "", file: StaticString = #file, line: UInt = #line) throws {
    if !condition {
        XCTFail(message, file: file, line: line)
        throw EarlyTestFailure()
    }
}

private struct EarlyTestFailure: Error {}
```

tothambrus11 (changes requested):
Very nice work! I really like how testable these APIs are. I'm a bit concerned about why we need `Memory` to be passed via an unsafe pointer. If you can explain the use case, maybe we can think of an alternative together.

When reviewing the test cases, I was not sure what properties they are testing, but that's fine for a draft. It would be nice if you could take the memory safety and other laws, derive some properties that should hold, and make some exhaustive or randomized property based tests that are convincing of correctness.

Sources/Interpreter/MemorySafetyValidator.swift
```diff
@@ -0,0 +1,230 @@
...
+
+/// A runtime validator that validates the safe use of memory within an allocation.
+///
+/// The validator does not manage memory. Instead, it observes and validates
+/// accesses to an existing allocation.
+struct MemorySafetyValidator {
+
+  /// Class Invariants:
+  ///   - If a region `d` has an active `.sink` access, then no strict ancestor
+  ///     of `d` is present in `composedRegions`.
```
tothambrus11: These could be `//` comments, otherwise they are interpreted as the docs of `Error` below.

Sources/Interpreter/MemorySafetyValidator.swift
```diff
@@ -0,0 +1,230 @@
...
+    allocation: Memory.Allocation.ID
+  ) {
+    self.memory = memory
+    self.allocationID = allocation
+    typeBindings = ReservedTypeRegions(memory: memory, allocation: allocation)
+    composedRegions = ComposedRegions(memory: memory, allocation: allocation)
+    regionAccesses = [:]
+  }
+
+  /// Starts an access of type `k` at `p`.
```
tothambrus11: Describe when it throws.

Sources/Interpreter/MemorySafetyValidator.swift
```diff
@@ -0,0 +1,230 @@
...
+/// accesses to an existing allocation.
+struct MemorySafetyValidator {
+
+  /// Class Invariants:
+  ///   - If a region `d` has an active `.sink` access, then no strict ancestor
+  ///     of `d` is present in `composedRegions`.
+
+  public enum Error: Regular, Swift.Error {
+    case noTypeBound(at: Memory.Address)
+    case notContains(Memory.Place, in: Memory.Place)
```
tothambrus11:
```suggestion
    case notContained(Memory.Place, in: Memory.Place)
```

Sources/Interpreter/AccessTracker.swift
RishabhRD: Regarding let access, it is always active as we can't derive inout, set or sink access from it.

Sources/Interpreter/AccessTracker.swift
RishabhRD: I believe it is not aligned with past [Dave's comment](https://github.com/hylo-lang/hylo/pull/1753#discussion_r2532324043).

Sources/Interpreter/AccessTracker.swift
RishabhRD: demand seems like demanding some resource owner for nodes. However, the nodes are owned by this data structure only, so we are not demanding that way.

Sources/Interpreter/AccessTracker.swift
RishabhRD: But it is not aligned with [documenting runtime error](https://github.com/stlab/better-code/blob/main/better-code/src/chapter-3-errors.md#how-to-document-runtime-errors) section of better-code book. Why do we think it is necessary?

Sources/Interpreter/AccessTracker.swift
RishabhRD: see if this is still a problem for readability.

Sources/Interpreter/AccessTracker.swift
RishabhRD: I was seeing subtree as one entity though from function boundary. Also was trying to match similar style from stdlib `Collection.all(satisfies: predicate)`.

Sources/Interpreter/AccessTracker.swift
RishabhRD: No, but any node must have atleast one children or atleast one active access.

Sources/Interpreter/ComposedRegions.swift
RishabhRD: Memory is going to contain these structures. I separated these structures to make the invariants of these separate and reasonable. All of these structures assume that corresponding allocation is alive.

Sources/Interpreter/ComposedRegions.swift
RishabhRD: Did it for avoiding copies of allocation

Sources/Interpreter/ComposedRegions.swift
RishabhRD: This one comes from Dave. AFAICT ascending order of offset has more words than necessary as only comparable thing in composed region is offset.

> It could be also interesting to describe if this is a list of all typed regions in the hierarchy put in a flat list, or the top-level non-overlapping set of regions.

I believe the doc for `ComposedRegion` provides info for that.

Sources/Interpreter/Memory.swift
RishabhRD: I don't have problem with this. But changing it in this PR would create unnecessary diff people have to go through. If project is aligned with this convention, will raise a different PR for the same.

Sources/Interpreter/Memory.swift
RishabhRD: Not aligned with better-code book for error document.

Sources/Interpreter/MemorySafetyValidator.swift
RishabhRD: Not aligned with better-code book.

Sources/Interpreter/AccessTracker.swift
RishabhRD: Adding type information is not aligned with [Dave's past comment](https://github.com/hylo-lang/hylo/pull/1753#discussion_r2532324043).

RishabhRD (commented):
Thank You @tothambrus11 for the comments. Resolved some of those. Need to resolve the rest.

Sources/Interpreter/AccessTracker.swift
tothambrus11: There is a little difference in that the surrounding verb gave a bit more meaning. I don't have a very strong opinion, and consistency may be better than having to evaluate these things every time.

Tests/InterpreterTests/ComposedRegionsTests.swift
RishabhRD: This is an interesting take. In safe language like swift, precondition failure usually means bug and not undefined behavior. But my argument is, test failure itself says there is a bug in the code and that implies it might be possible somewhere preconditions might not be satisfied.

Regarding crash of testing process, I don't have problem with that tbh as crashing due to buggy code would help me reach the root cause more easily as somewhere nearby there should be a precondition failure, right?

Sources/Interpreter/AccessTracker.swift
tothambrus11: NFE uses `demand` specifically for creating objects on demand in the given data structure if it didn't exist yet, so I think the use would be consistent. Anyway, no need to rename if you think this name is clearer.

Sources/Interpreter/AccessTracker.swift
tothambrus11: This is like a precondition that correct usage should uphold, but we are modeling lenience for incorrect use as part of the domain, which necessitates the use of thrown errors. Somewhere in the code we must document what guarantees the interpreter wants to uphold, and I think it's nice to also put these on methods so that we can have local information about what each operation requires.

> Because a runtime error report indicates a failure to fulfill postconditions, information about errors—including that they are possible—does not belong in a function's postcondition documentation, whose primary home is the summary sentence fragment.

@dabrahams I don't fully agree with this recommendation. In order to use a function correctly, one should know it's preconditions. Sometimes thrown errors are meant to be used as recoverable precondition violation checks for conditions that may happen outside of the developer's control in unfortunate cases, but should usually hold true as a happy path if care is taken.

For example, invoking a compiler with an output folder that doesn't exist yet will definitely become a failure, so it is useful information for the caller (e.g. writer of a build script) to know what the conditions of the happy path are, even though there is possibility of file system errors otherwise. IMO hiding relevant information about how to use a throwing function to stay on the happy path is comparably counterproductive as missing to add precondition documentation.

I'm not saying we should always document when something throws, as that's indeed unmanageable at higher levels, but there are cases when it's helpful.

Tests/InterpreterTests/ComposedRegionsTests.swift
tothambrus11: Yeah, it is perfectly safe to crash in Swift, it's just not very friendly for the purposes of testing. After noticing the first problem that a later part of the test depends on as a precondition, it's not useful to do so because:
- we already know about the problem at hand
- whatever is tested later on is likely working on garbage data anyways

If you fail a test early, it allows the test runner to continue without crashing the whole pipeline, so you can get a whole picture of what tests are failing. You could have both AdvancedFeature and BasicCoreFeature failing, and it's probably not worth debugging AdvancedFeature until you fixed BasicCoreFeature. Having an early crash in AdvancedFeature would hide the fact that the problem is introduced in BasicCoreFeature.

Sources/Interpreter/AccessTracker.swift
dabrahams: You have to look at the API anyway

Sources/Interpreter/AccessTracker.swift
dabrahams: This is not the place for the general argument. In this case though the interpreter is going to have blanket documentation that any UB in the program being interpreted will result in a UB error being thrown.

Sources/Interpreter/ComposedRegions.swift
tothambrus11: > This one comes from Dave. AFAICT ascending order of offset has more words than necessary as only comparable thing in composed region is offset.

It's hard for me to infer that because it's some distant information, and also I wouldn't be sure that I can safely assume this implication. If ComposedRegion was itself comparable, then it's more obvious, but if it's not an intrinsic property of ComposedRegion that it is Comparable, we could just document the extrinsic relationship between objects that's established by their parent array. 

> >     It could be also interesting to describe if this is a list of all typed regions in the hierarchy put in a flat list, or the top-level non-overlapping set of regions.

> I believe the doc for ComposedRegion provides info for that.

I guess it cannot be just top-level because ComposedRegion doesn't provide any nested ComposedRegion members. The type documentation on ComposedRegion still left me wondering, as I was someone without prior high-level overview of the system. Maybe some high-level architecture diagrams or descriptions could be useful to illustrate the main concepts in the future.

Tests/InterpreterTests/ComposedRegionsTests.swift
dabrahams: I went back and forth about this one.  But fundamentally @RishabhRD is correct that bugs can cause precondition failures when testing.  In the middle of a function that fails to establish some precondition of a function it calls, there's no opportunity for the way you write your test code to intervene and prevent it from reaching the precondition check.  So I think worrying about this kind of thing inside the tests themselves is a typical kind of programmer-obsessive behavior that's actually full of holes and not worth the effort.

tothambrus11: Anecdotally, this proved useful to me when refactoring some asset management system that was originally written by someone else, and they used `EXPECT_*` (fall-through) assertions everywhere. The test suite was crashing at random places depending on which test ran first, so I had to change a couple critical places to be short-circuiting rather than fall-through. Generally, most things could be fall-through except some nil-checks and array element count checks. In Google Tests, making the decision between `ASSERT`/`EXPECT` is almost free, so even if it's not complete, it can help a lot. But XCTest doesn't have much support for short-circuiting assertions, except `XCTTryUnwrap` for optionals, so the cost is indeed higher. 

The best solution would be if the test runner was resilient against crashes, and could restart a new process after crashing in one. XCTest's parallel testing is capable of doing this, except that its [reported success/failure summary is currently broken](https://github.com/swiftlang/vscode-swift/issues/2218). The sequential test runner just exits on the first crash. We should fix all these in Hylo's future testing framework.

Sources/Interpreter/AccessTracker.swift
dabrahams: Please use demand until the project settles on a different convention. It exists to get rid of the createIfNeeded boilerplate

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,25 @@
+import FrontEnd
+import Foundation
+import Utils
+
+extension AccessEffect: Regular {}
```
dabrahams: Can this be moved to the AccessEffect declaration?

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,25 @@
...
+public struct Access: Regular {
+
+  /// A unique `Access` identifier.
+  public typealias ID = UUID
+
+  /// A unique `Access` identifier.
+  public let id: ID
+
+  /// Kind of access.
+  public let kind: AccessEffect
```
dabrahams: Suggest `effect` or `mode`

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,25 @@
...
+
+  /// An access of kind `k`.
```
dabrahams: I don't remember 100% but we recently had a discussion in a meeting, where @kyouko-taiga gave her rationale for going back to the old Swift convention of a phrase that starts with "creates" and I thought we settled on doing that. (@RishabhRD you missed that one I think)

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
+import Utils
+import FrontEnd
+
+/// Tracks active accesses to parts of an object.
```
dabrahams: This should say what the type _is_ not what it _does_. Something like "a record of the active accesses..."

You should be able to get `PathComponent` into this sentence fragment. Think of a generic type parameter like a function, parameter and describe the role that it plays in the type.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+  // Class invariants:
+  //   1. Every node has at least one access or one child.
+  //
+  //   2. A `let` access is always active.
+  //
+  //   3. An `inout`/`set`/`sink` access is active iff:
+  //      - it's the last element of `accessStack` of the corresponding node, and
+  //      - the node has no descendants.
+
+  /// A node representing a part of an object identified by a path prefix.
```
dabrahams: It's a red flag, that you began the description of a thing called `Node` with the phrase "a node"

Also a red flag that you used the word "representing." Usually the name you're looking for shows up in the phrase after that word. For example, in this case, it might be "part"

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+  ///
+  /// Nodes form a tree where each node corresponds to a sequence of components
+  /// (a prefix of a `Path`). Each node stores the accesses at that part
+  /// and links to nodes representing its subparts.
```
dabrahams: Another argument for the word "part": you don't have to tell anyone that parts form a tree

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+  private struct Node {
+
+    /// The component that identifies this node relative to its parent.
+    let component: PathComponent
```
dabrahams: It's a red flag that you're repeating part of the type name. What about "`step`"?

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+
+    /// Stack of accesses on the node.
```
dabrahams: Try to tell me something that isn't entirely represented in the name. For example, you might say that the top of the stack is the only active access, if that's true. Another thing this repetition might push you to do is find a different name.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+    var accessStack: [Access]
+
+    /// The indices of this node's children, each representing a further
```
dabrahams: It's not clear at this point what these are indices into. Might make sense to "in the Global node storage" Or something. 

The phrase "each representing a further refinement…" tends to imply that these are descendants, not children. Which is it?

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+    /// refinement of the current path.
+    var children: [Index]
+
+    /// Creates a node for `c` with no accesses or children.
```
dabrahams: "For" is pretty semantically empty. In this case, I think you mean to say that the resulting node will have `c` as it's component property, no?

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+    init(_ c: PathComponent) {
+      self.component = c
+      self.accessStack = []
+      self.children = []
+    }
+
+  }
+
+  /// Storage of all nodes.
```
dabrahams: Try to avoid using the word storage here.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+  private var storage: [Node] = []
+
+  /// The index of a node in storage.
+  private typealias Index = Int
```
dabrahams: If this thing isn't conforming to collection, he probably shouldn't use index. Isn't this just a node ID at this level of abstraction?

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+
+  /// Indices in `storage` that are currently unused and may be reused.
+  private var freeNodes: [Index] = []
+
+  /// Creates a tracker for `object` with an initial access of kind `a`.
+  public init(_ object: PathComponent, with a: AccessEffect) {
```
dabrahams: "With" is a vacuous label. If you can't do better, a noun like "effect" or "mode" would be better.

dabrahams (commented):
There's more to review, but I have to take a break now so I thought I would leave you what I have

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+    self.storage.append(Node(object))
+    storage[0].accessStack.append(Access(kind: a))
+  }
+
+  /// An invalid access operation.
+  public enum Error: Swift.Error, Regular {
+    case overlappingExclusiveAccessExists(for: PathComponent)
```
dabrahams:
```suggestion
    case overlappingExclusiveAccess(PathComponent)
```

"for" is vacuous.  "to" or "of" might not be.  But don't we need to report an entire path?

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+  ///
+  /// - Precondition: `p` corresponds to a valid path.
+  public func accesses(along p: Path) -> [[Access]] {
+    indexPath(p).map { storage[$0].accessStack }
+  }
+
+  /// Returns the node corresponding to path `p`, creating any missing
+  /// intermediate nodes along the path if needed.
+  private mutating func createNodesIfNeeded(for p: Path) -> Index {
+    p.reduce(0) { i, c in
```
dabrahams: Algorithms like `reduce` are not great for things with side-effects.  It would be better to use a `for` loop here.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+        storage[i].children.first { storage[$0].component == c }
```
dabrahams: Should I not be concerned about the cost of a linear search in this loop?

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+        ?? addChild(c, to: i)
+    }
+  }
+
+  /// Returns the index of the first node along `p` that contains `a`.
+  ///
+  /// - Precondition: Some node visited while traversing `p` from the root
+  ///   contains `a`.
+  private func nodeIndex(_ a: Access, in p: Path) -> Index {
```
dabrahams:
```suggestion
  private func firstIndex(in p: Path, containing a: Access) -> Index {
```
Just expressing the summary directly.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+    p.reduce(0) { i, c in
+        storage[i].children.first { storage[$0].component == c }
+        ?? addChild(c, to: i)
+    }
+  }
+
+  /// Returns the index of the first node along `p` that contains `a`.
+  ///
+  /// - Precondition: Some node visited while traversing `p` from the root
+  ///   contains `a`.
```
dabrahams: The precondition is implied by the summary and  doesn't need to be stated.

```suggestion
```

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,232 @@
...
+    }
+    return i
+  }
+
+  /// Returns the indices of nodes along the longest existing prefix of `p`,
+  /// starting from the root.
+  ///
+  /// The returned array always begins with the root index and has length equal
+  /// to the number of matched path components plus one.
+  private func indexPath(_ p: Path) -> [Index] {
```
dabrahams:
```suggestion
  private func nodeIDs(_ p: Path) -> [NodeID] {
```

dabrahams:
We already use XCTest wrappers to improve error reporting so you could just integrate throwing into all of themSent from my iPhoneOn May 10, 2026, at 1:27 AM, Ambrus Tóth ***@***.***> wrote:﻿
@tothambrus11 commented on this pull request.



In Tests/InterpreterTests/ComposedRegionsTests.swift:
> +    XCTAssertTrue(c.canCompose(void_, at: 0))
+    c.compose(void_, at: p.offset + layouts[voidPair].parts[0].offset)

Anecdotally, this proved useful to me when refactoring some asset management system that was originally written by someone else, and they used EXPECT_* (fall-through) assertions everywhere. The test suite was crashing at random places depending on which test ran first, so I had to change a couple critical places to be short-circuiting rather than fall-through. Generally, most things could be fall-through except some nil-checks and array element count checks. In Google Tests, making the decision between ASSERT/EXPECT is almost free, so even if it's not complete, it can help a lot. But XCTest doesn't have much support for short-circuiting assertions, except XCTTryUnwrap for optionals, so the cost is indeed higher.
The best solution would be if the test runner was resilient against crashes, and could restart a new process after crashing in one. XCTest's parallel testing is capable of doing this, except that its reported success/failure summary is currently broken. The sequential test runner just exits on the first crash. We should fix all these in Hylo's future testing framework.

Sources/Interpreter/ComposedRegions.swift
dabrahams: If you mean that `Memory` is going to contain `ComposedRegion`s and this is a back-pointer, that's very bad juju that can easily lead to exclusivity violations.  We're trying to write this thing entirely with value semantics, and this is a reference.

Sources/Interpreter/AccessTracker.swift
RishabhRD: This one is children.

Sources/Interpreter/AccessTracker.swift
RishabhRD: Yes, changed it to report entire path.

Sources/Interpreter/AccessTracker.swift
RishabhRD: Made it a binary search, however the point we insert a subpart to a part is still linear due to array's insertion complexity. The first instinct says we might need SortedSet for optimizing the same.

Sources/Interpreter/ComposedRegions.swift
RishabhRD: Fixed the same, now we don't use pointers at all.

Sources/FrontEnd/AccessEffect.swift
```diff
@@ -1,3 +1,5 @@
+import Utils
+
 /// An access effect, specifying how a parameter, receiver, or remote part is accessed.
```
dabrahams: Try not to use the name of the thing in the comment.  It's _obviously_ an access effect!  Use the comment as an opportunity to say something that's _not_ in the name.  In this case you have it in the rest of the sentence


```suggestion
/// How a parameter or binding, or capture may (or must) be accessed.
```

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,23 @@
+import FrontEnd
+import Foundation
+import Utils
+
+/// A unique access participating in access control.
```
dabrahams: What's the meaning of “unique?”
“Participating in access control” seems like blather.  Don't all accesses do that?  Even unsafe ones participate w.r.t. the receiver.

```suggestion
/// An access occuring during execution.
```

This description is designed to distinguish it from an access in the IR, which is a different thing.  But working from the comment I wrote I would expect to see a memory region stored in it, since temporal accesses are always for some region.  So then I wonder if we have the right name or are we describing something else?

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,23 @@
...
+public struct Access: Regular {
+
+  /// A unique `Access` identifier.
+  public typealias ID = UUID
+
+  /// A unique `Access` identifier.
+  public let id: ID
+
+  /// Effect of access.
```
dabrahams: “of access“ is redundant

```suggestion
  /// The associated permissions and obligations.
```

Maybe?  Feels a little strained but it's a way to say something that's not in the name

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,23 @@
...
+  public let effect: AccessEffect
+
+  /// Creates an instance of access of effect `e`.
```
dabrahams: “of…of” doesn't parse.


```suggestion
  /// Creates an instance having effect `e`.
```

Note: “With” can be ambiguous here so I used “having”

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+  // Class invariants:
+  //   1. Every part has at least one access or one child.
+  //
+  //   2. A `let` access is always active.
+  //
+  //   3. An `inout`/`set`/`sink` access is active iff:
+  //      - it's the last element of `associatedAccesses` of the corresponding part, and
+  //      - the part has no descendants.
+
+  /// A subobject together with its associated accesses and immediate subparts.
```
dabrahams: **O**mit **N**eedless **W**ords

```suggestion
  /// A subobject's accesses and immediate subparts.
```

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+  private struct Part {
+
+    /// The step from parent subobject to this subobject.
+    let step: PathComponent
+
+    /// Stack of accesses in creation order, where `let` accesses are always
+    /// active and other accesses are active only at the top of the stack.
+    var associatedAccesses: [Access]
```
dabrahams: What value is “associated” adding?

```suggestion
    var accesses: [Access]
```

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+
+    /// IDs of immediate subparts in global parts storage.
```
dabrahams: **A**void **R**epeating **T**ype **I**nformation

```suggestion
    ///The immediate subparts.
```

It's not really “global” storage is it?  There's usually no point in distinguishing, in comments, between a thing's ID and the thing itself.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+    var subparts: [PartID]
+
+    /// Creates a `Part` reached from its parent by `step` with no accesses or subparts.
```
dabrahams:
```suggestion
    /// Creates a `Part` reached from its parent by `step`, having no accesses or subparts.
```

Slightly less awkward I think.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+    init(_ step: PathComponent) {
+      self.step = step
+      self.associatedAccesses = []
+      self.subparts = []
+    }
+
+  }
+
+  /// Allocated parts, including reusable entries.
```
dabrahams:
```suggestion
  /// Allocated parts, including free entries no longer in use.
```

All parts are reusable, even if currently in use.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+  private var storage: [Part] = []
+
+  /// Identity of a `Part` in `storage`.
+  private typealias PartID = Int
+
+  /// IDs of reusable entries in `storage`.
```
dabrahams:
```suggestion
  /// Elements of Storage not currently in use.
```

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+      self.subparts = []
+    }
+
+  }
+
+  /// Allocated parts, including reusable entries.
+  private var storage: [Part] = []
+
+  /// Identity of a `Part` in `storage`.
+  private typealias PartID = Int
```
dabrahams:
```suggestion
  private typealias PartID = Array<PartID>.Index
```

Can also think about 

```swift
private typealias PartStorage = Array<PartID>
private typealias PartID = PartStorage.Index
```

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+
+  /// IDs of reusable entries in `storage`.
+  private var freeParts: [PartID] = []
```
dabrahams: People know what a freelist is.

```suggestion
  private var freeList: [PartStorage.Index] = []
```

Using `Index` as the element type communicates these are inactive and not `Part`s in any useful sense.  Could reuse the ID fields of parts to create a linked freelist; not sure how much that saves though.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+
+  /// Creates a tracker for `object` with an initial access having capability `c`.
```
dabrahams:
```suggestion
  /// Creates an instance storing a ` c` access to `object`.
```

Where does the `object` value come from at the top of an access tree?

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+///
+/// `PathComponent` describes one step in a path to a subobject. For example,
+/// components for a tuple might be field indices, while components for a
+/// struct might be property names.
+///
+/// Invariants:
+///  - Accesses are active when they are created.
+///  - If `x` is an active exclusive access (`sink`, `set`, or `inout`) of an
+///    object `y`, no other access that includes `y` is active.
+struct AccessTracker<PathComponent: Regular & Comparable> {
```
dabrahams: Just like the summary, the name should reflect what it _is_.  It doesn't do the tracking by itself, that's partly the role of the thing calling `begin`/`end`.  You could say it's a _record_ of the accesses, but words like “record” should be avoided in type descriptions.

Maybe:

```suggestion
struct ObjectAccesses<PathComponent: Regular & Comparable> {
```

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+    self.storage.append(Part(object))
+    storage[0].associatedAccesses.append(Access(effect: c))
+  }
+
+  /// An invalid access operation.
+  public enum Error: Swift.Error, Regular {
+    case overlappingExclusiveAccess(Path)
+  }
+
+  /// A path describing a part of an object.
```
dabrahams: Avoid words like “describing,” “representing,” “record” (as a noun) etc, which are redundant with the fact that it is a type.

```suggestion
  /// A route to a subpart of some object.
```

“Subpart” isn't precise, but tends to imply that it could be N levels down.

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+  ///
+  /// A `Path` is a sequence of `Component` values.
+  ///
+  /// - An empty sequence identifies the whole object being tracked.
+  /// - Each subsequent element identifies a component of the region selected
```
dabrahams:
```suggestion
  /// - Each subsequent element identifies a part of the region identified
```

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+  /// Starts a new access having capability `a` at `p`.
+  public mutating func begin(_ a: AccessEffect, at p: Path) throws -> Access {
+    let n = demandParts(p)
+    guard let r = begin(a, at: n) else {
+      throw Error.overlappingExclusiveAccess(p)
+    }
+    return r
+  }
+
+  /// Starts and returns new access having capability `a` for part at `i`;
```
dabrahams: Just talk about the thing, not its ID when practical.

```suggestion
  /// Starts and returns new access having capability `a` for `i`;
```

Reconsider variable names

Sources/Interpreter/AccessTracker.swift
```diff
@@ -0,0 +1,239 @@
...
+  ///   by the previous elements.
+  ///
+  /// A path therefore describes a sequence of nested components, starting
+  /// from a specific object and refining down to a subcomponent.
+  ///
+  /// For example, given a structure `(b: (c: C, d: D), e: E)`, the path
+  /// `[b, d]` identifies `d`.
+  public typealias Path = ArraySlice<PathComponent>
+
+  /// Starts a new access having capability `a` at `p`.
```
dabrahams:
```suggestion
  /// Starts a new access having effect `a` at `p`.
```

It's not just _a_ capability, because `set` comes with an obligation, and `inout` gives read _and_ write capability.

dabrahams (commented):
Great progress! Please go through all the changes and try to apply the principles behind my comments before I do another review.  Thanks!

Sources/Interpreter/AccessTracker.swift
dabrahams: Ugh, I'm not sure I believe what I wrote.  The role of this thing is really to _validate_ accesses and the fact that it stores information about them is an implementation detail.  You don't  `add` and `remove` accesses, you `begin` and `end` them. So maybe `AccessValidator` and the doc comment summary should change accordingly.  Think about it!

Sources/Interpreter/AccessTracker.swift
RishabhRD: `MemorySafetyValidator` is going to contain `AccessValidator`, which would create this structure with sink access, as from discussion: `A var declaration is a sink access initially uninitialized.`

Sources/FrontEnd/AccessEffect.swift
```diff
@@ -37,3 +39,5 @@ extension AccessEffect: Comparable {
   }
 
 }
+
+extension AccessEffect: Regular {}
```
dabrahams: Can this conformance be moved to the declaration of `AccessEffect`?

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,230 @@
+import FrontEnd
+import Utils
+
+/// Validates accesses to an object and its subobjects, where subobjects
+/// are identified by paths of `PathComponent` values.
+///
+/// Invariants:
+///  - Accesses are active when they are created.
+///  - If `x` is an active exclusive access (`sink`, `set`, or `inout`) of an
+///    object `o`, no other access that includes `o` is active.
```
dabrahams: If you want to write these type invariants at all, they should not be in a doc comment, because they are not for the consumption of a client, unless a client gets the ability to inspect which accesses are active.  They don't, do they?

Just reading it, it's a little odd that it tells us one condition under which an access becomes active but doesn't tell us how it can become inactive.

```suggestion
/// Operations that would otherwise cause an active exclusive access
/// to overlap another active access throw `AccessError`s.
```

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,230 @@
...
+  // The object `(b: (c: C, d: D), f: F)` may be represented as:
+  //
+  //             root
+  //            /    \
+  //           b      f
+  //         /   \
+  //        c     d
+  //
+  // Each part stores the accesses currently associated with it.
+  // Parts without accesses or accessed descendants does not exist.
```
dabrahams:
```suggestion
  // Parts without accesses or accessed descendants do not exist.
```

But a part having an accessed descendant implies that the part itself is accessed (albeit inactively), right?  So

```suggestion
  // Only parts with accesses are represented.
```

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,230 @@
...
+/// Validates accesses to an object and its subobjects, where subobjects
+/// are identified by paths of `PathComponent` values.
+///
+/// Invariants:
+///  - Accesses are active when they are created.
+///  - If `x` is an active exclusive access (`sink`, `set`, or `inout`) of an
+///    object `o`, no other access that includes `o` is active.
+struct AccessValidator<PathComponent: Regular & Comparable> {
+
+  // The object `(b: (c: C, d: D), f: F)` may be represented as:
```
dabrahams:
```suggestion
  // The object `(b: (c: C, d: D), f: F)`, with all parts accessed, may be represented as:
```

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,230 @@
...
+  ///
+  /// - Invariant: Has atleast one access or one child.
+  private struct Part {
+
+    /// The step from parent subobject to this subobject.
+    let step: PathComponent
+
+    /// Accesses ordered by creation time, where `let` accesses are
+    /// always active, and all other accesses are active only when they
+    /// are the most recently created access and this part has no subparts.
```
dabrahams, line 33: Is this suggestion accurate?

```suggestion
    /// are the most recently created access and this part has no accessed subparts.
```

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,230 @@
...
+
+  /// A subobject's accesses and immediate subparts.
+  ///
+  /// - Invariant: Has atleast one access or one child.
+  private struct Part {
+
+    /// The step from parent subobject to this subobject.
+    let step: PathComponent
+
+    /// Accesses ordered by creation time, where `let` accesses are
```
dabrahams: Is this suggestion accurate?

```suggestion
    /// Accesses ordered by increasing creation time, where `let` accesses are
```

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,230 @@
...
+
+  }
+
+  /// Allocated parts, including free entries no longer in use.
+  private typealias PartStorage = Array<Part>
+
+  /// Identity of a `Part` in `parts`.
+  private typealias PartID = PartStorage.Index
+
+  /// Allocated parts, including free entries no longer in use.
```
dabrahams, line 54:
```suggestion
  /// Allocated parts, including free entries not currently in use.
```

But allocated and free are normally opposites, so


```suggestion
  /// Storage for parts, including free entries no longer in use.
```

Sources/Interpreter/AccessValidator.swift
RishabhRD: Not necessarily. Let's say we have type: `(outer: (inner: I))`. Then if we do have access to root object, we can have create access to `inner` using `subfield_view` without need of creating access to `outer`.

Sources/Interpreter/AccessValidator.swift
RishabhRD: >   /// are the most recently created access and this part has no accessed subparts.

Yes, just these subparts not necessarily are immediate subparts. I am little worried if it can be confusing with `subparts` field that says "immediate subparts" in API summary.

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,27 @@
...
+import Foundation
+import Utils
+
+/// An access to a `Region` occurring during execution.
+public struct Access<Region: Regular>: Regular {
+
+  /// A unique `Access` identifier.
+  public typealias ID = UUID
+
+  /// A unique `Access` identifier.
```
dabrahams, line 11:
```suggestion
  /// The identity of this access.
```

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,27 @@
+import FrontEnd
+import Foundation
+import Utils
+
+/// An access to a `Region` occurring during execution.
+public struct Access<Region: Regular>: Regular {
+
+  /// A unique `Access` identifier.
```
dabrahams, line 8:
```suggestion
  /// A marker value that distinguishes separately-initiated accesses
  /// having identical `location` and `effect`.
```

dabrahams: There's an argument to be made that equality comparison should only compare `id`s.  I don't think it matters much though

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,27 @@
...
+  public let id: ID
+
+  /// The associated permissions and obligations.
+  public let effect: AccessEffect
+
+  /// The location to which access applies.
+  public let location: Region
+
+  /// Creates an instance of access to `r` having effect `e`.
```
dabrahams, line 20:
```suggestion
  /// Creates a newly-initiated instance of access 
  /// to `r` having effect `e, unequal to to any other
  /// instance with the same properties.
```

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,229 @@
+import FrontEnd
+import Utils
+
+/// Validates accesses to an object and its subobjects, where subobjects
+/// are identified by paths of `PathComponent` values.
```
dabrahams, line 5:
```suggestion
/// are identified by strings of `PathComponent` values.
```

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,229 @@
...
+///
+/// Operations that would otherwise cause an active exclusive access
+/// to overlap another active access throw `AccessValidator.Error`s.
```
dabrahams, line 8:
```suggestion
/// to overlap another active access instead throw `AccessValidator.Error`s.
```

Sources/Interpreter/AccessValidator.swift
dabrahams: That's a separate issue from the addition of the word “accessed” (which AFAICT is needed for accuracy) is it not?

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,229 @@
...
+  //         /   \
+  //        c     d
+  //
+  // Each part stores the accesses currently associated with it.
+  // Parts without accesses or accessed descendants do not exist.
+
+  /// A subobject's accesses and immediate subparts.
+  ///
+  /// - Invariant: Has atleast one access or one child.
+  private struct Part {
```
dabrahams, line 26:
```suggestion
  private struct AccessedPart {
```

on the basis that parts not being accessed are not represented.  WDYT?

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,229 @@
...
+      return nil
+    }
+    let r = Access<PathComponent>(to: parts[p].step, effect: e)
+    self.parts[p].accesses.append(r)
+    return r
+  }
+
+  /// Ends `a` at `p`.
+  ///
+  /// - Postcondition: Provides strong exception safety guarantee.
```
dabrahams, line 103: I don't think we want to say this here.  
When one of these things throws, we are stopping the execution of the interpreter, and there's nothing useful we can do with its state IIUC, so making the strong guarantee promise might not be useful.  And to be useful at all it would have to apply to all the operations, in which case you can make a blanket statement in the type's doc comment.

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  /// Returns the accesses encountered along `p`, including the root.
+  public func accesses(along p: Path) -> [[Access<PathComponent>]] {
+    partIDs(p).map { parts[$0].accesses }
+  }
+
+  /// Returns the part corresponding to `p`, creating any missing
+  /// intermediate parts along `p` if needed.
+  private mutating func demandParts(_ p: Path) -> PartID {
```
dabrahams, line 129:
```suggestion
  private mutating func demandPart(at p: Path) -> PartID {
```

If it's returning a single PartID, it's name has to be singular, even though it may create multiple Part instances, because it returns what you're demanding.

Sources/Interpreter/AccessValidator.swift
```diff
@@ -0,0 +1,229 @@
...
+    var i = 0
+    for c in p {
+      i = part(reachedBy: c, from: i) ?? addChild(c, to: i)
+    }
+    return i
+  }
+
+  /// Returns first part along `p` containing `a`.
+  private func firstPart(in p: Path, containing a: Access<PathComponent>) -> PartID {
```
dabrahams, line 138: Especially if we're not going to explicitly document the precondition, we probably ought to have a trapping check for it.

PR #1867, opened 2026-04-07
Extract frame for projections and projection callers.

lucteo:
The frames will contain all the variables that are used cross regions, which will end up in different functions after elaboration.

Adds a first set of tests for projections.


Sources/Utils/Array+Extensions.swift
```diff
@@ -0,0 +1,28 @@
+extension Array where Element: Equatable {
+
+  /// Appends `element` to `self` iff it is not already contained in `self`, returning the index
+  /// where the element is located.
+  ///
+  /// - Complexity: O(n) where n is the length of `self`.
+  public mutating func appendUnlessContained(_ element: Element) -> Int {
+     if let index = self.firstIndex(of: element) {
```
tothambrus11:
```suggestion
  public mutating func appendUnlessContained(_ element: Element) -> Int {
    if let index = self.firstIndex(of: element) {
```

Sources/Utils/Array+Extensions.swift
```diff
@@ -0,0 +1,28 @@
+extension Array where Element: Equatable {
+
+  /// Appends `element` to `self` iff it is not already contained in `self`, returning the index
+  /// where the element is located.
+  ///
+  /// - Complexity: O(n) where n is the length of `self`.
+  public mutating func appendUnlessContained(_ element: Element) -> Int {
```
tothambrus11: Name suggestion: `appendIfAbsent`.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,171 @@
...
+  /// Materialization metadata for instructions that must become frame-backed.
+  fileprivate var instructionInfo: [InstructionID: InstructionMaterializationInfo] = [:]
+
+  /// Collects instructions in `region` from `f` that must become frame-backed, excluding `ignored`.
+  ///
+  /// We collect all instructions from `region` that are used outside `region`, and all the
+  /// instructions outside of `region` that are used by instructions in `region`.
+  public mutating func collectCrossRegionInstructions<C: Sequence>(
+    in f: Function, from region: C, ignoring ignored: [InstructionID] = []
+  ) where C.Element == InstructionID {
```
tothambrus11: You could rename `C` to `Instructions` (or `S`).

Tests/EndToEndTests/TestCases/Projection/CallerFrame.hylo
```diff
@@ -0,0 +1,12 @@
...
+
+subscript neg(_ x: Int): Int {
+  yield -x
+}
+
+public fun main() {
+  let x = -5 // requires frame to store this, making it accessible in the plateau
+  let p = neg[5]
+  precondition(p == x)
+  let _ = p.copy()
```
tothambrus11, line 11: Is this line necessary? I guess `precondition` would already extend the use after the projection.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -20,31 +20,59 @@ extension IR.Program {
...
+      return
+    }
 
     modules[m]![f].removeGhostDefinitions()
 
+    // Determine the scopes that need lowering.
     let source = modules[m]![f]
-    for d in source.projectionCallingScopes(id: f) {
-      elaborateCallerScope(d, in: m)
+    var scopes = source.projectionCallingScopes(id: f)
```
tothambrus11: We are never mutating the scopes, we just reassign them once. Consider removing the variable and inlining the call to `projectionCallingScopes` into the for loops.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
tothambrus11, line 221: It would be useful to describe what makes a scope required to be elaborated, and what is going to happen scopes that are not `projectionCallingScopes`. Are they completely untouched?

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -20,31 +20,59 @@ extension IR.Program {
...
+
+    // Create the frame; one frame for the entire function.
+    var e = FrameMaterializationInfo()
+    for d in scopes {
+      // Look only in plateau regions; that is skip the first and last regions, which are outside of the plateau.
+      for r in 1 ..< d.regionsCount - 1 {
```
tothambrus11: We could introduce a computed property for this as `d.plateauRegions: Range` in `ScopeDetails`.

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,166 @@
...
+    _ i: InstructionID, in f: Function, region r: C
+  ) -> [InstructionID] where C.Element == InstructionID {
+    f[i].operands.compactMap({ (o) -> InstructionID? in
+      switch o {
+      case .register(let def):
+        return r.contains(def) ? nil : def
+      case .parameter:
+        // We assume parameters are always defined inside the region; emitter will generate local
+        // code for them in the entry block.
+        return nil
```
tothambrus11: I don't quite understand this yet. To my understanding, all plateau regions end up as a new function, so they lose access to the `.parameter` registers of the original function. Or at least the register would refer to the wrong thing. I didn't see code that would pass each original function parameter as inputs to the resume functions, so I suppose they should be part of the frame. What am I missing?

tothambrus11: Is this taking care of this in DictionaryInstructionTransformer?

```swift
  /// Returns a transformed copy of `o`.
  func transform(_ o: Operand) -> Operand {
    switch o {
    case .constant(let c):
      return .constant(c)
    case .parameter(let b, let i):
      return rewrittenParameters[o] ?? .parameter(rewrittenBlocks[b]!, i)
```

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,166 @@
...
+  let subfield: RecordPath
+
+  /// The instruction that allocates the storage.
+  var storageInstruction: InstructionID
+
+  /// Creates metadata for materializing `id` in the frame.
+  public init(id: InstructionID, in f: Function) {
+    self.id = id
+    (self.subfield, self.storageInstruction) = Self.storageTrace(from: id, in: f)
+  }
```
tothambrus11: Some form of precondition or description of the role of `id` would be helpful here. Can it be only a `SubfieldView` or `AllocStack`? The parameter label could be `allocator` or `of i`.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+      }
+    })
+  }
+
+}
+
+/// Materialization metadata for a value that crosses region boundaries.
+///
+/// If instruction `j` uses instruction `i` and they are in different regions, then we need to
+/// materialize `i` in the frame, and not `j`.
```
tothambrus11:
```suggestion
/// If instruction `j` uses instruction `i` and they are in different regions,
/// the storage associated with `i` gets materialized in the frame.
```

I was initially confused by the term "Materialization metadata" and thought that when two SubfieldView instructions with materialization info pointing to the same storage, they would get materialized twice, but in fact they are uniqued by the origin storage. Maybe this phrasing helps a bit.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+  /// Returns `nil` if no frame-backed values are required.
+  mutating func materialize(
+    _ frame: inout FrameMaterializationInfo, in f: Function.ID
+  ) -> Operand? {
+    // Get the storage instructions that need to be placed in the frame.
+    // and the dictionary of their indices in the frame.
+    var storageInstructions: [InstructionID] = []
+    var indices: [InstructionID: Int] = [:]
+    for i in frame.instructionInfo.values {
+      indices[i.id] = storageInstructions.appendUnlessContained(i.storageInstruction)
```
tothambrus11: This is quadratic with respect to the number of elements stored in the frame. Hopefully that number will be usually low (below 100), so I guess that's fine. Otherwise, we could consider OrderedSet which has stable indices.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+///
+/// If instruction `j` uses instruction `i` and they are in different regions, then we need to
+/// materialize `i` in the frame, and not `j`.
+///
+/// Includes the subfield path to get to the storage of `i` from the frame, and the instruction
+/// that allocates the storage.
+private struct InstructionMaterializationInfo {
+
+  /// The ID of the instruction that must be materialized in the frame.
+  let id: InstructionID
```
tothambrus11: Having a property named `id` feels like it's the identity of `self` at use sites. It was not confusing because InstructionMaterializationInfo is used with the `i` abbreviation convention, which feels like it's just an Instruction, but it might be still worth to rename it. Some alternatives could be `subject` (or `source`).

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+        instructionInfo[j] = instructionInfo[j] ?? InstructionMaterializationInfo(id: j, in: f)
+      }
+    }
+  }
+
+  /// Returns `true` if instruction `i` in `f` is used outside the region `r`.
+  private func isUsedOutsideRegion<C: Sequence>(
+    _ i: InstructionID, in f: Function, region r: C
+  ) -> Bool where C.Element == InstructionID {
+    f.allUses(of: i).contains(where: { !r.contains($0.user) })
```
tothambrus11: It would be nice if we could write this in a lazy way without performance overhead, so that the first `contains` could early-return and we wouldn't have to enumerate all uses of an instruction. It's not worth refactoring at this point though, both because the expected element count is very low, and because Swift lazy collections have some overhead.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+  /// Materialization metadata for instructions that must become frame-backed.
+  fileprivate var instructionInfo: [InstructionID: InstructionMaterializationInfo] = [:]
+
+  /// Collects instructions in `region` from `f` that must become frame-backed, excluding `ignored`.
+  ///
+  /// After this call, `instructionInfo` contains all instructions from `region` that are used
+  /// outside `region`, and all the instructions outside of `region` that are used by instructions
+  /// in `region`.
+  public mutating func collectCrossRegionInstructions<C: Sequence>(
+    in f: Function, from region: C, ignoring ignored: [InstructionID] = []
```
tothambrus11: We could pass `ingored` as a lambda, or rather its negation, `includeIf include`. Then it's a bit nicer to use it in the function body `where !ignored.contains(r)` can become `.filter(include)` or `where include(i)`, and it can be slightly more trivial when we only ignore a single or no instruction.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+  ) where C.Element == InstructionID {
+    for i in region where !ignored.contains(i) {
+      // Check if this is used outside the region, and thus it must become frame-backed.
+      if isUsedOutsideRegion(i, in: f, region: region) {
+        instructionInfo[i] = InstructionMaterializationInfo(id: i, in: f)
+      }
+
+      // Check if the defining instructions of operands must become frame-backed.
+      for j in operandsOutsideOfRegion(i, in: f, region: region) where !ignored.contains(j) {
+        instructionInfo[j] = instructionInfo[j] ?? InstructionMaterializationInfo(id: j, in: f)
```
tothambrus11: We don't need to perform the reassignment if the instruction is already present at `j`. We could use this extension:

```swift
extension Dictionary {

  /// Assigns the value computed by `v` to `k` if `k` is not already present in the dictionary.
  ///
  /// - Returns true iff the value had to be inserted.
  @discardableResult
  mutating func insertIfAbsent(forKey k: Key, @autoclosure _ v: () -> Value) -> Bool {
    if !keys.contains(k) { 
      self[k] = v()
      return True
    }
    return False
  }

}
```

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+  /// After this call, `instructionInfo` contains all instructions from `region` that are used
+  /// outside `region`, and all the instructions outside of `region` that are used by instructions
+  /// in `region`.
+  public mutating func collectCrossRegionInstructions<C: Sequence>(
+    in f: Function, from region: C, ignoring ignored: [InstructionID] = []
+  ) where C.Element == InstructionID {
+    for i in region where !ignored.contains(i) {
+      // Check if this is used outside the region, and thus it must become frame-backed.
+      if isUsedOutsideRegion(i, in: f, region: region) {
+        instructionInfo[i] = InstructionMaterializationInfo(id: i, in: f)
```
tothambrus11: `instructionInfo[i]` may be already present, so we could use `insertIfAbsent`.

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,166 @@
...
+
+  /// Returns the type of the frame needed to store `storageInstructions`.
+  private static func frameType(
+    for storageInstructions: [InstructionID], in f: Function
+  ) -> AnyType {
+    var elements: [TupleType.Element] = []
+    elements.append(
+      contentsOf: storageInstructions.enumerated().lazy.map({ (index, i) -> TupleType.Element in
+        TupleType.Element(label: "t\(index)", type: f[i].result!.ast)
+      }))
```
tothambrus11, line 167: This will be a really interesting place for optimizations in the future. E.g. we don't have to store everything sequentially, after the last use of a slot for a certain origin storage, we can repurpose it for subsequent allocations.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+    return AnyType(TupleType(elements))
+  }
+
+  /// Replaces instruction `i` with a subfield view into the `index`th field of `frame`.
+  private mutating func replaceInstruction(
+    _ i: InstructionMaterializationInfo, frame: Operand, index: Int, in f: Function.ID
+  ) {
+    // Add a SubfieldView before each use of `i`, to load the value from the frame.
+    for u in self[f].allUses(of: i.id) {
+      // For other instructions, add a SubfieldView, to access the value from the frame.
```
tothambrus11: What does "other instructions" refer to? Is there any difference between this comment and the one outside the loop?

Sources/IR/Analysis/Module+CheckYieldCoherence.swift
```diff
@@ -58,13 +58,13 @@ extension Function {
...
         // Move the direct successors to the slide.
-        for s in successors { slide.appendUnlessContained(s) }
+        for s in successors { _ = slide.appendUnlessContained(s) }
         // Don't follow this path any further.
         return .skip
       }
 
       // If this is not a yield block, then it's part of the ramp (we don't explore past yield blocks).
-      ramp.appendUnlessContained(b)
+      _ = ramp.appendUnlessContained(b)
```
tothambrus11: We could add `@discardableResult` to `appendUnlessContained`, it seems we usually ignore the result.

tothambrus11 (changes requested):
Great work! This was very fun to review. For a long time, I was just reading the surrounding code, drawing stuff and after any doubt, realizing that in fact the code makes a lot of sense:D 

I can't claim to understand every part of the instruction rewriting yet, specifically around parameter rewriting. I posted a question somewhere in the comments.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
+import FrontEnd
+import Utils
+
+/// Tracks cross-region values that must be materialized in a function frame
```
kyouko-taiga: Are "materialization" and "frame-backing" related?

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,166 @@
...
+import Utils
+
+/// Tracks cross-region values that must be materialized in a function frame
+/// and provides the metadata required to rewrite their uses.
+struct FrameMaterializationInfo {
+
+  /// Materialization metadata for instructions that must become frame-backed.
+  fileprivate var instructionInfo: [InstructionID: InstructionMaterializationInfo] = [:]
+
+  /// Collects instructions in `region` from `f` that must become frame-backed, excluding `ignored`.
```
kyouko-taiga: I find it odd that `region` is singular yet it is a sequence. What's a region exactly?

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+/// and provides the metadata required to rewrite their uses.
+struct FrameMaterializationInfo {
+
+  /// Materialization metadata for instructions that must become frame-backed.
+  fileprivate var instructionInfo: [InstructionID: InstructionMaterializationInfo] = [:]
```
kyouko-taiga:
```suggestion
  fileprivate var instructionMetadata: [InstructionID: InstructionMaterializationInfo] = [:]
```
Not necessarily the best suggestion but I really don't like "xxxInfo".

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
+import FrontEnd
+import Utils
+
+/// Tracks cross-region values that must be materialized in a function frame
+/// and provides the metadata required to rewrite their uses.
+struct FrameMaterializationInfo {
```
kyouko-taiga:
```suggestion
struct FrameMaterialization {
```
Every data type contains information.

Sources/IR/Analysis/FrameMaterialization.swift
kyouko-taiga: It is shorter to write the context bound as `C: Sequence<InstructionID>`

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,166 @@
...
+  public mutating func collectCrossRegionInstructions<C: Sequence>(
+    in f: Function, from region: C, ignoring ignored: [InstructionID] = []
+  ) where C.Element == InstructionID {
+    for i in region where !ignored.contains(i) {
+      // Check if this is used outside the region, and thus it must become frame-backed.
+      if isUsedOutsideRegion(i, in: f, region: region) {
+        instructionInfo[i] = InstructionMaterializationInfo(id: i, in: f)
+      }
+
+      // Check if the defining instructions of operands must become frame-backed.
```
kyouko-taiga: I don't understand this comment. What operands are we talking about?

Sources/IR/Analysis/FrameMaterialization.swift
kyouko-taiga: I would call this method `assignIfAbsent`. Insertion methods typically take a position.

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,166 @@
...
+        return nil
+      case .constant:
+        return nil
+      }
+    })
+  }
+
+}
+
+/// Materialization metadata for a value that crosses region boundaries.
```
kyouko-taiga: I have yet to understand what materialization means.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+}
+
+/// Materialization metadata for a value that crosses region boundaries.
+///
+/// If instruction `j` uses instruction `i` and they are in different regions, then we need to
+/// materialize `i` in the frame, and not `j`.
+///
+/// Includes the subfield path to get to the storage of `i` from the frame, and the instruction
+/// that allocates the storage.
+private struct InstructionMaterializationInfo {
```
kyouko-taiga:
```suggestion
private struct InstructionMaterialization {
```

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+
+  /// The ID of the instruction that must be materialized in the frame.
+  let id: InstructionID
+
+  /// The subfield path to get to the storage of `id` from the frame, if any.
+  let subfield: RecordPath
```
kyouko-taiga:
```suggestion
  private let subfield: RecordPath
```
Explicit is better than implicit.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+
+  /// Creates metadata for materializing `id` in the frame.
+  public init(id: InstructionID, in f: Function) {
+    self.id = id
+    (self.subfield, self.storageInstruction) = Self.storageTrace(from: id, in: f)
+  }
+
+  /// Returns the subfield path and the `AllocStack` instruction at the base of the storage chain
+  /// rooted at `id` in `f`.
+  private static func storageTrace(
```
kyouko-taiga:
```suggestion
  private static func storagePath(
```

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+    from id: InstructionID, in f: Function
```
kyouko-taiga:
```suggestion
    _ id: InstructionID, in f: Function
```
I am not sure `from` is the right label. If I understand the docs correctly, `id` identifies some storage whose origin is defined elsewhere.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,166 @@
...
+extension Module {
+
+  /// Creates storage for `frame` in `f` and rewrites collected instructions to access it.
+  ///
+  /// Returns `nil` if no frame-backed values are required.
+  mutating func materialize(
+    _ frame: inout FrameMaterializationInfo, in f: Function.ID
+  ) -> Operand? {
+    // Get the storage instructions that need to be placed in the frame.
+    // and the dictionary of their indices in the frame.
```
kyouko-taiga:
```suggestion
    // Get the storage instructions that need to be placed in the frame and the dictionary of their
    // indices in the frame.
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -20,31 +20,59 @@ extension IR.Program {
...
+    modules[m]!.replacePlateau(d, region: 1, withPlateau: plateau, frame: frame)
   }
 
   /// Generates the plateau function for region `index` of caller scope `d` in module `m`,
   /// returning its identity.
   private mutating func generatePlateauFunction(
-    for d: ScopeDetails, region index: Int, plateauIndex: Int, in m: Module.ID
+    for d: ScopeDetails,
+    region index: Int,
+    plateauIndex: Int,
```
```suggestion
    plateau: Int,
```

Sources/IR/Analysis/ProjectionElaboration.swift
```diff
@@ -2,6 +2,12 @@ import FrontEnd
 
 extension Module {
 
+  // TODO: remove this
```
kyouko-taiga: When? Under what circumstances?

Sources/Utils/Array+Extensions.swift
```diff
@@ -0,0 +1,28 @@
+extension Array where Element: Equatable {
+
+  /// Appends `element` to `self` iff it is not already contained in `self`, returning the index
```
kyouko-taiga: What's the motivation for using an array rather than a set? I think we already depend on swift-collections so we could simply use an `OrderedSet` so that we won't need this extension.

Sources/IR/Analysis/FrameMaterialization.swift
tothambrus11: Even shorter: `: Sequence<some InstructionID>`

Sources/IR/Analysis/FrameMaterialization.swift
tothambrus11: What about `startingFrom`?

Sources/Utils/Array+Extensions.swift
tothambrus11: The element count is expected to be quite low (the number of distinct storages used across a region boundary). I'm not familiar with the performance characteristics of OrderedSet, but my guess is that any hashing + lookup through additional indirection would be more expensive than a linear scan in the typical case.

Sources/IR/Analysis/FrameMaterialization.swift
lucteo: Materialization, in general is the turning of something abstract into something concrete (or potentiality to actuality). My understanding is also used in backend engineering, but I can't find good references at this point; see: https://discourse.llvm.org/t/about-the-concept-of-materialization/45389)

In this context, I use it as we have some abstract value that needs to be places into a concrete thing, i.e., "materialized".

Yes, frame-backing is the same thing in this here. I used it always like "frame-backed" as the result of materialization.

(Naming these was very tricky to me; although I spent some time on this, I couldn't come up with better names)

Sources/IR/Analysis/FrameReification.swift
lucteo: a "region" is a contiguous sequence of instructions.

Sources/IR/Analysis/FrameMaterialization.swift
lucteo: fixed. thank you!

Sources/IR/Analysis/FrameMaterialization.swift
lucteo: fixed, thank you!

Sources/IR/Analysis/FrameReification.swift
lucteo: This instruction has some operands; these operands might refer to instructions defined outside of the current region. In that case we need to materialize the instructions referred to.

Sources/IR/Analysis/FrameMaterialization.swift
lucteo: fixed, thank you!

Sources/IR/Analysis/FrameMaterialization.swift
lucteo: Calling `allUses` shouldn't be costly, as it returns an already stored array. And the iteration stops the first time the predicate is `true`. So, I'm not sure if we can optimize this easily.

lucteo: ... unless we convert the region into a set. I think that makes a lot of sense, as we are checking a lot of instructions to see if they are in the set.

Fixed, thank you!

Sources/IR/Analysis/FrameReification.swift
lucteo: please see the comment at the top.

Sources/IR/Analysis/FrameMaterialization.swift
lucteo: I like `source`. Thank you!

Sources/IR/Analysis/FrameMaterialization.swift
lucteo: Applied `fileprivate` to all fields (they are all used outside the struct)

Sources/IR/Analysis/FrameReification.swift
lucteo: Added comment, thank you!

Sources/IR/Analysis/FrameMaterialization.swift
lucteo: Indeed. But the number of items stored in the frame should be small. Imagine a frame of 100 elements. That requires an extra-large projection.

While I expect an array to be faster here, this may be premature optimization. So, I'll switch to `OrderedSet`.

Sources/IR/Analysis/FrameReification.swift
lucteo: Yes, we might find ways to optimize this in the future.

Sources/IR/Analysis/FrameMaterialization.swift
lucteo: old comment. Removed. Thank you!

Sources/IR/Analysis/Module+CheckYieldCoherence.swift
lucteo: Reverting the changes related to `appendUnlessContained`.

But thank you, I didn't think about using `@discardableResult`.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: Great catch, thank you!

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: fixed, thank you

Sources/IR/Analysis/ProjectionElaboration.swift
lucteo: fixed, thank you

Sources/Utils/Array+Extensions.swift
lucteo: Switched to `OrderedSet`. For a small number of elements, array should be the fastest option, but that's premature optimization on my end.

Sources/Utils/Array+Extensions.swift
lucteo: reverted this.

Sources/IR/Analysis/FrameMaterialization.swift
```diff
@@ -0,0 +1,172 @@
+import OrderedCollections
+import FrontEnd
+import Utils
+
+/// Tracks cross-region values that must be materialized in a function frame
+/// and provides the metadata required to rewrite their uses.
+struct FrameMaterialization {
```
lucteo: FrameReification

lucteo: describe in the doc of the class: instruction is frame-backed => result of the instruction has been stored into a frame

Sources/IR/Analysis/FrameReification.swift
lucteo: also document the term at the beginning

lucteo: Sorry, I take the "contiguous sequence" part back. It doesn't need to be like that. It can be any set of instructions, the order doesn't matter. For now, it doesn't matter, but it will in the future, when we deal with projections and projection callers that deal with more than one block per region.

Tests/EndToEndTests/TestCases/Projection/CallerFrame.hylo
lucteo: strictly speaking NO. But the generated code is widely different.

If we don't add this line, then `precondition` will be executed in the slide, but its argument, a `Bool` will be put in the frame.

I do like having this line here, as it makes it very clear that `x` extends past the `end_project` instruction. (it took me some time to just look at the difference)

So, I prefer to keep it like this :)

Sources/IR/Analysis/FrameReification.swift
lucteo: good catch! this is a bug. I changed the comment and put a TODO in there. I suspect the changes will be slightly more involved.

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+/// ramp, the value must be reified in a frame that is passed from the ramp to the slide. The frame
+/// reification is the process of detecting which values need to be reified into a frame and
+/// performing the appropriate transformations, such as creating the frame storage and replacing
+/// the relevant instructions to access the frame.
+///
+/// We say that an instruction is "frame-backed" if its value is stored in the frame.
+struct FrameReification {
+
+  /// Materialization metadata for instructions that must become frame-backed.
+  fileprivate var instructionMetadata: [InstructionID: InstructionMaterialization] = [:]
```
tothambrus11:
```suggestion
  /// Reification metadata for instructions that must become frame-backed.
  fileprivate var instructionMetadata: [InstructionID: InstructionReification] = [:]
```

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+      case .register(let def):
+        return r.contains(def) ? nil : def
+      case .parameter:
+        // TODO: implement this
+        return nil
+      case .constant:
+        return nil
+      }
+    })
+  }
```
tothambrus11: Nit: these could be static or placed outside of the struct.

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+
+}
+
+/// Materialization metadata for a value that crosses region boundaries.
```
tothambrus11: We could replace all uses of Materialization with Reification in this PR.

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+      }
+    }
+  }
+}
+
+extension Module {
+
+  /// Creates storage for `frame` in `f` and rewrites collected instructions to access it.
+  ///
+  /// Returns `nil` if no frame-backed values are required.
```
tothambrus11:
```suggestion
  /// Returns the frame allocation's operand, or `nil` if no frame-backed values were required.
```
What the function returns in the non-nil case was non-evident from the docs. Alternatively, we could say "Creates and returns storage for..., iff frame-backed values were required` although that might be a bit too long for a summary.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -20,31 +32,58 @@ extension IR.Program {
...
+    var e = FrameReification()
     for d in source.projectionCallingScopes(id: f) {
-      elaborateCallerScope(d, in: m)
+      // Look only in regions that are plateau.
+      for r in d.plateauRegions {
+        let s = d.splitInstruction(startingRegion: r)
+        let a = d.instructions(region: r)
+        e.collectCrossRegionInstructions(in: source, from: a, ignoring: { (i) -> Bool in
+          return i == s
+        })
```
tothambrus11:
```suggestion
        e.collectCrossRegionInstructions(in: source, from: a, ignoring: { $0 == s })
```
Not so important to change, I guess this is also subjective. I would use the short-hand or omit the `return` keyword for brevity. In my algorithm library, I have a higher order function `equal_to_(v) -> ([V](V) -> Bool)` that could be used to make this predicate as `equal_to_(s)`.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -59,15 +98,27 @@ extension IR.Program {
       d.blocks[0]: entry
     ]
 
-    let rewrittenRegisters: [Operand: Operand] = [
+    var rewrittenRegisters: [Operand: Operand] = [
       .register(oldProject): .parameter(entry, 0)
     ]
 
+    // If we have a frame, take it from the function parameter.
+    if let f = frame {
```
tothambrus11:
```suggestion
    if let frame {
```
I find this helpful so I can avoid introducing a new name for something with essentially the same role. This introduces a shadowed variable with the same name.

tothambrus11:
In the future we could think about if we can decrease the frame size by only reifying the data that's actually needed, not the whole originating allocation. Saving space alone is not a good argument, because we would have the same data on stack anyways, but I could imagine that LLVM optimizations work better if they can shuffle around allocations locally instead of having some big allocation reified into a frame that's going to be used by another function.

tothambrus11 (commented):
LGTM! My main comment is that we could now rename materialization to reification everywhere. I left some stylistic comments, but they might be subjective, so feel free to ignore them (or comment and resolve).

Sources/IR/Analysis/FrameReification.swift
lucteo: fixed, thank you!

Sources/IR/Analysis/FrameReification.swift
lucteo: OMG. I'm sorry for this :(

tothambrus11 (approved):

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -103,7 +101,7 @@ extension IR.Program {
     ]
 
     // If we have a frame, take it from the function parameter.
-    if let f = frame {
+    if let frame {
       let t = modules[m]![d.id].type(of: f).ast
```
tothambrus11:
```suggestion
      let t = modules[m]![d.id].type(of: frame).ast
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
tothambrus11:
```suggestion
      rewrittenRegisters[frame] = ourFrame
```

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+/// performing the appropriate transformations, such as creating the frame storage and replacing
+/// the relevant instructions to access the frame.
+///
+/// We say that an instruction is "frame-backed" if its value is stored in the frame.
+struct FrameReification {
+
+  /// Reification metadata for instructions that must become frame-backed.
+  fileprivate var instructionMetadata: [InstructionID: InstructionReification] = [:]
+
+  /// Collects instructions in `region` from `f` that must become frame-backed, excluding `ignored`.
```
kyouko-taiga:
```suggestion
  /// Populates `instructionMetadata` with the instructions in `region` from `f` that have to be
  /// frame-backed, excluding `ignored`.
```

Sources/IR/Analysis/FrameReification.swift
kyouko-taiga: Ok, then I suggest the following:

```suggestion
      // Check if the definitions of the operands must become frame-backed.
```

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+      }
+
+      // Check if the defining instructions of operands must become frame-backed.
+      for j in Self.operandsOutsideOfRegion(i, in: f, region: regionElements) where !predicate(j) {
+        instructionMetadata.assignIfAbsent(forKey: j, InstructionReification(source: j, in: f))
+      }
+    }
+  }
+
+  /// Returns `true` if instruction `i` in `f` is used outside the region `r`.
```
kyouko-taiga:
```suggestion
  /// Returns `true` iff the instruction `i`, which is in `f`, is used outside the region `r`.
```

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+  private static func isUsedOutsideRegion(
+    _ i: InstructionID, in f: Function, region r: Set<InstructionID>
+  ) -> Bool {
+    f.allUses(of: i).contains(where: { (u) in !r.contains(u.user) })
+  }
+
+  /// Returns the instructions outside `r` that define operands of `i` in `f`.
```
kyouko-taiga:
```suggestion
  /// Returns the instructions outside `r` that define the operands of `i` in `f`.
```

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+  private static func operandsOutsideOfRegion(
+    _ i: InstructionID, in f: Function, region r: Set<InstructionID>
+  ) -> [InstructionID] {
```
kyouko-taiga:
```suggestion
  private static func operands(
    of i: InstructionID, outsideRegion r: Set<InstructionID>, in f: Function
  ) -> [InstructionID] {
```
I'm suggesting putting `f` at the end since both the instruction and the region are in `f`.

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+      }
+    })
+  }
+
+}
+
+/// Reification metadata for a value that crosses region boundaries.
+///
+/// If instruction `j` uses instruction `i` and they are in different regions,
+/// the storage associated with `i` gets reified in the frame.
```
kyouko-taiga:
```suggestion
/// If instruction `j` uses instruction `i` and they are in different regions, the storage
/// associated with `i` gets reified in the frame.
```

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+///
+/// Includes the subfield path to get to the storage of `i` from the frame, and the instruction
+/// that allocates the storage.
+private struct InstructionReification {
+
+  /// The ID of the instruction that must be reified in the frame.
+  fileprivate let source: InstructionID
+
+  /// The subfield path to get to the storage of `id` from the frame, if any.
```
kyouko-taiga: I typically write "if any" for optionals. What's the "none" case of this property?

Sources/IR/Analysis/FrameReification.swift
```diff
@@ -0,0 +1,184 @@
...
+    var i = source
+    while true {
+      switch f[i] {
+      case is AllocStack:
+        return (path, i)
+      case let s as SubfieldView:
+        path.insert(contentsOf: s.subfield, at: 0)
+        i = s.recordPlace.instruction!
+      default:
+        fatalError("Unexpected instruction in storage trace: \(f[i])")
```
kyouko-taiga:
```suggestion
        fatalError("unexpected instruction in storage trace: \(f[i])")
```

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -4,6 +4,18 @@ import FrontEnd
 extension IR.Program {
 
   /// Elaborates all functions in `self` that call projections to use regular function calls.
+  ///
+  /// This pass is similar to `elaborateProjections`, but it may be more complex, from multiple
+  /// perspectives:
+  ///  - A simpler projection caller has three regions: ramp, plateau, and slide.
+  ///  - Multiple projections can be called in the same "scope", and the calls may be interleaved.
+  ///  - A single function might contain multiple "scopes" calling projections.
```
kyouko-taiga: This description is great but it does not really tell me how the pass is similar or dissimilar to `elaborateProjections`.

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
```diff
@@ -20,31 +32,56 @@ extension IR.Program {
...
     modules[m]![f].removeGhostDefinitions()
 
+    // Determine the scopes that need lowering.
     let source = modules[m]![f]
+
+    // Create the frame; one frame for the entire function.
+    var e = FrameReification()
     for d in source.projectionCallingScopes(id: f) {
-      elaborateCallerScope(d, in: m)
+      // Look only in regions that are plateau.
```
```suggestion
      // Look only in regions that are part of the plateau.
```

Sources/Utils/Dictionary+Extensions.swift
```diff
@@ -17,3 +17,19 @@ extension Dictionary where Value: Equatable {
...
 
 }
+
+extension Dictionary {
+
+  /// Assigns the value computed by `v` to `k` if `k` is not already present in the dictionary.
+  ///
+  /// - Returns `true` iff the value had to be inserted.
+  @discardableResult
+  public mutating func assignIfAbsent(forKey k: Key, _ v: @autoclosure () -> Value) -> Bool {
```
kyouko-taiga:
```suggestion
  public mutating func assignValueIfAbsent(forKey k: Key, _ v: @autoclosure () -> Value) -> Bool {
```
For consistency with Swift's standard library

kyouko-taiga (commented):
Very nice!

Sources/IR/Analysis/Program+ElaborateProjectionCallers.swift
lucteo: the three items here show the main distinctions

Sources/IR/Analysis/FrameReification.swift
lucteo: clarified

kyouko-taiga (approved):
LGTM

PR #1881, opened 2026-05-08
Allows `hc` to be built with static linking

simonvarey:
Allows `hc` to be built with static linking, so the `hc` file can be moved to another directory.

Running `swift build -c release --static-swift-stdlib` will build `hc` as a statically linked executable.

The changes that needed to be made to allow `hc` to be built as a statically linked executable involved making `TestUtils` a test target. This is necessary as its dependency `XCTest` cannot be (easily) linked statically. As `TestUtils` has no non-test dependents, this only involved declaring `TestUtils` as a test target in `Package.swift` and moving the `TestUtils` source files to the `Tests` folder.

I have been building Hylo in a devcontainer on Windows so I haven't had a chance to test this, but it does build and pass the tests and it seems to work when I move it out of the `.build` folder.


tothambrus11:
Hi, welcome and thanks a lot for the PR, building a relocatable package is very useful! The changes related to moving the target wouldn't be necessary if we pass `--product hc` as a flag, as that's supposed to build only the specified product and its dependencies. Moving the target as a test target achieved a similar effect, as test targets are not built by default.

Statically linking the standard library doesn't work on Windows yet (see https://github.com/hylo-lang/hylo-new/pull/152 for details), we need to copy some of the standard library DLL dependencies of `hc.exe` to get a fully relocatable package that works on other computers. In the [new compiler's repo](https://github.com/hylo-lang/hylo-new) we use https://github.com/hylo-lang/swift-portable-tool-bundler/ for this. I just copied the same workflow here and it worked. See the release at https://github.com/hylo-lang/hylo/releases/tag/v-old-0.0.42

We are now transitioning to a new compiler, so most of our efforts go towards hylo-new. If you are interested in contributing to anything (compiler, tooling, website), let me know, I'd be happy to chat and give an introduction to whatever is needed!

simonvarey:
That makes sense about building `hc` as a product. That also makes sense about linking the standard library on Windows. I got some errors about `libswiftCore.so` when I tried to move `hc` to another devcontainer, so that might have been the issue.

The bundle release works really well. Thanks for that!

I'll have a look at the new compiler and let you know if I have any questions or if there is anything I'd like contribute.

tothambrus11:
Thanks for the report!

PR #1882, opened 2026-05-10
Interpreter: integrate safety data structures with interpreter runtime

RishabhRD:
The PR integrates the data structures introduced in #1865 for memory safety to be used in Interpreter runtime. That means any read/write/access/end_access operation in interpreter runtime is going to be validated using `MemorySafetyValidator` in `Memory` to detect UB in unsafe code.

**Going file wise instead of commit wise is going to be helpful in review**.


dabrahams:
I don't understand 

> We don't have pointers for the time being to provide unsafe code directly, and thus we use Memory and Validators to check the behavior of the same in tests.

RishabhRD:
> I don't understand
> 
> > We don't have pointers for the time being to provide unsafe code directly, and thus we use Memory and Validators to check the behavior of the same in tests.

I meant that currently interpreter doesn't support pointers. So, we can't directly express hylo programs that contain UB as pointers are needed for the same.

Instead, the tests directly call `Memory` in bad ways like copying from uninitialized region or access to wrong type, to ensure we detect UB from interpreter.

dabrahams:
I suggest not using “the same” and instead writing the noun you mean, because in the first instance it seems you're saying our tests check the behavior of pointers, which we don't have... a contradiction.  In the second instance it seems you mean that pointers are needed for UB checking.  OK.

I don't know that the remark in question is needed at all, though.  That's exactly how I would expect the validator to be unit tested.

RishabhRD:
Understood. Edited the description based on your suggestions.

Sources/Interpreter/Access.swift
```diff
@@ -24,4 +24,11 @@ public struct Access<Region: Regular>: Regular {
     effect = e
   }
 
+  /// Creates an instance of access to `r` having effect `e` and the given `id`.
```
dabrahams, line 27: “instance of access” is redundant


```suggestion
  /// Creates an instance to `r` having effect `e` and the given `id`.
```

Sources/Interpreter/ComposedRegions.swift
```diff
@@ -73,20 +73,22 @@ struct ComposedRegions {
...
 
-  /// Replaces the initialization records starting at `a` for the
-  /// parts of a `t` instance with the initialization record for a
-  /// `t` instance.
+  /// Records initialization of the `t` instance at `a`, replacing any
+  /// initialization records for its parts or subparts.
   ///
+  /// - Precondition: Any initialization records in the region occupied by
+  ///   the `t` instance at `a` describe parts of that instance or subparts
+  ///   thereof.
```
dabrahams, line 81: Initialization records are an implementation detail


```suggestion
  /// - Precondition: Any objects in the region occupied by
  ///   the instance are parts or subparts of that instance.
```

Sources/Interpreter/ComposedRegions.swift
```diff
@@ -73,20 +73,22 @@ struct ComposedRegions {
...
   /// - Precondition: The composed regions in `self` are consistent
```
dabrahams, line 82: Composed regions are no longer a thing, right?  Shouldn't this precondition be deleted?

Sources/Interpreter/ComposedRegions.swift
dabrahams: Isn't it also a precondition that all parts of the `t` instance have been initialized?

Sources/Interpreter/Memory.swift
dabrahams, line 9: “Memory types in p...” is not grammatical (it would have to be “memory for storing types in p”), and you don't run a program “for” an ABI.

Looking at the usage sites in this PR, I suggest

```suggestion
  public init(forRunning p: TypedProgram, on: any TargetABI) {
```

Or

```suggestion
  public init(forRunning p: TypedProgram, using: any TargetABI) {
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -165,11 +165,18 @@ public struct Interpreter {
       module: entryModuleID,
       functionInModule: entryFunctionID,
       instructionInFunction: entryFunction.firstInstruction(in: entryBlockID)!)
+    memory = Memory(typesIn: p.base, for: UnrealABI())
+
+    // First argument of main.
+    let a: Access<Memory.Place>
```
dabrahams, line 171: So there are no parameterless functions and the ones declared without parameters `sink` a `Void`?  I find that pretty surprising! @kyouko-taiga ?

Sources/Interpreter/Interpreter.swift
```diff
@@ -165,11 +165,18 @@ public struct Interpreter {
...
+    do { a = try memory.begin(.sink, to: memory.allocate(AnyType.void)) } catch {
+      unreachable()
+    }
```
dabrahams, line 174:
```suggestion
    // First argument of main.
    let a = try! memory.begin(.sink, to: memory.allocate(AnyType.void))
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -165,11 +165,18 @@ public struct Interpreter {
...
 
     // The return address of the bottom-most frame will never be used,
     // so we fill it with something arbitrary.
```
dabrahams, line 177: Storing program counters rather than return addresses in the stack will eliminate this problem.

Sources/Interpreter/Interpreter.swift
```diff
@@ -196,14 +203,17 @@ public struct Interpreter {
     print("\(currentInstruction.site): \(currentInstruction)")
     switch currentInstruction {
     case let x as IR.Access:
-      return .value(.init(payload: asPlace(x.source)))
+      let e = x.capabilities.elements.first!  // reifed IR has only one access
```
dabrahams, line 206: The comment is obscure.

Are you saying that there is only one capability in each `Access` by the time the interpreter runs?

If so IMO there should be a method or property on `Access` for retrieving it whose precondition is that there is exactly one capability.

```suggestion
      let e = x.finalCapability
```

-----

Further afield:

In fact IMO `capabilities` should be `potentialCapabilities` and then we could just use `capability` for the property posited above.

Also `capability` is an awfully cumbersome name to use.  I realize there's probably precedent but it's not just a capability but also (in the case of `set`) an obligation.  It would be nice to find a shorter word that we can use to capture these things, such as `mode`. Even `effect` would be more digestible @kyouko-taiga ?

Sources/Interpreter/Interpreter.swift
```diff
@@ -196,14 +203,17 @@ public struct Interpreter {
...
+      let a = try memory.begin(e, to: asPlace(x.source))
```
dabrahams, line 207: The naming in this line isn't working very well. It should be possible to make something like

```
let a = try memory.begin(x.effect, at: x.source)
```

read understandably. 

-----

@kyouko-taiga 

The fact that the access instruction has a `source` is really surprising to me, which hurts understandability here.  A `source` is normally paired with a `target` or `destination`, unless it means “source code,” which in this case it does not.  `location` seems more appropriate.

Sources/Interpreter/Interpreter.swift
```diff
@@ -196,14 +203,17 @@ public struct Interpreter {
...
+      return .value(.init(payload: a))
     case let x as AdvancedByBytes:
       _ = x
     case let x as AdvancedByStrides:
       _ = x
 
     case let x as AllocStack:
       let a = memory.allocate(x.allocatedType)
+      try memory.bind(a.address, to: x.allocatedType)
```
dabrahams, line 216: I guess I failed to create a world in which binding was unneeded :-(

I remember we discussed this at some point.

Sources/Interpreter/Interpreter.swift
```diff
@@ -245,12 +255,15 @@ public struct Interpreter {
...
     case let x as Load:
-      return try .value(.init(payload: memory.builtinValue(in: asPlace(x.source))));
-    case is MarkState:
-      // No effect on program state
+      return try .value(.init(payload: memory.builtinValue(in: asAccess(x.source))));
+    case let x as MarkState:
+      let p = asPlace(x.storage)
+      if !x.initialized {
+        try memory.markDeinitialized(p)
+      }
```
dabrahams, line 263: This code looks suspicious.  Doesn't `MarkState` have a state parameter?  Does it always deinitiailize?

dabrahams (commented):
Stopped partway through because I realized I'm in the old hylo repo, and this PR might be obsolete.  @RishabhRD please advise

Sources/Interpreter/Interpreter.swift
dabrahams, line 206: The comment is obscure.

Are you saying that there is only one capability in each `Access` by the time the interpreter runs?

If so IMO there should be a method or property on `Access` for retrieving it whose precondition is that there is exactly one capability.

```suggestion
      let e = x.finalCapability
```

-----

Further afield:

In fact IMO `capabilities` should be `potentialCapabilities` and then we could just use `capability` for the property posited above.

Also `capability` is an awfully cumbersome name to use.  I realize there's probably precedent but it's not just a capability but also (in the case of `set`) an obligation.  It would be nice to find a shorter word that we can use to capture these things, such as `mode`. @kyouko-taiga ?

Sources/Interpreter/Interpreter.swift
```diff
@@ -245,12 +255,15 @@ public struct Interpreter {
...
       return nil
     case let x as MemoryCopy:
-      try memory.copy(asPlace(x.source), to: asPlace(x.target))
+      try memory.copy(asAccess(x.source), to: asAccess(x.target))
```
dabrahams, line 266: If this sort of thing happens often there should be APIs like `copy` on the `Interpreter` which encapsulate the `asAccess` calls and makes this read as:

```swift
try copy(x.source, to: x.target)
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -340,7 +353,22 @@ public struct Interpreter {
   func asPlace(_ x: Operand) -> Memory.Place {
     switch x {
     case .register(let instruction):
-      topOfStack.registers[instruction]!.payload as! Memory.Place
+      let p = topOfStack.registers[instruction]!.payload
+      return (p as? Access<Memory.Place>)?.location ?? (p as! Memory.Place)
```
dabrahams, line 357: These two lines are complex and hard to understand.  Seems like some abstraction might be missing.  It's surprising to see the “if it's an access use its place; else if it's a place use that directly” logic.

Sources/Interpreter/Interpreter.swift
```diff
@@ -340,7 +353,22 @@ public struct Interpreter {
...
+    case .parameter(_, let i):
+      return topOfStack.parameters[i].location
+    case .constant:
+      preconditionFailure("Constant operand is not a Place.")
```
dabrahams, line 361: The failure message should not simply state a fact.  It should describe what is likely wrong, e.g. `"instruction \(x) at \(x.site) supplies a constant operand where a place is required"` (not saying that's the right message; it's just an example)

Sources/Interpreter/Interpreter.swift
```diff
@@ -340,7 +353,22 @@ public struct Interpreter {
...
+    }
+  }
+
+  /// Interpret `x` as an `Access<Memory.Place>`.
```
dabrahams, line 365: Improperly phrased as an imperative


```suggestion
  /// Interprets `x` as an `Access<Memory.Place>`.
```

But also it should describe what it returns and  avoid encoding type information. 

```suggestion
  /// Returns the  access described by `x`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -340,7 +353,22 @@ public struct Interpreter {
...
+  ///
+  /// - Precondition: `x` is an `AccessedPlace`.
```
dabrahams, line 367: Is `AccessedPlace` different from `Access<Memory.Place>`?  If not, why do I see both all over the code?  If so, the fact that they are pronounced almost the same seems like a potential source of confusion.

Sources/Interpreter/Memory.swift
```diff
@@ -349,28 +368,53 @@ extension Memory {
...
+    let s = source.location
+    let d = destination.location
     precondition(
-      source.type == destination.type,
-      "Copy source type \(source.type) must match with destination type \(destination.type).")
-    // TODO: throw when source is not in composed regions and mark destination region composed.
-    self.withUnsafeBytes(source, havingLayout: typeLayouts[source.type]) { a in
-      self.withUnsafeMutableBytes(destination) {
+      s.type == d.type,
+      "Copy source type \(s.type) must match with destination type \(d.type).")
```
dabrahams, line 378:
```suggestion
      "Copy source type \(s.type) must match the destination type \(d.type).")
```

or

```suggestion
      "Copy source type \(s.type) must match destination type \(d.type).")
```

You can say “I matched A _with_ a corresponding B” but you say “A matches B” or “A and B match”

Sources/Interpreter/Memory.swift
```diff
@@ -349,28 +368,53 @@ extension Memory {
...
+    try allocationSafetyValidator[s.allocation]!
+      .requireCanRead(
+        from: source.accessedRegion, in: self[s.allocation],
+        typeLayouts: &typeLayouts
+      )
+    try allocationSafetyValidator[d.allocation]!
+      .requireCanWrite(
+        to: destination.accessedRegion, in: self[d.allocation],
+        typeLayouts: &typeLayouts
+      )
```
dabrahams, line 389: These read awkwardly.  I'm not sure what to do about it, if anything.  I guess `allocationSafetyValidator` is cumbersome and makes it sound like its job is to check that allocations and deallocations are paired, but here we are checking accesses.


```suggestion
    try accessChecker[s.allocation]!
      .prepareToRead(
        from: source.accessedRegion, in: self[s.allocation],
        typeLayouts: &typeLayouts
      )
    try accessChecker[d.allocation]!
      .prepareToWrite(
        to: destination.accessedRegion, in: self[d.allocation],
        typeLayouts: &typeLayouts
      )
```

Also, I'm not sure the word “accessed” is helping in `accessedRegion`.  Since these things are `Place`s it seems like they could be passed directly to the `prepare…` methods.

Sources/Interpreter/Memory.swift
```diff
@@ -349,28 +368,53 @@ extension Memory {
...
+
+    self.withUnsafeBytes(s, havingLayout: typeLayouts[s.type]) { a in
+      self.withUnsafeMutableBytes(d) {
         var b = $0
```
dabrahams, line 393:
```suggestion
      self.withUnsafeMutableBytes(d) { b in
```

Sources/Interpreter/Memory.swift
```diff
@@ -349,28 +368,53 @@ extension Memory {
...
         b.copyElements(from: a)
       }
     }
+
+    try allocationSafetyValidator[d.allocation]!.markInitialized(
+      d.typedRegion,
+      in:
+        self[d.allocation], typeLayouts: &typeLayouts)
```
dabrahams, line 401: I can imagine no good reason for this line break


```suggestion
    try allocationSafetyValidator[d.allocation]!.markInitialized(
      d.typedRegion,
      in: self[d.allocation], typeLayouts: &typeLayouts)
```

or even

```suggestion
    try allocationSafetyValidator[d.allocation]!.markInitialized(
      d.typedRegion, in: self[d.allocation], typeLayouts: &typeLayouts)
```

Sources/Interpreter/Memory.swift
```diff
@@ -349,28 +368,53 @@ extension Memory {
...
   /// Returns the builtin value stored in `p`.
-  func builtinValue(in p: Place) throws -> BuiltinValue {
-    precondition(p.type.isBuiltin);
-    precondition(allocation[p.allocation] != nil)
-
-    // TODO: throw if `p` doesn't have a composed builtin value.
-    let o = p.offset;
-    let a = allocation[p.allocation]!
-    let t = p.type.base as! BuiltinType
+  mutating func builtinValue(in p: Access<Place>) throws -> BuiltinValue {
```
dabrahams, line 405:
```suggestion
  /// Returns the builtin value stored at `p`.
  mutating func builtinValue(at p: Access<Place>) throws -> BuiltinValue {
```

Just a thought.  There are some subtle linguistic issues here:
1. We store things _in_ places
2. However, as phrased the comment tends to imply `p` itself holds the thing stored there.  It does not.  It really is more like an address, that cannot be accessed except by reference to a domain (a map if speaking physically) being addressed, and we store things _at_ addresses.
3. We don't even have a place anymore; we have an `Access<Place>`.

Point 2 really makes me think place is  the wrong name to use in the interpreter and we should go back to `Address`, or use `TypedAddress` to be more explicit.  I don't know what to do about 3.

Sources/Interpreter/Memory.swift
```diff
@@ -349,28 +368,53 @@ extension Memory {
...
+    precondition(p.location.type.isBuiltin);
+    precondition(allocation[p.location.allocation] != nil)
```
dabrahams, line 407: What kind of a place has no associated allocation?!

Sources/Interpreter/Memory.swift
```diff
@@ -349,28 +368,53 @@ extension Memory {
...
+
+    try allocationSafetyValidator[p.location.allocation]!
+      .requireCanRead(
+        from: p.accessedRegion, in: self[p.location.allocation],
+        typeLayouts: &typeLayouts
+      )
```
dabrahams, line 413: This pattern is going to happen over and over and should be encapsulated in an interpreter method

Sources/Interpreter/Memory.swift
```diff
@@ -381,5 +425,49 @@ extension Memory {
     default: fatalError("Unsupported builtin type \(t).")
     }
   }
+
+  /// Creates and returns new access to `p` having effect `e`.
```
dabrahams, line 429:
```suggestion
  /// Returns a new access to `p` having effect `e`.
```

Sources/Interpreter/Memory.swift
```diff
@@ -381,5 +425,49 @@ extension Memory {
...
+  public mutating func markDeinitialized(_ p: Place) throws {
+    allocationSafetyValidator[p.allocation]!.markDeinitialized(
+      p.typedRegion, typeLayouts: &typeLayouts
+    )
+  }
 }
 
+extension Memory.Place {
+  /// `TypedRegion` inside `allocation`.
+  var typedRegion: Memory.Allocation.TypedRegion {
```
dabrahams, line 453: I don't like that this property has the same name as the type.

Sources/Interpreter/Memory.swift
```diff
@@ -381,5 +425,49 @@ extension Memory {
...
+    .init(offset: self.offset, type: self.type)
+  }
+}
+
+extension Access<Memory.Place> {
+  /// The accessed region within `self.location.allocation`.
+  var accessedRegion: Access<Memory.Allocation.TypedRegion> {
```
dabrahams, line 460: How is this thing not just a repackaging of `self`?  Do we need a separate type?  I'm not sure I understand the landscape of this code.

dabrahams (commented):
I got most of the way through now.  Please LMK if you want to discuss interactively.  It looks like there might be some needless complexity.

PR #1888, opened 2026-05-15
Adds `Equatable` and `Collection` conformance for `String`

simonvarey:
Adds `Equatable` and `Collection` conformance for `String`. Two `String`s are equal if and only if their `utf8` properties are equal. `String`s are treated as a collection of single code point strings.

Also adds a new initializer `(code_point:)` to `UTF8Array` which creates an `UTF8Array` instance with a single UTF8 code point stored inline.

Also makes the `memberwise init` of `String` public.

Also adds tests for `String` and `UTF8Array`.

### DIFF
--- a/StandardLibrary/Sources/Core/String.hylo
+++ b/StandardLibrary/Sources/Core/String.hylo
@@ -22,3 +22,31 @@ public conformance String: Copyable {
   }
 
 }
+
+public conformance String: Collection {
+
+  /// A position in an String.
+  public typealias Position = Int
+
+  /// A string containing a single code unit 
+  public typealias Element = String
+
+  public fun start_position() -> Int { 0 }
+
+  public fun end_position() -> Int { count() }
+
+  public fun position(after p: Int) -> Int { p + 1 }
+
+  /// Returns the number of elements in `self`.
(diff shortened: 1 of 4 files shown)


tothambrus11:
Hi! Thanks for the PR, it's really cool that you could implement this. Our approach to strings is similar to Swift. The basic String API should work correctly for non-english characters, and iteration should yield extended grapheme clusters. That will require linking some unicode processing library with the standard library.

I'm not sure how equality and hashing should be implemented, since a string can have multiple different utf8 representation (e.g. á can be composed of "a" + accent). Should we choose performance or correctness by default?

I think the semantics you have proposed should now be accessed through `myString.utf8[i]` and `myString.utf8 == "abc".utf8`, since that makes it explicit that you are operating on the utf8 representation as a collection. I think this should already work.

simonvarey:
That makes sense. Thanks for the feedback! I was surprised by the idea of O(n) string indexing, but I see that Swift does not index strings by integers, which I hadn't considered. I was thinking that `utf8` was just a implementation detail whose functionality would need to be exposed, but if it is to be used as the UTF8 view of the string then I can see how these conformances would not be needed.

PR #1896, opened 2026-05-29
Interpreter Memory Safety: Optimize canBegin complexity to O(1)

RishabhRD:
Previously `canBegin` method needed to traverse the whole subtree for a let access. However, that is too compute intensive and thus the PR optimizes the same by maintaining a record for exclusive accesses in subtree.

### DIFF
--- a/Sources/Interpreter/AccessValidator.swift
+++ b/Sources/Interpreter/AccessValidator.swift
@@ -37,11 +37,15 @@ struct AccessValidator<PathComponent: Regular & Comparable> {
     /// The immediate subparts.
     var subparts: [PartID]
 
+    /// Number of non-`let` accesses in `self`'s descendant parts.
+    var descendantExclusiveAccessCount: Int
+
     /// Creates an instance reached from its parent by `step`, having no accesses or subparts.
     init(_ step: PathComponent) {
       self.step = step
       self.accesses = []
       self.subparts = []
+      descendantExclusiveAccessCount = 0
     }
 
   }
@@ -85,6 +89,9 @@ struct AccessValidator<PathComponent: Regular & Comparable> {
     guard let r = begin(e, at: n) else {
       throw Error.overlappingExclusiveAccess(p)
     }
+    if e != .let {
+      incrementExclusiveAccessInAncestors(of: p)
+    }
     return r
   }
 
@@ -108,6 +115,9 @@ struct AccessValidator<PathComponent: Regular & Comparable> {
     let ns = partIDs(p)
     parts[ns.last!].accesses.removeAll { $0 == a }
     releaseMaximalEmptySuffix(ns)
+    if a.effect != .let {
+      decrementExclusiveAccessInAncestors(of: p)
+    }
   }
 
   /// Throws iff `a` at `p` is not active.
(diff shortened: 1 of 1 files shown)


Sources/Interpreter/AccessValidator.swift
```diff
@@ -85,6 +89,9 @@ struct AccessValidator<PathComponent: Regular & Comparable> {
     guard let r = begin(e, at: n) else {
       throw Error.overlappingExclusiveAccess(p)
     }
+    if e != .let {
+      incrementExclusiveAccessInAncestors(of: p)
```
dabrahams:
```suggestion
      beginExclusiveAccess(at: p)
```

Sources/Interpreter/AccessValidator.swift
```diff
@@ -108,6 +115,9 @@ struct AccessValidator<PathComponent: Regular & Comparable> {
     let ns = partIDs(p)
     parts[ns.last!].accesses.removeAll { $0 == a }
     releaseMaximalEmptySuffix(ns)
+    if a.effect != .let {
+      decrementExclusiveAccessInAncestors(of: p)
```
dabrahams:
```suggestion
      endExclusiveAccess(at: p)
```

Sources/Interpreter/AccessValidator.swift
```diff
@@ -210,14 +202,22 @@ struct AccessValidator<PathComponent: Regular & Comparable> {
...
+      parts[p].descendantExclusiveAccessCount == 0
     } else {
-      return parts[p].subparts.isEmpty
+      parts[p].subparts.isEmpty
+    }
+  }
+
+  /// Calls `f` with each existing part along `p` in `Path` order, stopping at
+  /// the first missing part.
+  private func forEachExistingPart(_ p: Path, _ f: (PartID) -> ()) {
```
dabrahams, line 211:
```suggestion
  private func forEachPart(along p: Path, _ f: (PartID) -> ()) {
```

Or

```suggestion
  private func forEachCompletePart(along p: Path, _ f: (PartID) -> ()) {
```

If that's what Existing really means.

Sources/Interpreter/AccessValidator.swift
```diff
@@ -176,20 +182,6 @@ struct AccessValidator<PathComponent: Regular & Comparable> {
...
-  {
-    // TODO: this is compute intensive task and would be really bad for performance.
-    try predicate(p)
-      && (try parts[p].subparts.allSatisfy {
-        try allSubparts($0, satisfies: predicate)
-      })
-  }
-
   /// Releases the maximal suffix of parts along `ps` that have neither
   /// `accesses` nor `subparts`, making those parts reusable by
```
dabrahams: If you could try to phrase this more semantically and make it less like an english description of the function body, that would be great.  Does the reader know what “release” means?

Sources/Interpreter/AccessValidator.swift
```diff
@@ -227,4 +227,20 @@ struct AccessValidator<PathComponent: Regular & Comparable> {
     if j == parts[p].subparts.endIndex { return nil }
     return parts[p].subparts[j]
   }
+
+  /// Increment exclusive access count in ancestors of `p.last`, if any.
+  private mutating func incrementExclusiveAccessInAncestors(of p: Path) {
```
dabrahams: A summary that reads exactly like the name is usually a sign that one or both should be improved.

Sources/Interpreter/AccessValidator.swift
RishabhRD: `Path` can have subparts that are not currently existing in the representation. For example, in type: `(outer: (inner: I))`
The tree might currently be represented as:
```
    root
     /
outer
````
However, the function can accept the path `[outer, inner]` and would call `f` for `outer` only as that's the prefix currently existing.

